PR #28, opened 2017-09-05
[Nette] forms setRequired(false)

### LINKED ISSUE #27: [Nette] setRequire(false) on form inputs
Ref https://forum.nette.org/cs/26672-missing-setrequired-true-false-on-field-abc-in-form

TomasVotruba:
Closes #27 

image

### What is the function?

I really didn't get it from [the forum](https://forum.nette.org/cs/26672-missing-setrequired-true-false-on-field-abc-in-form), so at the moment it works like this:

Find on variable of `Nette\Application\UI\Form` type:

```php
->addCondition($form::FILLED)
```

and replace it by:

```php
->setRequired(false)
```

Could you confirm/help me with that @dg && @petrvacha?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![Build Status](https://img.shields.io/travis/TomasVotruba/Rector/master.svg?style=flat-square)](https://travis-ci.org/TomasVotruba/Rector)
 [![Coverage Status](https://img.shields.io/coveralls/TomasVotruba/Rector/master.svg?style=flat-square)](https://coveralls.io/github/TomasVotruba/Rector?branch=master)
 
-This tool will *reconstruct* (change) your code - **run it only in a new clean git branch**.
+This tool will **upgrade your application** for you.
 
 
 ## All Reconstructors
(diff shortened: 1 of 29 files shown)


dg:
Just note: in specific cases, this may be an incompatible change.

TomasVotruba:
I don't actually understand it... this is just first example.

Could you provide cases to make it compatible?

dg:
I mean rare cases. For example:

```php
$form->addText('name')
    ->addCondition($form::FILLED)
        ->addRule(...)
        ->addRule(...);

$form['name']
        ->addRule(...); // this rule is called regardless of whether it is filled in (Nette 2.4 triggers warning here)
```

Automatically changed code:

```php
$form->addText('name')
    ->setRequired(false)
    ->addRule(...)
    ->addRule(...);

$form['name']
        ->addRule(...); // now is this rule called only when input is filled (no warning)
```

TomasVotruba:
Thanks. I haven't seen such use case yet. 

So there should be added another refactoring, that adds  `setRequired()` if there is only `addRule()`?

dg:
Yes, when there is addRule(), there must be setRequired(). 

But every such changes must be checked by the programmer, because in some cases it can change the function.

TomasVotruba:
Sure, the aim is to keep programmer out of manually work and keep only checking. Something like code review.

Thanks, I will look on that later

PR #37, opened 2017-09-07
[Flow] init

TomasVotruba:


TomasVotruba:
Use similar approach to https://github.com/sensiolabs-de/deprecation-detector/blob/master/src/Visitor/Deprecation/FindDeprecatedTagsVisitor.php

Just add inside trigger error checks

TomasVotruba:
Work WIP

PR #46, opened 2017-09-23
added information for install on older PHP version

grogy:
Hi,

I am trying how add this tool to project with older PHP. Here is described setup. What do you think about it?


README.md
```diff
@@ -21,6 +21,21 @@ At the moment these packages are supported:
 composer require rector/rector --dev
 ```
 
+# Install on PHP < 7.1
+
+You must have separated environment with PHP 7.1 (for example in Docker container). When you have it then run next command. The reason for this difficulty setup is your dependencies are only PHP <7.1 compatible and you cannot add 7.1.x package.
```
TomasVotruba: - run next => run following
- "The reason for this difficulty setup is your dependencies are only PHP <7.1 compatible and you cannot add 7.1.x package." 
   - I think this information is already included in the title.

README.md
```diff
@@ -21,6 +21,21 @@ At the moment these packages are supported:
...
+
+```
+composer create-project rector/rector your-path-to-rector
```
TomasVotruba: I like this!

README.md
```diff
@@ -21,6 +21,21 @@ At the moment these packages are supported:
...
+```
+
+When do you have it then you can run all commands like
+
+```
+your-path-to-rector/bin/rector reconstruct src --level=nette24
+your-path-to-rector/bin/rector reconstruct src --level=symfony40
```
TomasVotruba: In this case maybe more common use case will be: 

```bash
your-path-to-rector/bin/rector reconstruct /var/www/old-project --level=nette24
your-path-to-rector/bin/rector reconstruct /var/www/another-old-project --level=symfony40
```

In words: reconstruct more code projects that are not co-located with Rector directory.

README.md
grogy: Fixed

README.md
grogy: Fixed

TomasVotruba:
Thank you, merged :+1:

PR #47, opened 2017-09-23
search files recursive in directories

grogy:
When we set a directory then make sense search recursive in subdirectories too.

Reference to `Nette\Utils` API - https://api.nette.org/2.4/Nette.Utils.Finder.html#_from


TomasVotruba:
Nice catch, thanks!

PR #54, opened 2017-09-29
added rector for Environment in Nette 0.9

grogy:
Rector convert from `Environment` class from Nette 0.9.x to 2.0.x

I added configuration for Nette 0.9.x too. :-)


TomasVotruba:
Wow, amazing work :+1: :clap: 

Thank you

Btw, that's the project you need to upgrade?

grogy:
Yes, it is that ;-)

PR #56, opened 2017-09-30
added rector for Image in Nette 0.9

grogy:


tests/Rector/Contrib/Nette/Image/NetteEnvironmentRector/Wrong/wrong.php.inc
```diff
@@ -0,0 +1,3 @@
+<?php
```
TomasVotruba: All looks good. Just rename the directory to `NetteImageRector`, so autoloading works and the namespace and PSR-4 is kept

grogy: Thanks, I fixed it

TomasVotruba:
Thanks

PR #59, opened 2017-10-02
Merge class naming Nette 2.0 Rectors

TomasVotruba:


TomasVotruba:
@grogy This is the rector you can use to rename classes in groups

PR #93, opened 2017-10-17
Nette extracted rules + GetToMethodCallRector

TomasVotruba:


packages/DeprecatedAnnotation/README.md
```diff
@@ -0,0 +1,7 @@
+# Deprecated Annotation
+
+Idea to normalize and automate deprecations, thanks @JanTvrdik for pushing further
+
+Inspire:
+
+- https://hackernoon.com/how-kotlins-deprecated-relieves-pain-of-colossal-refactoring-8577545aaed
```
TomasVotruba: @jantvrdik I think this post shows the idea you had. Implemented right in IDE :) bit better than dump annotation, but still not automated.

PR #134, opened 2017-11-05
Added changed classes in Nette 2.0

grogy:
Added classes whose have namespace or are renamed in Nette 2.0 from 0.9


src/config/level/nette/nette20.yml
```diff
@@ -2,5 +2,9 @@ rectors:
     Rector\Rector\Dynamic\ClassReplacerRector:
         'Environment': 'Nette\Environment'
         'Image': 'Nette\Image'
+        'IRouter': '\Nette\Application\IRouter'
```
TomasVotruba: No need to prefix those with `\`, or is it causing some troubles?

TomasVotruba:
I've removed the slashes so it's ready to merge.
Let me know if that caused any troubles.

grogy:
It looks and work ok. Thank you

PR #145, opened 2017-11-06
Prefer level shortcuts insetad of config files in readme

OndraM:
I read from top to bottom, meaning I first struggled to get the proper path to yml file for `--config` option working (the readme contained a wrong example - which I also fixed).

Only after I made it, I found on next line in the readme I could just use the shortcut 🙃 .

IMHO the shortcut for `--level` is more simpler usage, and the `--config` is meant for some specific use-cases, so I would first suggest using this one, which will IMHO work for most of people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,13 +34,18 @@ E.g. Do you need to upgrade to Nette 2.4?
 1. Run rector on your `/src` directory
 
 ```bash
-vendor/bin/rector process src --config vendor/bin/rector/src/config/level/nette/nette24.yml
+vendor/bin/rector process src --level nette24
 ```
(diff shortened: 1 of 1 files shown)


TomasVotruba:
The goal of this show functionality then magic. It's less obvious now.

Could you add not that `level` is just shortucs to  `src/config/level/*` directory?

Thanks for fixing wrong path!

OndraM:
Ok, what about like this?

(Or feel free to edit the commit manually on merge :-) ).

TomasVotruba:
Perfect! :+1: 

Thanks

PR #158, opened 2017-11-12
namespace rector fix

mssimi:
Hey, this PR fixes bug, where namespace rector generated wrong class names. I was unable fix one test error, which was already present as it wants me to "Use service and constructor injection rather than instantiation", but service container is not instantiated yet. Have a nice day:)

### DIFF
--- a/src/Rector/Dynamic/NamespaceReplacerRector.php
+++ b/src/Rector/Dynamic/NamespaceReplacerRector.php
@@ -185,11 +185,10 @@ final class NamespaceReplacerRector extends AbstractRector
     {
         /** @var FullyQualified $resolvedName */
         $resolvedName = $nameNode->getAttribute(Attribute::RESOLVED_NAME);
-        $fullyQualifiedName = $resolvedName->toString();
         $completeNewName = $this->resolveNewNameFromNode($resolvedName);
 
         // first dummy implementation - improve
-        $cutOffFromTheLeft = strlen($fullyQualifiedName) - strlen($nameNode->toString());
+        $cutOffFromTheLeft = strlen($completeNewName) - strlen($nameNode->toString());
 
         return substr($completeNewName, $cutOffFromTheLeft);
     }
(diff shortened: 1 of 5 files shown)


TomasVotruba:
Hey, thanks for the PR. I don't much understand the issue from the change.

Could you provide minimal code for

- Before
- After
- and After expected 

That would make easier for me to get it :) Thanks

mssimi:
I have namespace rector like this

```
rectors:
    Rector\Rector\Dynamic\NamespaceReplacerRector:
        'fs\admin\AppBundle': 'able\AppBundle'
```
Without fix namespaces regenerated are correct but class short names are not.

Before:

```
<?php

namespace fs\admin\AppBundle\Controller;

use fs\admin\AppBundle\Entity\TomasVotruba;
use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
use Symfony\Bundle\FrameworkBundle\Controller\Controller;

class DefaultController extends Controller
{
    /**
     * @Route("/", name="homepage")
     */
    public function indexAction()
    {
        $tomasVotruba = new TomasVotruba();die;
    }
}
```
After:

```
<?php

namespace able\AppBundle\Controller;

use able\AppBundle\Entity\TomasVotruba;
use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
use Symfony\Bundle\FrameworkBundle\Controller\Controller;

class DefaultController extends Controller
{
    /**
     * @Route("/", name="homepage")
     */
    public function indexAction()
    {
        $tomasVotruba = new sVotruba();die;
    }
}
```
Notice this line  `$tomasVotruba = new sVotruba();die;` it is missing "toma" after new keyword.

TomasVotruba:
Thank you, I understand this much better.

Could I ask you to add test for this? Your code should be just fine.

Adding wrong/correct combination here: https://github.com/rectorphp/rector/blob/master/tests/Rector/Dynamic/NamespaceReplacerRector/Test.php#L27-L30

And your setup to this configuration file: https://github.com/rectorphp/rector/blob/master/tests/Rector/Dynamic/NamespaceReplacerRector/Test.php#L35

It would be ready to merge!

mssimi:
Tests added.

TomasVotruba:
Awesome, thank you!

Issue #133, opened 2017-11-05
Simple readme


grogy:
Task for me - create PR for siple readme. Now it is too long for read.

TomasVotruba:
Resolved

PR #163, opened 2017-11-14
docs - extract install for old PHP into docs folder

grogy:
It is part of my ideas for issue #133.

What do you mean about it? For now are installation steps in one place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ Rector **upgrades your application** for you, with focus on open-source projects
 composer require --dev rector/rector @dev nikic/php-parser 'dev-master#5900d78 as v3.1.1'
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -23,6 +23,8 @@ Rector **upgrades your application** for you, with focus on open-source projects
 composer require --dev rector/rector @dev nikic/php-parser 'dev-master#5900d78 as v3.1.1'
 ```
 
+Do you have old PHP on project? Ok, [it is not problem](/docs/HowUseWithOldPhp.md).
```
TomasVotruba: While we're at it, I'd add note for other dependenies in confllict:
"Do you have old PHP or dependencies in conflict?"

Because some people can have Symfony 3.0 and want to use this. They should find link to *composer project* way here.

grogy: Thank you, you have right!

TomasVotruba:
Thanks, that's exactly what I meant

PR #165, opened 2017-11-14
fix typo

grogy:


TomasVotruba:
There is case that failed tests

grogy:
Thank you, fixed

TomasVotruba:
Thank you

PR #173, opened 2017-11-21
[Dynamic] Value object remover 

### LINKED ISSUE #172: [Draft] Rector to remove value objects
Hi, this is draft of, how we can use Rector in our project. I like it really much 👍 

```php
$variable = new MyObject('hello world');
```
After:
```php
$variable = 'hello world';
```

```php
<?php declare (strict_types=1);

namespace Rector\Contrib\Symfony;

use PhpParser\Node;
use Rector\Rector\AbstractRector;

class RemoveValueObjectRector extends AbstractRector
{
	/** @var array */
	private $classNamesToRefactor;


	public function __construct(array $classNames)
	{
		$this->classNamesToRefactor = $classNames;
	}


	public function isCandidate(Node $node): bool
	{
		if (!$node instanceof Node\Expr\New_) {
			return FALSE;
		}

		if (count($node->args) !== 1) {
			return FALSE;
		}

		$nodeClassName = (string) $node->class;

		if (!in_array($nodeClassName, $this->classNamesToRefactor, TRUE)) {
			return FALSE;
		}

		return TRUE;
	}


	/**
	 * @param Node\Expr\New_ $node
	 */
	{
		return $node->args[0];
	}
}
```

TomasVotruba:
Closes #172


TomasVotruba:
@lexinek This how whole Rector with test would look like

Feel free to add more use cases to test

TomasVotruba:
Add support for property annotation changer:

```php
/**
 * @var Name
 */
private $property;
```

into


```php
/**
 * @var string
 */
private $property;
```

TomasVotruba:
Add support for typehting change:

```php
public function someMethod(Name $name)
{
}
```

to

```php
public function someMethod(string $name)
{
}
```

TomasVotruba:
@lexinek I've covered property and argument type, since I'm on da :train: 
Could you check it?

TomasVotruba:
Merging for now to use in other PRs. Still open to comments

PR #181, opened 2017-12-02
Twig prepared set

greg0ire:


README.md
```diff
@@ -13,6 +13,8 @@ Rector **upgrades your application** for you, with focus on open-source projects
     <a href="/src/config/level/phpunit"><img src="/docs/images/phpunit.jpg"></a>
     <img src="/docs/images/space.png">
     <a href="/src/config/level/roave"><img src="/docs/images/roave.png"></a>
+    <img src="/docs/images/space.png">
+    <a href="/src/config/level/twig"><img src="/docs/images/twig.png"></a>
```
greg0ire: Hmm... the logo is big...

TomasVotruba: I'll handle it, no woriies

TomasVotruba:
Could you please rebase on current master?

Those fails should have been already resolved

greg0ire:
Sure

TomasVotruba:
Awesome work, thank you :clap: 

Just one detail, the config name should be *target version* to upgrade to from one right bellow it.
In this case `twig20.yml`. I'll handle it

TomasVotruba:
Just one question, I've dig throught Twig CHANGELOG and it looks this change was done in:  1.34.0 and 2.4.0 too: https://github.com/twigphp/Twig/blob/2.x/CHANGELOG#L27

Are they compatible?
If so, maybe we should add `twig24.yml` and just include this config. What do you think?

greg0ire:
I though since 1.34 < 2.4, this was how to deal with it, but maybe this reasoning is flawed. If people use 2, they will not consider using files starting with 1?

TomasVotruba:
Well according to change log, version 2.3.x didn't have namespaces yet. Adding namespace was probably backported to 1.3 branch.

So if people use 2.3 and move to 2.4, they will use this config too

Could you confirm @nicolas-grekas?

nicolas-grekas:
I don't remember the precise version number, but namespaces appeard at the same time in v1 and v2.
In a BC/optional way.

greg0ire:
Maybe use a symlink?

greg0ire:
Or maybe using version numbers in the filename is not right? Maybe describe the migration instead?

TomasVotruba:
@nicolas-grekas Thanks, I assumed so.

@greg0ire Symlink would make sense only if  there were no other differences between version.
Using `import` includes single file and allows further extending the config.

TomasVotruba:
Resolved: https://github.com/rectorphp/rector/commit/69105ffcb2487957d0affd0b4bc6c92c4ec55ab1

PR #193, opened 2017-12-19
Add more tests for SpecificMethodsRector

carusogabriel:
@TomasVotruba Tomáš, I was trying to use `rector` in some `PHPUnit` refractories, but `assertEmpty` and `assertArrayHasKey` wasn't working properly. Adding some tests I could verify that:
```diff
1) Rector\Tests\Rector\Contrib\PHPUnit\SpecificMethodRector\Test::test
Original file "/var/www/rector/tests/Rector/Contrib/PHPUnit/SpecificMethodRector/Wrong/wrong.php.inc" did not match the result.
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-        $this->assertArrayHasKey('...', ['...'], 'second argument');\n
-        $this->assertEmpty('...', 'second argument');\n
+        $this->assertArrayHasKey('...', 'second argument');\n
+        $this->assertTrue(empty('...'), 'second argument');\n
```
What do I need to study to fix this? Nikitas's PHP-Parser?

### DIFF
--- a/tests/Rector/Contrib/PHPUnit/SpecificMethodRector/Correct/correct.php.inc
+++ b/tests/Rector/Contrib/PHPUnit/SpecificMethodRector/Correct/correct.php.inc
@@ -5,5 +5,13 @@ final class MyTest extends \PHPUnit\Framework\TestCase
     public function test()
     {
         $this->assertIsReadable('...', 'second argument');
+        $this->assertArrayHasKey('...', ['...'], 'second argument');
+        $this->assertEmpty('...', 'second argument');
+        $this->assertFileExists('...', 'second argument');
+        $this->assertDirectoryExists('...', 'second argument');
+        $this->assertInfinite('...', 'second argument');
+        $this->assertNull('...', 'second argument');
+        $this->assertIsWritable('...', 'second argument');
+        $this->assertNan('...', 'second argument');
     }
 }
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thanks for your work and the diff, that's super easy to read!

Did you managed to check https://github.com/rectorphp/rector/blob/master/src/Rector/Contrib/PHPUnit/SpecificMethodRector.php, where could be the problem?

carusogabriel:
I did. In this [`if`](https://github.com/rectorphp/rector/blob/master/src/Rector/Contrib/PHPUnit/SpecificMethodRector.php#L65):
```php
/** @var MethodCall $node */
if (! isset($node->args[0])) {
    return false
}
```
the execution stops. This is why I asked about learn how `PHP-Parser` works, because maybe we will need to check something there :sweat_smile:

TomasVotruba:
Oh, sorry, I missed the question.

> What do I need to study to fix this? Nikitas's PHP-Parser?

Yep. But not as whole. 

In this case it is tricky, because `empty()` is not `FuncCall`, but `Empty_` node.

See:
https://github.com/rectorphp/rector/pull/194/files#diff-740e97ade92e0a9fb1358a5595da4d41R72

TomasVotruba:
@carusogabriel I've commented the PR, it would be faster and better to explain on specific PR rather than study PHP-Parser :)

Ask anything you need to expalin better there.

carusogabriel:
Gonna open an issue in an hour to share what I've learned while refactoring tests and we can automatize!

PR #194, opened 2017-12-19
[PHPUnit] SpecificMethodsRector fix for empty()

TomasVotruba:


src/Rector/Contrib/PHPUnit/SpecificMethodRector.php
```diff
@@ -104,9 +103,11 @@ private function renameMethod(MethodCall $methodCallNode): void
         [$trueMethodName, $falseMethodName] = $this->oldToNewMethods[$this->activeFuncCallName];
 
         if ($oldMethodName === 'assertTrue' && $trueMethodName) {
-            $methodCallNode->name = $trueMethodName;
+            /** @var string $trueMethodName */
+            $methodCallNode->name = new Identifier($trueMethodName);
```
TomasVotruba: This is not related, just best practise. In PHP-Praser the string would be converted to `Identifier` anyway, this makes it explicit.

src/Rector/Contrib/PHPUnit/SpecificMethodRector.php
```diff
@@ -67,14 +68,12 @@ public function isCandidate(Node $node): bool
         }
 
         $firstArgumentValue = $node->args[0]->value;
-        if (! $firstArgumentValue instanceof FuncCall) {
+
+        $funcCallName = $this->resolveFunctionName($firstArgumentValue);
```
TomasVotruba: Here everything that is not `FuncCall` node is skipped. So `empty()` was too, since it's `Empty_` node.

src/Rector/Contrib/PHPUnit/SpecificMethodRector.php
```diff
@@ -116,4 +117,20 @@ private function moveArgumentUp(MethodCall $methodCallNode): void
...
+        if ($node instanceof FuncCall) {
+            /** @var Name $nameNode */
+            $nameNode = $node->name;
+
+            return $nameNode->toString();
+        }
+
+        if ($node instanceof Empty_) {
+            return 'empty';
+        }
```
TomasVotruba: This is helper function, that can deal with both `FuncCall` and `Empty_`.
Always getting the function name as result.

src/Rector/Contrib/PHPUnit/SpecificMethodRector.php
```diff
@@ -104,16 +103,40 @@ private function renameMethod(MethodCall $methodCallNode): void
...
-        $funcCall = $methodCallNode->args[0]->value;
-        $methodCallNode->args[0] = $funcCall->args[0];
+        $funcCallOrEmptyNode = $methodCallNode->args[0]->value;
+        if ($funcCallOrEmptyNode instanceof FuncCall) {
+            $methodCallNode->args[0] = $funcCallOrEmptyNode->args[0];
+        }
+
+        if ($funcCallOrEmptyNode instanceof Empty_) {
+            // what should be here on the right part instead to fix this? :)
+            $methodCallNode->args[0] = $funcCallOrEmptyNode->args[0];
```
TomasVotruba: @carusogabriel I got question for you. Try to figure this out :)

Now it fails, because we try to get args from `Empty_` node. The [`Empty_` node class might help](https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Expr/Empty_.php)

```
1) Rector\Tests\Rector\Contrib\PHPUnit\SpecificMethodRector\Test::test
Undefined property: PhpParser\Node\Expr\Empty_::$args

/var/www/Rector/src/Rector/Contrib/PHPUnit/SpecificMethodRector.php:123
```

carusogabriel: Okay, let me see. We get `args`, that contains what in inside the function, right?

So we need to find a method that returns what in inside `empty`? :man_technologist:

carusogabriel: Find it! We call `$funcCallOrEmptyNode->expr`, with is the expression passed to `empty()` :smile:

carusogabriel: I gonna try to work on `array_key_exists` fix now.

TomasVotruba: @carusogabriel That is...


... correct :+1:  Wow, you're getting better than me faster than I thought :1st_place_medal: 


I'll finish this `empty()` issue now and merge this PR.

Feel free to send PR for the `array_key_exists` yourself.

PR #196, opened 2017-12-19
First try to fix assertArrayHasKey

carusogabriel:
@TomasVotruba I got something, but now the custom message has been overwritten. I guess we will need to create a new `SpecificMethodArrayRector` or something, so we can keep the Singular Responsibility for each class, and also when I implement new Rectors for `array`s, it will be easier. Actually, I think we'll need to work in `AbstractRector` as well. If you can guide me :+1:


TomasVotruba:
Surpassed by #200

PR #198, opened 2017-12-19
Extend SpecificMethodCountRector

carusogabriel:
I extended `SpecificMethodCountRector` and now we can work with `assertEquals`, `assertNotEquals` and `assertNotSame`. Also, `sizeof`, an alias to `count` is also detected now.

### DIFF
--- a/src/Rector/Contrib/PHPUnit/SpecificMethodCountRector.php
+++ b/src/Rector/Contrib/PHPUnit/SpecificMethodCountRector.php
@@ -13,9 +13,23 @@ use Rector\Rector\AbstractRector;
 /**
  * Before:
  * - $this->assertSame(5, count($anything));
+ * - $this->assertNotSame(5, count($anything));
+ * - $this->assertEquals(5, count($anything));
+ * - $this->assertNotEquals(5, count($anything));
+ * - $this->assertSame(5, sizeof($anything));
+ * - $this->assertNotSame(5, sizeof($anything));
+ * - $this->assertEquals(5, sizeof($anything));
+ * - $this->assertNotEquals(5, sizeof($anything));
  *
  * After:
  * - $this->assertCount(5, $anything);
+ * - $this->assertNotCount(5, $anything);
+ * - $this->assertCount(5, $anything);
+ * - $this->assertNotCount(5, $anything);
+ * - $this->assertCount(5, $anything);
+ * - $this->assertNotCount(5, $anything);
+ * - $this->assertCount(5, $anything);
+ * - $this->assertNotCount(5, $anything);
  */
 final class SpecificMethodCountRector extends AbstractRector
 {
@@ -41,7 +55,7 @@ final class SpecificMethodCountRector extends AbstractRector
         if (! $this->methodCallAnalyzer->isTypesAndMethods(
             $node,
             ['PHPUnit\Framework\TestCase', 'PHPUnit_Framework_TestCase'],
-            ['assertSame']
+            ['assertSame', 'assertEquals', 'assertNotSame', 'assertNotEquals']
         )) {
             return false;
         }
@@ -60,7 +74,9 @@ final class SpecificMethodCountRector extends AbstractRector
             return false;
         }
 
-        return $secondArgumentValue->name->toString() === 'count';
+        $coutableMethod = $secondArgumentValue->name->toString();
+
+        return $coutableMethod === 'count' || $coutableMethod === 'sizeof';
     }
 
     /**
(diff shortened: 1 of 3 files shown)


TomasVotruba:
Btw, could you leave all commits here? I can read changes between than more easily.
Rather 20 commits with story than 1 perfect commit with no background ;)

src/Rector/Contrib/PHPUnit/SpecificMethodCountRector.php
```diff
@@ -60,15 +74,25 @@ public function isCandidate(Node $node): bool
             return false;
         }
 
-        return $secondArgumentValue->name->toString() === 'count';
+        $coutableMethod = $secondArgumentValue->name->toString();
+
+        return $coutableMethod === 'count' || $coutableMethod === 'sizeof';
```
TomasVotruba: For 2+ items to check one value agains, `in_array()` is more useful and easier to add more method names

src/Rector/Contrib/PHPUnit/SpecificMethodCountRector.php
```diff
@@ -60,15 +74,25 @@ public function isCandidate(Node $node): bool
...
-        $methodCallNode->name = new Identifier('assertCount');
+        $oldMethodName = $methodCallNode->name->toString();
+
+        if ($oldMethodName === 'assertEquals' || $oldMethodName === 'assertSame') {
+            /** @var string $trueMethodName */
+            $methodCallNode->name = new Identifier('assertCount');
+        } elseif ($oldMethodName === 'assertNotEquals' || $oldMethodName === 'assertNotSame') {
+            /** @var string $falseMethodName */
+            $methodCallNode->name = new Identifier('assertNotCount');
+        }
```
TomasVotruba: I'm thinking, there should be some service that handle services rename. It's quite often the case.

So this hard to read code could be replaced by:

```php
$renameMap = [
    'assertEquals' => 'assertCount',
    'assertSame' => 'assertCount'
    'assertNotEquals' => 'assertNotCount',
    'assertNotSame' => 'assertNotCount',
];

$this->methodRenamer->renameNode($node, $renameMap);
```


What do you think? :wink:

carusogabriel: Good Idea, it also makes easier to add more assertions in the future.

tests/Rector/Contrib/PHPUnit/SpecificMethodCountRector/Correct/correct.php.inc
```diff
@@ -5,5 +5,12 @@ final class MyTest extends \PHPUnit\Framework\TestCase
     public function test()
     {
         $this->assertCount(5, $something);
```
carusogabriel: @TomasVotruba Gonna add a 3º argument, so we can test the custom error message isn't overwrite.

src/Rector/Contrib/PHPUnit/SpecificMethodCountRector.php
TomasVotruba: Allright, do you want to try it out or should I handle it?

Since we have `Rector\NodeAnalyzer` for analyzing (read-only),
the place for this one could be `Rector\NodeChanger` for changin (write). This one would be  `Rector\NodeChanger\MethodNameChanger` service

carusogabriel: I'm trying. I let you know if everything blow up :fire:

src/Rector/Contrib/PHPUnit/SpecificMethodCountRector.php
```diff
@@ -60,15 +77,21 @@ public function isCandidate(Node $node): bool
...
 
     /**
      * @param MethodCall $methodCallNode
      */
     public function refactor(Node $methodCallNode): ?Node
     {
-        $methodCallNode->name = new Identifier('assertCount');
+        $oldMethodName = $methodCallNode->name->toString();
+
+        $methodCallNode->name = new Identifier($this->renameMethodsMap[$oldMethodName]);
```
carusogabriel: @TomasVotruba I handled it by calling the key in `$this->renameMethodsMap`, since we already filtered the assertions methods in [`L53`](https://github.com/rectorphp/rector/pull/198/files#diff-a7ca84c531c86cadc201c1d5354dd0baR58). Is it right? :confused:

TomasVotruba: Nice one! :+1:

carusogabriel: Sometimes I forget this `PHP` methods :joy:

src/Rector/Contrib/PHPUnit/SpecificMethodCountRector.php
carusogabriel: Sorry, now I understood this comment. We should delegate this function to a class, right? `SOLID` :smile:

TomasVotruba:
Looks ready to merge, is it?

carusogabriel:
@TomasVotruba Yep :tada:

TomasVotruba:
Thanks you :bowing_man:

PR #203, opened 2017-12-20
[PHPUnit] Add in_array and array_search to SpecificMethodRector

carusogabriel:
While refactoring tests, I discovery that `assertTrue(in_array('foo', $anything))` could be refactored to `assertContains('foo', $anything)`. Soo, why not have a `Rector` for that? :sweat_smile: 

@TomasVotruba May I have you opinions before stand this to also support `array_search`? :innocent:

### DIFF
--- a/src/Rector/Contrib/PHPUnit/SpecificMethodRector.php
+++ b/src/Rector/Contrib/PHPUnit/SpecificMethodRector.php
@@ -27,6 +27,8 @@ final class SpecificMethodRector extends AbstractRector
     private $oldToNewMethods = [
         'is_readable' => ['assertIsReadable', 'assertNotIsReadable'],
         'array_key_exists' => ['assertArrayHasKey', 'assertArrayNotHasKey'],
+        'array_search' => ['assertContains', 'assertNotContains'],
+        'in_array' => ['assertContains', 'assertNotContains'],
         'empty' => ['assertEmpty', 'assertNotEmpty'],
         'file_exists' => ['assertFileExists', 'assertFileNotExists'],
         'is_dir' => ['assertDirectoryExists', 'assertDirectoryNotExists'],
--- a/tests/Rector/Contrib/PHPUnit/SpecificMethodRector/Correct/correct.php.inc
+++ b/tests/Rector/Contrib/PHPUnit/SpecificMethodRector/Correct/correct.php.inc
@@ -4,14 +4,16 @@ final class MyTest extends \PHPUnit\Framework\TestCase
 {
     public function test()
     {
-        $this->assertIsReadable('...', 'second argument');
-        $this->assertArrayHasKey('...', ['...'], 'second argument');
-        $this->assertEmpty('...', 'second argument');
-        $this->assertFileExists('...', 'second argument');
-        $this->assertDirectoryExists('...', 'second argument');
-        $this->assertInfinite('...', 'second argument');
-        $this->assertNull('...', 'second argument');
-        $this->assertIsWritable('...', 'second argument');
-        $this->assertNan('...', 'second argument');
+        $this->assertArrayHasKey('...', ['...'], 'argument');
+        $this->assertContains('...', ['...'], 'argument');
+        $this->assertNotIsReadable('...');
+        $this->assertEmpty('...');
+        $this->assertFileNotExists('...');
+        $this->assertDirectoryExists('...');
+        $this->assertFinite('...');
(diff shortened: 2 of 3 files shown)


TomasVotruba:
Thanks for the PR! This will make PHPUnit test fails reporting mucassertTrueh clearer, gj :+1: 

I wonder, could this be added to https://github.com/rectorphp/rector/blob/master/src/Rector/Contrib/PHPUnit/SpecificMethodRector.php#L29

like this? 

```php
 'array_key_exists' => ['assertArrayHasKey', 'assertArrayNotHasKey'],
'in_array' => ['assertContains', 'assertNotContains']
```

The code looks 1:1 at first sight.

carusogabriel:
Actually, this is valid. Just one question, shouldn't we slipt `SpecificMethodRector`? This `Rector` contains `arrays`, `files`, `directories`, etc, and for SOLID and extendability would be easier, wouldn't?

carusogabriel:
Furthermore, in the future, I what to expand this `Rector` to handles `in_array` with strict searching. [See documentation, example number 2](http://php.net/manual/en/function.in-array.php#refsect1-function.in-array-examples).

TomasVotruba:
How would you split it and why?

carusogabriel:
I was wondering in this table:

Rector | Handles
--------- | ---------
`SpecificMethodArrayKeyExistsRector` | `array_key_exists`, `isset($anything['foo'])`
`SpecificMethodsFileRector` | `file_exists`, `file_get_contents`
`SpecificMethodsFileIsRector` | `is_readable`, `is_null`, `is_writable`
`SpecificMethodsIsRector` | `is_nan`, `is_null`, `is_infinite`
`SpecificMethodsDirectoryRector` | `is_dir`
`SpecificMethodsEmptyRector` | `empty`, `empty($anything->foo)`

Do you think is overwork? Maybe we could extend a base class?

TomasVotruba:
And what would be the advantage to do so for us or end-user?

Think of 6x more code, which is 90 % duplicated and also tests.

carusogabriel:
End-user: they can customize what they want to be refactored or no. Current, they call `SpecificMethodRector`, and all the methods in `$oldToNewMethods` will be refactored.

For us: we can extend this `Rector`s easily, without filing `SpecificMethodRector` with `if`s and `else`s, import more `Node`s, etc. Also, is **`S`**`OLID` (Single Responsibility Principle) to each `Rector` handles a refactor, isn't?

But is up to you, if you want to keep `SpecificMethodRector`, no problem :sweat_smile:

TomasVotruba:
Maybe I can't image the use case, because I'd go only the one way that is used now  :sweat_smile: 

Could you provide 2 concurrent use cases in code?

carusogabriel:
Imagine that old test developer, that always use `assertTrue(file_exists($file))`, and still wants to use it. But, he discovers `Rector`, and wants to refactories his tests, except by `file_exists`. He can create his on config file without `SpecificMethodsFileRector`. Did I imagine it too far?

TomasVotruba:
I see, I'm not sure anyone would like create own config and pick which methods to rename and which not... but let's say there will be someone.

In such case, this Rector can be configurable via constructor options:
https://github.com/rectorphp/rector/blob/master/src/Rector/Dynamic/ClassReplacerRector.php#L14-L25

Like Dynamic rectors now are: https://github.com/rectorphp/rector#simple-setup-with-dynamic-rectors


So just add a constructor, where methods to change will be listed:

```php
public function  __construct(array $methodsToChange = [])
{
    $this->methodsToChange = [] ? $this->defaults;
}
```

And then add configurable options to the docs:

```yaml
rectors:
    Rector\Rector\Contrib\PHPUnit\SpecificMethodCountRector:
        - 'is_file'
        - 'count'
```

What do you think?

carusogabriel:
Hm, interesting :thinking: 
Gonna try this and see what we get. Gonna also add `array_search` on this

carusogabriel:
I've added `in_array` and `array_search` to `$oldToNewMethods` array in `SpecificMethodRector`. Also, reduced the verbosity of its tests.
Gonna work on dynamic method, as [suggested](https://github.com/rectorphp/rector/pull/203#issuecomment-353078751) by @TomasVotruba later.

TomasVotruba:
Thank you, look forward to it :+1:

TomasVotruba:
Merging as original proposal in this PR is finished

Thanks for you contribution

PR #213, opened 2017-12-23
[PHPUnit] Improve Rectors class comments

carusogabriel:


src/Rector/Contrib/PHPUnit/SpecificMethod/AssertSameBoolNullToSpecificMethodRector.php
```diff
@@ -11,14 +11,10 @@
...
- * - $this->assertSame(null, $anything);
- * - $this->assertSame(true, $anything);
- * - $this->assertSame(false, $anything);
+ * - $this->assertSame({keyword}}, $anything);
  *
  * After:
- * - $this->assertNull($anything);
- * - $this->assertTrue($anything);
- * - $this->assertFalse($anything);
+ * - $this->assert{keyword}($anything);
```
TomasVotruba: This could mean that any `assert*` function is changed, which is not true.
Only these 3 method are touched and it's even in class name

src/Rector/Contrib/PHPUnit/SpecificMethod/AssertTrueFalseToSpecificMethodRector.php
```diff
@@ -14,10 +14,12 @@
...
 /**
  * Before:
- * - $this->assertTrue(is_readable($readmeFile), 'messag');
+ * - $this->assertTrue({function}($something), 'messag');
+ * - $this->assertFalse({function}($something), 'messag');
  *
  * After:
- * - $this->assertIsReadable($readmeFile, 'message'));
+ * - $this->assert{$function}($something, 'message'));
+ * - $this->assertNot{$function}($something, 'message'));
```
TomasVotruba: These comments should be before/after examples. As less generic as possible, so people understand it with away.

`assert{$function}` could be anything

src/Rector/Contrib/PHPUnit/SpecificMethod/AssertTrueIssetToObjectHasAttributeRector.php
```diff
@@ -15,9 +15,11 @@
 /**
  * Before:
  * - $this->assertTrue(isset($anything->foo));
+ * - $this->assertFalse(isset($anything->foo));
  *
  * After:
  * - $this->assertObjectHasAttribute('foo', $anything);
+ * - $this->assertObjectNotHasAttribute('foo', $anything);
```
TomasVotruba: Nice, I like this one!

src/Rector/Contrib/PHPUnit/SpecificMethod/AssertSameBoolNullToSpecificMethodRector.php
carusogabriel: Gonna fix them in #214!

src/Rector/Contrib/PHPUnit/SpecificMethod/AssertTrueIssetToObjectHasAttributeRector.php
carusogabriel: I keep this one, the other two I couldn't figure out better comments

TomasVotruba:
Thanks!

PR #215, opened 2017-12-23
[PHPUnit] Add alias for is_int and is_float

carusogabriel:
PHP has some alias for `is_int` and `is_float`, and [`PHPUnit` does support them](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/IsType.php#L107). So, why not to add it? :nerd_face:


TomasVotruba:
Thank you!

PR #220, opened 2017-12-25
Add ProgressBar 

TomasVotruba:
Checking many files deserves nice informative progress bar, how far long the progress bar is.

### Before

image

### After

image

### Support for debug verbosity!

image


src/Console/Command/ProcessCommand.php
```diff
@@ -137,6 +139,7 @@ private function ensureSomeRectorsAreRegistered(): void
     private function processFiles(array $fileInfos): void
     {
         $this->symfonyStyle->title('Processing files');
+        $this->symfonyStyle->progressStart(count($fileInfos));
```
carusogabriel: So easy to add this functionality :tada:

TomasVotruba: It was pain before. Gotta know the right people ;)

PR #222, opened 2017-12-25
[NodeChanger] Use IdentifierRenamer by Rector

carusogabriel:
Let's give one step at time :innocent: 

Before extends `NodeChanger\IdentifierRenamer` to support `PHPUnit`, let's use it by `Rector` and see what we need :santa:


src/NodeChanger/MethodNameChanger.php
```diff
@@ -3,20 +3,21 @@
...
 use PhpParser\Node\Identifier;
 
 final class MethodNameChanger
 {
     /**
-     * @param string[] $renameMethodMap
+     * @param string|string[] $newMethodName
      */
-    public function renameNode(MethodCall $node, array $renameMethodMap): ?Node
+    public function renameNode(Node $node, $newMethodName): void
```
TomasVotruba: This should never happen, always strict type agains `string` or `array`.

src/NodeChanger/MethodNameChanger.php
```diff
@@ -3,20 +3,21 @@
...
     {
-        $oldNodeMethodName = $node->name->toString();
+        if (is_array($newMethodName)) {
```
TomasVotruba: `if (is_<something>)` is code smell that suggest method decoupling

src/Rector/Contrib/Nette/Application/TemplateMagicInvokeFilterCallRector.php
```diff
@@ -72,7 +80,7 @@ private function changeToInvokeFilterMethodCall(MethodCall $methodCallNode): voi
         $filterName = $identifierNode->toString();
         $filterArguments = $methodCallNode->args;
 
-        $methodCallNode->name = new Identifier('invokeFilter');
+        $this->methodNameChanger->renameNode($methodCallNode, 'invokeFilter');
```
TomasVotruba: I can imagine this use case as valid :+1:

src/NodeChanger/MethodNameChanger.php
TomasVotruba: The other method could be `renameNodeWithMap(Node $node, array $renameMethodMap)`

TomasVotruba:
Nice refactoring, great job!

src/NodeChanger/MethodNameChanger.php
carusogabriel: Yep, I though in `renameToNode` in case of a `string` given and `renameNodeWithMap` with an `array` was passed!

TomasVotruba: That is ambiguous behavior for outside scope.

Usually I'd recommend you *Clean Code* or [Principles of Package Design](https://leanpub.com/principles-of-package-design), but you see this short version ;):

- https://github.com/jupeter/clean-code-php#avoid-conditionals
- https://github.com/jupeter/clean-code-php#dont-use-flags-as-function-parameters
-

carusogabriel: @TomasVotruba Couldn't have a better teacher. I'm learning a lot working with you!

src/NodeChanger/MethodNameChanger.php
```diff
@@ -3,20 +3,22 @@
 namespace Rector\NodeChanger;
 
 use PhpParser\Node;
-use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Identifier;
 
 final class MethodNameChanger
 {
+    public function renameNode(Node $node, string $newMethodName): void
```
TomasVotruba: Is there some case when `MethodCall` typehint fails?

src/NodeChanger/MethodNameChanger.php
```diff
@@ -3,20 +3,22 @@
...
      * @param string[] $renameMethodMap
      */
-    public function renameNode(MethodCall $node, array $renameMethodMap): ?Node
+    public function renameNodeWithMap(Node $node, array $renameMethodMap): void
     {
         $oldNodeMethodName = $node->name->toString();
 
         $node->name = new Identifier($renameMethodMap[$oldNodeMethodName]);
-
-        return $node;
```
TomasVotruba: Nice decoupling :mrs_claus:

src/NodeChanger/MethodNameChanger.php
carusogabriel: Yep, see [this commit](https://github.com/rectorphp/rector/commit/4eb9c376dc25f953009831f9a1c5da1cad9c70a2), I explained why on its title

src/NodeChanger/MethodNameChanger.php
carusogabriel: Thanks 🤓

src/NodeChanger/MethodNameChanger.php
TomasVotruba: This is exactly the case where this strict typing and decoupling to this service helps :)

It should be fixed on place where `Expr` is passed. `MethodCall` should be passed instead.

Could you change that on responsible place?

src/NodeChanger/MethodNameChanger.php
```diff
@@ -2,21 +2,23 @@
...
 namespace Rector\NodeChanger;
 
-use PhpParser\Node;
-use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr;
 use PhpParser\Node\Identifier;
 
 final class MethodNameChanger
 {
+    public function renameNode(Expr $node, string $newMethodName): void
```
carusogabriel: @TomasVotruba I could force it to be an `Expr`. Some places, `Rector/Dynamic/PropertyNameReplacerRector` for example, passes a `PropertyFetch` to `renameNode`. What should we do?

TomasVotruba: Well, this is `MethodNameChanger`, so it takes care of `MethodCall`.

`PropertyFetch` should be handled by `PropertyNameChanger` class.

It's similar to [NodeAnalyzers](https://github.com/rectorphp/rector/tree/master/src/NodeAnalyzer), where each node has own service that has own rules for them.

So I'd create `PropertyNameChanger`, if I guessed that corretly.

carusogabriel: Hm, one more `Changer`, got! Gonna create it.

If an abstraction arise, I'll handle it :man_technologist:

carusogabriel:
@TomasVotruba 3 more `NameChanger`s were created:
- `PropertyNameChanger`;
- `StaticCallNameChanger`;
- `ClassConstFetchNameChanger`;

src/Rector/Dynamic/MethodNameReplacerRector.php
```diff
@@ -118,7 +134,7 @@ public function refactor(Node $node): ?Node
             return $this->resolveClassRename($node, $oldToNewMethods, $methodName);
         }
 
-        $node->name = new Identifier($oldToNewMethods[$methodName]);
+        $this->methodNameChanger->renameNode($node, $oldToNewMethods[$methodName]);
```
carusogabriel: @TomasVotruba The current errors are coming from here. Should we change some logic?

src/Rector/Dynamic/MethodNameReplacerRector.php
```diff
@@ -134,7 +134,11 @@ public function refactor(Node $node): ?Node
             return $this->resolveClassRename($node, $oldToNewMethods, $methodName);
         }
 
-        $this->methodNameChanger->renameNode($node, $oldToNewMethods[$methodName]);
+        if ($node instanceof MethodCall) {
```
carusogabriel: @TomasVotruba I know we should not check for instances, but I could refactor better this method to fix the [errors](https://travis-ci.org/rectorphp/rector/builds/321886423).

TomasVotruba:
Ready to merge?

carusogabriel:
Ready! The remains renames we need to do I’ll do it seperly to create a `PHPUnitChangeName` 😁

TomasVotruba:
On second read... maybe it would be better to have one rename service, that will check available types in method, **since they share 100% the code**.


```php
class IdentifierRenamer
{

    public function renameNode(Node $node, string $name)
    {
        $this->ensureHasNameIdentifier($node);
        // ...
    }

   private function ensureHasNameIdentifier(Node $node)
   {
       if (get_class($node), ['MethodCall', 'StaticCall', ... ]) {
       }

       throw NodeWithoutIdentifierException(sprintf('....'));
    }
}
```

What do you think?

carusogabriel:
Ok. I'll squash the commits, create this one service called `IdentifierRenamer` and use it everywhere :man_technologist:

TomasVotruba:
Thank you!

carusogabriel:
@TomasVotruba Done. Could you please help me solve the `PHPStan` error?

src/NodeChanger/IdentifierRenamer.php
```diff
@@ -0,0 +1,41 @@
...
+
+        $node->name = new Identifier($renameMethodMap[$oldNodeMethodName]);
+    }
+
+    private function ensureNodeHasIdentifier(Node $node): void
+    {
+        $typeNode = Strings::after(get_class($node), '\\', 3);
+
+        if (! in_array($typeNode, [
+            'MethodCall', 'StaticCall', 'PropertyFetch', 'ClassConstFetch',
```
TomasVotruba: Do you know `MethodCall::class` trick?
That allows you to check `get_class($node)` right away.

TomasVotruba:
@carusogabriel Sure.

`was not matched in reported errors` - this means that ignored error is no more present.
You can remove it from `phpstan.neon` > `paremeters > ignoreErrors` section.

src/NodeChanger/IdentifierRenamer.php
carusogabriel: Does it work within `array`s? I didn't know :scream: 

Gonna use it!

TomasVotruba: It does, since it's only "is string in strings?".

```php
in_array(get_class($node), [MethodCall::class, StaticCall::class...]
```

For `instanceof` it doesn't work for arrays.

src/Rector/Contrib/PHPUnit/SpecificMethod/AssertTrueFalseInternalTypeToSpecificMethodRector.php
```diff
@@ -93,17 +93,12 @@ public function isCandidate(Node $node): bool
...
         $this->moveFunctionArgumentsUp($methodCallNode);
 
         return $methodCallNode;
     }
 
-    private function renameMethod(MethodCall $methodCallNode): void
-    {
-        $this->methodNameChanger->renameNode($methodCallNode, $this->renameMethodsMap);
-    }
-
```
TomasVotruba: Nice cleanup!

src/NodeChanger/IdentifierRenamer.php
```diff
@@ -0,0 +1,42 @@
...
+
+    private function ensureNodeHasIdentifier(Node $node): void
+    {
+        if (! in_array(get_class($node), [
+            ClassConstFetch::class, MethodCall::class, PropertyFetch::class, StaticCall::class,
+        ])) {
+            throw new Exception('The given Node does not have a valid Identifier.');
+        }
+    }
+}
```
TomasVotruba: Simple and clear solution :heart:

TomasVotruba:
Thank you for you work, patience with my comments and all the refactoring to improve this code base :+1:

Issue #208, opened 2017-12-21
[PHPUnit] Can not get PHPUnit assert methods refactored


carusogabriel:
@TomasVotruba As suggested to you, I'm opening this issue to track this. I'll use [`Phalcon Framework`](https://github.com/phalcon/cphalcon) as an example. Steps to reproduce:
### 1º: Fork
I forked and checkout the `4.0.x` branch.

### 2º Run Rector
I installed `Rector` globally, with #207, and I ran:
```
rector process tests/ --level phpunit60
```
Only the `tests/_ci/phpunit.php` file got changed:
```diff
- PHPUnit_TextUI_Command::main();
+ PHPUnit\TextUI\Command::main();
```
There are many `file_exists`, `is_*` (not merged wet here #199), that didn't got affected. I tried to reproduce a PR I did a while ago there [`#13192`](https://github.com/phalcon/cphalcon/pull/13192).

TomasVotruba:
I recall I had many troubles with `compser global` at @apigen, let's narrow allowed installitions to `composer create-project *` and `composer require` to not waste time with that.

Could you retry with that?

TomasVotruba:
Also, have you run `composer update *` on phalcon? It's vendor deps might be needed.

carusogabriel:
I tried. Here as the steps:
### 1º: Fork
I forked and checkout the `4.0.x` branch.
### 2º Install cphalcon dependencies
I installed cphalcon dependencies with `composer install`.
### 3º Install rector with composer
I ran:
```sh
composer require --dev rector/rector @dev nikic/php-parser '4.0.x-dev'
```
And got some errors:
```sh
 Problem 1
    - rector/rector dev-master requires symplify/easy-coding-standard ^3.0 -> satisfiable by symplify/easy-coding-standard[v3.0.0, v3.0.1, v3.0.2, v3.0.3, v3.0.4, v3.0.5, v3.0.6].
    - rector/rector dev-rector-phpunit-simpler requires symplify/easy-coding-standard ^3.0 -> satisfiable by symplify/easy-coding-standard[v3.0.0, v3.0.1, v3.0.2, v3.0.3, v3.0.4, v3.0.5, v3.0.6].
    - symplify/easy-coding-standard v3.0.6 requires squizlabs/php_codesniffer ^3.2 -> satisfiable by squizlabs/php_codesniffer[3.2.0, 3.2.1, 3.2.2, 3.x-dev] but these conflict with your requirements or minimum-stability.
[...]

PR #224, opened 2017-12-25
Use Tolerant ReflectionDocblock to prevent mailformed annotation exceptions

TomasVotruba:
Ref https://github.com/rectorphp/rector/issues/208#issuecomment-353876429

+ Extra 

### Before

before

### After

after


TomasVotruba:
That's way to go! Thanks to `-vvv` we can also see the whole stack trace.

Wait for any reports, I'm still investigating the laravel case :mag:

carusogabriel:
@TomasVotruba I'm as well. With this help me see what files are causing issues and fix our `Rector`s!

PR #225, opened 2017-12-26
[NodeTypeResolver] performance optimization

TomasVotruba:
Make NodeTypeResolver lazy, since it's the heaviest part of code, that might not even be used.

Only class types, method return types and property types can be used.

### Related Changes

- make use of polymorphic dispatching - https://github.com/nikic/PHP-Parser/commit/1c11626f0a2fda566d9db9f388feb0682093955a
- remove `ClassMethodTypeResolver.php`, only delegates to parent class/interface/trait
- use classes over string in TypeResolver dispatcher, performance difference is barely noticeable: 
image

- add feature list to README, so people get can better idea of what Rector can do


TomasVotruba:
Wait for it! I want to make TypeResolver lazy, it takes 80% of performance at the moment.

packages/NodeTypeResolver/src/NodeTypeResolver.php
```diff
@@ -15,27 +15,25 @@
...
-                continue;
-            }
-
-            // resolve just once
-            if ($node->getAttribute(Attribute::TYPES)) {
-                return $node->getAttribute(Attribute::TYPES);
-            }
+        if (! isset($this->perNodeTypeResolvers[$node->getType()])) {
+            return [];
+        }
```
TomasVotruba: I finally found use of https://github.com/nikic/PHP-Parser/commit/1c11626f0a2fda566d9db9f388feb0682093955a

This should be much faster than foreach

carusogabriel: Let’s see how it goes 🙏🏻

packages/NodeTypeResolver/tests/LazyTypeAttributeTest.php
```diff
@@ -0,0 +1,46 @@
...
+
+        $nodes = $this->getNodesForFileOfType(__DIR__ . '/LazyTypeAttributeSource/MethodCall.php.inc', Variable::class);
+        $this->assertCount(3, $nodes);
+
+        $this->variableNode = $nodes[2];
+    }
+
+    public function testLazyPrototype(): void
+    {
+        $this->variableNode->setAttribute('someKey', function () {
```
carusogabriel: We need to search a CS for return types in anonymous functions :thinking:

carusogabriel: @TomasVotruba [Is this](https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Sniffs/TypeHints/TypeHintDeclarationSniff.php#L26) what we are looking for? 

I don't know how PHP_CodeSniffer works and how we could include it, but if you can guide me :man_teacher:

TomasVotruba:
I'll prepare `NodeTypeResolver` as service approach in separate PR to prevent his one growing any further

Issue #216, opened 2017-12-23
[Proposal][PHPUnit] AssertTrueFalseComparisonToSpecificMethodRector


carusogabriel:
There are a lot of comparisons that could be refactor to `PHPUnit` assertion method:

Current | Better
-- | -
`assertTrue($foo === $bar)` | `assertSame($bar, $foo)`
`assertTrue($foo <> $bar)` | `assertNotEquals($bar, $foo)`
`assertTrue($foo >= $bar)` | `assertGreaterThanOrEqual($bar, $foo)`

@TomasVotruba What do you think? Can I propose this `Rector`? :innocent: 

Nikita's `PHP-Parser` [does handle it](https://github.com/nikic/PHP-Parser/tree/master/lib/PhpParser/Node/Expr/BinaryOp) :+1:

TomasVotruba:
Sure, that make sense :mrs_claus:

TomasVotruba:
Feel free to do these in PR from the start, no need to create issue for it.

carusogabriel:
I'm just finishing to work on #217, I'll start this one!

TomasVotruba:
Take your time. Better 1 finished PR than 5 opened issues :1st_place_medal:

PR #227, opened 2017-12-26
[PHPUnit] Create AssertComparisonToSpecificMethodRector

carusogabriel:
As suggested in #216, I've created a new `AssertComparisonToSpecificMethodRector`, so we can refactor comparisons to better `PHPUnit` assertion methods. 

I just need to add a new error to `ignoreErrors` section in `phpstan.neon`, because the `PhpParser\Node\Expr\BinaryOp::getOperatorSigil()` method [is abstract](https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Expr/BinaryOp.php#L39), so, it raises this error. Otherwise, everything should be ready to go :mrs_claus:


TomasVotruba:
Looks great, not much to comment on :+1:

PR #228, opened 2017-12-26
Add PHPStan to composer all command

carusogabriel:
As the Contribute Guidelines [says](https://github.com/rectorphp/rector#how-to-contribute): `Tests, coding standard and` **`PHPStan`** `checks must pass`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,7 @@
         }
     },
     "scripts": {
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Yea, that was missed because PHPStan was unable to work with php-parser 4-dev we use.
Now it is thanks to phar file.

Thank you!

PR #229, opened 2017-12-26
[PHPUnit] Add get_class to AssertCompareToSpecificMethodRector

carusogabriel:
An easy refactor when checking instances. This assertion:
```php
$this->assertSame('Foo', get_class($foo));
```
Can became:
```php
$this->assertInstanceOf('Foo', $foo);
```


TomasVotruba:
I love how simple this is now thanks to your abstraction skills :zap:

PR #230, opened 2017-12-26
[PHPUnit] Support variables as argument in AssertCompareToSpecificMethodRector

carusogabriel:
Most of the time, people already store what they expected in a `$variable`, and them pass it to the `PHPUnit` assert method. This PR adds support to this.

@TomasVotruba Should we create a method `resolveFirstArgument`, or `Variable`, `LNumber` and `String` have something in common that we could check? Maybe [`Cast`](https://github.com/nikic/PHP-Parser/tree/master/lib/PhpParser/Node/Expr/Cast)?


TomasVotruba:
I think this is ok - readable and clear.

Thanks for updating the test

PR #231, opened 2017-12-26
[PHPUnit] Support keyword in uppercase at AssertSameBoolNullToSpecificMethodRector

carusogabriel:
As the purpose of `Rector` is to upgrade legacy apps, some of them do not follow the `PSR-2` recommendation to [define keywords in lowercase](http://www.php-fig.org/psr/psr-2/#25-keywords-and-truefalsenull). This PR fix some cases that we could face it :+1:


TomasVotruba:
:+1:

PR #233, opened 2017-12-26
Add return type to anonymous function

carusogabriel:


TomasVotruba:
Thanks! Isn't this possible to cover with fixer/sniff? It would save us work

carusogabriel:
@TomasVotruba I'll search it!

PR #234, opened 2017-12-26
travis: split phpstan and cs

TomasVotruba:
To separate reponsibility and possibly speed up test runs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,9 @@ language: php
 matrix:
   include:
     - php: 7.1
-      env: STATIC_ANALYSIS=true
+      env: CODING_STANDARD=true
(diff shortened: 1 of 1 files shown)


carusogabriel:
Why not use `Travis CI` [Build Stages](https://docs.travis-ci.com/user/build-stages) feature? :smile:

TomasVotruba:
This runs in parallel = faster

**Stages run in series** - one waits on another. [Rector had them before, but after waiting 15 minutes for build to finished, they were removed](https://github.com/rectorphp/rector/search?q=stage&type=Commits&utf8=%E2%9C%93).

Stages doesn't make sense for open-source projects, unless you do some deployment and need everything to pass.

PR #235, opened 2017-12-26
[CS] Add Sniff to check missing return type-hinting

carusogabriel:
As suggested in #233 by @TomasVotruba, we must check if we type-hinted our functions.

Did I configure it correctly?

### DIFF
--- a/easy-coding-standard.neon
+++ b/easy-coding-standard.neon
@@ -7,6 +7,8 @@ includes:
 checkers:
     - SlevomatCodingStandard\Sniffs\Classes\UnusedPrivateElementsSniff
 
+    - SlevomatCodingStandard\Sniffs\TypeHints\TypeHintDeclarationSniff
+
     # Metrics
     PHP_CodeSniffer\Standards\Generic\Sniffs\Files\LineLengthSniff:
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Yes, you did. But this one is already active in `php70.neon` set

You can check all loaded checkers by:

```bash
vendor/bin/ecs show
```

carusogabriel:
So we need to find one that checks `anonymous functions`?

TomasVotruba:
I see. Did you find why it doesn't work? 
Maybe test just 1 rule on 1 separate file.

TomasVotruba:
Ux miss click

TomasVotruba:
Also check if there is some invalid case that could be fixed. It's possible everything is correct so it can't find any issue

carusogabriel:
I'm gonna try revert #233 and run `composer check-cs` to see if it detects something. Otherwise, I'll open an issue in `slevomat/coding-standard` reporting it.

carusogabriel:
Closing as is something related to `slevomat/coding-standard`

TomasVotruba:
:+1:

PR #236, opened 2017-12-27
[MethodCallTypeResolver] Extend method resolveMethodCallName

carusogabriel:
One more bug discovered in #208:
```sh
PHP Recoverable fatal error:  Object of class PhpParser\Node\Scalar\MagicConst\Class_ could not be converted to string in /home/carusogabriel/.composer/vendor/rector/rector/packages/NodeTypeResolver/src/PerNodeTypeResolver/MethodCallTypeResolver.php on line 83
```

### DIFF
--- a/packages/NodeTypeResolver/src/PerNodeTypeResolver/MethodCallTypeResolver.php
+++ b/packages/NodeTypeResolver/src/PerNodeTypeResolver/MethodCallTypeResolver.php
@@ -5,9 +5,8 @@ namespace Rector\NodeTypeResolver\PerNodeTypeResolver;
 use PhpParser\Node;
 use PhpParser\Node\Expr\Assign;
 use PhpParser\Node\Expr\MethodCall;
-use PhpParser\Node\Expr\PropertyFetch;
 use PhpParser\Node\Expr\Variable;
-use PhpParser\Node\Scalar\MagicConst\Class_;
+use PhpParser\Node\Identifier;
 use Rector\BetterReflection\Reflector\MethodReflector;
 use Rector\Node\Attribute;
 use Rector\NodeTypeResolver\Contract\PerNodeTypeResolver\PerNodeTypeResolverInterface;
@@ -73,21 +72,13 @@ final class MethodCallTypeResolver implements PerNodeTypeResolverInterface
         return $this->resolveMethodReflectionReturnTypes($methodCallNode, $methodCallVariableType, $methodCallName);
     }
 
-    private function resolveMethodCallName(MethodCall $methodCallNode): string
+    private function resolveMethodCallName(MethodCall $methodCallNode): ?string
     {
-        if ($methodCallNode->name instanceof Variable) {
-            return (string) $methodCallNode->name->name;
+        if ($methodCallNode->name instanceof Identifier) {
+            return $methodCallNode->name->toString();
         }
 
-        if ($methodCallNode->name instanceof PropertyFetch) {
-            return (string) $methodCallNode->name->name;
-        }
-
-        if ($methodCallNode->name instanceof Class_) {
-            return $methodCallNode->name->getName();
-        }
-
-        return (string) $methodCallNode->name;
+        return null;
     }
 
     private function getVariableToAssignTo(MethodCall $methodCallNode): ?string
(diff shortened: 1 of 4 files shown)


packages/NodeTypeResolver/src/PerNodeTypeResolver/MethodCallTypeResolver.php
```diff
@@ -69,15 +70,18 @@ public function resolve(Node $methodCallNode): array
...
             return (string) $methodCallNode->name->name;
         }
 
         if ($methodCallNode->name instanceof PropertyFetch) {
-            // not implemented yet
-            return null;
+            return (string) $methodCallNode->name->name;
+        }
+
+        if ($methodCallNode->name instanceof Class_) {
```
carusogabriel: @TomasVotruba Should we check the instance of `Class_` or we can generalize to `MagicConst`? As they all have the [`getName` method](https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Scalar/MagicConst.php#L27)?

TomasVotruba:
Could you please add failing test case? I'm not sure what this change should do

carusogabriel:
I'll try, because I discovered it in #208, running in Laravel.

carusogabriel:
The only thing that I discovered was the class that causes the failure: [`CacheRepositoryTest`](https://github.com/laravel/framework/blob/master/tests/Cache/CacheRepositoryTest.php). I'm still investigating why.

packages/NodeTypeResolver/src/PerNodeTypeResolver/MethodCallTypeResolver.php
```diff
@@ -72,15 +73,18 @@ public function resolve(Node $methodCallNode): array
         return $this->resolveMethodReflectionReturnTypes($methodCallNode, $methodCallVariableType, $methodCallName);
     }
 
-    private function resolveMethodCallName(MethodCall $methodCallNode): ?string
+    private function resolveMethodCallName(MethodCall $methodCallNode): string
```
nikic: This whole method doesn't make a lot of sense to me. It looks like it picks rather random "names". E.g. for `$foo->$bar()` it gives `"bar"`, for `$foo->{$bar->baz}` it gives `"baz"`, and with this change, for `$foo->{__CLASS__}` it gives `"__CLASS__"`.

What you probably want is to check for `Identifier` only and return null in all other cases (unless you can resolve a dynamic name, of course).

TomasVotruba: I've made some cleanup here - https://github.com/rectorphp/rector/pull/237/commits/1d435beffeb8f3b16c9a69ee71f9311b2686066d - maybe it would be more clear.

I didn't test these cases yet. It should only match non-dynamic values.

> What you probably want is to check for Identifier only and return null in all other cases.

Could you provide some real code so I could better understand?

TomasVotruba:
What needs to be done here to merge this?

carusogabriel:
I'll work in @nikic suggestions, to avoid duplicate code!

packages/NodeTypeResolver/src/PerNodeTypeResolver/MethodCallTypeResolver.php
carusogabriel: @TomasVotruba If you can guide me, please: what this method should do to `Rector`s? What should it return?

TomasVotruba: This class detects `{something}->methodCall()` type.

So in code:

```php
$variable = {something}->methodCall()
```

we know `$variable` is of certain type

Like in this test: https://github.com/rectorphp/rector/blob/7f66efa095e41b0ee0681efc8b569bd6114f90bc/packages/NodeTypeResolver/tests/PerNodeTypeResolver/AssignTypeResolver/Source/MethodCall.php.inc#L7

I reliazed it could be simpler and standalone tested. I will probably look on it

carusogabriel: Okay. I'll try to test and fix everthing!

carusogabriel:
@TomasVotruba I think we are ready to go :man_technologist:

TomasVotruba:
Thank you!

PR #239, opened 2017-12-27
[PHPUnit] Create AssertFalseStrposToContainsRector

carusogabriel:
We commonly assert for substring in `PHPUnit` with `strpos`:
```php
$this->assertTrue(strpos('foo', 'f') !== false);
```
But we can simplify this with `assertContains`, as [it handles it](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/StringContains.php#L82):
```php
$this->assertContains('f', 'foo');
```
As we already have `AssertComparisonToSpecificMethodRector`, that changes the first code to:
```php
$this->assertNotSame(false, strpos('foo', 'f'));
```
And `AssertSameBoolNullToSpecificMethodRector` that changes to:
```php
$this->assertNotFalse(strpos('foo', 'f'));
```
I've created `AssertFalseStrposToContainsRector` to finish this refaction :nerd_face:


TomasVotruba:
Thanks!

PR #240, opened 2017-12-27
Clean comments that do not break PHPStan

carusogabriel:
From what I've studied, `PHPStan` errors can be fixed with `Strong Type-Hinting` via comments. But, actually, getting into `Rector`'s source code, I could remove several of them, as well some namespaces that were only used by this comments. Even rules in `phpstan.neon` I could remove.

@TomasVotruba Is this a valid refactor, or should we keep them?

### DIFF
--- a/src/Rector/Contrib/Nette/DI/ExpandFunctionToParametersArrayRector.php
+++ b/src/Rector/Contrib/Nette/DI/ExpandFunctionToParametersArrayRector.php
@@ -5,6 +5,7 @@ namespace Rector\Rector\Contrib\Nette\DI;
 use Nette\Utils\Strings;
 use PhpParser\Node;
 use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Scalar\String_;
 use Rector\Node\Attribute;
 use Rector\Node\NodeFactory;
 use Rector\NodeAnalyzer\MethodArgumentAnalyzer;
@@ -73,7 +74,7 @@ final class ExpandFunctionToParametersArrayRector extends AbstractRector
      */
     public function refactor(Node $methodCallNode): ?Node
     {
-        /** @var Node\Scalar\String_ $argument */
+        /** @var String_ $argument */
         $argument = $methodCallNode->args[0]->value;
         $argument->value = Strings::trim($argument->value, '%');
 
--- a/bin/rector.php
+++ b/bin/rector.php
@@ -34,11 +34,8 @@ try {
     }
 
     // 3. Run Console Application
-    /** @var Application $application */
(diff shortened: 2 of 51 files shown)


TomasVotruba:
Well, most of them are for IDE, resp. PHPStorm, which needs to know type to autocomplete.
Many values in Node are `mixed`, so e.g. for Identifier it doesn't autocomplete `->toString()` method.

Does PHPStorm work clearly for any case you've removed?

Cleaning `phpstan.neon` for unused ignore errors is ok.

src/Rector/Contrib/Nette/DI/ExpandFunctionToParametersArrayRector.php
```diff
@@ -73,7 +74,7 @@ public function isCandidate(Node $node): bool
      */
     public function refactor(Node $methodCallNode): ?Node
     {
-        /** @var Node\Scalar\String_ $argument */
+        /** @var String_ $argument */
```
TomasVotruba: :+1: 

This could be covered by fixer in Symplify\CodingStandard: https://github.com/Symplify/Symplify/tree/master/packages/CodingStandard#types-should-not-be-referenced-via-a-fullypartially-qualified-name-but-via-a-use-statement-wrench

But it now lacks support for doc types. Would you be able to add so this is CI checked? That'd be great!

carusogabriel:
@TomasVotruba I use `Atom`. Could you please check this branch in `PHPStorm`? Because was a lot of clean up, and I guess having them just for auto-complete is strange, isn't?

src/Rector/Contrib/Nette/DI/ExpandFunctionToParametersArrayRector.php
carusogabriel: Gonna add it in another PR to keep one feature per :smile:

TomasVotruba:
@carusogabriel I checked that briefly and all those annotation are required for PHPStorm, unless new version changed that, which is less likely.

> I guess having them just for auto-complete is strange

Why? Do you prefer code without autocompletion?

carusogabriel:
> Why? Do you prefer code without autocompletion?

No that I do not prefer, but having comments just for that :sweat_smile: 

But okay, now I understand why they are all here. Also, I'll police myself to use them when necessary :policeman:

TomasVotruba:
Thanks!

TomasVotruba:
> No that I do not prefer, but having comments just for that :sweat_smile:

Well there is no other real reason to have them except this one :) 

Btw, have you tried PHPStorm? It saves so much work when used longer than few months

PR #241, opened 2017-12-27
Use NodeFactory by Rector

carusogabriel:


src/Rector/Contrib/PHPUnit/SpecificMethod/AssertTrueFalseInternalTypeToSpecificMethodRector.php
```diff
@@ -110,10 +119,8 @@ private function moveFunctionArgumentsUp(MethodCall $methodCallNode): void
         unset($oldArguments[0]);
 
         $methodCallNode->args = array_merge([
-            new Arg(new String_(
-                $this->oldMethodsToTypes[$isFunctionName]
-            )),
-            new Arg($argument),
+            $this->nodeFactory->createArg(new String_($this->oldMethodsToTypes[$isFunctionName])),
```
carusogabriel: @TomasVotruba As we have `createTrueConstant` and like methods in `NodeFactory`, could we have `createString`? :thinking:

src/Rector/Contrib/PhpParser/SetLineRector.php
```diff
@@ -29,10 +28,19 @@
      */
     private $identifierRenamer;
 
-    public function __construct(MethodCallAnalyzer $methodCallAnalyzer, IdentifierRenamer $IdentifierRenamer)
```
carusogabriel: @TomasVotruba Is there a Fixer for `$identifierRenamer` instead of `$IdentifierRenamer`?

TomasVotruba: You can extend this one: https://github.com/Symplify/Symplify/blob/master/packages/CodingStandard/README.md#property-name-should-match-its-type-if-possible-wrench

TomasVotruba:
Thank you

PR #246, opened 2017-12-28
[ProcessCommand] Improvement by reporting changed files

carusogabriel:
I've made a little improvement to our `ProcessCommand`: report _how many_, and **what files**, were changed :tada:

### Changes:
- Now the `printToFile` and `processFile` methods in `FormatPerservingPrinter` and `FileProcessor` return the result of `file_put_contents`, and as this function [returns `false` in cause of failure](http://php.net/manual/en/function.file-put-contents.php#refsect1-function.file-put-contents-returnvalues), I've also added `return false` in case the file didn't get changed, so we can check it;
- I've used the [`listing` method](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Console/Style/SymfonyStyle.php#L102-L111) from `SymfonyStyle`, to avoid `for` loops in our code. As suggested by @javiereguiluz in [`symfony/symfony#25616`](https://github.com/symfony/symfony/pull/25616#issuecomment-354243140), I've created `ConsoleStyle`, extending from `SymfonyStyle`, so we can continue to get the dashes `-` while listing, instead of stars `*`;
- The `newLine` method, also from `SymfonyStyle`, allows us to pass the number of new lines we want, so we can avoid repeating that command.

### Screenshots:
Before:
before
After:
after

### DIFF
--- a/src/Application/FileProcessor.php
+++ b/src/Application/FileProcessor.php
@@ -44,11 +44,14 @@ final class FileProcessor
         return $this->processFileToString($file);
     }
 
-    public function processFile(SplFileInfo $fileInfo): void
+    /**
+     * @return mixed
+     */
+    public function processFile(SplFileInfo $fileInfo)
     {
         [$newStmts, $oldStmts, $oldTokens] = $this->nodeTraverserQueue->processFileInfo($fileInfo);
 
-        $this->formatPerservingPrinter->printToFile($fileInfo, $newStmts, $oldStmts, $oldTokens);
+        return $this->formatPerservingPrinter->printToFile($fileInfo, $newStmts, $oldStmts, $oldTokens);
     }
 
     /**
--- a/src/Application/FileProcessor.php
+++ b/src/Application/FileProcessor.php
@@ -45,7 +45,7 @@ final class FileProcessor
     }
 
     /**
-     * @return mixed
+     * @return int|bool
      */
     public function processFile(SplFileInfo $fileInfo)
     {
--- a/src/Application/FileProcessor.php
+++ b/src/Application/FileProcessor.php
@@ -44,10 +44,7 @@ final class FileProcessor
         return $this->processFileToString($file);
     }
 
-    /**
-     * @return int|bool
-     */
-    public function processFile(SplFileInfo $fileInfo)
+    public function processFile(SplFileInfo $fileInfo): bool
     {
         [$newStmts, $oldStmts, $oldTokens] = $this->nodeTraverserQueue->processFileInfo($fileInfo);
 
--- a/src/Console/Command/ProcessCommand.php
+++ b/src/Console/Command/ProcessCommand.php
@@ -67,6 +67,11 @@ final class ProcessCommand extends Command
      */
     private $differAndFormatter;
 
+    /**
+     * @var string[]
+     */
+    private $changedFiles = [];
+
     public function __construct(
         FileProcessor $fileProcessor,
         RectorCollector $rectorCollector,
@@ -118,6 +123,11 @@ final class ProcessCommand extends Command
         $this->processCommandReporter->reportLoadedRectors();
 
         $this->processFiles($files);
+
+        if (count($this->changedFiles) > 0) {
+            $this->processCommandReporter->reportChangedFiles($this->changedFiles);
+        }
+
         $this->symfonyStyle->success('Rector is done!');
 
         return 0;
@@ -140,8 +150,9 @@ final class ProcessCommand extends Command
      */
     private function processFiles(array $fileInfos): void
     {
-        $this->symfonyStyle->title('Processing files');
-        $this->symfonyStyle->progressStart(count($fileInfos));
+        $totalFiles = count($fileInfos);
+        $this->symfonyStyle->title(sprintf('Processing %s Files', $totalFiles));
+        $this->symfonyStyle->progressStart($totalFiles);
 
         $i = 1;
         foreach ($fileInfos as $fileInfo) {
@@ -158,14 +169,14 @@ final class ProcessCommand extends Command
             $this->symfonyStyle->progressAdvance();
         }
 
-        $this->symfonyStyle->newLine();
-        $this->symfonyStyle->newLine();
+        $this->symfonyStyle->newLine(2);
     }
 
     private function processFile(SplFileInfo $fileInfo, int &$i): void
     {
+        $oldContent = $fileInfo->getContents();
+
         if ($this->parameterProvider->provideParameter(self::OPTION_DRY_RUN)) {
-            $oldContent = $fileInfo->getContents();
             $newContent = $this->fileProcessor->processFileToString($fileInfo);
 
             if ($newContent !== $oldContent) {
@@ -174,7 +185,11 @@ final class ProcessCommand extends Command
                 ++$i;
             }
         } else {
-            $this->fileProcessor->processFile($fileInfo);
+            $newContent = $this->fileProcessor->processFile($fileInfo);
+
+            if ($newContent !== $oldContent) {
+                $this->changedFiles[] = $fileInfo->getPathname();
+            }
         }
     }
 }
--- a/src/Console/Command/ProcessCommand.php
+++ b/src/Console/Command/ProcessCommand.php
@@ -151,7 +151,7 @@ final class ProcessCommand extends Command
     private function processFiles(array $fileInfos): void
     {
         $totalFiles = count($fileInfos);
-        $this->symfonyStyle->title(sprintf('Processing %s Files', $totalFiles));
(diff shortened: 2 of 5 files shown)


src/Console/ConsoleStyle.php
```diff
@@ -0,0 +1,20 @@
...
+     */
+    public function listing(array $elements): void
+    {
+        $elements = array_map(function ($element) {
+            return sprintf(' - %s', $element);
+        }, $elements);
+        $this->writeln($elements);
+        $this->newLine();
+    }
+}
```
TomasVotruba: Nice move! Personally I'd use compsition over inheritance, but in this case it would create over 100 lines of duplicated code.

So :+1:  for this one

src/Console/Command/ProcessCommand.php
```diff
@@ -118,6 +123,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $this->processCommandReporter->reportLoadedRectors();
 
         $this->processFiles($files);
+
+        if (count($this->changedFiles) > 0) {
+            $this->processCommandReporter->reportChangedFiles($this->changedFiles);
```
TomasVotruba: :+1: for the right place to put this

src/Console/Command/ProcessCommand.php
```diff
@@ -140,8 +150,9 @@ private function ensureSomeRectorsAreRegistered(): void
      */
     private function processFiles(array $fileInfos): void
     {
-        $this->symfonyStyle->title('Processing files');
-        $this->symfonyStyle->progressStart(count($fileInfos));
+        $totalFiles = count($fileInfos);
+        $this->symfonyStyle->title(sprintf('Processing %s Files', $totalFiles));
```
TomasVotruba: Here it should be probably `%d`, due to numeric value

TomasVotruba: And `Process x **f**iles`, since it's informative message and not a headline in post-way

src/Printer/FormatPerservingPrinter.php
```diff
@@ -22,17 +22,18 @@ public function __construct(BetterStandardPrinter $betterStandardPrinter)
      * @param Node[] $newStmts
      * @param Node[] $oldStmts
      * @param Node[] $oldTokens
+     * @return mixed
      */
-    public function printToFile(SplFileInfo $fileInfo, array $newStmts, array $oldStmts, array $oldTokens): void
+    public function printToFile(SplFileInfo $fileInfo, array $newStmts, array $oldStmts, array $oldTokens)
```
TomasVotruba: Could it return `bool` in case of change? It's easier to work with than `mixed`

src/Console/Output/ProcessCommandReporter.php
```diff
@@ -28,18 +28,23 @@ public function reportLoadedRectors(): void
         $this->symfonyStyle->title(sprintf(
             '%d Loaded Rector%s',
             $this->rectorCollector->getRectorCount(),
-            $this->rectorCollector->getRectorCount() === 1 ? '' : 's'
+            $this->rectorCollector->getRectorCount() === 1 ?: 's'
```
TomasVotruba: Nice :+1:

src/Console/Output/ProcessCommandReporter.php
```diff
@@ -28,18 +28,23 @@ public function reportLoadedRectors(): void
...
-        foreach ($rectorList as $rector) {
-            $this->symfonyStyle->writeln(sprintf(
-                ' - %s',
-                get_class($rector)
-            ));
-        }
-
-        $this->symfonyStyle->newLine();
+
+        $this->symfonyStyle->listing(array_map(function ($rector): string {
```
TomasVotruba: Missing typehint on `function (Rector $rector)`

src/Console/Output/ProcessCommandReporter.php
```diff
@@ -43,7 +43,11 @@ public function reportLoadedRectors(): void
      */
     public function reportChangedFiles(array $changedFiles): void
     {
-        $this->symfonyStyle->title(sprintf('%s Changed Files', count($changedFiles)));
+        $this->symfonyStyle->title(sprintf(
+            '%s Changed File%s',
```
TomasVotruba: Change **f**iles

src/Application/FileProcessor.php
```diff
@@ -44,11 +44,14 @@ public function processFileWithRectorsToString(SplFileInfo $file, array $rectorC
         return $this->processFileToString($file);
     }
 
-    public function processFile(SplFileInfo $fileInfo): void
+    /**
+     * @return int|bool
+     */
+    public function processFile(SplFileInfo $fileInfo)
```
TomasVotruba: Here `bool` is again preferred

src/Console/Command/ProcessCommand.php
```diff
@@ -169,12 +179,15 @@ private function processFile(SplFileInfo $fileInfo, int &$i): void
...
-                $this->symfonyStyle->writeln($this->differAndFormatter->diffAndFormat($oldContent, $newContent));
+                $this->consoleStyle->writeln(sprintf('<options=bold>%d) %s</>', $i, $fileInfo->getPathname()));
+                $this->consoleStyle->writeln($this->differAndFormatter->diffAndFormat($oldContent, $newContent));
                 ++$i;
             }
         } else {
-            $this->fileProcessor->processFile($fileInfo);
+            $newContent = $this->fileProcessor->processFile($fileInfo);
+            if ($newContent !== false) {
+                $this->changedFiles[] = $fileInfo->getPathname();
```
TomasVotruba: Could use immidiate report as in the code above?

```php
$this->consoleStyle->writeln(sprintf('<options=bold>%d) %s</>', $i, $fileInfo->getPathname()));
```

It is preffereable to see any output right away than wait till the end to see all changed files. So people can see the Rector is doing some real work

It would also simplify rest of code

TomasVotruba:
I like the idea! :+1: 

I've added notes to change

src/Application/FileProcessor.php
carusogabriel: But as the `file_put_content` returns `int`, is correct to force bool?

src/Console/Output/ProcessCommandReporter.php
carusogabriel: We definitely need to find a Sniff for our closures :-1:

src/Printer/FormatPerservingPrinter.php
carusogabriel: As I said above: as the `file_put_contents` return `int`, should we force `bool`?

src/Console/Command/ProcessCommand.php
carusogabriel: That a good idea actually. In case `Rector` fails, it had already shown what have already changed :+1:

src/Console/Command/ProcessCommand.php
carusogabriel: Always forget that `sprintf` works like `C` printf :sweat_smile:

src/Console/Command/ProcessCommand.php
carusogabriel: As we are updating the progress bar together with the process of `processFile`, we'll need to figure out a way to do not conflict the two commands:
captura de tela de 2017-12-28 15-19-15

src/Console/Output/ProcessCommandReporter.php
carusogabriel: Hm, why?
As we are reporting the _changed_ files, in the past, we should use the verb in past. `x Change files` would sound like: You have X file of change, and not, You have X changed files :smile:

src/Application/FileProcessor.php
TomasVotruba: I think it is ok to retype, when you need to get answer:

-  `true` = file has changed
- `false` = file has not changed

carusogabriel: Ok, gonna use true and false

src/Console/Command/ProcessCommand.php
TomasVotruba: Ah, I had a scratch in the back of my head it could be the problem.

In that case it would be the best to do it like you suggest - collect first, than render second. Even for the code above.

Like PHPUnit - progress bar first, result second. Maybe that's the reason why it is done this way :sweat_smile:

TomasVotruba:
Ping me here please when this is ready to merge

carusogabriel:
@TomasVotruba 100% ready 💯

TomasVotruba:
Thanks, merged!

Could you please verify, if `--dry-run` works well with progress bar?

carusogabriel:
@TomasVotruba Sure. If I found any problem, I'll use the same logic as we used here: collect, them show :smile:

TomasVotruba:
:+1:

PR #250, opened 2017-12-28
Have only one rector file

carusogabriel:
Same as [`Laravel Artisan`](https://github.com/laravel/laravel/blob/master/artisan), we can have our `PHP` code inside a non `PHP` file.


bin/rector
```diff
@@ -1,5 +1,55 @@
 #!/usr/bin/env php
 <?php
```
TomasVotruba: Could you also `strict_types()` here?

TomasVotruba:
There were some issues for this, but I don't remember what it was :)

Like it!

bin/rector
carusogabriel: I got a message when trying to do it:
`Fatal error: strict_types declaration must be the very first statement in the script` :cold_sweat:

carusogabriel: Did it. Sorry, I forgot :sweat_smile:

TomasVotruba: That was probably the reason.
But it works here: https://github.com/phpstan/phpstan/blob/master/bin/phpstan

Have you managed to figure that out?
It might be related to include file

carusogabriel: Oh, if `PHPStan` has, we can have it too. Gonna search how to fix it :man_technologist:

carusogabriel: For some weird reason, it was related to my IDE, not to PHP itself. Gonna duplicate this to `Symplify/EasyCodingStandard`

TomasVotruba: Thanks for investigation!

PR #251, opened 2017-12-28
[Travis CI] Sudo false may make it faster

carusogabriel:
[Once I've heard](https://github.com/laravel/framework/pull/6084#issuecomment-67198487) that `sudo: false` inside `Travis CI` makes it faster. Let's see :shipit:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: php
 
 matrix:


TomasVotruba:
AFAIK it's actually `false` by default for repositories after 2016

See: https://docs.travis-ci.com/user/reference/overview/

carusogabriel:
Yes, [they are](https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments), good catch :+1:

PR #259, opened 2018-01-02
[ProcessCommand] Add one line between the progress bar and Exception trace

carusogabriel:
Add one line between the progress bar and Exception trace.
Before:
captura de tela de 2018-01-02 03-01-54
After:
captura de tela de 2018-01-02 03-03-09

### DIFF
--- a/src/Console/Command/ProcessCommand.php
+++ b/src/Console/Command/ProcessCommand.php
@@ -160,7 +160,7 @@ final class ProcessCommand extends Command
                 $this->processFile($fileInfo, $i);
             } catch (Throwable $throwable) {
                 throw new FileProcessingException(
-                    sprintf('Processing of "%s" file failed.', $fileInfo->getRealPath()),
+                    sprintf('Processing of "%s" file failed.', $fileInfo->getRelativePathname()),
                     $throwable->getCode(),
                     $throwable
                 );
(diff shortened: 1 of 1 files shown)


src/Console/Command/ProcessCommand.php
```diff
@@ -159,8 +159,9 @@ private function processFiles(array $fileInfos): void
             try {
                 $this->processFile($fileInfo, $i);
             } catch (Throwable $throwable) {
+                $this->consoleStyle->newLine();
```
TomasVotruba: :+1:

src/Console/Command/ProcessCommand.php
```diff
@@ -159,8 +159,9 @@ private function processFiles(array $fileInfos): void
...
                 throw new FileProcessingException(
-                    sprintf('Processing of "%s" file failed.', $fileInfo->getRealPath()),
+                    sprintf('Processing of "%s" file failed.', $fileInfo->getRelativePathname()),
```
TomasVotruba: This path should be relative to `getcwd()`, so that inludes the `tests` directory and is eaier to find.

carusogabriel: I tried to simplify the pathname to do not include `tests`, as we passed to the `process` argument:
```bash
rector process tests/ --level phpunit-specific-method
```

TomasVotruba: That is true, but user has to find file itselt and we don't want to force him to remember what was passed as argument and prepend it to each file.

It's common for CLI tools (PHPUnit, PHP CS Fixer, ECS, PHPStan) to report due to current working directory.

TomasVotruba:
Thank you!

PR #260, opened 2018-01-02
[ProcessCommand] Fix --dry-run option

carusogabriel:
As [foreseen](https://github.com/rectorphp/rector/pull/246#issuecomment-354471135) by @TomasVotruba in #246, the `--dry-run` option should also follow the process of collecting the changes, and them render.
Before:
captura de tela de 2018-01-02 04-03-55
After:
captura de tela de 2018-01-02 04-12-00

I've also added a title before the diffs: `X files with changes` :man_technologist:

### DIFF
--- a/src/Console/Command/ProcessCommand.php
+++ b/src/Console/Command/ProcessCommand.php
@@ -72,6 +72,11 @@ final class ProcessCommand extends Command
      */
     private $changedFiles = [];
 
+    /**
+     * @var string[][]
+     */
+    private $diffFiles = [];
+
     public function __construct(
         FileProcessor $fileProcessor,
         RectorCollector $rectorCollector,
@@ -124,6 +129,10 @@ final class ProcessCommand extends Command
 
         $this->processFiles($files);
 
+        if (count($this->diffFiles) > 0) {
+            $this->processCommandReporter->reportDiffFiles($this->diffFiles);
+        }
+
         if (count($this->changedFiles) > 0) {
             $this->processCommandReporter->reportChangedFiles($this->changedFiles);
         }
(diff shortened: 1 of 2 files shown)


src/Console/Command/ProcessCommand.php
```diff
@@ -179,9 +188,10 @@ private function processFile(SplFileInfo $fileInfo, int &$i): void
             $newContent = $this->fileProcessor->processFileToString($fileInfo);
 
             if ($newContent !== $oldContent) {
-                $this->consoleStyle->writeln(sprintf('<options=bold>%d) %s</>', $i, $fileInfo->getPathname()));
-                $this->consoleStyle->writeln($this->differAndFormatter->diffAndFormat($oldContent, $newContent));
-                ++$i;
+                $this->diffFiles[] = [
+                    'file' => sprintf('<options=bold>%d) %s</>', ++$i, $fileInfo->getRelativePathname()),
```
TomasVotruba: `++$i` used in the code take some thinking what the value is.
Could you seperate it before/after this line and use `$i` only?

src/Console/Output/ProcessCommandReporter.php
```diff
@@ -52,6 +52,23 @@ public function reportChangedFiles(array $changedFiles): void
...
+            '%d file%s with changes',
+            count($diffFiles),
+            count($diffFiles) === 1 ? '' : 's'
+        ));
+
+        foreach ($diffFiles as $diffFile) {
+            $this->consoleStyle->writeln($diffFile['file']);
+            $this->consoleStyle->writeln($diffFile['diff']);
+        }
+    }
```
TomasVotruba: Nice decoupling, thanks!

src/Console/Command/ProcessCommand.php
carusogabriel: I initialize it as `0` fell lines above. Should I revert as well?

TomasVotruba: On second though, this look quite ok. I checked on mobile and it was too many changes.

So :+1:  for the way it is now!

TomasVotruba:
Thank you!

I shall check the tests now :)

PR #261, opened 2018-01-02
Welcome 2018 :tada:

carusogabriel:


TomasVotruba:
Same as in Symplify PR. This change would add extra maintenance work with much less added value

PR #263, opened 2018-01-02
Bump to Symplify 3.1 and fix related code, make tests pass

TomasVotruba:


src/NodeChanger/IdentifierRenamer.php
```diff
@@ -33,10 +40,15 @@ public function renameNodeWithMap(Node $node, array $renameMethodMap): void
...
+        if (in_array(get_class($node), $this->nodeClassesWithIdentifier, true)) {
+            return;
         }
+
+        throw new NodeMissingIdentifierException(sprintf(
+            'Node "%s" does not contain a "$name" property with "%s". Pass only one of "%s".',
+            get_class($node),
+            Identifier::class,
+            implode('", "', $this->nodeClassesWithIdentifier)
+        ));
```
carusogabriel: :+1: For custom Exception

TomasVotruba: Thanks for inspiration!

carusogabriel: Why not invert the logic here?
```php
if (!in_array(get_class($node), $this->nodeClassesWithIdentifier, true)) {
     throw new NodeMissingIdentifierException(sprintf(
            'Node "%s" does not contain a "$name" property with "%s". Pass only one of "%s".',
            get_class($node),
            Identifier::class,
            implode('", "', $this->nodeClassesWithIdentifier)
))};
```
So we can avoid empty `return;`

carusogabriel: Why not invert the logic here? So we can avoid empty `return;`

TomasVotruba: It's called *early return statement*. The goal is to resolve the easy part first.

Than resolve more difficult part.

So the complexity of call is related to complexity of algorithm. From easiest to most difficult.

src/NodeVisitor/Collector/NodeCollector.php
```diff
@@ -1,47 +0,0 @@
...
-    }
-
-    /**
-     * @return Expression[]|Node[]
-     */
-    public function getNodesToRemove(): array
-    {
-        return $this->nodesToRemove;
-    }
-}
```
TomasVotruba: Found by https://www.tomasvotruba.cz/blog/2018/01/01/new-in-symplify-3-4-time-saving-coding-standard-checkers/#4-possible-unused-public-method

src/NodeChanger/IdentifierRenamer.php
TomasVotruba: Here it's explained better: https://github.com/jupeter/clean-code-php#avoid-nesting-too-deeply-and-return-early-part-1

carusogabriel: _Early return_, I have heard about them. But in this case, we have `in_array` check. Just denying it, by prefix with `!`, isn't better? Isn't less or more complexity, just denying :thinking:

TomasVotruba: Well, what if we need to add another check, which requires the `$name` property exist on the object?
(Which is the case now, because we call `$node->name' but rather implicit).

carusogabriel: Now I understood your point. Thanks for teach me :man_teacher:

PR #266, opened 2018-01-05
[docs] How it works added

TomasVotruba:


carusogabriel:
@TomasVotruba How is the cycle of life of a file in Rector? I mean, does it pass thru all Rectors? When I propose `AssertFalseStrposToContainsRector`, I imagined that the files passes by `AssertComparisonToSpecificMethodRector`, then `AssertTrueFalseToSpecificMethodRector`, after `AssertSameBoolNullToSpecificMethodRector` and so on. But sound like isn't like that. I've tested it with [this repo](https://github.com/Mastercard/sdk-core-php).

TomasVotruba:
@carusogabriel So you ask *In what order are those Rectors applied*?

If so, the order is now in natural order in config. First goes first.
So you can configure the order as you need, without any editation of Rector classes

TomasVotruba:
@carusogabriel I've added piece of code to explain better now

carusogabriel:
@TomasVotruba if you try to run `phpunit-specific-method` on https://github.com/Mastercard/sdk-core-php maybe we can understand what I asked :sweat_smile: 
We need to run Rector 3 times to get everything fixed.

TomasVotruba:
> We need to run Rector 3 times to get everything fixed.

That might happen, if the order of Recnd tors depends on each other and now are reversed.

E.g.

1st rector changes annotation to method: @expectedAnnotation => `expectedAnnotation()`
2nd rector changes method name `expectedAnnotation` to `expectAnnotation`

If you switch them, first the method will be changed, then new method with old name would be created.

So I'd try to play around with their order in `phpunit60.yml`

carusogabriel:
@TomasVotruba Now at the computer, let me explain better!
Let's use the following assertion:
```php
$this->assertTrue(strpos('foo', 'f') !== false);
```
What do we want in the end?
```php
$this->assertContains('f', 'foo');
```
In [the repo](https://github.com/Mastercard/sdk-core-php) I used as and example, if we run Rector for the first time, we have this:
```diff
-$this->assertTrue(strpos($headers['Authorization'][0], 'oauth_body_hash') !== false);
+$this->assertNotSame(false, strpos($headers['Authorization'][0], 'oauth_body_hash'));
```
And this is my complain: do we need to run Rector for a second time to get:
```diff
-$this->assertTrue(strpos($headers['Authorization'][0], 'oauth_body_hash') !== false);
+$this->assertContains('oauth_body_hash', $headers['Authorization'][0]);
```
Since our `config/levels/phpunit/phpunit-specific-method.yml` are configured:
```yml
Rector\Rector\Contrib\PHPUnit\SpecificMethod\AssertComparisonToSpecificMethodRector: ~
Rector\Rector\Contrib\PHPUnit\SpecificMethod\AssertTrueFalseToSpecificMethodRector: ~
Rector\Rector\Contrib\PHPUnit\SpecificMethod\AssertFalseStrposToContainsRector: ~
```
I tried to get deep into Rector to understand his cycle, but I couldn't :disappointed:

TomasVotruba:
**The order of rectors in cofig matters**. The simler first, more complex latter.

Try this or another order:

```yaml
Rector\Rector\Contrib\PHPUnit\SpecificMethod\AssertTrueFalseToSpecificMethodRector: ~
Rector\Rector\Contrib\PHPUnit\SpecificMethod\AssertFalseStrposToContainsRector: ~
Rector\Rector\Contrib\PHPUnit\SpecificMethod\AssertComparisonToSpecificMethodRector: ~
```

TomasVotruba:
It was the order :)

See https://github.com/rectorphp/rector/pull/269

docs/HowItWorks.md
```diff
@@ -0,0 +1,71 @@
...
+                $rector->reconstruct($node);
+            }
+        }
+    }
+}
+```
+
+### 2.1 Prepare Phase
+
+- File is parsed by [`nikic/php-parser`](https://github.com/nikic/PHP-Parser), 4.0-dex (this is important, because this version support writing modified tree back to file)
```
carusogabriel: Misspelling 4.0-dev

TomasVotruba: Got it!

PR #265, opened 2018-01-04
Code Cleanup, mostly AbstractRector and Travis speedup

TomasVotruba:
### Related Changes

- make use of all Travis builds for coverage performance improvement
- simplify  and decouple `ExpressionPrependingNodeVisitor` from `AbstractRector`
- drop `Rector\Rector\Contrib\Nette\Forms\FormSetRequiredRector`, since it's very weak and requires human-check; this change is out of scope of Rector
- move `PropertyAdder` to one place - `AbstratRector` => remove `ShutdownTraverser `

PR #274, opened 2018-01-07
Extended Autoloading, ClassAndMethodResolver, ProcessCommand cleanup

TomasVotruba:
Follow up to #265 

Integrating improvements after testing `src/config/level/phpunit/phpunit-specific-method.yml` on `symfony\symfony`
  

## Related Changes

- merge `MethoCallCallerTypes` and `StaticCallCallerType` to `NodeTypeResolver`
- get rid of `Attribute::CALLER_TYPES`, use `MethodCallTypeResolver` instead


bin/rector_bootstrap.php
```diff
@@ -1,14 +1,10 @@
 <?php declare(strict_types=1);
 
-/**
- * This allows to load "vendor/autoload.php" both from
- * "composer create-project ..." and "composer require" installation.
- */
 $possibleAutoloadPaths = [
+    // load from nearest vendor
+    getcwd() . '/vendor/autoload.php',
```
carusogabriel: `getcwd` may help us solve your issues with autoloader :pray:

src/Console/Output/ProcessCommandReporter.php
```diff
@@ -44,6 +44,10 @@ public function reportLoadedRectors(): void
      */
     public function reportChangedFiles(array $changedFiles): void
     {
+        if (count($changedFiles) <= 0) {
```
carusogabriel: This is perfect, didn't think about that. 
:+1: for early returns

TomasVotruba: Thanks! It's from DDD (Doman Driven Design), the responsibiltiy should be coupled to the classes that are responsible.

If sb would call this method with empty array (now knowing he or she should check for item count), it would run headline and empty list. Now it's safe and keeps responsbility from programmer in the code.

packages/NodeTypeResolver/src/PerNodeTypeResolver/MethodCallTypeResolver.php
```diff
@@ -48,76 +55,70 @@ public function getNodeClasses(): array
...
         }
 
-        $variableName = (string) $variableNode->name;
-
-        $methodCallVariableTypes = $this->typeContext->getTypesForVariable($variableName);
-
-        $methodCallName = $this->resolveMethodCallName($methodCallNode);
-
-        if (! $methodCallVariableTypes || ! $methodCallName) {
+        $methodName = (string) $methodCallNode->name;
```
carusogabriel: @TomasVotruba I guess we will have problems with this like #236

TomasVotruba: I see, we will. Could you send failing test to `master` for this?

carusogabriel: Sure, sending

TomasVotruba: Oh, I thought you're reporting some error. In that case it's allright

PR #279, opened 2018-01-09
[Fix] Object of class PhpParser\Node\{} could not be converted to string

carusogabriel:
[As predicted](https://github.com/rectorphp/rector/pull/274#discussion_r160067629) in #274

### DIFF
--- a/packages/NodeTypeResolver/src/PerNodeTypeResolver/MethodCallTypeResolver.php
+++ b/packages/NodeTypeResolver/src/PerNodeTypeResolver/MethodCallTypeResolver.php
@@ -70,11 +70,11 @@ final class MethodCallTypeResolver implements PerNodeTypeResolverInterface, Node
     {
         $parentCallerTypes = $this->resolveMethodCallVarTypes($methodCallNode);
 
-        if ($methodCallNode->name instanceof Identifier) {
-            $methodName = $methodCallNode->name->toString();
-        } else {
-            $methodName = (string) $methodCallNode->name;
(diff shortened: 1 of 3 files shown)


packages/NodeTypeResolver/src/PerNodeTypeResolver/MethodCallTypeResolver.php
```diff
@@ -68,7 +69,12 @@ public function getNodeClasses(): array
     public function resolve(Node $methodCallNode): array
     {
         $parentCallerTypes = $this->resolveMethodCallVarTypes($methodCallNode);
-        $methodName = (string) $methodCallNode->name;
+
+        if ($methodCallNode->name instanceof Identifier) {
+            $methodName = $methodCallNode->name->toString();
+        } else {
+            $methodName = (string) $methodCallNode->name;
```
TomasVotruba: I wonder, does it make sense to rename `__CLASS__`?

I think we should return empty array, if not `Identifier`:

```php
if (! $methodCallNode->name instanceof Identifier) {
    return [];
}

$methodName = $methodCallNode->name->toString();
```

carusogabriel: Yeah, makes sense. I'll change the logic :man_technologist:

TomasVotruba:
Thank you!

CS fails seems to be related to bug in new PHP CS Fixer 2.10, so merging

PR #281, opened 2018-01-10
[DocBlockAnalyzer] Fix tests

carusogabriel:
In DocBlockAnalyzer line 88 we `count` the return of `getTagsByName`. [In PHP 7.2](https://travis-ci.org/rectorphp/rector/jobs/326743440#L607), if we try to count `null`, an error arises: 
```
count(): Parameter must be an array or an object that implements Countable
```
So, instead of null, just return an empty array.


TomasVotruba:
Thanks shorter, better, faster, cleaner

PR #280, opened 2018-01-09
Add vendor-bin to $possibleAutoloadPaths

carusogabriel:
I was able to install Rector in my Composer fork with [the solution in `README.md`](https://github.com/rectorphp/rector#do-you-have-conflicts), but a Fatal Error arrises:
```
PHP Fatal error:  Uncaught Error: Class 'Symplify\PackageBuilder\Console\Style\SymfonyStyleFactory' not found in /var/www/composer/vendor-bin/rector/vendor/rector/rector/bin/rector:47
Stack trace:
#0 {main}
  thrown in /var/www/composer/vendor-bin/rector/vendor/rector/rector/bin/rector on line 47
```
After add `vendor-bin` as the first option in the `$possibleAutoloadPaths` array at `bin/rector_bootstrap.php`, Rector ran perfectly.

PR #288, opened 2018-01-14
Add vendor-bin to $possibleAutoloadPaths

carusogabriel:
This suppress #280, with a branch on Rector allowing local tests.

@TomasVotruba I'll test and see with position `vendor-bin` is better.


TomasVotruba:
Thank you!

PR #289, opened 2018-01-14
[README] Improve "Do you Have Conflicts" section

carusogabriel:


README.md
```diff
@@ -40,16 +40,21 @@ composer require --dev rector/rector 'dev-master' nikic/php-parser '4.0.x-dev'
 
 ### Do you Have Conflicts?
 
-It may also happen your dependencies are in conflict, e.g. PHPStan requires `nikic/php-parser 3.0`, or older PHPUnit etc. This might be solved in the future, when `nikic/php-parser` 4 becomes stable.
+It may also happen your dependencies are in conflict, e.g. PHPStan requires PHP-Parser version 3, or older PHPUnit etc. This might be solved in the future, when PHP-Parser version 4 becomes stable.
```
TomasVotruba: Why this change?

README.md
```diff
@@ -40,16 +40,21 @@ composer require --dev rector/rector 'dev-master' nikic/php-parser '4.0.x-dev'
...
 ```bash
 composer require bamarni/composer-bin-plugin --dev
-composer bin rector require rector/rector:dev-master --dev nikic/php-parser 4.0.x-dev
 ```
 
-And Rector is still accessible in:
+Then, require Rector using `composer bin`:
+
+```bash
+composer bin rector require --dev rector/rector 'dev-master' nikic/php-parser '4.0.x-dev'
```
TomasVotruba: It more clear what to do with `composer bin`, thanks

Could you separate package name and version with `:`? Now it looks like 4 packages

README.md
carusogabriel: I think to write `PHP-Parser` instead of `nikic/php-parser` is better, as we are explaining something, not require or so :smile:

README.md
carusogabriel: Sure, and will replicate in line 38 as well

README.md
TomasVotruba: As this is less popular package, it might be harder to find. Composer package name is fqn for it.

But if you insist, at least add link to the GitHub repository for it.

carusogabriel: Let's also link PHPStan? As isn't so popular as well

TomasVotruba: Yes, thanks

TomasVotruba:
:+1:

PR #292, opened 2018-01-14
better tests

mssimi:
I added one test and corrected other tests, since my first post was not correct.
Format `return $this->render('AppBundle:ClassWithNamedService1:index.html.twig');` is without breaking code of bundles, I guess conversion to bundle less apps should be different rector.

`return $this->render('@App/App/index.html.twig');`php is also correct maybe we should let user choose format.


TomasVotruba:
Just awesome, that's what I needed. Thank you for the PR to my PR!

PR #297, opened 2018-01-18
Fixes from PHPStan

carusogabriel:
@TomasVotruba What I send to you on Hangouts. I let you know when I finish, so you can review and merge!

### DIFF
--- a/src/Rector/Dynamic/ArgumentReplacerRector.php
+++ b/src/Rector/Dynamic/ArgumentReplacerRector.php
@@ -91,9 +91,9 @@ final class ArgumentReplacerRector extends AbstractRector
                 }
             } else {
                 // replace old value with new one
+                /** @var Arg $argumentOrParameter */
                 $argumentOrParameter = $argumentsOrParameters[$position];
 
-                /** @var Arg $argumentOrParameter */
                 $resolvedValue = $this->nodeValueResolver->resolve($argumentOrParameter->value);
 
                 if ($resolvedValue === $key) {
--- a/src/Rector/Dynamic/ArgumentReplacerRector.php
+++ b/src/Rector/Dynamic/ArgumentReplacerRector.php
@@ -29,9 +29,9 @@ final class ArgumentReplacerRector extends AbstractRector
     private $methodCallAnalyzer;
 
     /**
-     * @var mixed[][]|null
+     * @var mixed[][]
      */
-    private $activeArgumentChangesByPosition;
+    private $activeArgumentChangesByPosition = [];
 
     /**
      * @var ClassMethodAnalyzer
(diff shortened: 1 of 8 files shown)


src/Rector/Dynamic/ArgumentReplacerRector.php
```diff
@@ -108,12 +108,12 @@ public function refactor(Node $node): ?Node
     }
 
     /**
-     * @return mixed[][]|null
+     * @return mixed[][]
      */
-    private function matchArgumentChanges(Node $node): ?array
+    private function matchArgumentChanges(Node $node): array
```
TomasVotruba: Here it's different from other arrays that return 0 or more items.

It should be null, since it returns array with 1 item.

Null means miss match. Or maybe it should be false?

src/Rector/Dynamic/MethodNameReplacerRector.php
```diff
@@ -172,7 +172,7 @@ private function isMethodName(Node $node, array $types): bool
         $parentClassName = $node->getAttribute(Attribute::PARENT_CLASS_NAME);
 
         /** @var Identifier $node */
-        if (! isset($this->perClassOldToNewMethods[$parentClassName][$node->toString()])) {
+        if (! isset($this->perClassOldToNewMethods[$parentClassName][$node->name])) {
```
TomasVotruba: It weird this passed before. Was it covered?
We should cover it

src/Rector/Dynamic/ClassReplacerRector.php
```diff
@@ -56,7 +56,9 @@ private function resolveNewNameFromNode(Node $node): string
     {
         $nameNode = $this->resolveNameNodeFromNode($node);
 
-        return $this->oldToNewClasses[$nameNode->toString()];
+        if ($nameNode !== null) {
+            return $this->oldToNewClasses[$nameNode->toString()];
+        }
```
TomasVotruba: This should be probably covered by tests too. PHPStan doesn't fix that

src/Rector/Dynamic/ArgumentReplacerRector.php
carusogabriel: Returning `false` would complicated our return type, wouldn't?

src/Rector/Dynamic/ClassReplacerRector.php
carusogabriel: Yes, I thought about that too. Let's stay with PHPStan fixes in this PR, and expand coverage in the next PR?

src/Rector/Dynamic/ArgumentReplacerRector.php
TomasVotruba: It though it would, that's why I went with nullable here.

carusogabriel:
@TomasVotruba Ready to merge. I'll cover your request in the following PR!

TomasVotruba:
Thanks :+1:

PR #299, opened 2018-01-23
[Proposal] Remove ECS shipping with Rector

carusogabriel:
With this PR, I want to propose the removal of the ship of EasyCodingStandard with Rector, and here are some points:
- 1º (and the most important): ECS requires several Symfony dependencies, Squizlabs' PHP_codesniffer, Slevomat's coding-standard and FriendsOfPHP's PHP-CS-Fixer in their latest versions, with cause **_a lot_** of conflicts with old versions (and as the purpose of Rector is update legacy projects, we are going to face this conflicts for sure)!
- 2º: We just have it to encourage its use. Well, we already have it in `README.md`.

I know the intentions are the best, but we already have a lot of conflicts to solve :disappointed:


TomasVotruba:
I'm open to discussing here, thanks for feedback

It's needed to fix cleanups after Rector work.
Like import namespaces, remove empty docblocks etc.

I got feedback from community, that it's not user friendly to recommend running script that is missing and shipping by default is must have.

Moreover when you write about conflicts, we should know about them and how to solve them first. If this would be responsibility of the user, it could cause him trouble leading to not using this project.

What do you think?

carusogabriel:
> This package has no intention in formatting your code, as coding standard tools handle this much better.

So we need to elaborate more this phase, right? With Rector **does need** ECS to work, this should be explicit write. 

> If this would be responsibility of the user...

Yes, I agree with you, but, we must also agree that they might want Rector just for a work, or other, right?

> Moreover when you write about conflicts, we should know about them and how to solve them first.

I'm talking about version conflicts. Rector and ECS use the last of all packages versions, and they, in most of the cases, are requiring >=PHP 7, and legacy projects, in 99% are using <=PHP 5.6. Isn't conflicts that we can solve, like `PHP-Parser`, but only conflicts that we can avoid to have and provide a better user experience.

TomasVotruba:
> So we need to elaborate more this phase, right? With Rector does need ECS to work, this should be explicit write.

Yes, exactly.

> 
Yes, I agree with you, but, we must also agree that they might want Rector just for a work, or other, right?

In both cases, user should not be bothered by this knowledge. It should be ready to go right away after install.

> ...only conflicts that we can avoid to have and provide a better user experience.

For these reasons package removal is not a solution. This one is: https://github.com/rectorphp/rector/blob/master/README.md#do-you-have-conflicts

If it doesn't work for some reason, it should be fixed there.

Soullivaneuh:
I don't see why Rector does need ECS here. ECS is a coding standard tool AFAIK, why Rectore should needs it?

TomasVotruba:
That's best question! `php-parser` add and changes code without any coding standard rules. So it creates properties with no spaces in between them, fully qualified namespaces etc.

See these examples:

-
 [correct](https://github.com/rectorphp/rector/blob/master/tests/Rector/Contrib/Symfony/HttpKernel/GetterToPropertyRector/Correct/correct2.php.i-nc)
- [wrong](https://github.com/rectorphp/rector/blob/master/tests/Rector/Contrib/Symfony/HttpKernel/GetterToPropertyRector/Wrong/wrong2.php.inc)

It's job of coding standard making this right

- import namespace
- remove unused namespaces
- add spaces between elements
- remove redundant spaces from empty doc blocks 

etc.

carusogabriel:
Closing this as @TomasVotruba is right:

> `php-parser` add and changes code without any coding standard rules. So it creates properties with no spaces in between them, fully qualified namespaces etc.

TomasVotruba:
Should we improve Readme?

carusogabriel:
@TomasVotruba I'm finishing a new Rector, I'll propose some improvements for discussing

TomasVotruba:
Awesome, thanks

PR #300, opened 2018-01-23
[PHPUnit] Create AbstractPHPUnitRector

carusogabriel:
This PR solves some problems with PHPUnit Rectors:
- Some assertions weren't being refactored if they extended a `BaseTestCase` class, not PHPUnit's;
- Unify how we candidate our classes to look up for assertions: end with `Test`, as it [is the standard in PHPUnit](https://phpunit.de/getting-started-with-phpunit.html). In the feature, we can remove this check, looking up just for the assertions, but only when improving the speed of Rector.

After that, a bug arises calling `toString` in FuncCall that are actually variables: `$method($argument)`, also fixed with this PR.

### DIFF
--- a/src/Rector/AbstractPHPUnitRector.php
+++ b/src/Rector/AbstractPHPUnitRector.php
@@ -1,0 +1,17 @@
+<?php declare(strict_types=1);
+
+namespace Rector\Rector;
+
+use Nette\Utils\Strings;
+use PhpParser\Node;
+use Rector\Node\Attribute;
+
+abstract class AbstractPHPUnitRector extends AbstractRector
+{
+    protected function isInTestClass(Node $node): bool
+    {
+        $className = $node->getAttribute(Attribute::CLASS_NAME);
+
+        return Strings::endsWith((string) $className, 'Test');
+    }
+}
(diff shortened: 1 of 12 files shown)


src/Rector/AbstractPHPUnitRector.php
```diff
@@ -0,0 +1,17 @@
...
+use PhpParser\Node;
+use Rector\Node\Attribute;
+
+abstract class AbstractPHPUnitRector extends AbstractRector
+{
+    protected function isInTestClass(Node $node): bool
+    {
+        $className = $node->getAttribute(Attribute::CLASS_NAME);
+
+        return Strings::endsWith((string) $className, 'Test');
```
TomasVotruba: Could you add comment there, that this is only approximate assumption?
It might be important in some edge cases, where these Rectorswould fail, so it's easier to debug.

TomasVotruba:
Like it overall!
 
Have you tested that on real project? Does it pass?

src/Rector/AbstractPHPUnitRector.php
TomasVotruba: Also the note that from PR would be great: https://phpunit.de/getting-started-with-phpunit.html

carusogabriel:
@TomasVotruba `Composer` is a proof of concept of this PR. Run with master branch, and them with this fork, and you'll see the results!

TomasVotruba:
Awesome, thanks :+1:

PR #305, opened 2018-02-02
PHPUnit 7.0

carusogabriel:


carusogabriel:
Waiting for https://github.com/GeckoPackages/GeckoDiffOutputBuilder/pull/6

SpacePossum:
Hi @carusogabriel and @TomasVotruba :)

IIUC you're updating to PHPUnit 7 here and need `gecko`'s `diff` to follow, however:
`PHPUnit7` ships with [`diff/3.0`](https://github.com/sebastianbergmann/phpunit/blob/7.0.0/composer.json#L40) , this is important because `diff/3.0` ships with a new outputbuilder: [StrictUnifiedDiffOutputBuilder](https://github.com/sebastianbergmann/diff/blob/3.0.0/src/Output/StrictUnifiedDiffOutputBuilder.php) which is the same as gecko's.
I've moved the `gecko` one to `sebastianbergmann/diff` for a bunch of reasons, but most important to know is that now the `gecko` one has become redundant the whole package will be removed (after a period of time, giving people time to update to the `sebastianbergman/diff` version).

Let me know if you want to know more about the (namespace) changes.

carusogabriel:
@SpacePossum Thanks for that information, I really appreciated :+1: 

@TomasVotruba I try to work on that :construction:

packages/ConsoleDiffer/src/config/services.yml
```diff
@@ -5,12 +5,12 @@ services:
     Rector\ConsoleDiffer\:
         resource: ..
 
-    GeckoPackages\DiffOutputBuilder\UnifiedDiffOutputBuilder:
+    SebastianBergmann\Diff\Output\UnifiedDiffOutputBuilder:
```
carusogabriel: @TomasVotruba Can I ask your help here?

TomasVotruba:
@carusogabriel Wow, is that soo simple?

Could you please share screens for diff before/after this change?

packages/ConsoleDiffer/src/config/services.yml
```diff
@@ -5,12 +5,8 @@ services:
...
-        arguments:
-            $options:
-                'fromFile': 'Original'
-                'toFile': 'New'
+    SebastianBergmann\Diff\Output\UnifiedDiffOutputBuilder:
 
     SebastianBergmann\Diff\Differ:
         arguments:
-            - '@GeckoPackages\DiffOutputBuilder\UnifiedDiffOutputBuilder'
+            - '@SebastianBergmann\Diff\Output\UnifiedDiffOutputBuilder'
```
SpacePossum: I think you want;
`'@SebastianBergmann\Diff\Output\StrictUnifiedDiffOutputBuilder'`

and keep the arguments you had for `GeckoPackages\DiffOutputBuilder\UnifiedDiffOutputBuilder` and use those for `SebastianBergmann\Diff\Output\StrictUnifiedDiffOutputBuilder`

TomasVotruba: Thanks for help!

> I think you want;
'@SebastianBergmann\Diff\Output\StrictUnifiedDiffOutputBuilder'

What are difference between those 2?

SpacePossum: There are two differences between the builders:
- the `strict` one will give you more control and options (see https://github.com/sebastianbergmann/diff for all configurations options)
- `StrictUnifiedDiffOutputBuilder` will generate strict `udiff` format output that is compat for the `patch` and `git apply` commands (outputs like `diff -u` command). The non-strict `UnifiedDiffOutputBuilder` has some exceptions on the output (like additional warnings about line end differences) making it more suitable for PHPUnit.

TomasVotruba: I see, thanks!

- https://github.com/sebastianbergmann/diff#strictunifieddiffoutputbuilder

carusogabriel: Actually, with my current change, everything is working :sweat_smile:
captura de tela de 2018-02-05 13-54-05

TomasVotruba: Could you also add the previous form? There should be numbers

SpacePossum: yeah, you lost the "precious" line numbers ;)

carusogabriel: Oh, small missing, big importance.

I'll try one more time before asking @TomasVotruba help :fire:

TomasVotruba: Here is solution: https://github.com/rectorphp/rector/pull/305#discussion_r166007142

carusogabriel: @TomasVotruba Fell free to finish this, I'm not handling this some some reason :cry:

TomasVotruba:
## Result

image

TomasVotruba:
Thank you @carusogabriel  and @SpacePossum

TomasVotruba:
@carusogabriel I think the same approach can be applied to https://github.com/Symplify/Symplify, right?

TomasVotruba:
@SpacePossum Great job with getting this OutputDiffer to PHPUnit! :clap:

carusogabriel:
@TomasVotruba Maybe, I'll try this there!

PR #308, opened 2018-02-03
Redirect to route rector

mssimi:
Hey added rector simplifying redirection to route. Helper method redirectToRoute was introduced in SF 2.6.

```php
/**
 * Before:
 * return $this->redirect($this->generateUrl("homepage"));
 * After:
 * return $this->redirectToRoute("homepage");
 */
```


TomasVotruba:
Very nice PR, thank you

PR #313, opened 2018-02-06
Fix some misspellings

carusogabriel:
Found with https://github.com/vlajos/misspell_fixer


TomasVotruba:
Awesome :+1:  :+1: 

We should make Travis CI check from this.

PR #315, opened 2018-02-08
[PHPUnit] Scenario Annotation replacer Rector

### LINKED ISSUE #306: PHPUnit 7 level
PHPUnit 7 [was released](https://phpunit.de/announcements/phpunit-7.html), and all the `@scenario` annotations should be renamed to `@test`.

@TomasVotruba Is there a Rector for this already, or we should create/modify one?

TomasVotruba:
Closes #306


src/Rector/Contrib/PHPUnit/ScenarioToTestAnnotationRector.php
```diff
@@ -0,0 +1,51 @@
...
+use Rector\ReflectionDocBlock\NodeAnalyzer\DocBlockAnalyzer;
+
+/**
+ * Before:
+ * - @scenario
+ *
+ * After:
+ * - @test
+ */
+final class ScenarioToTestAnnotationRector extends AbstractPHPUnitRector
```
carusogabriel: A whole Rector for changing an annotation? Can't we create a Dynamic one, like `ClassReplacerRector`? :thinking:

TomasVotruba: I think we can.

How would you differ `@scenario` in Codeception and `@scenario` in PHPUnit?

carusogabriel: We can receive a class, and look at it parents, and the annotation name, like:
```yml
PHPUnit\Framework\TestCase:
   - 'test': 'scenario'
```

TomasVotruba: That could work!
And how would you limit annotations to nodes? To `ClassMethod` and `Property` nodes or all of them?

I'll try to handle this later

TomasVotruba:
I've pushed new proposal

TomasVotruba:
Ready to review

README.md
```diff
@@ -259,29 +259,53 @@ rectors:
...
 private $name;
 ```
 
 ```diff
-- public function someMethod(ValueObjects\Name $name) { ...
+- public function someMethod(ValueObject\Name $name) { ...
 + public function someMethod(string $name) { ...
 ```
 
+## Replace Property and Methods Annotations
```
carusogabriel: Plural or singular should be consistent here 👍🏻

Issue #177, opened 2017-11-29
Create PHAR with prefixes to avoid class and package version conflicts


soullivaneuh:
Is a `.phar` build planned for this tool? I would be interested. :-)

TomasVotruba:
Why not :)

I like this approach: https://github.com/phpstan/phpstan-shim

Soullivaneuh:
Well, as I can have a link for a `wget`, I'm ok. :stuck_out_tongue:

TomasVotruba:
Let's do this then :) could you prepare such repository and test it?
We can move the repo here afterwards and generate phar to GitHub pagest

Soullivaneuh:
Not sure to get time to learn how to build a .phar now. But if nobody does it, I'll do it one time.

About the .phar story. What is the interest of a `-shim` repository? Why not publish the phar on a website or on the release file instead like php-cs-fixer?

TomasVotruba:
I think it's `composer update` approach. Install and forget.

I never update my phar deps

TomasVotruba:
Feel free to re-open with PR or reference to demo repository.
Closing now to keep only valid issues open only

TomasVotruba:
Reoped due to: https://github.com/rectorphp/rector/issues/302#issuecomment-361648078

TomasVotruba:
### Resources

- https://github.com/box-project/box2

TomasVotruba:
I'm making this priority, since the bin plugin doesn't work as expected

inverse:
Once we have phar support could we add this to phive?

TomasVotruba:
Useful source

- https://andreas.heigl.org/2017/01/18/building-a-phar-automated/

TomasVotruba:
As discusses, think of prefixing only:

- https://github.com/humbug/php-scoper

mssimi:
Tried php-scoper atm I cannot whitelist namespace easily.

https://github.com/humbug/php-scoper/issues/192

TomasVotruba:
Consider using it, as it cover 80 % cases.

The rest need to be resolve our way:

- whitelist can be run done with string replace via cli (Rector => Prefix123Rector by php-scoper, back by [`sed` in cli](https://askubuntu.com/a/20416/325651))
- config files prefixing - simple script should handle that

TomasVotruba:
[...]

PR #319, opened 2018-02-14
[Help wanted] Add PHAR

TomasVotruba:
Closes https://github.com/rectorphp/rector/issues/177

### On the Menu

- [ ] prepare composer script to install and make phar
- [ ] prepare own phar build script
- [ ] make local `rector.phar` run work

- box and php-scoper shown as projects WIP not usable in state to this day

### Related Sources

- https://github.com/fprochazka/phpstan-compiler
- https://github.com/composer/composer/blob/master/src/Composer/Compiler.php
- https://github.com/ApiGen/ApiGen/pull/788 (revert of PHAR)
- http://php.net/manual/en/phar.construct.php
- https://stackoverflow.com/questions/15750913/generating-a-phar-for-a-simple-application
- https://github.com/koto/phar-util/blob/master/phar-build.php


bin/rector
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env php
-<?php declare(strict_types=1);
+<?php
```
carusogabriel: Why? :cry:

TomasVotruba: Unfortunatelly it's needed for PHAR to make work.

See PHPUnit:

- https://github.com/sebastianbergmann/phpunit/issues/2791#issuecomment-334953382
- https://github.com/sebastianbergmann/phpunit/commit/70ccf0370f0b99c22819d3730167a7d448911b9e

composer.json
```diff
@@ -24,23 +24,24 @@
         "webmozart/assert": "^1.2"
     },
     "require-dev": {
+        "humbug/box": "^3.0@dev",
```
theofidry: If you are using `bamarni/composer-bin-plugin` I would recommend you to install `humbug/box` with it as the later doesn't need to execute any of your code

composer.json
```diff
@@ -67,7 +68,12 @@
             "vendor/bin/ecs check bin packages src tests --fix",
             "bin/clean_levels.sh"
         ],
-        "phpstan": "vendor/bin/phpstan.phar analyse packages src tests --level max --configuration phpstan.neon"
+        "phpstan": "vendor/bin/phpstan.phar analyse packages src tests --level max --configuration phpstan.neon",
+        "build-phar": [
+            "vendor/bin/php-scoper add-prefix --output-dir=build --force",
```
theofidry: I would remove the dev dependencies first there

scoper.inc.php
```diff
@@ -0,0 +1,57 @@
...
+    // PHP-Scoper's goal is to make sure that all code for a project lies in a distinct PHP namespace. However, you
+    // may want to share a common API between the bundled code of your PHAR and the consumer code. For example if
+    // you have a PHPUnit PHAR with isolated code, you still want the PHAR to be able to understand the
+    // PHPUnit\Framework\TestCase class.
+    //
+    // A way to achieve this is by specifying a list of classes to not prefix with the following configuration key. Note
+    // that this does not work with functions or constants neither with classes belonging to the global namespace.
+    //
+    // Fore more see https://github.com/humbug/php-scoper#whitelist
+    'whitelist' => [
```
theofidry: I think you can remove that one

box.json
```diff
@@ -0,0 +1,16 @@
+{
```
theofidry: You can optimise it later but I would recommend to enable compression (`"compression": "GZ"`)

TomasVotruba:
So first step - building phar file - is done.

Now we need to test it, that it works :)

composer.json
TomasVotruba: How would you automate that to don't bother user with that?
I've added it to `composer.json` script, but it takes too much perfromance.

box.json
TomasVotruba: That failed for me before, but now works. Thanks :+1:

TomasVotruba:
@theofidry Thanks for the feedback, much appreciated from the package's maintainer/author.

Do you have some simple idea how to test `rector.phar`? What is the best practise to do so in the wild?

composer.json
```diff
@@ -67,7 +68,13 @@
             "vendor/bin/ecs check bin packages src tests --fix",
             "bin/clean_levels.sh"
         ],
-        "phpstan": "vendor/bin/phpstan.phar analyse packages src tests --level max --configuration phpstan.neon"
+        "phpstan": "vendor/bin/phpstan.phar analyse packages src tests --level max --configuration phpstan.neon",
+        "build-phar": [
+            "composer bin php-scoper require --dev humbug/php-scoper",
```
TomasVotruba: @theofidry Here is the bin install used

box.json
theofidry: IIRC this can fail if the PHAR is too big (with the current implementation) but this should be ok unless you have over 10K files

composer.json
theofidry: you could do `composer bin {box|php-scoper} install` in your composer script. Even if unnecessary once already installed it's like not even half a second. Alternatively I'm using Makefiles lately for this kind of stuff. You can see some examples on alice, AliceDataFixture, php-scoper or box: just do `make test` or `make build` and everything is taken care of

theofidry:
You're welcome.

My recommendation for testing is to have a couple of e2e tests, i.e. tests where you run your application with the binary (be it from the source `bin/rector` or the PHAR) as a black box and test the output. You can then easily switch between the binary of the PHAR when you need to debug.

The way I did it with PHP-Scoper is relatively simple: I have a few directories where you have the classes/composer.json and stuff and a file containing the expected output. I run PHP-Scoper and Box to build a PHAR and test if the output is the same as the expected one. You can find the examples [there](https://github.com/humbug/php-scoper/blob/master/Makefile#L61).

For box the e2e testing is a bit lighter, I have a few integration tests to make sure Box works and for testing the PHAR I simply try to build the Box PHAR with the PHAR. In other words I generate the PHAR once (with the regular binary `bin/box`) and a second time with the generated PHAR. See [this](https://github.com/humbug/box/blob/master/Makefile#L59) for more details

composer.json
TomasVotruba: That would still bother the user.

Maybe add bash script, that would check if vendor-bin directory for package exist, and install it only if not. What do you think?

box.json
TomasVotruba: There was missing argument in constructor error, but I can reproduce it

composer.json
theofidry: Yeah you can definitely have a `compile` bash script or something instead.

Remember most of your users/contributors won't have to build the PHAR so I don't think it's gonna bother them much. But otherwise a bash/Makefile gives more flexibility on how to do things

TomasVotruba: I'll look on it.

> Remember most of your users/contributors won't have to build the PHAR so I don't think it's gonna bother them much. But otherwise a bash/Makefile gives more flexibility on how to do things

Speed matter for us as well.

theofidry: Yeah but I think the time taken by composer to dump the autoloader (which is the only thing it will do if the packages are already installed) is negligible compared to the rest.

IMO the places of this process where perf. can be improved:

- The scoping itself, i.e. the core of PHP-Scoper
- PHP-Scoper as standalone as we could leverage parallelism for example. But I don't think I'll invest much time on it tbh. I'm more interesting in integrating with Box which case the parallelisation already happens there
- Box itself
- PHP-Parser

Any optimisation in any of those above is more likely to have a higher impact that the half second spent on dumping the autoloader.

That said if it's bugging you, there's nothing wrong with using bash/Makefile to handle this

composer.json
TomasVotruba: I did that before but:

- it took so much time to remove, install dev, then install all
- it's covered by finder already
- it's not possible due to php-cs-fixer, that has `/src` file in `/tests` dir; which fails on `composer dump`

TomasVotruba:
@theofidry Thank you, that seems like great simple trick! I'll investigave those sources.

composer.json
theofidry: The Finder doesn't cover that. You are ignoring a library test files with the current config but not the whole dev library.

Yeah PHP-CS-Fixer is a bit annoying here cf. https://github.com/api-platform/schema-generator/blob/master/scoper.inc.php#L39-L42 and https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3462

Although you are right you can also ignore this. This is something that will be handled in Box at some point which will be way more performant for PHP-Scoper as well as it will only scope the files filtered by Box

tests/Phar/PharTest.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Rector\Tests\Phar;
+
+use PHPUnit\Framework\TestCase;
+
+final class PharTest extends TestCase
+{
+    public function test(): void
+    {
+        $this->assertTrue(file_exists(__DIR__ . '/../../rector.phar'));
```
carusogabriel: Maybe `assertFileExists`?

TomasVotruba: This is WIP

carusogabriel: I'll follow your discussion with @theofidry and see what I can help :smile:

TomasVotruba:
@theofidry I've added detailed steps with prefixing and phar building to Travis to show the error.

I have no idea how to fix that or what it means, being stuck for hours.

Could you give me some hints? It would save me 

https://travis-ci.org/rectorphp/rector/builds/342723839#L817

theofidry:
I'll have a look in a bit

theofidry:
Actually maybe a bit later. I need to fix a couple of things in php-scoper and box but that won't take too long. I'll have a look at Rector today still

TomasVotruba:
@theofidry No problem, thank you for any help you find time for

TomasVotruba:
My last finding...


1. I run `php-scoper` with output to `build

2. Then I run `composer dump`

```php
composer dump --working-dir=build
```

But no classes from `build/vendor` is loaded.

theofidry:
I'm checking that now :)

TomasVotruba:
Is it possible, that `composer dump` loads only packages from main `composer.json` and skips all those from `/vendor`?

theofidry:
No that part definitely works. I think I got a lead on the issue it's just frigging slow (the scoping takes 1-2min each time...)

theofidry:
Ok I found too bugs I'll need to fix in PHP-Scoper:

- When collecting the files, I retrieve the finders and merge them in a `AppendIterator()`. First I should ditch it for `nikic/inter::chain()`. Second there's an issue when appending files. Indeed if you use `Finder::append([])` twice, for both the key will be a numeric key so when merging the two finders one element will be lost.
- Second is more serious but it's good to have found it. When scoping a file, before prefixing anything I resolve the names/symbols into their fully-qualified form whenever possible. For example for:

```php
use Foo\Bar;

$x = new Bar();
```

The name resolver will resolve the `'Bar'` for the statement into `'\Foo\Bar'`. Then it's more stuff is done to know if the name should be prefixed or not.

The bug here is that the name resolver doesn't take into account the types. It looks at the namespace and use statement, but without any further check, so if you have:

```php
use Foo\Bar();

bar(); // Call the function not the class
```

It will work in PHP (provided `bar` is registered properly), but the name resolver will assume the use statement above is for it so will resolve `bar` into `\Foo\bar` which is a mistake since it's a class use statement not a function use statement.

TomasVotruba:
Thanks for getting into this.

I found another case:

### Error

```
PHP Fatal error:  Uncaught LogicException: Passed array does not specify an existing static method (class 'ComposerAutoloaderInit77d7c9b76b143be69b30c174e4a33892' not found) in /var/www/Rector/build/vendor/composer/autoload_real.php:20
```

### Origin


```php
<?php

// autoload_real.php @generated by Composer

class ComposerAutoloaderInit77d7c9b76b143be69b30c174e4a33892
{
    private static $loader;

    public static function loadClassLoader($class)
    {
        if ('Composer\Autoload\ClassLoader' === $class) {
            require __DIR__ . '/ClassLoader.php';
        }
    }

    public static function getLoader()
    {
        if (null !== self::$loader) {
            return self::$loader;
        }

        spl_autoload_register(array('ComposerAutoloaderInit77d7c9b76b143be69b30c174e4a33892', 'loadClassLoader'), true, true);
```

### Prefixed

```php
namespace PhpScoper5a885c7e13dc5;

// autoload_real.php @generated by Composer
class ComposerAutoloaderInit77d7c9b76b143be69b30c174e4a33892
{
    private static $loader;
    public static function loadClassLoader($class)
    {
        if ('Composer\\Autoload\\ClassLoader' === $class) {
            require __DIR__ . '/ClassLoader.php';
        }
    }
    public static function getLoader()
    {
        if (\null !== self::$loader) {
            return self::$loader;
        }
        \spl_autoload_register(array('ComposerAutoloaderInit77d7c9b76b143be69b30c174e4a33892', 'loadClassLoader'), \true, \true);
```

The latter calls `ComposerAutoloaderInit77d7c9b76b143be69b30c174e4a33892`, `loadClassLoader`, and not the namespaced version.

tests/Phar/PharTest.php
TomasVotruba: Done :+1:

TomasVotruba:
I see, I guess this would be fixed by the other 2 issues you mentioned.

theofidry:
Did some PRs so it should work better now. Also the patcher I registered shouldn't be needed anymore eiehter

TomasVotruba:
Thanks for the work! 
When I rerun [the script](https://github.com/rectorphp/rector/pull/319/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R71)  It fails on PHP error for me:

```php
PHP Fatal error:  Namespace declaration statement has to be the very first statement or after any declare call in the script in phar:///var/www/Rector/rector.phar/bin/rector on line 3
```

Related code in Phar `bin/rector` looks like this:

```php
#!/usr/bin/env php
<?php # ← here is extra space, that might cause that, at least what I saw in Stackoverflow answer
namespace PhpScoper5a89bf1332dcc;
```

theofidry:
That’s weird, it’s not what I’m getting on my machine at all. Do you try my
bin/compile?

TomasVotruba:
I run the script I linked. What should I run instead?

theofidry:
No that’s what I did as well. Very strange

theofidry:
I actually had this issue with PHPUnit when trying to run it from phpStorm, removing the shebang line made it work for phpStorm but then I needed to do `php vendor/bin/phpunit`. I dunno what can cause this but I doubt it's the transformation itself. You can verify by removing the extra space (also check that this is an actual space not some weird unicode character).

TomasVotruba:
> You can verify by removing the extra space

How to do that in PHAR file? I didn't find a way.

theofidry:
Does that happens only when it's in the PHAR? You could do that by doing something like that with [PsySH](psysh.org):

```php
$x = new Phar('rector.phar');

$path = 'bin/rector';
$newContents = implode("\n", array_map('trim', explode("\n", file_get_contents($x[$path])));

$x->addFileFromString($path, $newContents);
```

Alternatively you can test directory with the scoped files. You just need to adjust the permissions of `build/rector/bin/rector`, modify it and try to execute it

TomasVotruba:
When I run the script without shebang, I'm back to autoloading error:

```php
PHP Fatal error:  Uncaught Error: Class 'PhpScoper5a8af4c82ef05\Symplify\PackageBuilder\Configuration\LevelConfigShortcutFinder' not found in phar:///var/www/Rector/rector.phar/bin/rector:19
```

Could I maybe send similar test case to your project to help you find the spot?

theofidry:
I was getting an error along those lines. So now why is the shebang causing this error I have no idea... It looks like a _really_ obscure bug.

As for your error, yes please :) PHP-Scoper is still quite young so I do except all kind of weird cases to fix. But the only what to get it better is try it and spot the issues :P

TomasVotruba:
@theofidry Coudl you navigate me there?

theofidry:
Sure. The way I go about those kind of issues usually is by trying to play directly with the scoped content (there could be an issue with the PHAR as well but it's easier to make sure the scoped content is working fine first).

So I try to make it work with the scoped content: check the source to see where the issues is and see how it should be fixed: if it's just a config thing missing or an actual bug.

If I identified the issue as a bug in the scoper, then you have tests that can be written in [specs](https://github.com/humbug/php-scoper/tree/master/specs). You have two format for the tests inside, [the short syntax](https://github.com/humbug/php-scoper/blob/master/specs/misc.php#L23) and [the long one](https://github.com/humbug/php-scoper/blob/master/specs/misc.php#L45). So if you have a bug, step one is to fill it, but if I can have the failing test case that's even better (once you created the file just type `make test` and the rest is taken care of)

bin/rector
TomasVotruba: I found a way to revert this :+1:

packages/PharBuilder/src/PharBuilder.php
```diff
@@ -0,0 +1,29 @@
...
+
+    public function build(string $buildDirectory): void
+    {
+        $this->init();
+        $this->compiler->compile($buildDirectory);
+    }
+
+    private function init(): void
+    {
+        ini_set('phar.readonly', '0');
```
theofidry: I don't think this is working unless `phar.readonly` is actually already enabled

TomasVotruba: I have no idea, just used it from some other PHAR build tool. Could you confirm? I can remove it then

phpunit.dist.xml
```diff
@@ -0,0 +1,42 @@
+<phpunit
```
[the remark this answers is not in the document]
TomasVotruba: Nice, I though that's automated in some way like `.json` validation is.

Thank for the tip

packages/PharBuilder/src/config/config.yml
```diff
@@ -0,0 +1,26 @@
...
+    Rector\PharBuilder\:
+        resource: '../'
+        public: true
+
+    # Symfony\Console
+    Symfony\Component\Console\Style\SymfonyStyle:
+        factory: ['Symplify\PackageBuilder\Console\Style\SymfonyStyleFactory', 'create']
+
+    Symfony\Component\Console\Input\ArgvInput: ~
+    Symfony\Component\Console\Input\InputInterface:
```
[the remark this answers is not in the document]
TomasVotruba: What is more clear to most people?

docs/BuildingRectorPhar.md
```diff
@@ -0,0 +1,55 @@
...
+
+To build `rector.phar` just run:
+
+```bash
+composer build phar
+```
+
+That includes 3 steps:
+
+## 1. Preffix Classes to Remove Conflicts
```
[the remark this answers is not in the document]
TomasVotruba: I'll add WIP and update after bare `rector.phar` works

packages/PharBuilder/src/config/config.yml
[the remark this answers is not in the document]
theofidry: No idea. I find the `@` notation clearer but I can only speak for myself

packages/PharBuilder/src/PharBuilder.php
theofidry: I can confirm:

```
$ psysh
>>> ini_get('phar.readonly')
=> "1"
>>> ini_set('phar.readonly', 0)
=> false
```

And [the doc](https://secure.php.net/manual/en/phar.configuration.php#ini.phar.readonly) confirms it as well:

>This setting can only be unset in php.ini due to security reasons. If phar.readonly is disabled in php.ini, the user may enable phar.readonly in a script or disable it later. If phar.readonly is enabled in php.ini, a script may harmlessly "re-enable" the INI variable, but may not disable it.

packages/PharBuilder/src/config/config.yml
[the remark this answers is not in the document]
TomasVotruba: I ask before it's first time I see it and it looks like an argument to me.

`alias` is clear and recommended solution. Though yours is shorter, it look unclear, so I'll stick with `alias`

packages/PharBuilder/src/PharBuilder.php
TomasVotruba: Thanks!

Let me know if you handle that in some way

composer.json
```diff
@@ -20,21 +20,17 @@
         "symfony/console": "^4.0",
         "symfony/dependency-injection": "^4.0",
         "symfony/finder": "^4.0",
-        "symplify/easy-coding-standard": "^3.2",
-        "symplify/package-builder": "^3.2",
-        "webmozart/assert": "^1.3"
+        "symplify/package-builder": "^3.2"
     },
     "require-dev": {
+        "symplify/easy-coding-standard": "^3.2",
```
TomasVotruba: @carusogabriel This might make you happy :) 
After bit of thinking this would be too much to add to phar at the moment, since ECS depends on 2 other big packages

composer.json
```diff
@@ -20,21 +20,17 @@
...
         "latte/latte": "^2.4",
         "nette/application": "^2.4",
         "nette/di": "^2.4",
         "nette/forms": "^2.4",
         "phpstan/phpstan-shim": "^0.9",
         "phpunit/phpunit": "^7.0",
         "slam/php-cs-fixer-extensions": "^1.13",
-        "symfony/expression-language": "^4.0",
-        "symfony/form": "^4.0",
-        "symplify/coding-standard": "^3.2",
```
TomasVotruba: Already covered in ECS

carusogabriel:
@TomasVotruba Let me know when you finish. I guess will be easier to review and understand :smile:

TomasVotruba:
I don't have time to continue on this, as I focus on testing Rectors on Sylius and CakePHP.

Anyone feel free to take over or start from scratch.

mssimi:
Hey @TomasVotruba , what is missing to make it work? Prefixing?

TomasVotruba:
Hey, well, build doesn't work yet neither.

mssimi:
Building phar itself works for me.

TomasVotruba:
@mssimi  Really? What does it mean?

We can close this then. Could you send new PR please?

mssimi:
@TomasVotruba can we discuss it on google hangouts?

TomasVotruba:
@mssimi Sure :+1:

mssimi:
@TomasVotruba check [this](https://travis-ci.org/mssimi/rector/jobs/359082325) and let me know what you think.

TomasVotruba:
@mssimi That looks awesome :+1:  PR please!

mssimi:
@TomasVotruba I was loading wrong autoloader. Now I noticed problem which we need to solve to make it work. Method `glob()` does not work in PHAR and Symfony uses it for autowiring... I am looking for alternatives, we will have to replace all occurrences to make it work:(

mssimi:
Symfony is definitely not compatible with PHAR, Loaders use glob, Kernel uses glob, Finder uses glob. What do you propose @TomasVotruba to overcome this issue?

TomasVotruba:
Not sure, after bit Googling: https://github.com/cakephp/phinx/pull/1042/files

mssimi:
I found this solution too, but Symfony logic is more complex as it uses patterns, we will have to re-implement [this](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Config/Resource/GlobResource.php) class.

TomasVotruba:
That is doable. I do simalar changes in ECS: https://github.com/Symplify/Symplify/blob/master/packages/EasyCodingStandard/src/DependencyInjection/EasyCodingStandardKernel.php#L96

bendavies:
have just built this phar manually and it work on my symfony project! performance is a problem, but that's another issue i guess.

mssimi:
@bendavies How you solved autowiring issue?

bendavies:
I'm not seeing any...

mssimi:
Are you sure PHAR is autoloading files inside PHAR not outside?

bendavies:
no idea. I just know it's working and upgrading my symfony 2.8 project to 3.0 very nicely...

mssimi:
Maybe you could provide PHAR you compiled so we can check it.

bendavies:
i checked out this branch and ran `packages/PharBuilder/bin/compile`. that's it

mssimi:
This branch is not working yet, compilator does not include vendor at all, you actually probably run non-phar files.

mssimi:
Same thing happend to me.

bendavies:
you are right. it didn't autoload from the phar.

TomasVotruba:
Continues in #389

Issue #322, opened 2018-02-22
Move few static rectors to dynamic ones


TomasVotruba:
## Inspiration

- https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/QueryBuilder.php

### Start Case

- method name change + arg change, like Control redrawControl
    - https://github.com/rectorphp/rector/blob/bb4938db6e940eccfc88547b7c2dbc0ae42d801c/src/Rector/Contrib/Nette/Application/ValidateControlRector.php
    - https://github.com/nette/application/commit/afd8d321dd9eecd61fb96957d7a088933a7945ca#diff-c1f2adfe9e39de4486c15b301dc871eb

![image](https://user-images.githubusercontent.com/924196/36592755-54a57894-1897-11e8-9fe7-67126f13d380.png)

TomasVotruba:
An approach inspired by Google's [ClangMR](https://research.google.com/pubs/pub41342.html) could look like:

```php
$rectorBuilder = new RectorBuilder;

$rectorBuilder->matchMethodCallByName('validateControl')
    ->changeTo('redrawControl')
    ->addArgument(2, false);
```

What do you think @carusogabriel ?

carusogabriel:
@TomasVotruba Great ideas, I'll take a look this weekend :smile:

carusogabriel:
I really like this, make the creation of a new Rector by a simple method chaining API, looks like a game changer :muscle:

TomasVotruba:
Closing in favor of #435

PR #324, opened 2018-02-23
BuilderRector with simpler API to create new Rectors

TomasVotruba:
Ref #322

Writing Rectors could be easy as:

```php
return $this->builderRectorFactory->create()
    ->matchMethodCallByType('Nette\Application\UI\Control')
    ->matchMethodName('validateControl')
    ->changeMethodNameTo('redrawControl')
    ->addArgument(1, false);
```

### Todo

- [x] document in `/doc/RectorBuilder.md`

### Related Changes

- #326


TomasVotruba:
This section is ready to review. Now I only have to fix failng tests due to dependency on remove `nette/application` package. That is not related to this Builder concept

TomasVotruba:
**Ready to review**

packages/RectorBuilder/src/DependencyInjection/CompilerPass/RectorProvidersCompilerPass.php
```diff
@@ -0,0 +1,35 @@
...
+        );
+
+        // see https://symfony.com/doc/current/service_container/expression_language.html
+        foreach ($rectorProviderDefinitions as $rectorProviderDefinition) {
+            $rectorService = new Expression(
+                sprintf('service("%s").provide()', addslashes($rectorProviderDefinition->getClass()))
+            );
+
+            $rectorCollectorDefinition->addMethodCall('addRector', [$rectorService]);
+            $rectorNodeTraverserDefinition->addMethodCall('addVisitor', [$rectorService]);
```
TomasVotruba: Consider config only approach in tests instead of mixing configs with this

carusogabriel:
@TomasVotruba Oh, this should be an important change from what I can first see. I’ll review it tonight 😁

docs/HowItWorks.md
```diff
@@ -69,3 +69,7 @@ rectors:
 
 - After this Rector displays list of changed files
 - Or with `--dry-run` option the diff of these files
+
+### Similar Projects
+
+- [ClangMR](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/41342.pdf) for C++ by Google (closed source) - almost idential workflow, develope independly though
```
carusogabriel: I wish we could see the real project. Would give us so many ideas :pray:

packages/RectorBuilder/src/BuilderRector.php
```diff
@@ -0,0 +1,120 @@
...
+namespace Rector\RectorBuilder;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\MethodCall;
+use Rector\Node\NodeFactory;
+use Rector\NodeAnalyzer\MethodCallAnalyzer;
+use Rector\NodeChanger\IdentifierRenamer;
+use Rector\Rector\AbstractRector;
+
+final class BuilderRector extends AbstractRector
```
carusogabriel: This is the new API?

packages/RectorBuilder/tests/BuilderRector/RectorProvider.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * @return RectorInterface[]|NodeVisitor[]
+     */
+    public function provide(): array
+    {
+        $validateToRedrawControlRector = $this->builderRectorFactory->create()
+            ->matchMethodCallByType('Stub_Nette\Application\UI\Control')
+            ->matchMethodName('validateControl')
+            ->changeMethodNameTo('redrawControl')
+            ->addArgument(1, false);
```
carusogabriel: How easier :open_mouth:

tests/Rector/Contrib/Nette/Application/InjectPropertyRector/InjectPropertyRectorTest.php
```diff
@@ -25,11 +27,8 @@ public function provideWrongToFixedFiles(): array
...
     }
 
-    /**
-     * @return string[]
-     */
-    protected function getRectorClasses(): array
+    protected function provideConfig(): string
     {
-        return [InjectPropertyRector::class];
+        return __DIR__ . '/config.yml';
```
carusogabriel: Isn't a simple `return config.yml` enough? All config files live in the same test directory, right?

packages/RectorBuilder/src/BuilderRector.php
TomasVotruba: Yep!

tests/Rector/Contrib/Nette/Application/InjectPropertyRector/InjectPropertyRectorTest.php
TomasVotruba: Relative paths can fail more often than absolute ones. Explicit over implicit.

Also better IDE supported with click right to the file

There is even fixer for that: https://github.com/Symplify/CodingStandard/blob/master/README.md#includerequire-should-be-followed-by-absolute-path

docs/HowItWorks.md
TomasVotruba: I've added jscodeshift right under it. I saw the fluent API there as well.

Feel free to check it to dig some ideas

PR #326, opened 2018-02-24
Move to config-based testing

TomasVotruba:
### The Problem

- some tests used config
- other used array of classes to filter all loaded Rectors
- combined with `RectorCollector` adding new test required
    - adding tests class 
    - adding rector in config level
    - adding array of used classes Rectors

### Changes

- do not load Rectors via `FileProcessor->processFileWithRectorsToString()` with `RectorNodeTraverser`, that could be forgotten and is hard to debug
- rather use per config container, that is nice and clear
- put config to root test (possibly enforce?); protected method is checked by FileSystem guard now

### Related Changes

- drop `RectorCollector`, merge to `RectorNodeTraverser`


phpstan.neon
```diff
@@ -28,6 +28,8 @@ parameters:
         - '#Method Rector\\BetterReflection\\Reflector\\SmartClassReflector::reflect\(\) should return Rector\\BetterReflection\\Reflection\\ReflectionClass\|null but returns Rector\\BetterReflection\\Reflection\\Reflection#' # 1
         - '#Parameter \#1 \$node of method Rector\\NodeTypeResolver\\NodeTypeResolver::resolve\(\) expects PhpParser\\Node, PhpParser\\Node\\Expr|string given#' # 1
         - '#Cannot call method render\(\) on phpDocumentor\\Reflection\\DocBlock\\Tag\|string#' # 1
+        - '#Parameter \#1 \$visitor of method PhpParser\\NodeTraverser::addVisitor\(\) expects PhpParser\\NodeVisitor, PhpParser\\NodeVisitor|Rector\\Contract\\Rector\\RectorInterface given#'
+        - '#Method Rector\\NodeTraverser\\RectorNodeTraverser::getRectors\(\) should return array<PhpParser\\NodeTraverser\|Rector\\Contract\\Rector\\RectorInterface> but returns array<PhpParser\\NodeVisitor>#'
```
carusogabriel: This PHPStan exceptions are growing fast :cry:

TomasVotruba: Well it's not perfect.

In this particular case we might extends from `NodeVisitor` interface, but IIRC that would meant some coupling and empty methods. Shall verify

TomasVotruba: Done, it works :+1:

carusogabriel: Yes, I know, just comment that I'll try to reduce it again in the future :smile:

PR #331, opened 2018-02-28
Use toLowerString method

carusogabriel:
Ref: https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Identifier.php#L45-L52


TomasVotruba:
This looks like hiding API to external package. 
What if we use something else or this method would be removed?

carusogabriel:
> This looks like hiding API to external package.

Didn't get it :confused: 

> What if we use something else or this method would be removed?

What do you suggest?

TomasVotruba:
> > This looks like hiding API to external package.
>
> Didn't get it confused

I need to use `strtolower` here for some reason (unite ids). It could be hashing, it could be `strtoupper`.

I don't need `php-parser` to do it and it's not his job here. From reading such code you suggest, I'd think **I need `php-parser` here**, which is not true.

PR #335, opened 2018-03-03
Rector as Services to remove dump proxy "rectors" to "services" mechanism

TomasVotruba:
### Why?

- using `rectors` key in config instead of `services` was just to **save 1 extra line of config**:

	### Before

	```yml
	# rector.yml
	rectors:
	    SomeRector:
	        key: value
	```

	### After

	```yml
	# rector.yml
	services:
	    SomeRector:
	        $config:
	            key: value

	    # which is in full syntax
	    SomeRector:
	        arguments:
	            $config:
	                key: value
	```

Which is clear to first time user, that doesn't explore Rector DI in depth?

- to make this work - [RectorBundle](https://github.com/rectorphp/rector/pull/335/files#diff-59f9726089512850a088f57a73242974) was needed and [RectorsExtension](https://github.com/rectorphp/rector/pull/335/files#diff-c75c58fa52563ec19a2dd3f6d24da013)
-  extra classes that duplicated Symfony DI were needed: [ConfigMerger](https://github.com/rectorphp/rector/pull/335/files#diff-af08ddc6ca2a93dcbef230e826f4a117), [RectorClassNormalizer](https://github.com/rectorphp/rector/pull/335/files#diff-3aa4eb1d089cf6db7a4a2ac2c9981654) and [RectorClassValidator](https://github.com/rectorphp/rector/pull/335/files#diff-77c15770c60fb59b16065ad165e9bd8f)

- **I don't like reinventing the wheel**, moreover in Symfony\DI case

- I also recall, how frustrated I was when writing [ECS](https://github.com/symplify/easyCodingStandard/) and exploring PHP_CodeSniffer and PHP CS Fixer when I found out - all sniffs/fixers are just statically added services; it would save me dozens of hours if `services` like here would be used

- Also writing a [post about Rector and Collector](https://github.com/TomasVotruba/tomasvotruba.cz/blob/master/source/_posts/2018/2018-03-08-why-is-collector-pattern-so-awesome.md) helped me to realize this WTF

    
image

- Each step of such 1-level magic slows down the understanding of the code, [as stringly defined classes in Symfony used to](https://github.com/symfony/symfony/issues/11881).

For all these reasons we should get rid of this extra layer that only came from historical background now and adds more comlexity then help.

### Benefits

- 5 classes replicating Symfony\DI behavior was removed
- everyone can see from the config that Rector is service, **right away from the config**
- better understanding how Rector get dependencies and how they can be configured
- also allows to use any Symfony\DI features right away from the config; lot of hacking would be required to do it the old way
- removing extra config validation like this: https://github.com/rectorphp/rector/pull/335/files#diff-0e65def841bad669b99fc89ef2257117L173 and let Symfony\DI handle it instead

### Todo

- [x] fix post https://github.com/TomasVotruba/tomasvotruba.cz/blob/master/source/_posts/2018/2018-03-08-why-is-collector-pattern-so-awesome.md


carusogabriel:
@TomasVotruba Gonna review this afternoon. Sounds like an amazing improvement.

TomasVotruba:
Thank you, it definitely will :+1:

README.md
```diff
@@ -135,132 +135,142 @@ You can:
...
     Rector\Rector\Dynamic\PseudoNamespaceToNamespaceRector:
-        # old namespace prefix
-        - 'PHPUnit_'
-        # exclude classes
-        - '!PHPUnit_Framework_MockObject_MockObject'
+        $configuration:
+            # old namespace prefix
+            - 'PHPUnit_'
+            # exclude classes
+            - '!PHPUnit_Framework_MockObject_MockObject'
```
TomasVotruba: Now it could be another param with explicit name, like `$skipClasses`
To show this feature to the user without commenting everything

TomasVotruba:
Makes use of this awesome feature: https://github.com/symfony/symfony/pull/21313

TomasVotruba:
Just a note, there is quite simple and clear way to write *custom service formats* in `services.yml`

All is here: https://github.com/Symplify/Symplify/blob/adbf304c01a41c5d42593840436fa2450d364d2a/packages/EasyCodingStandard/src/Yaml/CheckerTolerantYamlFileLoader.php#L27-L81

```php

namespace Symplify\EasyCodingStandard\Yaml;

use Nette\Utils\Strings;
use Symfony\Component\Config\FileLocatorInterface;
use Symfony\Component\DependencyInjection\ContainerBuilder;
use Symfony\Component\DependencyInjection\Loader\YamlFileLoader;

/**
 * The need: https://github.com/symfony/symfony/pull/21313#issuecomment-372037445
 */
final class CheckerTolerantYamlFileLoader extends YamlFileLoader
{
    /**
     * @var CheckerConfigurationGuardian
     */
    private $checkerConfigurationGuardian;

    public function __construct(ContainerBuilder $containerBuilder, FileLocatorInterface $fileLocator)
    {
        $this->checkerConfigurationGuardian = new CheckerConfigurationGuardian();

        parent::__construct($containerBuilder, $fileLocator);
    }

    /**
     * @param string $file
     * @return array|mixed|mixed[]
     */
    protected function loadFile($file)
    {
        $decodedYaml = parent::loadFile($file);

        if (isset($decodedYaml['services'])) {
            return $this->moveArgumentsToPropertiesOrMethodCalls($decodedYaml);
        }

        return $decodedYaml;
    }

    /**
     * @param mixed[] $yaml
     * @return mixed[]
     */
    private function moveArgumentsToPropertiesOrMethodCalls(array $yaml): array
    {
        foreach ($yaml['services'] as $checker => $serviceDefinition) {
            if (empty($serviceDefinition)) {
                continue;
            }

            // is checker service?
            if (! Strings::endsWith($checker, 'Fixer') && ! Strings::endsWith($checker, 'Sniff')) {
                continue;
            }

            if (Strings::endsWith($checker, 'Fixer')) {
                $this->checkerConfigurationGuardian->ensureFixerIsConfigurable($checker, $serviceDefinition);
                // move parameters to "configure()" call
                $yaml['services'][$checker]['calls'] = [
                    ['configure', [$serviceDefinition]],
                ];
            }

            if (Strings::endsWith($checker, 'Sniff')) {
                // move parameters to property setters
                foreach ($serviceDefinition as $key => $value) {
                    $this->checkerConfigurationGuardian->ensurePropertyExists($checker, $key);
                    $yaml['services'][$checker]['properties'][$key] = $this->escapeValue($value);
                }
            }

            // cleanup parameters
            foreach ($serviceDefinition as $key => $value) {
                unset($yaml['services'][$checker][$key]);
            }
        }

        return $yaml;
    }

    /**
     * @param mixed $value
     * @return mixed
     */
    private function escapeValue($value)
    {
        if (is_numeric($value)) {
            return $value;
        }

        return Strings::replace($value, '#@#', '@@');
    }
}
```

PR #337, opened 2018-03-03
[POC] Laravel Controller to Symfony

TomasVotruba:
Playing around with first MVC FW => MVC FW migration

### Sources

- https://stackoverflow.com/questions/38859132/migrating-from-laravel-to-symfony
- https://stackoverflow.com/questions/44083131/do-i-need-to-switch-laravel-to-symfony

### Related Changes

- add `levels` command to show nice list of all levels

image


tests/Rector/Contrib/FrameworkMigration/LaravelToSymfony/Wrong/wrong.php.inc
```diff
@@ -0,0 +1,14 @@
+<?php declare(strict_types=1);
+
+namespace App\Controller;
+
+use App\Http\Controllers\Controller;
+use Illuminate\Http\Response;
+
+class UserController extends Controller
+{
+    public function show(): Response
```
carusogabriel: This return type is wrong. It should be `Illuminate\View\View`

TomasVotruba: Ok. I just copied example from the docs: https://laravel.com/docs/5.6/controllers#basic-controllers

Could you maybe send gist of what is common Laravel controller looks like? It'd be really helpful, since I don't use it at all

carusogabriel: Strange. The `view` helper method returns an `Illuminate\View\View`. The `json` a `Illuminate\Http\JsonResponse`. The `response` that only returns a `Response`

src/Console/Command/LevelsCommand.php
```diff
@@ -0,0 +1,59 @@
...
+    public function __construct(ConsoleStyle $consoleStyle)
+    {
+        $this->consoleStyle = $consoleStyle;
+
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this->setName(CommandNaming::classToName(self::class));
```
carusogabriel: Why not something like `--levels`? Like `--help`? Or no? :)

TomasVotruba: It's common approach in CLI command to name them.

Like 

```
git clone ...
bin/rector process ...
bin/rector levels
```

Options you used for to extend command. Help is exception in this.

TomasVotruba: Is that what you meant?

carusogabriel: Yes, but good explanation, TIL

TomasVotruba:
Merging as simple POC works + `Levels` command is needed in master.
Will be improved on iterations of course

PR #339, opened 2018-03-04
Fix AssertTrueFalseToSpecificMethodRector

### LINKED ISSUE #334: [PHPUnit] in_array with third is wrong refactory
Bug report:
```diff
-assertTrue(in_array($foo, [$bar], true))
+assertContains($foo, $bar, true)
```

The third arguments shouldn't be include.

carusogabriel:
Fixes #334

### DIFF
--- a/tests/Rector/Contrib/PHPUnit/SpecificMethod/AssertTrueFalseToSpecificMethodRector/Wrong/wrong.php.inc
+++ b/tests/Rector/Contrib/PHPUnit/SpecificMethod/AssertTrueFalseToSpecificMethodRector/Wrong/wrong.php.inc
@@ -5,7 +5,7 @@ final class MyTest extends \PHPUnit\Framework\TestCase
     public function test()
     {
         $this->assertTrue(array_key_exists('...', ['...']), 'argument');
-        $this->assertTrue(in_array('...', ['...']), 'argument');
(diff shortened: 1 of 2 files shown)


tests/Rector/Contrib/PHPUnit/SpecificMethod/AssertTrueFalseToSpecificMethodRector/Wrong/wrong.php.inc
```diff
@@ -5,7 +5,7 @@ final class MyTest extends \PHPUnit\Framework\TestCase
     public function test()
     {
         $this->assertTrue(array_key_exists('...', ['...']), 'argument');
-        $this->assertTrue(in_array('...', ['...']), 'argument');
+        $this->assertTrue(in_array('...', ['...'], true), 'argument');
```
TomasVotruba: Could you also add 

```php
$this->assertTrue(in_array('...', ['...']), 'argument');
```

so we know this case is still covered?

carusogabriel: I'll revert the `array_search`, so we know that both cases are covered, without expand our tests :)

PR #354, opened 2018-03-11
[Dynamic] Value object remover fixes

### LINKED ISSUE #351: [idea] Refactoring enum value objects
Using this rector:
```yml
services:
    Rector\Rector\Dynamic\ValueObjectRemoverRector:
        $valueObjectsToSimpleTypes:
            'OdbavTo\Domain\Model\Ticket\Input\GuestField': 'string'
```

Preview of what can/can not be done:

- [x] `foo(GuestField $guestField)` -> `foo(string $guestField)`
- [x] /** @var GuestField|null */ -> `/** @var string */`
@TomasVotruba this seems to be a bug as it introduces bc break -> it was nullable before, now in annotation is only `string` instead of `null|string`

- [x] `public function guestField(): ?GuestField` -> `public function guestField(): ?string` @TomasVotruba rector did not find and fix this one

- [x] `public function updateGuestField(?GuestField $guestField): void` ->`public function updateGuestField(?string $guestField): void`
@TomasVotruba rector seems to have problems with nullable, as this one was not fixed

---

Edited by @tomasvotruba

### Another Refactoring

These are not related to valueObjects, but rather custom Enums.

- `$guestField->equals($ticketInputValue->guestField())` -> `$guestField === $ticketInputValue->guestField()`
Any idea how this could be done via rector? I am thinking about regex find/replace right now, it should not be hard, but the idea is to teach rector how to do this, let's change world together

- `GuestField::LAST_NAME()` -> `GuestField::LAST_NAME`
This one can be easily done via regex find/replace

TomasVotruba:
Closes #351

### DIFF
--- a/easy-coding-standard.yml
+++ b/easy-coding-standard.yml
@@ -18,6 +18,7 @@ services:
     Symplify\CodingStandard\Sniffs\DependencyInjection\NoClassInstantiationSniff:
         extraAllowedClasses:
             - 'PhpParser\Node\*'
+            - 'Symplify\PackageBuilder\Reflection\*'
             - 'PhpParser\Comment\Doc'
             - 'PhpParser\NodeTraverser'
             - 'Rector\DeprecationExtractor\Deprecation\*'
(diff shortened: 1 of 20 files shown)


TomasVotruba:
Continuing...

TomasVotruba:
DocBlock changes are messy, but there is nothing better wa can do untill AST-formatpreserving tool is ready for docblocks

PR #366, opened 2018-03-16
Fix 3th argument of is_a

### LINKED ISSUE #364: [PHPUnit] is_a wrong refactor with 3th arg
Bug report:
```diff
-assertTrue(in_a($object, ClassName::class, true))
+assertInstanceOf(true, ClassName::class, $object)
```
The third arguments shouldn't be included.

Also, maybe an improvement idea is to check if `$object` is actually an object. Ref: https://github.com/zendframework/zend-expressive/pull/590

carusogabriel:
Fix #364


TomasVotruba:
Thanks

PR #369, opened 2018-03-17
Rename level folder to levels

carusogabriel:
I'd like to rename the `level` folder to `levels`.

IMHO makes sense as we have a command called `levels`, and the folder holds all the _levels_ :smile:


TomasVotruba:
It's standard to name all directories in singular. E.g. Exception, Provider, Rector...

That prevents checking number of 0 - m elements inside particular dir and jump renaming.

Smart applies for db tables

PR #370, opened 2018-03-18
Move config to root level

### LINKED ISSUE #357: config should be root level
config/ inside src/ dir is not obvious or expected - and hard to spot there.
Instead it should be a root dir besides bin/ and the others.
Usually only classes are inside /src IMO.

It will make it a bit easier to work with.
Thanks for the great work btw.

carusogabriel:
This resolves #357 by moving the `config` folder to the root level.


TomasVotruba:
There are 2 types of configs:

- user configs - the levels
- internal config - services and packages

This should move only the first one.

carusogabriel:
@TomasVotruba Should we keep `config` as the folder name?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

TomasVotruba:
Yep, that's converion as well (e.g. in Symfony https://github.com/symfony/framework-bundle/tree/master/Resources/config, though this one is nested).

carusogabriel:
@TomasVotruba Done, should be RTM after Travis :)

PR #373, opened 2018-03-20
Remove not nullable params

carusogabriel:
/cc @ondrejmirtes Should PHPStan alert us about this?

### DIFF
--- a/packages/RectorBuilder/src/BuilderRector.php
+++ b/packages/RectorBuilder/src/BuilderRector.php
@@ -17,17 +17,17 @@ final class BuilderRector extends AbstractRector
     private $methodCallAnalyzer;
 
     /**
-     * @var string|null
+     * @var string
      */
     private $methodCallType;
 
     /**
-     * @var string|null
+     * @var string
      */
     private $methodName;
 
     /**
-     * @var string|null
+     * @var string
      */
     private $newMethodName;
 
--- a/src/Rector/Contrib/PHPUnit/SpecificMethod/AssertTrueFalseToSpecificMethodRector.php
+++ b/src/Rector/Contrib/PHPUnit/SpecificMethod/AssertTrueFalseToSpecificMethodRector.php
@@ -61,7 +61,7 @@ final class AssertTrueFalseToSpecificMethodRector extends AbstractPHPUnitRector
     private $nodeFactory;
 
     /**
-     * @var string|null
+     * @var string
      */
     private $activeFuncCallName;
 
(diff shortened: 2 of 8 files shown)


packages/RectorBuilder/src/BuilderRector.php
```diff
@@ -17,17 +17,17 @@
...
-     * @var string|null
+     * @var string
      */
     private $methodName;
 
     /**
-     * @var string|null
+     * @var string
      */
     private $newMethodName;
```
TomasVotruba: What if the Builder matches only a class?

src/Rector/Dynamic/Configuration/ArgumentReplacerRecipe.php
```diff
@@ -110,7 +110,7 @@ public function getType(): string
     }
 
     /**
-     * @return mixed|null
+     * @return mixed
      */
     public function getDefaultValue()
```
TomasVotruba: What about `someFunction($argument);`?

src/Rector/Dynamic/NamespaceReplacerRector.php
```diff
@@ -96,7 +96,7 @@ private function resolveNameFromNode(Node $node): string
         }
 
         if ($node instanceof Name) {
-            /** @var FullyQualified|null $resolveName */
+            /** @var FullyQualified $resolveName */
             $resolveName = $node->getAttribute(Attribute::RESOLVED_NAME);
```
TomasVotruba: What if the value was not resolved for some special case?
It happened before and broke Rector

src/Rector/Contrib/Sensio/FrameworkExtraBundle/TemplateAnnotationRector.php
```diff
@@ -88,7 +88,7 @@ public function isCandidate(Node $node): bool
      */
     public function refactor(Node $classMethodNode): ?Node
     {
-        /** @var Return_|null $returnNode */
+        /** @var Return_ $returnNode */
         $returnNode = $this->betterNodeFinder->findLastInstanceOf((array) $classMethodNode->stmts, Return_::class);
```
TomasVotruba: What if no `return` is found? Not every method returns something

src/Rector/Contrib/PHPUnit/SpecificMethod/AssertTrueFalseToSpecificMethodRector.php
```diff
@@ -61,7 +61,7 @@
     private $nodeFactory;
 
     /**
-     * @var string|null
+     * @var string
      */
     private $activeFuncCallName;
```
TomasVotruba: This value is null on construct, isn't it?

carusogabriel:
I'll add tests and fix some of them then :sweat_smile:

carusogabriel:
Closing as invalid

PR #374, opened 2018-03-20
[PHPUnit] Safe way to check if a class is a Test class

carusogabriel:
The current logic checks if the TestCase ends with `Test`, but this isn't the safest way to do it. For example:
```php
<?php
class TestingCase extends \PHPUnit\Framework\TestCase {}

class MyTestCase extends TestingCase {}
```

### DIFF
--- a/src/Rector/AbstractPHPUnitRector.php
+++ b/src/Rector/AbstractPHPUnitRector.php
@@ -3,7 +3,6 @@
 namespace Rector\Rector;
 
 use PhpParser\Node;
-use PHPUnit\Framework\TestCase;
 use Rector\NodeTypeResolver\NodeTypeResolver;
 
 abstract class AbstractPHPUnitRector extends AbstractRector
@@ -27,6 +26,6 @@ abstract class AbstractPHPUnitRector extends AbstractRector
     {
         $nodeResolved = $this->nodeTypeResolver->resolve($node);
 
-        return ! array_intersect([TestCase::class, 'PHPUnit_Framework_TestCase'], $nodeResolved);
+        return ! array_intersect(['PHPUnit\Framework\TestCase', 'PHPUnit_Framework_TestCase'], $nodeResolved);
     }
 }
--- a/src/Rector/AbstractPHPUnitRector.php
+++ b/src/Rector/AbstractPHPUnitRector.php
@@ -8,12 +8,25 @@ use Rector\NodeTypeResolver\NodeTypeResolver;
 
 abstract class AbstractPHPUnitRector extends AbstractRector
 {
+    /**
+     * @var NodeTypeResolver
+     */
+    private $nodeTypeResolver;
+
+    /**
+     * Nasty magic, unable to do that in config autowire _instanceof calls.
+     *
+     * @required
+     */
+    public function setNodeTypeResolver(NodeTypeResolver $nodeTypeResolver): void
+    {
+        $this->nodeTypeResolver = $nodeTypeResolver;
+    }
+
     protected function isInTestClass(Node $node): bool
     {
-        $nodeTypeResolver = new NodeTypeResolver();
+        $nodeResolved = $this->nodeTypeResolver->resolve($node);
 
-        $nodeResolved = $nodeTypeResolver->resolve($node);
-
-        return (bool) ! array_intersect([TestCase::class, 'PHPUnit_Framework_TestCase'], $nodeResolved);
+        return ! array_intersect([TestCase::class, 'PHPUnit_Framework_TestCase'], $nodeResolved);
     }
 }
--- a/src/Rector/AbstractPHPUnitRector.php
+++ b/src/Rector/AbstractPHPUnitRector.php
@@ -4,14 +4,16 @@ namespace Rector\Rector;
 
 use PhpParser\Node;
 use PHPUnit\Framework\TestCase;
-use Rector\Node\Attribute;
+use Rector\NodeTypeResolver\NodeTypeResolver;
 
 abstract class AbstractPHPUnitRector extends AbstractRector
 {
     protected function isInTestClass(Node $node): bool
     {
-        $parentClassName = (string) $node->getAttribute(Attribute::PARENT_CLASS_NAME);
+        $nodeTypeResolver = new NodeTypeResolver();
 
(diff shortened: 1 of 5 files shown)


carusogabriel:
Do not merge wet, some cases fail if the test class extends a class that actually extends PHPUnit. I'll add tests to ensure that.

src/Rector/AbstractPHPUnitRector.php
```diff
@@ -2,21 +2,16 @@
...
 {
-    /**
-     * Check if the file is a PHPUnit TestCase. By default, it should end with "Test", as it is the standard.
-     *
-     * @see https://phpunit.de/getting-started-with-phpunit.html
-     */
     protected function isInTestClass(Node $node): bool
     {
-        $className = (string) $node->getAttribute(Attribute::CLASS_NAME);
+        $parentClassName = (string) $node->getAttribute(Attribute::PARENT_CLASS_NAME);
```
carusogabriel: @TomasVotruba Do you know any Attribute that indicates all class parents?

TomasVotruba: All available `Attribute` items are in the `Attribute` class.

This should be resolved by TypeResolver, as it's used only sometimes. So I'd get `CLASS_NODE` and then get its types.

What do you think?

src/Rector/AbstractPHPUnitRector.php
```diff
@@ -2,21 +2,16 @@
...
+use PHPUnit\Framework\TestCase;
 use Rector\Node\Attribute;
 
 abstract class AbstractPHPUnitRector extends AbstractRector
 {
-    /**
-     * Check if the file is a PHPUnit TestCase. By default, it should end with "Test", as it is the standard.
-     *
-     * @see https://phpunit.de/getting-started-with-phpunit.html
-     */
```
TomasVotruba: Why removing this?

What if sb will come and try to analyze nodes (much worse performance) instead of simple and fast PHPUnit approach.

carusogabriel: Because I test Rector against an application that didn't follow this rule, plus, all its test classes extends a based one :(

TomasVotruba: I see

src/Rector/AbstractPHPUnitRector.php
carusogabriel: @TomasVotruba Do you have an example of `TypeResolver` that I can study and update this PR?

src/Rector/AbstractPHPUnitRector.php
```diff
@@ -2,21 +2,30 @@
...
 {
     /**
-     * Check if the file is a PHPUnit TestCase. By default, it should end with "Test", as it is the standard.
-     *
-     * @see https://phpunit.de/getting-started-with-phpunit.html
+     * @var MethodCallAnalyzer
      */
+    private $methodCallAnalyzer;
+
+    public function __construct(MethodCallAnalyzer $methodCallAnalyzer)
```
carusogabriel: @TomasVotruba Should we revert 691eb10? Because doing this way, as the classes that extends it, will overwrite this method :(

src/Rector/AbstractPHPUnitRector.php
```diff
@@ -2,21 +2,18 @@
...
 {
-    /**
-     * Check if the file is a PHPUnit TestCase. By default, it should end with "Test", as it is the standard.
-     *
-     * @see https://phpunit.de/getting-started-with-phpunit.html
-     */
     protected function isInTestClass(Node $node): bool
     {
-        $className = (string) $node->getAttribute(Attribute::CLASS_NAME);
+        $nodeTypeResolver = new NodeTypeResolver();
```
carusogabriel: @TomasVotruba Should we revert 691eb10? Because doing this way, isn't working by our CS :(

carusogabriel: @TomasVotruba ping.

Is there a way to DI without `__construct`? If not, revert #300 will be the solution (unfortunately)

TomasVotruba: I missed this, sorry.

You can inject dependency as in Abstract class: https://github.com/rectorphp/rector/blob/9f4ca8071a2c06aeaae336f84c7cf5293fc211db/src/Rector/AbstractRector.php#L20-L48

That's Symfony approach to this

carusogabriel:
Finally done, we can now merge it :tada:

src/Rector/AbstractPHPUnitRector.php
```diff
@@ -2,21 +2,31 @@
...
+        $this->nodeTypeResolver = $nodeTypeResolver;
+    }
+
     protected function isInTestClass(Node $node): bool
     {
-        $className = (string) $node->getAttribute(Attribute::CLASS_NAME);
+        $nodeResolved = $this->nodeTypeResolver->resolve($node);
 
-        return Strings::endsWith($className, 'Test');
+        return ! array_intersect([TestCase::class, 'PHPUnit_Framework_TestCase'], $nodeResolved);
```
TomasVotruba: What happens, if sb installs a Rector but doesn't have PHPUnit with `TestCase` class installed?

carusogabriel: I didn’t get it :(

Without `TestCase` isn’t possible to execute tests, is?

TomasVotruba: Well, somebody can use `PHPUnit_Framework_TestCase`, right?

carusogabriel: We are checking it, aren't we? :thinking: 
```php
return ! array_intersect([TestCase::class, 'PHPUnit_Framework_TestCase'], $nodeResolved);
```

TomasVotruba: Yes, but that depends on `TestCase` to exist. And it does in our code.

Sb might download Rector and have only `PHPUnit_Framework_TestCase` in their code.
Soe they `TestCase` will end un as unexisting class.

That's why it should string here.

carusogabriel: So I'll ignore this file from our CS. I used `::class` here because CS complained about it

TomasVotruba:
Awesome, thank you :+1:

PR #375, opened 2018-03-20
Use short flags for PHPStan

carusogabriel:


TomasVotruba:
What is the motivation?

Just example: do you know what `-p` is for in PHP CS Fixer?

carusogabriel:
@TomasVotruba Gotcha, we use the long one here to be descriptive :smile:

PR #376, opened 2018-03-21
Add shortcuts for levels and config

carusogabriel:
Following https://github.com/Symplify/Symplify/pull/705.

IMHO we should add this, to give the user all possible options to run their commands :)

### DIFF
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -42,7 +42,7 @@ final class Application extends SymfonyApplication
     {
         $inputDefinition->addOption(new InputOption(
             'config',
-            null,
+            'c',
             InputOption::VALUE_REQUIRED,
             'Path to config file.',
             getcwd() . '/rector.yml'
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Could you also verify they work? They're both related to Config resolver in bin file

carusogabriel:
@TomasVotruba Looks like https://github.com/Symplify/Symplify/blob/master/packages/PackageBuilder/src/Configuration/LevelConfigShortcutFinder.php will need to be changed as well :smile:

carusogabriel:
@TomasVotruba Should we wait to require the next version of Symplify or is RTM?

TomasVotruba:
Merge would break the use actually. I'd wait for the stable release, since there is already enough troubles with other dev dependencies

TomasVotruba:
ECS 4 is behind the corner, so this can be merged

PR #378, opened 2018-03-24
AbstractToConstructorInjectionRector fix

mssimi:
Fix AbstractToConstructorInjectionRector to work with class names in get arguments.

### DIFF
--- a/src/Rector/Contrib/Symfony/FrameworkBundle/AbstractToConstructorInjectionRector.php
+++ b/src/Rector/Contrib/Symfony/FrameworkBundle/AbstractToConstructorInjectionRector.php
@@ -6,7 +6,6 @@ use PhpParser\Node;
 use PhpParser\Node\Expr\ClassConstFetch;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Name;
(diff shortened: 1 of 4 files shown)


src/Rector/Contrib/Symfony/FrameworkBundle/AbstractToConstructorInjectionRector.php
```diff
@@ -78,4 +77,27 @@ public function refactor(Node $methodCallNode): ?Node
...
+
+        if (! $argument instanceof ClassConstFetch) {
+            return null;
+        }
+
+        if ($argument->class instanceof Name) {
+            return $argument->class->getAttribute(Attribute::RESOLVED_NAME)->toString();
+        }
+
+        return null;
```
carusogabriel: Can't we move the `if ($argument->class instanceof Name) {` before the `if (! $argument instanceof ClassConstFetch) {` to clean up an `if`?

mssimi: PHPStan wont let as $argument->class may not be available.

TomasVotruba:
Thank you Marek :+1:

PR #379, opened 2018-03-25
Make assertInstanceOf Rector smarter

### LINKED ISSUE #377: [PHPUnit] Improve assertInstanceOf Rector
Current, imported classes that are used with `instanceof` aren't refactored correctly:

```php
<?php
use Foo\Bar;

$this->assertTrue($foo instanceof Bar\Baz);
```

```diff
-$this->assertInstanceOf('Bar\Baz', $foo);
+$this->assertInstanceOf('Foo\Bar\Baz', $foo);
```

carusogabriel:
This resolves #377 by resolving the class name with `ClassName::class` instead of `'ClassName'`

### DIFF
--- a/src/Rector/Contrib/PHPUnit/SpecificMethod/AssertInstanceOfComparisonRector.php
+++ b/src/Rector/Contrib/PHPUnit/SpecificMethod/AssertInstanceOfComparisonRector.php
@@ -8,6 +8,7 @@ use PhpParser\Node\Expr\MethodCall;
 use Rector\Node\NodeFactory;
 use Rector\NodeAnalyzer\MethodCallAnalyzer;
 use Rector\NodeChanger\IdentifierRenamer;
+use Rector\NodeTypeResolver\PerNodeTypeResolver\NameTypeResolver;
 use Rector\Rector\AbstractPHPUnitRector;
 
 /**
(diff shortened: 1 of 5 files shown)


src/Rector/Contrib/PHPUnit/SpecificMethod/AssertInstanceOfComparisonRector.php
```diff
@@ -82,7 +90,7 @@ public function changeOrderArguments(MethodCall $methodCallNode): void
         /** @var Instanceof_ $comparison */
         $comparison = $oldArguments[0]->value;
 
-        $className = $comparison->class->toString();
+        $className = $this->nameTypeResolver->resolve($comparison->class)[0];
```
carusogabriel: @TomasVotruba Is there a better way to resolve it? If no, should be RTM :)

carusogabriel:
@TomasVotruba Please review and merge :smile:

src/Rector/Contrib/PHPUnit/SpecificMethod/AssertInstanceOfComparisonRector.php
```diff
@@ -81,13 +81,13 @@ public function changeOrderArguments(MethodCall $methodCallNode): void
         /** @var Instanceof_ $comparison */
         $comparison = $oldArguments[0]->value;
 
-        $className = $comparison->class->toString();
+        $className = $comparison->class->toCodeString();
```
TomasVotruba: This is some new method? What does it do?

TomasVotruba:
Thank you :

src/Rector/Contrib/PHPUnit/SpecificMethod/AssertInstanceOfComparisonRector.php
carusogabriel: Instead of `Class`, it returns `\Class` if it so. Isn't something new I guess https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Name.php#L99-L107

TomasVotruba: I see, cool.

Btw, with hindsight I think we can use `toLowerString()` as well. I think the php-parse dependency would not be a problem :)

carusogabriel: I'll look into that

PR #389, opened 2018-03-29
PHAR improvements

mssimi:
Related to #319 

Simplified and greatly improved performance of PHAR building script by using rsync.
Add composer script to build PHAR.
You can now ignore PHAR files in .pharignore.
Included an example for prefixing vendors using rector(performance is bad here so example is just for small folder).

Atm there is problem with autowring in PHAR as Symfony uses glob and it is not supported by PHAR.
At least [this](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Config/Resource/GlobResource.php) class will have to be re-implemented. I may add test for it later.


packages/PharBuilder/src/config/rector.yml
```diff
@@ -0,0 +1,7 @@
+parameters:
+    prefix: 'RectorPhar'
+
+services:
+    Rector\Rector\Dynamic\NamespaceReplacerRector:
+        $oldToNewNamespaces:
+            'Jean85': '%prefix%Jean85'
```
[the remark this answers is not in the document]
TomasVotruba: First things first. When PHAR is done, we can improve performance.

Doing both at same time improved neither of both, tried for you :D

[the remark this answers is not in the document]
carusogabriel: I'm with Tomas here. Let's first fix this PHAR issue, than we can join forces together and make Rector as fast as we can :muscle:

mssimi:
Added test for` GlobResource`. I could use some help here.

mssimi:
`realpath()` also seems buggy in PHAR.

tests/Phar/GlobResourceTest.php
```diff
@@ -0,0 +1,16 @@
...
+
+final class GlobResourceTest extends TestCase
+{
+    public function testGlobResource(): void
+    {
+        $resource = new GlobResource('phar://' . __DIR__ . '/Source/rector.phar/src/', '', true);
+
+        $this->assertNotCount(0, $resource);
+    }
+}
```
TomasVotruba: Here it's needed to override `getContainerLoader()` in `AppKernel` with own `PharAwaretGlobFileLoader` class:

```php
final class RectorKernel extends Kernel
{

    /**
     * @param ContainerInterface|ContainerBuilder $container
     */
    protected function getContainerLoader(ContainerInterface $container): DelegatingLoader
    {
        $kernelFileLocator = new FileLocator($this);

        $loaderResolver = new LoaderResolver([
            // @todo - create this clsas
            return PharAwaretGlobFileLoader($container, $kernelFileLocator));
            // ...
        ]);

        return new DelegatingLoader($loaderResolver);
    }

}
```

As done in here https://github.com/symplify/packagebuilder#10-do-you-need-to-merge-parameters-in-yaml-files-instead-of-override

---

`PharAwaretGlobFileLoader` might look like this:


```php
namespace Symfony\Component\Config\Loader\GlobFileLoader;

final class PharAwaretGlobFileLoader extends GlobFileLoader
{
    protected function glob(string $pattern, bool $recursive, &$resource = null, bool $ignoreErrors = false)
    {
        // override parent glob() method logic to not-use "glob" function
    }
}

TomasVotruba: What do you think @mssimi ?

mssimi: Check GlobResource usages, I think it is not only used in loader.

TomasVotruba:
Follow up in #449

PR #390, opened 2018-03-31
Ignore PHPStorm .idea folder

carusogabriel:


mssimi:
What about ignoring .idea in your git global ignore config?

carusogabriel:
@mssimi Can I do it? :open_mouth:

grogy:
@carusogabriel yes, it is possible. You can read it in [documentation](https://git-scm.com/docs/gitignore) and inspire in [GitHub repository](https://github.com/github/gitignore)

TomasVotruba:
@carusogabriel Could you always provide some reasoning for PRs, like what issue do you struggle with?
There might be some better solutions, but it's hard to that tell without knowing your troubles.

Btw, global gitignore is way to go :)

Here is the solution I use: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore
It will just create a file you need to edit:

```bash
git config --global core.excludesfile ~/.gitignore_global
echo ".idea" > ~/.gitignore_global
```

PR #400, opened 2018-04-02
Fix #387

### LINKED ISSUE #387: [BUG] PHP crashes in VariableTypeResolver
Finding running against https://github.com/nelmio/alice

```
PHP Recoverable fatal error:  Object of class PhpParser\Node\Expr\Variable could not be converted to string in /home/carusogabriel/.composer/vendor/rector/rector/packages/NodeTypeResolver/src/PerNodeTypeResolver/VariableTypeResolver.php on line 54
```

carusogabriel:
This PR is a fixes #387. Unfortanly I couldn't find the code that was breaking, only the file: `tests/Loader/LoaderIntegrationTest.php` :disappointed: 

I just invert the logic to first check for a `Variable`, as the error was:
```
PHP Recoverable fatal error:  Object of class PhpParser\Node\Expr\Variable could not be converted to string in /home/carusogabriel/.composer/vendor/rector/rector/packages/NodeTypeResolver/src/PerNodeTypeResolver/VariableTypeResolver.php on line 54
```

### DIFF
--- a/packages/NodeTypeResolver/src/PerNodeTypeResolver/VariableTypeResolver.php
+++ b/packages/NodeTypeResolver/src/PerNodeTypeResolver/VariableTypeResolver.php
@@ -51,13 +51,13 @@ final class VariableTypeResolver implements PerNodeTypeResolverInterface, NodeTy
             return $this->nodeTypeResolver->resolve($classNode);
         }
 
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Thank you, let's validate it in practice

PR #402, opened 2018-04-03
Remove minimum stability "beta" as Symplify 4 was realeased

carusogabriel:
As well this was causing a lot of messages like `"but these conflict with your requirements or minimum-stability"` :)


TomasVotruba:
Thank you

PR #404, opened 2018-04-03
Use phpstan-strict-rules

carusogabriel:
I'd like to propose the usage of PHPStan Strict Rules in Rector. This is a selection of rules to improve strict programming and make the code even more readable.

All the rules can be found in its [`README.md`](https://github.com/phpstan/phpstan-strict-rules/blob/master/README.md).

### DIFF
--- a/packages/BetterReflection/src/Reflector/SmartClassReflector.php
+++ b/packages/BetterReflection/src/Reflector/SmartClassReflector.php
@@ -79,7 +79,7 @@ final class SmartClassReflector
     {
         // anonymous class
         if ($className === null) {
-            if ($classLikeNode && property_exists($classLikeNode, 'extends')) {
+            if ($classLikeNode !== null && property_exists($classLikeNode, 'extends')) {
                 return [$classLikeNode->extends->toString()];
             }
 
@@ -94,7 +94,7 @@ final class SmartClassReflector
             // intentionally empty
         }
 
-        if ($classLikeNode) {
+        if ($classLikeNode !== null) {
             return $this->resolveClassParentsFromNode($classLikeNode);
         }
 
@@ -158,7 +158,7 @@ final class SmartClassReflector
             return $this->classReflector;
         }
 
-        if ($currentSource) {
+        if ($currentSource !== null) {
             $this->lastSource = $currentSource;
             return $this->classReflector = $this->classReflectorFactory->createWithSource($currentSource);
         }
@@ -168,7 +168,7 @@ final class SmartClassReflector
 
     private function isValidClassName(string $className): bool
     {
-        if (empty($className)) {
+        if ($className === '') {
             return false;
         }
 
--- a/packages/BetterReflection/src/SourceLocator/SourceLocatorFactory.php
+++ b/packages/BetterReflection/src/SourceLocator/SourceLocatorFactory.php
@@ -47,7 +47,7 @@ final class SourceLocatorFactory
         $sourceLocators = [];
         [$directories, $files] = $this->splitToDirectoriesAndFiles($source);
 
-        if (count($directories)) {
+        if (count($directories) > 0) {
             $sourceLocators[] = new DirectoriesSourceLocator($directories, $this->locator);
         }
 
@@ -66,7 +66,7 @@ final class SourceLocatorFactory
      */
     private function createCommonLocators(): array
     {
-        if ($this->commonLocators) {
+        if (count($this->commonLocators) > 0) {
             return $this->commonLocators;
         }
 
--- a/packages/NodeTraverserQueue/src/BetterNodeFinder.php
+++ b/packages/NodeTraverserQueue/src/BetterNodeFinder.php
@@ -57,7 +57,7 @@ final class BetterNodeFinder
     public function findLastInstanceOf($nodes, string $type): ?Node
     {
         $foundInstances = $this->nodeFinder->findInstanceOf($nodes, $type);
-        if (! $foundInstances) {
+        if (count($foundInstances) === 0) {
             return null;
         }
 
--- a/packages/NodeTypeResolver/src/PerNodeTypeResolver/AssignTypeResolver.php
+++ b/packages/NodeTypeResolver/src/PerNodeTypeResolver/AssignTypeResolver.php
@@ -50,7 +50,7 @@ final class AssignTypeResolver implements PerNodeTypeResolverInterface, NodeType
 
         $variableTypes = $this->resolveTypeForRightSide($assignNode);
 
-        if ($variableTypes) {
+        if (count($variableTypes) > 0) {
             /** @var Variable $variableNode */
             $variableNode = $assignNode->var;
             $variableName = (string) $variableNode->name;
--- a/packages/NodeTypeResolver/src/TypesExtractor/ConstructorPropertyTypesExtractor.php
+++ b/packages/NodeTypeResolver/src/TypesExtractor/ConstructorPropertyTypesExtractor.php
@@ -38,7 +38,7 @@ final class ConstructorPropertyTypesExtractor
     public function extractFromClassNode(Class_ $classNode): array
     {
         $constructorParametersWithTypes = $this->getConstructorParametersWithTypes($classNode);
-        if (! count($constructorParametersWithTypes)) {
+        if (count($constructorParametersWithTypes) === 0) {
             return [];
         }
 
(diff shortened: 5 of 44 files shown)


src/Console/Command/ProcessCommand.php
```diff
@@ -185,7 +185,7 @@ private function processFile(SplFileInfo $fileInfo): void
     {
         $oldContent = $fileInfo->getContents();
 
-        if ($this->parameterProvider->provideParameter(Option::OPTION_DRY_RUN)) {
+        if ($this->parameterProvider->provideParameter(Option::OPTION_DRY_RUN) !== null) {
```
TomasVotruba: I don't see point in turning everything `if ($value)` into `if ($value !== null)`
It the return value would be -5, false, null and some value, it would make sense. Now it only forces more writing and more thinking while reading the code: "what other false value apart `null` that method returns?"

Could you provide some reasoning?

src/FileSystem/PhpFilesFinder.php
```diff
@@ -25,7 +25,7 @@ public function findInDirectoriesAndFiles(array $source): array
             }
         }
 
-        if (count($directories)) {
+        if (count($directories) > 0) {
```
TomasVotruba: This suggested there might be minus count of directories and that it should be tested.

:-1:

TomasVotruba: This suggested there might be minus count of directories and that it should be tested.

:-1:

src/NodeAnalyzer/ClassLikeAnalyzer.php
```diff
@@ -95,7 +95,7 @@ private function resolveNameNode(Node $node): string
      */
     private function resolveExtendsTypes(ClassLike $classLikeNode, ?string $className = null): array
     {
-        if (! $classLikeNode->extends) {
+        if ($classLikeNode->extends === null) {
```
TomasVotruba: This change makes sense :+1:

src/Rector/Contrib/Symfony/Controller/RedirectToRouteRector.php
```diff
@@ -91,15 +91,15 @@ private function resolveArguments(MethodCall $node): array
...
 
-        if (! $generateUrlNode->args[1] && $node->args[1]) {
+        if (! isset($generateUrlNode->args[1]) && isset($node->args[1])) {
             $arguments[] = [];
         }
 
-        if ($node->args[1]) {
+        if (isset($node->args[1])) {
             $arguments[] = $node->args[1];
         }
```
TomasVotruba: Could you add tests that prove this would fail without issets? So we know it has meaning to add and is also covered.

src/Rector/Dynamic/AnnotationReplacerRector.php
```diff
@@ -84,11 +84,8 @@ private function shouldSkip(Node $node): bool
 
         /** @var Node|null $parentNode */
         $parentNode = $node->getAttribute(Attribute::PARENT_NODE);
-        if (! $parentNode) {
-            return true;
-        }
 
-        return false;
+        return $parentNode === null;
```
TomasVotruba: Nice :+1:

TomasVotruba:
Thank you. I've added few comments.

Not sure it's worth the code readbility. The code is now slower to understand for me, because of from false/true comparision there are even 4 options the code could go.

Also did you fix this manually? I imagine this could be blocker for contributors. I've already encountered few people that send PR but didn't understand what they should do about the PHPStan error. DX and code readability should be our priority.

bendavies:
did you forget to commit the include in the neon file?

carusogabriel:
> I imagine this could be blocker for contributors. I've already encountered few people that send PR but didn't understand what they should do about the PHPStan error. DX and code readability should be our priority.

@TomasVotruba Be a blocker? Why? There are just best practices to make our code better :(

> did you forget to commit the include in the neon file?
@bendavies No, I didn't, is included :+1:

src/FileSystem/PhpFilesFinder.php
carusogabriel: "Minous count of directories" :thinking: 

So it should be 
```php
if (count($directories) >= 1) {
```
?

src/Console/Command/ProcessCommand.php
carusogabriel: PHPStan complained about that nothing more than booleans should be passed to an `if` statement.

The `ParameterProvider::provideParameter` method returns the given parameter **or null**, so we need to check if the parameter is different from `null`, right?

TomasVotruba:
Because "I've already encountered few people that send PR but didn't understand what they should do about the PHPStan error". 
Tools should help, not discourage new contributors. If it would automatically fix those, that would be no troubles for newcommers.

Imagine the whole coding standard here would only throw errors at people and no fixes.

carusogabriel:
@TomasVotruba Should I continue to work on this one, or you won't approve it?

packages/BetterReflection/src/Reflector/SmartClassReflector.php
```diff
@@ -168,7 +168,7 @@ private function getClassReflector(): ClassReflector
 
     private function isValidClassName(string $className): bool
     {
-        if (empty($className)) {
+        if ($className === '') {
```
TomasVotruba: This rule seem good

packages/BetterReflection/src/SourceLocator/SourceLocatorFactory.php
```diff
@@ -66,7 +66,7 @@ public function createWithSource(array $source): SourceLocator
      */
     private function createCommonLocators(): array
     {
-        if ($this->commonLocators) {
+        if (count($this->commonLocators) > 0) {
```
TomasVotruba: This rule seems good, but only for adding `count()`

The `> 0` part is not context aware, thus not needed

packages/NodeTraverserQueue/src/BetterNodeFinder.php
```diff
@@ -57,7 +57,7 @@ public function findFirstInstanceOf($nodes, string $type): ?Node
     public function findLastInstanceOf($nodes, string $type): ?Node
     {
         $foundInstances = $this->nodeFinder->findInstanceOf($nodes, $type);
-        if (! $foundInstances) {
+        if (count($foundInstances) === 0) {
```
TomasVotruba: This rule seems good, since it makes array explicit

packages/NodeTypeResolver/src/PerNodeTypeResolver/AssignTypeResolver.php
```diff
@@ -50,7 +50,7 @@ public function resolve(Node $assignNode): array
 
         $variableTypes = $this->resolveTypeForRightSide($assignNode);
 
-        if ($variableTypes) {
+        if (count($variableTypes) > 0) {
```
TomasVotruba: This rule seems good, but again `count($iterable)` only

packages/NodeTypeResolver/src/TypesExtractor/ConstructorPropertyTypesExtractor.php
```diff
@@ -38,7 +38,7 @@ public function __construct(MethodReflector $methodReflector, TypeAnalyzer $type
     public function extractFromClassNode(Class_ $classNode): array
     {
         $constructorParametersWithTypes = $this->getConstructorParametersWithTypes($classNode);
-        if (! count($constructorParametersWithTypes)) {
+        if (count($constructorParametersWithTypes) === 0) {
```
TomasVotruba: This rule seems good

packages/NodeTypeResolver/src/TypesExtractor/ConstructorPropertyTypesExtractor.php
```diff
@@ -68,18 +68,16 @@ private function getConstructorParametersWithTypes(Class_ $classNode): array
...
-                if ($this->typeAnalyzer->isBuiltinType($parameterType)) {
-                    continue;
-                }
-
-                $parametersWithTypes[$parameterName] = [$parameterType];
+            if ($this->typeAnalyzer->isBuiltinType($parameterType)) {
+                continue;
             }
+
+            $parametersWithTypes[$parameterName] = [$parameterType];
```
TomasVotruba: This looks like bigger refactoring. What happened here?

src/Builder/ConstructorMethodBuilder.php
```diff
@@ -44,9 +44,9 @@ public function addParameterAndAssignToConstructorArgumentsOfClass(
         VariableInfo $variableInfo,
         Expression $assignNode
     ): void {
+        /** @var ClassMethod|null $constructorMethod */
         $constructorMethod = $classNode->getMethod('__construct');
-        /** @var ClassMethod $constructorMethod */
-        if ($constructorMethod) {
+        if ($constructorMethod !== null) {
```
TomasVotruba: This rule seems good, mailnly added nullable annotation

TomasVotruba:
@carusogabriel I've added comments to particular use cases, that might improve readability without much contributing overhead. Mostly `count()` around iterables.

Could you add only those?

src/Rector/Contrib/Symfony/Controller/RedirectToRouteRector.php
TomasVotruba: Is this added somewhere?

src/FileSystem/PhpFilesFinder.php
TomasVotruba: Simply

```
if (count($directories)) {

```

src/Console/Command/ProcessCommand.php
TomasVotruba: Well, this is in application flow could not be used outside, so PHPStan can't correctly detect this.

src/FileSystem/PhpFilesFinder.php
carusogabriel: But `if (count($directories)) {` is wrong as `count` returns `int`, so I'll use `if (count($directories) >= 1) {`

TomasVotruba: That seems like super_strict_types. 

Why would you use that?

PR #405, opened 2018-04-03
Fix default config from string to array

mssimi:
After update to new Symplify `ConfigFileFinder::provide()` accepts array instead of string as second param.


TomasVotruba:
Thank you :+1:

PR #406, opened 2018-04-07
Support PHPUnit 7.1 iterable type in assertInternalType Rector

carusogabriel:
PHPUnit 7.1 [was released yesterday](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-7.1.md#710---2018-04-06) with support for the `iterable` type.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "require-dev": {
         "symplify/easy-coding-standard": "^4.0",
         "phpstan/phpstan-shim": "^0.9",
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thanks :+1: 

The fails is related to the Symplify package. Could you look at it and fix here?

carusogabriel:
@TomasVotruba Can you look into that? I'm not very familiar with Symplify :(

TomasVotruba:
First you need to identify, what Rector class and method is causing that

PR #409, opened 2018-04-08
[NodeTypeResolver] Remove dependency on 3rd party code

TomasVotruba:
It was a good start to make use of existing problems by type-resolving of 3rd party code.
But now it's unneccesary dependency and causes code to fail, e.g. https://github.com/rectorphp/rector/pull/395

This will decouple NodeTypeResolver from any Nette and Symfony classes.


TomasVotruba:
@carusogabriel Yep, see description :)

TomasVotruba:
Rebased

PR #412, opened 2018-04-09
Generalize AssertComparisonToSpecificMethodRector definition

carusogabriel:
Remove the `===`, as `AssertComparisonToSpecificMethodRector` have many comparions


TomasVotruba:
When I read "comparison", I have no idea what that means. It could be is_string(),  assertIsString

When I open the Rector, I see only =, <, >, ! chars. That's more narrow and easier to understand when reading 10 PHPUnit Rectors . 
https://github.com/rectorphp/rector/blob/c4fb76453ce9e51c9e14f4b1d638989145b8767f/src/Rector/Contrib/PHPUnit/SpecificMethod/AssertComparisonToSpecificMethodRector.php#L21-L29

But I see what you mean, that `===` is not helpful either.

I have a suggestion, to replace it with chars in brackets (`=`, `<`, `>`, `!`). That would give user better idea without reading code samples. What do you think?

carusogabriel:
@TomasVotruba Maybe "comparison operations", to match https://php.net/manual/en/language.operators.comparison.php?

PR #413, opened 2018-04-09
Make AssertComparisonToSpecificMethodRector smarter

carusogabriel:
Inspired by https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3540

With this PR, I want to make `AssertComparisonToSpecificMethodRector` smarter, recognizing Yoda Comparisons and check what member has a `string`, number, const (`true`, `false` or `null`), as well variables starting with `exp` (from expected). This will result in more refactoring when passing to other Rectors :smile: 

TODO:
- [x] Recognize `array`
- [x] Recognize `float`

### DIFF
--- a/src/Rector/Contrib/PHPUnit/SpecificMethod/AssertComparisonToSpecificMethodRector.php
+++ b/src/Rector/Contrib/PHPUnit/SpecificMethod/AssertComparisonToSpecificMethodRector.php
@@ -5,7 +5,11 @@ namespace Rector\Rector\Contrib\PHPUnit\SpecificMethod;
 use PhpParser\Node;
 use PhpParser\Node\Arg;
 use PhpParser\Node\Expr\BinaryOp;
+use PhpParser\Node\Expr\ConstFetch;
 use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Scalar\LNumber;
+use PhpParser\Node\Scalar\String_;
 use Rector\NodeAnalyzer\MethodCallAnalyzer;
 use Rector\NodeChanger\IdentifierRenamer;
 use Rector\Rector\AbstractPHPUnitRector;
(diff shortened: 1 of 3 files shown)


src/Rector/Contrib/PHPUnit/SpecificMethod/AssertComparisonToSpecificMethodRector.php
```diff
@@ -106,8 +110,16 @@ public function changeOrderArguments(MethodCall $methodCallNode): void
...
+        if ($this->isConstantValue($expression->right)) {
+            $firstArgument = new Arg($expression->right);
+            $secondArgument = new Arg($expression->left);
+        } elseif ($this->isConstantValue($expression->left)) {
+            $firstArgument = new Arg($expression->left);
+            $secondArgument = new Arg($expression->right);
+        } else {
+            $firstArgument = new Arg($expression->right);
+            $secondArgument = new Arg($expression->left);
+        }
```
mssimi: Can't we drop one condition here?

carusogabriel: Witch one? I’m searching from constant value in both members of the comparison, than I’m assuming a not Yoda Comparison.

mssimi: If you move  elseif `($this->isConstantValue($expression->left))` to be first condition you can drop elseif.

mssimi:
```php 
if ($this->isConstantValue($expression->left)) {
 $firstArgument = new Arg($expression->left);
 $secondArgument = new Arg($expression->right);
} else {
 $firstArgument = new Arg($expression->right);
 $secondArgument = new Arg($expression->left);
}
```

mssimi: I think this kind of issue can be automated maybe ECS could do that, what do you think @TomasVotruba ?

TomasVotruba: @mssimi How would you check it in some simple way? Coding Standard is not ready for this, since it has to check flow of control.
It's rather taks for PHPStan

mssimi: Simple workflow I have in my mind.

Consider code like this.

If
Code 1
elseif
Code 2
else
Code3

If Code 1 === Code3 then drop first condition and change second from elseif to if. ESC can't do this?

TomasVotruba: It is possible, just from my point of view not worth the work.

TomasVotruba:
Let me know if this is ready to merge

src/Rector/Contrib/PHPUnit/SpecificMethod/AssertComparisonToSpecificMethodRector.php
```diff
@@ -132,4 +141,11 @@ private function renameMethod(MethodCall $methodCallNode): void
             'assertFalse' => $falseMethodName,
         ]);
     }
+
+    private function isConstantValue(Node $node): bool
+    {
+        return in_array(get_class($node), [Array_::class, ConstFetch::class], true)
+              || is_subclass_of($node, Scalar::class)
```
carusogabriel: Using this check we can support all kind of [Scalar Nodes](https://github.com/nikic/PHP-Parser/tree/master/lib/PhpParser/Node/Scalar) :tada:

TomasVotruba: Very nice :1st_place_medal:

PR #417, opened 2018-04-12
remove unnecessary condition(NamespaceReplacerRector)

mssimi:
Hey, I think this condition is unnecessary as next method that retrieve name, also checks that:)

### DIFF
--- a/src/Rector/Dynamic/NamespaceReplacerRector.php
+++ b/src/Rector/Dynamic/NamespaceReplacerRector.php
@@ -43,10 +43,6 @@ final class NamespaceReplacerRector extends AbstractRector
 
     public function isCandidate(Node $node): bool
     {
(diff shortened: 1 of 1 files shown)


TomasVotruba:
I see. As `isCandidate()`  should filter out all invalid cases, the duplication should be removed in the second method.

mssimi:
No by next method I mean `resolveNameFromNode()`.

TomasVotruba:
I see. Link to method code would more helpful next time.

In that case, `resolveNameFromNode()` should return  `?string`, so the invalid type is more explicit. `null` beats "is string empty"

mssimi:
I am not fan of nullable typehints, I use them only when neccesary, coz if you make typehint nullable you have to deal with that later in code.

TomasVotruba:
I see, let's keep it for now.

mssimi:
Anytime bro:)

PR #419, opened 2018-04-13
Add autoload_files parameter

TomasVotruba:
E.g. for PHP_Codesniffer


src/FileSystem/FileGuard.php
```diff
@@ -6,9 +6,9 @@
 
 final class FileGuard
 {
-    public static function ensureFileExists(string $file, string $location): void
```
carusogabriel: Note to self: revise `static` methods

PR #423, opened 2018-04-14
apply ArrayToYieldDataProviderRector

TomasVotruba:
Ref #401 

Refactoring Rector with Rector :)

image


TomasVotruba:
I think there was some issue reported. Could you try it?

carusogabriel:
@TomasVotruba We can use. But, we'd need to rollback the annotations. What is your opinion?

TomasVotruba:
What would be added value of that?

carusogabriel:
@TomasVotruba We're just using the features we have. The decision is up to you :)

TomasVotruba:
I just don't see any added value in it. Only it's less precise, more code, less readable and ambiguous. But that's me. Do you think of any advantages?

TomasVotruba:
:+1: Thanks

Issue #424, opened 2018-04-20
Create CodeQuality level


carusogabriel:
## New Ideas

- [ ] early return

    ```diff
     if ($this->hasDocBlock($tokens, $index)) {
    -    $docToken = $tokens[$this->getDocBlockIndex($tokens, $index)];
    +    return $tokens[$this->getDocBlockIndex($tokens, $index)];
    -} else {
    -    $docToken = null;
    -}
    -return $docToken;
    +return null;
	``` 

- [ ] if/else array value set
    Maybe

    ```diff
    -if (empty($value)) {
    -   $this->arrayBuilt[][$key] = true;
    -} else {
    -  $this->arrayBuilt[][$key] = $value;
    -}
    +$this->arrayBuilt[][$key] = empty($value) ?: $value;
    ```

    Would be tricky for the value inside if

## Done

- [x] return value if not null

    ```diff
     $newNode = $this->refactor($node);
    -if ($newNode !== null) {
    -    return $newNode;
    -}
    -return null;
    +return $newNode;
    ```

    - can be applied for any value

    ```diff
     $newNode = $this->refactor($node);
    -if ($newNode !== 25) {
    -    return $newNode;
    -}
    -return 25;
    +return $newNode;
- [x] eliminate dead code

    ```diff
    -$eachFuncCall->args[0];
    ```
- [x] Simplify checks:

    ```diff
    -is_array($values) && empty($values)
    +$values === []
    ```

- [x] simplify If

    ```diff
    -if (strpos($docToken->getContent(), "\n") === false) {
    -    return true;
    -}
    -return false;
    +return strpos($docToken->getContent(), "\n") === false;
    ```

- [x] short return

    ```diff
    -$nodeClasses = array_keys($robotLoader->getIndexedClasses());
    -return $nodeClasses;
    +return array_keys($robotLoader->getIndexedClasses());
    ```
- [x] simplify foreach to coala null

    ```diff
    -foreach ($this->oldToNewFunctions as $oldFunction => $newFunction) {
    -    if ($currentFunction === $oldFunction) {
    -        return $newFunction;
    -    }
    -}
    -
    -return null;
    +return $this->oldToNewFunctions[$currentFunction] ?? null;
    ```
- [x] binary switch
    `switch` statements with one `case` and `default` can be simplified to `if/else`:

    ```diff
    -switch ($foo) {
    -    case 'my string':
    -        $result = 'ok';
    -    break;
    -    
    -    default:
[...]

PR #426, opened 2018-04-22
[CodeQuality] Create InArrayAndArrayKeysToArrayKeyExistsRector

carusogabriel:
As suggested in https://github.com/rectorphp/rector/issues/424#issuecomment-383379319.

Reference that the two logics are the same: https://3v4l.org/2ok4M

### DIFF
--- a/src/Rector/Contrib/CodeQuality/InArrayAndArrayKeysToArrayKeyExistsRector.php
+++ b/src/Rector/Contrib/CodeQuality/InArrayAndArrayKeysToArrayKeyExistsRector.php
@@ -23,18 +23,14 @@ final class InArrayAndArrayKeysToArrayKeyExistsRector extends AbstractRector
 
     public function isCandidate(Node $node): bool
     {
-        if (! $node instanceof FuncCall) {
+        if (! $this->isInArrayFunction($node)) {
             return false;
         }
 
-        /** @var Name $funcCallName */
-        $funcCallName = $node->name;
+        /** @var FuncCall $inArrayFunction */
+        $inArrayFunction = $node;
 
-        if ($funcCallName->toString() !== 'in_array') {
-            return false;
-        }
-
-        $secondArgument = $node->args[1]->value;
+        $secondArgument = $inArrayFunction->args[1]->value;
         if (! $secondArgument instanceof FuncCall) {
             return false;
         }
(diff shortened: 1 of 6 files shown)


TomasVotruba:
New PR :tada:  

Shall I review? :)

carusogabriel:
@TomasVotruba Please. I guess we can already start this new level with subfolders (sublevels), but I'd like your advice

src/Rector/Contrib/CodeQuality/InArrayAndArrayKeysToArrayKeyExistsRector.php
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        /** @var Name $funcCallName */
+        $funcCallName = $node->name;
+
+        if (! $funcCallName instanceof Name) {
+            return false;
+        }
+
+        if ($funcCallName->toString() !== 'in_array') {
```
TomasVotruba: What about shorter?

```php
return $funcCallName->toString() === 'in_array';
```

TomasVotruba:
> I guess we can already start this new level with subfolders (sublevels), but I'd like your advice

I'd wait until we have 20 or so rules. Now it has no added value

src/Rector/Contrib/CodeQuality/InArrayAndArrayKeysToArrayKeyExistsRector.php
carusogabriel: I'd like to keep the early returns. Is there any problem? :)

TomasVotruba: Is this early return?

carusogabriel: The above are

TomasVotruba: Above calls are fine :)

Just `if ($funcCallName->toString() !== 'in_array') {` is not neccesary. 
I recall you've send similar PRs to code here to get rid of there `$value ? true : false` calls

TomasVotruba:
Thank you, well done :+1:

PR #431, opened 2018-04-25
Apply ArrayToYieldDataProviderRector on packages as well

### LINKED ISSUE #430: YieldRector missed method
```php
     public function provideWrongToFixedFiles(): array
     {
         return [[__DIR__ . '/Wrong/wrong.php.inc', __DIR__ . '/Correct/correct.php.inc']];
     }
```

TomasVotruba:
Closes #430


TomasVotruba:
Replaced by #433, better decoupled

PR #439, opened 2018-04-27
Autodiscovery services via "kernel_class" parameter

TomasVotruba:
Ref #428

Issue #368, opened 2018-03-17
Idea - EntityForDoctrineRepositoryProvider automatic provider 


mssimi:
I was thinking about automatization of [this](https://github.com/rectorphp/rector/blob/master/tests/Rector/Architecture/DoctrineRepositoryAsService/Source/EntityForDoctrineRepositoryProvider.php#L13).

What about something like this to automate it?

```php
<?php declare(strict_types=1);

namespace App\Utils\rector;

use Doctrine\ORM\EntityManagerInterface;
use Rector\Contract\Bridge\EntityForDoctrineRepositoryProviderInterface;

final class EntityForDoctrineRepositoryProvider implements EntityForDoctrineRepositoryProviderInterface
{

    /**
     * @var EntityManagerInterface
     */
    private $entityManager;

    public function __construct(EntityManagerInterface $entityManager)
    {
        $this->entityManager = $entityManager;
        $this->populateMap();
    }

    /**
     * @var string[]
     */
    private $map = [];

    /**
     * @param string $name
     * @return null|string
     */
    public function provideEntityForRepository(string $name): ?string
    {
        return $this->map[$name] ?? null;
    }

    /**
     * @return array
     * @throws \Doctrine\ORM\ORMException
     */
    private function entities(): array
    {
        return $this->entityManager->getConfiguration()->getMetadataDriverImpl()->getAllClassNames();
    }

    private function populateMap(): void
    {
        foreach ($this->entities() as $entity) {
            $this->map[\get_class($this->entityManager->getRepository($entity))] = $entity;
        }
    }
}
```

Maybe we could add it to rector itself what do you think @TomasVotruba?

TomasVotruba:
How would you configure it in config?
What if sb has different version?

mssimi:
> How would you configure it in config?

I could use some help here, if I knew how to do that, I would create pull request, atm I cannot make it work even with simple provider probably due to #303. I posted it so we can discuss, if we can make it work.

> What if sb has different version?

What is sb?

TomasVotruba:
Sb = somebody (on mobile abroad, so shortcutting :)). Every user can depend on different Doctrine version. So what happens when I use this package with incompatible Doctrine version?

I'm not sure either, since I don't have the code to test on. I thought you've made it work for you code base.

Well, what happens when you register it as service? And go from there.

Using this approach you have to basically register whole Doctrine to Rector, which is crazy. I'd provide simple array instead , using your script in `README` as tip how to get that list.

[...]

PR #440, opened 2018-04-27
doctrine alias to entity class rector

### LINKED ISSUE #438: [Doctrine] alias to class Rector
Follow up to #434 

This alias support is [removed in Doctrine 2.7+](https://github.com/doctrine/doctrine2/commit/63cb8018dd1a889be4e08a31fad1e76f4cc876d0#diff-824ad31297dda8bacca99948535bc937), so this belongs to `doctrine27.yml`

There should be Rector that handles 'App:Post' => `AppBundle\Entity\Post` conversion, to cover this case
https://github.com/rectorphp/rector/pull/434/files#diff-1959c4870ae07888515b8c6ed0d52831

Also related to #428

mssimi:
Closes #438 

Check it and let me know what you think:)

### DIFF
--- a/src/NodeAnalyzer/MethodArgumentAnalyzer.php
+++ b/src/NodeAnalyzer/MethodArgumentAnalyzer.php
@@ -33,9 +33,6 @@ final class MethodArgumentAnalyzer
         return true;
     }
 
-    /**
-     * @param MethodCall $node
-     */
     public function isMethodFirstArgumentString(Node $node): bool
     {
         if (! $this->hasMethodFirstArgument($node)) {
--- a/src/Rector/Contrib/Doctrine/AliasToClassRector.php
+++ b/src/Rector/Contrib/Doctrine/AliasToClassRector.php
@@ -3,9 +3,8 @@
 namespace Rector\Rector\Contrib\Doctrine;
 
 use PhpParser\Node;
-use PhpParser\Node\Scalar\String_;
-use Rector\Node\Attribute;
 use Rector\Node\NodeFactory;
+use Rector\NodeAnalyzer\MethodArgumentAnalyzer;
 use Rector\NodeAnalyzer\MethodCallAnalyzer;
 use Rector\Rector\AbstractRector;
 use Rector\RectorDefinition\CodeSample;
@@ -29,52 +28,23 @@ final class AliasToClassRector extends AbstractRector
     private $nodeFactory;
 
     /**
+     * @var MethodArgumentAnalyzer
+     */
+    private $methodArgumentAnalyzer;
+
+    /**
      * @param string[] $aliasesToNamespaces
      */
     public function __construct(
         array $aliasesToNamespaces,
         MethodCallAnalyzer $methodCallAnalyzer,
+        MethodArgumentAnalyzer $methodArgumentAnalyzer,
         NodeFactory $nodeFactory
     ) {
         $this->methodCallAnalyzer = $methodCallAnalyzer;
         $this->aliasesToNamespaces = $aliasesToNamespaces;
         $this->nodeFactory = $nodeFactory;
-    }
-
-    public function isCandidate(Node $node): bool
-    {
-        if (! $this->methodCallAnalyzer->isMethod($node, 'getRepository')) {
-            return false;
-        }
-
-        $className = $node->getAttribute(Attribute::CLASS_NAME);
-
-        if ($className === null) {
-            return false;
-        }
-
-        /** @var MethodCall $node */
-        if (count($node->args) !== 1) {
-            return false;
-        }
-
-        if (! $node->args[0]->value instanceof String_) {
-            return false;
-        }
-
-        /** @var String_ $string */
-        $string = $node->args[0]->value;
-
-        return $this->isAlias($string->value) && $this->entityFqn($string->value) !== '';
-    }
-
-    public function refactor(Node $node): ?Node
-    {
-        $node->args[0]->value = $this->nodeFactory->createClassConstantReference(
-            $this->entityFqn($node->args[0]->value->value)
-        );
-
-        return $node;
+        $this->methodArgumentAnalyzer = $methodArgumentAnalyzer;
     }
 
     public function getDefinition(): RectorDefinition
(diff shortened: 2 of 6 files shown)


TomasVotruba:
I looked very briely. You wrote about general solution, but this doesn't seem like it.
What are you plans about it?

src/Rector/Contrib/Doctrine/AliasToClassRector.php
```diff
@@ -0,0 +1,102 @@
...
+
+        return $node;
+    }
+
+    public function getDefinition(): RectorDefinition
+    {
+        return new RectorDefinition('Replaces doctrine alias with class.', [
+            new CodeSample('$em->getRepository("AppBundle:Post");', '$em->getRepository(\App\Entity\Post::class);'),
+        ]);
+    }
```
TomasVotruba: This method should be consistently up, so reader know what is this about right from the 1st line.

src/Rector/Contrib/Doctrine/AliasToClassRector.php
```diff
@@ -0,0 +1,102 @@
...
+
+    private function isAlias(string $name): bool
+    {
+        return strpos($name, ':') !== false;
+    }
+
+    private function entityFqn(string $name): string
```
TomasVotruba: I have no diea what this means.
The method should be a verb. converThisToThat, getThisFromThat...

src/Rector/Contrib/Doctrine/AliasToClassRector.php
```diff
@@ -0,0 +1,102 @@
...
+    {
+        [$namespaceAlias, $simpleClassName] = explode(':', $name, 2);
+
+        if (array_key_exists($namespaceAlias, $this->aliasesToNamespaces)) {
```
TomasVotruba: Why not isset?

src/Rector/Contrib/Doctrine/AliasToClassRector.php
```diff
@@ -0,0 +1,102 @@
...
+            return sprintf('%s\%s', $this->aliasesToNamespaces[$namespaceAlias], $simpleClassName);
+        }
+
+        return '';
```
TomasVotruba: Maybe null, since it was invalid?

Could you add test case for this?

src/Rector/Contrib/Doctrine/AliasToClassRector.php
```diff
@@ -0,0 +1,102 @@
...
+    {
+        if (! $this->methodCallAnalyzer->isMethod($node, 'getRepository')) {
+            return false;
+        }
+
+        $className = $node->getAttribute(Attribute::CLASS_NAME);
+
+        if ($className === null) {
+            return false;
+        }
```
TomasVotruba: This should be added to `isMethod()` above to `isTypeAndMethod()`. It will check the type of variable and method name

src/Rector/Contrib/Doctrine/AliasToClassRector.php
```diff
@@ -0,0 +1,102 @@
...
+
+        /** @var MethodCall $node */
+        if (count($node->args) !== 1) {
+            return false;
+        }
+
+        if (! $node->args[0]->value instanceof String_) {
+            return false;
+        }
```
TomasVotruba: Maybe decouple these 2 conditions to `hasOneStringArgument()` method.  It would be easier to read.
There might be future potential to class decoupling as there are more places with such code

mssimi:
I did, but I actually realized, we have only to provide namespace alias not whole map.

src/Rector/Contrib/Doctrine/AliasToClassRector.php
```diff
@@ -0,0 +1,102 @@
...
+
+        /** @var String_ $string */
+        $string = $node->args[0]->value;
+
+        return $this->isAlias($string->value) && $this->entityFqn($string->value) !== '';
```
TomasVotruba: Maybe group to `isAliasWithConfiguredEntity()` method. This more complex to read then clear method name

src/Rector/Contrib/Doctrine/AliasToClassRector.php
mssimi: I use that method in candidate as well as refactor, if I return null phpstan won't let me unless I add condition to refactor, which is actually unneeded. PHPStan false positive.

src/NodeAnalyzer/MethodArgumentAnalyzer.php
```diff
@@ -33,9 +33,6 @@ public function hasMethodSecondArgument(Node $node): bool
         return true;
     }
 
-    /**
-     * @param MethodCall $node
-     */
     public function isMethodFirstArgumentString(Node $node): bool
     {
         if (! $this->hasMethodFirstArgument($node)) {
```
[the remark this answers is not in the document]
mssimi: I thought u gonna ask, it is actually checked by `if (! $this->hasMethodFirstArgument($node))`.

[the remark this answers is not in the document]
TomasVotruba: Yes, that's what I mean. So?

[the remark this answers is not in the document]
mssimi: That means any node can enter that, coz it is validated?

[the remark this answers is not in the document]
mssimi: It is validated [here](https://github.com/mssimi/rector/blob/cbcb9106c2094f7b8685e70fdae827c7f2239db1/src/NodeAnalyzer/MethodArgumentAnalyzer.php#L16)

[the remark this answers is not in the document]
TomasVotruba: Allright :+1:

mssimi:
I thought we already had discussion(#368) I cannot be done w/o adding doctrine as dependency?

TomasVotruba:
I didn't understand you mean manual configuratoin by that.

I though you mean like configurable "method argument replacer", where you put service that handles argument values and converts them to something else.

---

People won't do this manually, it has no added value. Yet still you can make it more abstract by using interface with default implementation that would make use of `kernel_class`.

The same way there is service name to type: 
 https://github.com/rectorphp/rector/blob/a04fcde93275995d8d81f22c1501330a3d8d2bd6/src/Rector/Contrib/Symfony/FrameworkBundle/AbstractToConstructorInjectionRector.php#L94

Anyone can overload it manually, if they ever want.

If that's too abstract (I recall it is), we can merge this. Then #439 and then I'll look on it.

PR #442, opened 2018-04-28
improvement - argument analyzer dynamic

mssimi:
I noticed ArgumentAnalyzer has methods like isFirstSomething, IsSecondSomething. I think we should generalize it. Check it and let me know what you think.


TomasVotruba:
Nth is really unreadable. What about classic count?

mssimi:
Do you mean method name or method param or both?:D

TomasVotruba:
"Nth" for naming fo anything.

mssimi:
I'm open for suggestions, but "isMethodCountArgument" sounds worse:D

TomasVotruba:
Thank you Marek

PR #445, opened 2018-04-28
Add format to describe command

### LINKED ISSUE #408: Add describe command with CLI/Markdown output
Make use of #407 and add a `describe` command that:

- collects all Rector from provided `--config`/`--level` option and displays their `RectorDefinition->getDescription`
- has `--format` option with CLI as default, and Markdown as possible value
- has `--with-diff` option to show `CodeSample[]` diffs

TomasVotruba:
Closes #408

orevuew


TomasVotruba:
The generated md file is here: https://github.com/rectorphp/rector/blob/2a69df4ff4b0af4ca92a666697f1ef768ed73639/docs/AllRectorsOverview.md

TomasVotruba:
The generated md file is here: https://github.com/rectorphp/rector/blob/2a69df4ff4b0af4ca92a666697f1ef768ed73639/docs/AllRectorsOverview.md

Made with

```bash
bin/rector describe --level all --format md > docs/AllRectorsOverview.md
```

.travis.yml
```diff
@@ -26,6 +26,8 @@ script:
   - if [[ $STATIC_ANALYSIS != "" ]]; then composer phpstan; fi
   # Rector demo run
   - if [[ $RUN_RECTOR != "" ]]; then bin/rector process src tests --level symfony40 --dry-run; fi
+  # Rector describe command
+  - if [[ $RUN_RECTOR != "" ]]; then bin/rector describe --level all; fi
```
carusogabriel: Why in Travis?

README.md
```diff
@@ -32,6 +32,8 @@ Rector can:
 - [Add typehints based on new types of parent class or interface](/docs/DynamicRectors.md#remove-a-value-object-and-use-simple-type)
 - And much more...
 
+...just look at overview of [all available Rectors](/docs/AllRectorsOverview.md)
```
carusogabriel: Maybe:
```markdown
All Rectors overview are available in [the documentation](/docs/AllRectorsOverview.md).
```

docs/AllRectorsOverview.md
```diff
@@ -0,0 +1,902 @@
...
+
+```diff
+-$value = new SomeOldClass;
++$value = new SomeNewClass;
+```
+
+
+## Rector\Rector\Contrib\CodeQuality\InArrayAndArrayKeysToArrayKeyExistsRector
+
+Simplify in_array and array_keys functions combination into array_key_exists when array_keys has one argument only
```
carusogabriel: Can we, someway, highlight this function names?
```diff
-Simplify in_array and array_keys functions combination into array_key_exists when array_keys has one argument only
+Simplify `in_array` and `array_keys` functions combination into `array_key_exists` when `array_keys` has one argument only
```

docs/AllRectorsOverview.md
```diff
@@ -0,0 +1,902 @@
...
+```
+
+
+## Rector\Rector\Contrib\Nette\Application\InjectPropertyRector
+
+Turns properties with @inject to private properties and constructor injection
+
+```diff
+/**
+                  * @var SomeService
```
carusogabriel: Unaligned phpDocs

docs/AllRectorsOverview.md
```diff
@@ -0,0 +1,902 @@
...
+## Rector\Rector\Contrib\Symfony\Process\ProcessBuilderInstanceRector
+
+Turns ProcessBuilder::instance() to new ProcessBuilder in Process in Symfony. Part of multi-step Rector.
+
+```diff
+-$processBuilder = Symfony\Component\Process\ProcessBuilder::instance($args);
++$processBuilder = new Symfony\Component\Process\ProcessBuilder($args);
+```
+
+
```
carusogabriel: One unnecessary line between all descriptions

docs/AllRectorsOverview.md
```diff
@@ -0,0 +1,902 @@
...
+## Rector\Rector\Dynamic\AnnotationReplacerRector
+
+[Dynamic] Turns defined annotations above properties and methods to their new values.
+
+```diff
+-/** @test */
++/** @scenario */
+                  public function someMethod() {};
```
carusogabriel: Unaligned function

src/Console/Command/DescribeCommand.php
```diff
@@ -93,21 +129,48 @@ private function describeRectorCodeSamples(array $codeSamples): void
...
+        $rectorDefinition = $rector->getDefinition();
+        if ($rectorDefinition->getDescription()) {
+            $this->consoleStyle->writeln(' * ' . $rectorDefinition->getDescription());
+        }
+
+        if (! $input->getOption(Option::OPTION_NO_DIFFS)) {
+            $this->describeRectorCodeSamples($rectorDefinition->getCodeSamples());
         }
+
+        $this->consoleStyle->newLine(2);
```
carusogabriel: Maybe my complaining about unnecessary lines can be solved here

README.md
TomasVotruba: Well, everything is documentation, and link should be clear about where it goes.

I'd exepect "the documention" go to global "/docs".

.travis.yml
TomasVotruba: The simplest test

docs/AllRectorsOverview.md
TomasVotruba: I thought of that too. Any ideas? 

- 1. manually 
- 2. have a list of words that should be highllighter

src/Console/Command/DescribeCommand.php
TomasVotruba: Of course, I'm used to use 2 lines before new headline, but I can change to standard 1

docs/AllRectorsOverview.md
carusogabriel: Manually, like here in GitHub if we want to highlight something

TomasVotruba: I'll update the definitions then.

src/Console/Command/DescribeCommand.php
```diff
@@ -93,21 +129,48 @@ private function describeRectorCodeSamples(array $codeSamples): void
...
     }
 
-    private function ensureSomeRectorsAreRegistered(): void
+    private function printWithMarkdownFormat(InputInterface $input, RectorInterface $rector): void
```
mssimi: I would decouple `printWithMarkdownFormat` and `printWithCliFormat` from DescribeCommand class. I suggest creating `class MardownFormatter implements Formatter{...}` and `class CliFormatter implements Formatter{...}`.

TomasVotruba: Since it's not needed on multiple places, this would only make maintaining harder.
Not everything that could be abstracted, is good idea to abstract.

PR #446, opened 2018-04-29
Migrate from BetterDocBlockReflection to BetterPhpDocParser - step 1

TomasVotruba:
### Related new Features in Symplify

...that are used here:

- https://github.com/Symplify/Symplify/pull/810
- https://github.com/Symplify/Symplify/pull/809
- https://github.com/Symplify/Symplify/pull/808
- https://github.com/Symplify/Symplify/pull/807
- https://github.com/Symplify/Symplify/pull/806
- https://github.com/Symplify/Symplify/pull/804
- https://github.com/Symplify/Symplify/pull/801
- https://github.com/Symplify/Symplify/pull/796

Another PR to remove ReflectionDocBlock completelly will follow


src/Php/TypeAnalyzer.php
```diff
@@ -8,7 +8,11 @@
 {
     public function isPhpReservedType(string $type): bool
     {
-        return in_array($type, ['string', 'bool', 'mixed', 'object', 'iterable', 'array', 'float', 'int'], true);
+        return in_array(
+            $type,
+            ['string', 'bool', 'null', 'false', 'true', 'mixed', 'object', 'iterable', 'array', 'float', 'int'],
```
mssimi: Isn't boolean missing here as it can be used as phpdoc value as well as bool?

TomasVotruba: Maybe. Where should be add test to cover it and be sure?

mssimi: I guess for uses cases you use in code, it won't fail.

Issue #447, opened 2018-04-29
[BUG] PHPUnit Rectors don't recognize static calls


carusogabriel:
PHPUnit Rectors don't recognize static calls:
```php
self::assertSame(true, $foo);
```
The call above should be refactored.

@TomasVotruba Any clue how to solve this?

TomasVotruba:
Could you send failing test?

PR #448, opened 2018-04-29
Failing test for #447

carusogabriel:
As requested https://github.com/rectorphp/rector/issues/447#issuecomment-385260542

### DIFF
--- a/src/NodeAnalyzer/StaticMethodCallAnalyzer.php
+++ b/src/NodeAnalyzer/StaticMethodCallAnalyzer.php
@@ -40,6 +40,27 @@ final class StaticMethodCallAnalyzer
     }
 
     /**
+     * @param string[] $methods
+     */
+    public function isMethods(Node $node, array $methods): bool
+    {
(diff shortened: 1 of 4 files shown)


TomasVotruba:
Just to be sure we didn't remove a test case, could you keep the old method and just add newline?

carusogabriel:
Why? I can't see anything been losing with this change. If it is the case, we need to have a static call in ever TestCase of PHPUnit Rectors and that'd be a overengineer :(

TomasVotruba:
Because this PR removes a test.
We need to add it only, removing tests doesn't help. Like this:

```diff
  $this->assertFalse(! ($something === false));
  $this->assertTrue(! $something);
+self::assertTrue(! $something);
```

carusogabriel:
@TomasVotruba Sure, I'll work on that, thanks for the tip.

TomasVotruba:
Fixed :+1:

carusogabriel:
@TomasVotruba Thanks

PR #449, opened 2018-04-29
PHAR alternative - build PrefixedRector

mssimi:
It is not complete, but already got working proof of concept:)

I'm taking a little break now that regular expressions killing me:)

rector-prefixer

What is missing:

1. [ ] Get tests working(php.inc is not prefixed)
2. [ ] Fix error found by tests
3. [ ] Git push build dir somewhere to github.


TomasVotruba:
Great job :+1: I can handle regulars, if that's pain for you.

It looks like wrong screenshot, since it replaces classes.

mssimi:
It is not wrong screen I wanted to make simple preview it is already somehow working:D Notice I run that under build directory.

TomasVotruba:
I see. Yesterday we talked about `NamespaceReplacer` and now you're using `ClassReplace`, it just does different thing :).

`NamespacePrefixer` is the goal and the screen doesn't do it.

php-scoper.phar.pubkey
```diff
@@ -0,0 +1,14 @@
+-----BEGIN PUBLIC KEY-----
+MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEAxKi/ABhiCMn7hqH9/W0x
```
TomasVotruba: PHP Scoper should be a dependnecy in `composer.json`

mssimi:
This pull request is not about `NamespacePrefixerRector` this is about PHAR alternative. I made working prefixed rector w/o creating custom Rector for that.

php-scoper.phar.pubkey
mssimi: In docs it is not recommended.

TomasVotruba:
Oh, the image it not for the PR you provided, but for the tool it generated. I get it now

php-scoper.phar.pubkey
TomasVotruba: Still, should be dependency in `composer.json`. Without that we have to maintain phar manually.

What advantages do you see?

mssimi: Conflict avoidance.

TomasVotruba:
@theofidry

> @mssimi what about doing that in PHP-Scoper directly? I think it would be more interesting to have it able to scope a Symfony application out of the box rather than requiring every user to meddle with regexes

What exactly do you mean by this?

theofidry:
I mean instead of trying to modify the YAML service files like [that](https://github.com/rectorphp/rector/pull/449/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R68), a support could be added to PHP-Scoper to be able to change the YAML files directly

TomasVotruba:
I see, that's more clear, thanks. I can look on that later.

At the moment I'd like skip `Rector\*` classes in prefixing, so this can be removed: https://github.com/rectorphp/rector/pull/449/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R69

How to do that?

I found only Whitelist: https://github.com/humbug/php-scoper#whitelist, but it seems only accepts fully class names only

theofidry:
>I found only Whitelist: https://github.com/humbug/php-scoper#whitelist, but it seems only accepts fully class names only

Yes for now. We need to find an easier way to whitelist stuff (see https://github.com/humbug/php-scoper/issues/192#issuecomment-382157399) and extend it to constants and functions as well.

TomasVotruba:
How to do it now?

php-scoper.phar.pubkey
TomasVotruba: I just tried `bamarni/composer-bin-plugin` but it cannot work due to php-parser 3/4 conflict.

image

TomasVotruba: So phar seems as the best option for the time being. It doesn't evolve much and the php-parser 4 + better reflectoin is main blocker there, so missing update-ability won't be such a problem.

PR #452, opened 2018-04-30
Fix AssertPropertyExistsRector

### LINKED ISSUE #420: [BUG] Call toString on null at AssertPropertyExistsRector
The `AssertPropertyExistsRector` needs to be update. It assumes that the first argument of `property_exists` is a class, when the following code is also possible:
```php
$response = new Response();
$response->foo = 'bar';

$this->assertFalse(property_exists($response, 'foo'));
```

mssimi:
Closes #420

### DIFF
--- a/src/Rector/Contrib/PHPUnit/SpecificMethod/AssertPropertyExistsRector.php
+++ b/src/Rector/Contrib/PHPUnit/SpecificMethod/AssertPropertyExistsRector.php
@@ -6,6 +6,7 @@ use PhpParser\Node;
 use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\FuncCall;
 use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Identifier;
 use PhpParser\Node\Name;
 use Rector\Builder\IdentifierRenamer;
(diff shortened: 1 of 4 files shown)


TomasVotruba:
@carusogabriel Could you review and merge it? You have better knowledge of these :)

src/Rector/Contrib/PHPUnit/SpecificMethod/AssertPropertyExistsRector.php
```diff
@@ -73,7 +82,7 @@ public function isCandidate(Node $node): bool
             return false;
         }
 
-        if (! $this->methodCallAnalyzer->isMethods($node, array_keys($this->renameMethodsMap))) {
+        if (! $this->methodCallAnalyzer->isMethods($node, ['assertTrue', 'assertFalse'])) {
```
carusogabriel: We can use either `$renameMethodsWithClassMap` or `$renameMethodsWithObjectMap` here

mssimi: Since there are now two maps, it puts me into position to choose one of those. I find this clearer and more efficient as I don't have to transform keys to array.

TomasVotruba:
Thank you both :+1:

PR #460, opened 2018-05-08
Bump to better-reflection PR running on php-parser 4

TomasVotruba:
Ref https://github.com/Roave/BetterReflection/pull/410


TomasVotruba:
This allows to drop https://github.com/rectorphp/BetterReflection

composer.json
```diff
@@ -13,11 +13,11 @@
         "nette/robot-loader": "^3.0",
         "nette/utils": "^2.5",
         "nikic/php-parser": "^4.0",
-        "rector/better-reflection": "^4.0",
         "sebastian/diff": "^3.0",
         "symfony/console": "^4.0",
         "symfony/dependency-injection": "^4.0",
         "symfony/finder": "^4.0",
+        "roave/better-reflection": "dev-master#31129a702cb15553b7260aca8df16e1d2a181d2b",
```
carusogabriel: We can use 6 digits commit here :smiley:

TomasVotruba: I know, I just could not make it work and tried full hash here.
Later I found, that it was the `composer.lock` who made the trouble.

Will revert :)

I still wait on the stable, if comming any time soon.

Issue #441, opened 2018-04-28
Performance


mssimi:
I did some testing with blackfire.io you can check it [here](https://blackfire.io/profiles/6503441a-e657-4a00-8ed3-90a0e935751e/graph).

It seems like performance bottleneck comes from https://github.com/rectorphp/BetterReflection.

I did some testing and I think we may improve performance by adding caching [here](https://github.com/rectorphp/BetterReflection/blob/368e2698e36e1ba1fe2a5d6c2cba8f33747d7a5c/src/SourceLocator/Type/AbstractSourceLocator.php#L41). I tested in on simple Symfony Form and I noticed it is actually doing same thing 3 times, if we cache it, we should improve performance by 300%.

TomasVotruba:
Why not send PR instead? :)

mssimi:
May be BC break.

carusogabriel:
> May be BC break.

@mssimi Don't worry, Rector isn't even in Alpha, we can break whatever we want :joy:

TomasVotruba:
Closing as unclear intentions

PR #467, opened 2018-05-13
[CodeQuality] Create UnnecessaryTernaryExpressionRector

carusogabriel:
Implements https://github.com/rectorphp/rector/issues/424#issuecomment-383398012

Roadmap:
- [x] Identify ternary expression that can be remove
- [x] Remove unnecessary ternary expressions
- [x] Fix expressions. e.g.:
```diff
-$foo === $bar ? false : true;
+$foo !== $bar;
```

### DIFF
--- a/docs/AllRectorsOverview.md
+++ b/docs/AllRectorsOverview.md
@@ -179,6 +179,15 @@ Simplify `in_array` and `array_keys` functions combination into `array_key_exist
 +array_key_exists("key", $array);
 ```
 
+## Rector\Rector\Contrib\CodeQuality\UnnecessaryTernaryExpressionRector
+
+Remove unnecessary ternary expressions.
+
+```diff
+-$foo === $bar ? true : false;
++$foo === $bar;
+```
+
 ## Rector\Rector\Dynamic\ParentTypehintedArgumentRector
 
 [Dynamic] Changes defined parent class typehints.
--- a/src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
+++ b/src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
@@ -95,20 +95,18 @@ final class UnnecessaryTernaryExpressionRector extends AbstractRector
 
     private function fixBinaryOperation(BinaryOp $operation): BinaryOp
     {
-        $operandsMap = [
-            '===' => NotIdentical::class,
-            '!==' => Identical::class,
-            '==' => NotEqual::class,
-            '!=' => Equal::class,
-            '<>' => Equal::class,
-            '>' => Smaller::class,
-            '<' => Greater::class,
-            '>=' => SmallerOrEqual::class,
-            '<=' => GreaterOrEqual::class,
+        $inverseOperandMap = [
+            Identical::class => NotIdentical::class,
+            NotIdentical::class => Identical::class,
+            Equal::class => NotEqual::class,
+            NotEqual::class => Equal::class,
+            Greater::class => Smaller::class,
+            Smaller::class => Greater::class,
+            GreaterOrEqual::class => SmallerOrEqual::class,
+            SmallerOrEqual::class => GreaterOrEqual::class,
         ];
 
-        $operand = $operation->getOperatorSigil();
-        $binaryOpClassName = $operandsMap[$operand];
+        $binaryOpClassName = $inverseOperandMap[get_class($operation)];
 
         return new $binaryOpClassName($operation->left, $operation->right);
     }
--- a/src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
+++ b/src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
@@ -76,9 +76,7 @@ final class UnnecessaryTernaryExpressionRector extends AbstractRector
         $ifExpression = $ternaryExpression->if;
         $elseExpression = $ternaryExpression->else;
 
-        if (! $ifExpression instanceof ConstFetch
-            || ! $elseExpression instanceof ConstFetch
-        ) {
+        if (! $ifExpression instanceof ConstFetch || ! $elseExpression instanceof ConstFetch) {
             return false;
         }
 
--- a/src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
+++ b/src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
@@ -83,17 +83,17 @@ final class UnnecessaryTernaryExpressionRector extends AbstractRector
      */
     public function refactor(Node $ternaryNode): ?Node
     {
-        /** @var BinaryOp $binaryOpOperation */
-        $binaryOpOperation = $ternaryNode->cond;
+        /** @var BinaryOp $binaryOperation */
+        $binaryOperation = $ternaryNode->cond;
 
         if ($this->ifValue === 'true' && $this->elseValue === 'false') {
-            return $binaryOpOperation;
+            return $binaryOperation;
         }
 
-        return $this->fixBinaryOperation($binaryOpOperation);
+        return $this->inverseBinaryOperation($binaryOperation);
     }
 
-    private function fixBinaryOperation(BinaryOp $operation): BinaryOp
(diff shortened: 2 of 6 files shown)


src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        return $ternaryExpression;
+    }
+
+    private function fixBinaryOperation(BinaryOp $operation): BinaryOp
+    {
+        $operandsMap = [
+            '===' => 'NotIdentical',
+            '!==' => 'Identical',
```
TomasVotruba: Why not use `Identical::class` over weak strings?

src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
```diff
@@ -0,0 +1,109 @@
...
+        $this->ifValue = $ifExpressionName->toLowerString();
+        $this->elseValue = $elseExpressionName->toLowerString();
+
+        return ! in_array('null', [$this->ifValue, $this->elseValue], true);
+    }
+
+    /**
+     * @param Ternary $ternaryExpression
+     */
+    public function refactor(Node $ternaryExpression): ?Node
```
TomasVotruba: It's convention to use ternaryNode here.

tests/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector/Correct/correct.php.inc
```diff
@@ -0,0 +1,40 @@
...
+    public function necessaryTernaryExpression(): bool
+    {
+        return $foo ?: $bar;
+    }
+
+    public function falsePositiveTernaryExpression(): string
+    {
+        return $foo === '' ? null : $foo;
+    }
+}
```
TomasVotruba: This should be split to more standalone files, so it's easier to fix when it will break in the future

src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
carusogabriel: Would import a lot of classes. Is that ok?

carusogabriel:
@tomasvotruba I’ve tried, but couldn’t found anyone to fix. This is good and bad :sweat_smile:

tests/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector/Correct/correct.php.inc
carusogabriel: How should we split?

src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
TomasVotruba: Of course. Why wouldn't be?

What if class name would change to `Identic`. Nor Rector nor PHPStorm would be able to refactor string construction

tests/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector/Correct/correct.php.inc
TomasVotruba: Oh, I see there should be kept untouched.
Let's keep it then as it is

TomasVotruba:
Well, what's the added value of this when no code will invoke it? :)

What about Symfony?

carusogabriel:
> Well, what's the added value of this when no code will invoke it? :)

I've tested in not legacy applications, but remember Rector's philosophy: **Rector - Upgrade your Legacy App to Modern Codebase**

TomasVotruba:
> I've tested in not legacy applications, but remember Rector's philosophy: Rector - Upgrade your Legacy App to Modern Codebase

Well, it should match at least 1 real world case, so we don't create a code that none uses.

> I've put Rector to run against it, but as its speed isn't good (#441), I let you know when finish it smile

That's weird, it usually takes ~10 seconds for 1 rule to finish. Isn't there some performance issue?

carusogabriel:
Ready to merge?

carusogabriel:
> That's weird, it usually takes ~10 seconds for 1 rule to finish. Isn't there some performance issue?

No way that is possible. It takes hours to run on my machine. While PHPStan, for example, take seconds.

TomasVotruba:
Could you share the full code you run?

src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
```diff
@@ -0,0 +1,117 @@
...
+            '!=' => Equal::class,
+            '<>' => Equal::class,
+            '>' => Smaller::class,
+            '<' => Greater::class,
+            '>=' => SmallerOrEqual::class,
+            '<=' => GreaterOrEqual::class,
+        ];
+
+        $operand = $operation->getOperatorSigil();
+        $binaryOpClassName = $operandsMap[$operand];
```
TomasVotruba: What about `<=>`?

carusogabriel:
> Could you share the full code you run?

`vendor/bin/rector process src/Symfony/Component --level code-quality`

src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
carusogabriel: How would you refactory it?

TomasVotruba:
You wrote you run it in on Rectors code. I see, I'll check it when online

src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
TomasVotruba: I'd add test first

TomasVotruba: Any updated on this or ETA?

carusogabriel: I couldn’t understand your point. `<=>` can’t be refactored.

TomasVotruba: Maybe I don't understand this part of code. Could you explain it?

TomasVotruba: Maybe I don't understand this part of code. Could you explain it?

src/Rector/Contrib/CodeQuality/UnnecessaryTernaryExpressionRector.php
```diff
@@ -0,0 +1,117 @@
...
+        return ! in_array('null', [$this->ifValue, $this->elseValue], true);
+    }
+
+    /**
+     * @param Ternary $ternaryNode
+     */
+    public function refactor(Node $ternaryNode): ?Node
+    {
+        /** @var BinaryOp */
+        $binaryOpOperation = $ternaryNode->cond;
```
TomasVotruba: Typo

```diff
-$binaryOpOperation;
+$binaryOperation;
```

TomasVotruba:
I'll look on this and merge it locally.

carusogabriel:
@TomasVotruba Okay, I'm working, this is why I couldn't look into that.

TomasVotruba:
@carusogabriel No problem, I think there is no extra work needed from your side. Just details.

docs/AllRectorsOverview.md
```diff
@@ -179,6 +179,15 @@ Simplify `in_array` and `array_keys` functions combination into `array_key_exist
...
 ```
 
+## Rector\Rector\Contrib\CodeQuality\UnnecessaryTernaryExpressionRector
+
+Remove unnecessary ternary expressions.
+
+```diff
+-$foo === $bar ? true : false;
++$foo === $bar;
+```
```
TomasVotruba: The generate command is super nice. Image we'd update this manually on every new Rector :cactus:

PR #479, opened 2018-05-30
Various NodeTypeResolver and tests improvements

TomasVotruba:
### Related changes

- move from strings "SomeClass" to autoloaded classes `Rector\Tests\ExistingClass` in tests
- drop string assumptions

PR #480, opened 2018-05-31
Decouple Rectors dependency on specific classes

TomasVotruba:
Follow up to #479 

### Goals

- drop custom stubs in packages/BetterReflectoin/stubs
- make run tests against abstraction instead of specific imlementation, e.g. "SomeObject" instead of "Nette\Object"

### Related Changes

- drop Stubber from BetterReflection extension


tests/Rector/Contrib/Nette/Utils/MagicMethodRector/MagicMethodRectorTest.php
```diff
@@ -15,10 +15,6 @@
      */
     public function test(string $wrong, string $fixed): void
     {
-        if (PHP_VERSION >= '7.2.0') {
-            $this->markTestSkipped('This test needs PHP 7.1 or lower.');
-        }
-
```
TomasVotruba: This allows to run all tests on various suites again :parasol_on_ground:

PR #481, opened 2018-05-31
Improve speed by using TypeResolver only when requested

TomasVotruba:


TomasVotruba:
Cherry-picked to master

Issue #503, opened 2018-06-07
Yaml regex experiment


TomasVotruba:
Since more people asked about this, it deserves reopening

Before we though about using parsing/dumping/modifying ast of yaml #17

Maybe it would be enough to use regular instead.
https://www.google.cz/search?q=regex+replace+word+folowed+by+another+word&oq=regex+replace+word+folowed+by+another+word&aqs=chrome..69i57.15559j0j7&client=ms-android-google&sourceid=chrome-mobile&ie=UTF-8

"key:\s+old_subkey" to "key:\s+new_subkey"

TomasVotruba:
- https://github.com/shopsys/shopsys/commit/c035785bccd6abbebbb4a0f1184a4a4e2857e25b#diff-b401dbbf5297ae59e7fbca6571583a14

- https://github.com/symfony/symfony/blob/3.0/UPGRADE-3.0.md#security

### Main cases?

- try preg_replace... on first match stop
- turn to dummy rettors, nothing complex

```yaml
key:
	another_key: old_value

key:
	anothe_key: new_value
```

# 2

```yaml
key:
	old_key: value

key:
	new_key: value
```

TomasVotruba:
Init by #509

PR #509, opened 2018-06-27
Add Yaml refactoring support

TomasVotruba:
Ref #503

## Preview

image


xabbuh:
How does this support the different ways to write YAML? I am thinking of for example the inline notation and literal or folded blocks. On a quick glance it looks like we could end up with invalid YAML here.

TomasVotruba:
This is just prototype. Vission is to support those format to end up with valid YAML file

xabbuh:
You may find some inspiration in the test suite of the Symfony YAML component where we had to deal with such edge cases.

TomasVotruba:
Could link those? Are they relevant to refactoring?

xabbuh:
You could scan the tests in https://github.com/symfony/symfony/tree/master/src/Symfony/Component/Yaml/Tests.

TomasVotruba:
Thanks, I'll get to it after 
https://github.com/rectorphp/rector/issues/514

Issue #514, opened 2018-07-03
Yaml deprecations overview


TomasVotruba:
# Yaml deprecations

### 5.0

- [ ] https://github.com/symfony/symfony/blob/master/UPGRADE-5.0.md#frameworkbundle bundle něco

- [ ] https://github.com/symfony/symfony/blob/master/UPGRADE-5.0.md#securitybundle - [ ] remove paramter

- [ ] https://github.com/symfony/symfony/blob/master/UPGRADE-5.0.md#twigbundle - [ ] default value changed 

### 4.2

- [ ] https://github.com/symfony/symfony/blob/master/UPGRADE-4.2.md#serializer from true to false default

- [ ] https://github.com/symfony/symfony/blob/master/UPGRADE-4.2.md#doctrinebridge remove lazy tags

### 4.1

- [ ] https://github.com/symfony/symfony/blob/master/UPGRADE-4.1.md#frameworkbundle

- [ ] https://github.com/symfony/symfony/blob/master/UPGRADE-4.1.md#securitybundle stateless

- [x] https://github.com/symfony/symfony/blob/master/UPGRADE-4.1.md#twigbundle - [ ] false to debug mode

### 3.4

- [x] https://github.com/symfony/symfony/blob/3.4/UPGRADE-3.4.md#frameworkbundle - [ ] drop defaults

- [x] https://github.com/symfony/symfony/blob/3.4/UPGRADE-3.4.md#yaml B-) remove !!!

### 3.2

- [x] https://github.com/symfony/symfony/blob/3.4/UPGRADE-3.2.md#yaml key followed by :

### 3.1

- [x] https://github.com/symfony/symfony/blob/3.4/UPGRADE-3.1.md#yaml - !!

TomasVotruba:
Closing as not priority ATM

PR #519, opened 2018-07-17
Add Yaml parse nth argument replacement

TomasVotruba:
Ref #514 

See https://github.com/symfony/symfony/blob/3.4/UPGRADE-3.1.md#yaml

Basically merge 2nd, 3rd, 4th and 5th bool, int args to 2nd with flags.

## <strike>`parse()`</strike> done!

image

image

## <strike>`dump()`</strike> done!

image

## Related Changes

- reduce complexity on `Argument*Rector`s from factories + abstract factory and recipes + abstract recipe 
to value objects with static factory; using them before was really cognitive overloading for me


TomasVotruba:
Maybe argument merger?

PR #523, opened 2018-07-23
Fix autoloads, add parameters autobinding

TomasVotruba:


bin/build-prefixed-rector.sh
```diff
@@ -53,7 +53,7 @@ chmod +x build/bin/rector
 
 # clear kernel cache to make use of this new one,
 # #todo? maybe prefix this cache as well?
-(find build/ -type f | xargs sed -i 's/_rector_cache/_prefixed_rector_cache/g')
+(find build/ -type f | xargs sed -i 's#_rector_cache#_prefixed_rector_cache#g')
```
TomasVotruba: This make \/ \/// \ \ \ more readable

PR #528, opened 2018-07-25
Improve test-suite run

TomasVotruba:


TomasVotruba:
Test locally on small set with Blackfire

PR #547, opened 2018-08-01
RectorInterface + DescribeCommand improvements

### LINKED ISSUE #539: Add configuration examples to RectorDefinition
Without any example the user has to check the test to really find out how to configure it.
That's just anoying.

See as in PHP CS Fixer: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/f75251ba3b38e9c22a3e051308ecfae93528a0a7/src/Fixer/ArrayNotation/ArraySyntaxFixer.php#L63

And include it in `DescribeCommand`, follow up to #545

TomasVotruba:
Closes #539

## Related Changes

- add `RectorInterface`, parent to both PHP and YAML Rector interfaces
- change `DescribeCommand` to `GenerateRectorOverviewCommand` that takes care of the docs only
- move Rectors from `Dynamic` namespace to more specific like `Typehint`, `Annotation`, `Namespace_` etc. that's consistent to the rest of Rectors, that are all in own category

## Inspiration

- https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1c10240da97479274fd40a136c3857ff94f7f93f/src/Console/Command/HelpCommand.php#L440

- https://stackoverflow.com/questions/29839194/php-cs-fixer-need-more-information-on-using-fix-level-option


TomasVotruba:
Rather than DescribeCommand might be better GenerateDocsCommand

See https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/8fe9602ba7281a96d0b967c2ae2d53278f818025/src/Console/Command/HelpCommand.php#L440

PR #550, opened 2018-08-03
Fixes + bump to Symplify 4.6

TomasVotruba:


phpstan.neon
```diff
@@ -72,6 +73,9 @@ parameters:
         - '#Constructor of class [\s\S]+ has an unused parameter \$html#' # 2
         - '#Call to an undefined method Nette\\Utils\\Html::add\(\)#' # 1
 
+        # A test class
+        - '#Class SomeClass not found#'
```
carusogabriel: More and more skipped errors in `phpstan.neon`. I don't like it :cry:

TomasVotruba: What about do something about it?

carusogabriel: @TomasVotruba Start to fix them instead of ignoring. If we use PHPStan and ignore the errors, there's no point in use it. I'll prepare a PR eliminating some of them

TomasVotruba:
It might be more effective if you focus on bugs in phostan. There is one with SplFileInfo that produced ~10 new error messages.

phpstan.neon
TomasVotruba: It might be more effective if you focus on bugs in phpstan. There is one with SplFileInfo that produced ~10 new error messages.

PR #562, opened 2018-08-08
[FrameworkBundle] Add test for services alias as well

TomasVotruba:


TomasVotruba:
@mssimi Alias is not the proble. We need more details on this

mssimi:
Example project [here](https://github.com/mssimi/rector-example)

TomasVotruba:
Thanks :+1:

TomasVotruba:
It seem the trait is the problm

mssimi:
We already had trait problem before, but I cannot remember where:D

TomasVotruba:
It was fixed before, but PHPStan kind of skips including traits. I though it's not important if PHPStan doesn't do it, so I didn't add them.

Now I see it IS important, so re-added them :) 

Short story: believe your guts and keep tests 1:1 :D

PR #564, opened 2018-08-09
[FrameworkBundle] Let GetToConstructorInjectionRector work with trait as well

### LINKED ISSUE #560: symfony-constructor-injection.yml does not cover all use-cases
Example of code to refactor which is not refactored:

```php
final class AnyController extends Controller
{
    public function anything()
    {
         $this->get('translator')->trans('something');
    }
}
```
I managed to find out rector ends up [here](https://github.com/rectorphp/rector/blob/02fa25d458d0e98946ed45d02a6df61c381a6c7f/packages/Symfony/src/Bridge/DefaultAnalyzedSymfonyApplicationContainer.php#L48).

TomasVotruba:
Closes #560


TomasVotruba:
@mssimi This should fix it. This the parent trait test case: https://github.com/rectorphp/rector/pull/564/files#diff-e5bdcf0566e5d578d1857ed74d5fd27e

mssimi:
Tested on example I posted, still does nothing.

TomasVotruba:
Thanks for feedback.

Got it :+1: 

image

mssimi:
Actually Interface should be used there. 

```php
\Symfony\Bundle\FrameworkBundle\Translation\Translator

\Symfony\Bundle\FrameworkBundle\Translation\TranslatorInterface
```

TomasVotruba:
Could you add it via PR?
I'm not sure it's compatible or how to determine the proopper Interface to make it always work

mssimi:
Well, if you don't use Interface, exception is thrown.

TomasVotruba:
I see. Could you send fix?

TomasVotruba:
Luckily, it was an easy pick. Your application helped me to find it :)
https://github.com/rectorphp/rector/pull/565

PR #584, opened 2018-08-14
Rector itself - merge isCandidate() to refactor() + add getNodeType()

### LINKED ISSUE #553: Consider changing isCandidate() to Node class subsriber
Since PHP_CodeSniffer, PHP CS Fixer and PHPStan use the same approach: https://github.com/spaze/phpstan-disallowed-calls/blob/master/src/MethodCalls.php

And its not often clear where to put condition and "active..." helper properties are used.

Early returns might work better and also faster, due to only token subscription.

TomasVotruba:
Closes #553


composer.json
```diff
@@ -45,7 +45,8 @@
             "Rector\\Doctrine\\": "packages/Doctrine/src",
             "Rector\\Utils\\": "packages/Utils/src",
             "Rector\\YamlRector\\": "packages/YamlRector/src"
-        }
+        },
+        "classmap": ["custom"]
```
carusogabriel: Why?

custom/MergeIsCandidateRector.php
```diff
@@ -0,0 +1,317 @@
...
+use Rector\Builder\StatementGlue;
+use Rector\NodeTypeResolver\NodeTypeResolver;
+use Rector\NodeTypeResolver\PhpDoc\NodeAnalyzer\DocBlockAnalyzer;
+use Rector\NodeTypeResolver\PHPStan\Type\TypeToStringResolver;
+use Rector\Rector\AbstractRector;
+use Rector\RectorDefinition\CodeSample;
+use Rector\RectorDefinition\RectorDefinition;
+use Rector\Utils\NodeTraverser\CallableNodeTraverser;
+
+final class MergeIsCandidateRector extends AbstractRector
```
carusogabriel: Will we drop it before merge this PR?

packages/PHPUnit/src/Rector/ExceptionAnnotationRector.php
```diff
@@ -73,30 +73,23 @@ public function test()
...
-        return false;
+        return [ClassMethod::class];
     }
 
     /**
      * @param ClassMethod $classMethodNode
      */
     public function refactor(Node $classMethodNode): ?Node
     {
+        if (! $classMethodNode instanceof ClassMethod) {
```
carusogabriel: I don't think we need it. If the refactor method was called, it means we have an instance of `ClassMethod`, like returned in `getNodeTypes`. This should be fixed in all Rectors

src/Rector/AbstractRector.php
```diff
@@ -50,17 +50,24 @@ public function setAbstractRectorDependencies(PropertyAdder $propertyAdder, Expr
...
+//        } elseif (method_exists($this, 'isCandidate')) {
+//            if (! $this->isCandidate($node)) {
+//                return null;
+//            }
+//        }
 
         $newNode = $this->refactor($node);
+
+        // nothing has changed, return original node
         if ($newNode !== null) {
```
carusogabriel: Wrong comment? If `$newNode` is different from null, it means we have a new Node

src/Rector/AbstractRector.php
```diff
@@ -50,17 +50,24 @@ public function setAbstractRectorDependencies(PropertyAdder $propertyAdder, Expr
...
-        } elseif (method_exists($this, 'isCandidate')) {
-            if (! $this->isCandidate($node)) {
-                return null;
-            }
+        $nodeClass = get_class($node);
+
+        if (! in_array($nodeClass, $this->getNodeTypes(), true)) {
+            return null;
         }
+//        if (method_exists($this, 'getNodeTypes')) {
```
carusogabriel: TODO: Remove it and add `getNodeTypes` as an abstract method

composer.json
TomasVotruba: So composer can load `custom/MergeIsCandidateRector.php`.

I'll move it to `/examples` in the end, so we have complex code to show people.

custom/MergeIsCandidateRector.php
TomasVotruba: Yea, I'll just move it to `/examples`, to show what Rector can do.

packages/PHPUnit/src/Rector/ExceptionAnnotationRector.php
TomasVotruba: Yes yes, working on it!

TomasVotruba:
- duplicated old `instanceof` removed
- rector `MergeIsCandidateRector` moved to examples
- added `Rector\Utils\NodeTraverser\CallableNodeTraverser` that helps with inner traversing
- rebased

Ready to merge :+1:

carusogabriel:
@TomasVotruba I'll review this afternoon and I'll merge if everything is all right :+1:

TomasVotruba:
:+1: Thanks @carusogabriel

Issue #619, opened 2018-09-08
Undefined index issues with upgrading a cake plugin


dereuromark:
I was trying to upgrade https://github.com/dereuromark/cakephp-setup to latest 3.6 of CakePHP.

Running the documented command

    vendor/bin/rector process src --level cakephp36 --dry-run

got me

> %PHP Notice:  Undefined index: App\Model\Table\UsersTable in /app.local/vendor/dereuromark/cakephp-setup/vendor/rector/rector/src/Rector/MethodCall/MethodNameReplacerRector.php on line 126

TomasVotruba:
Could you provide more info? Try with `--vvv`. The processed file causing the file if possible.

I'll try it on the application, thanks for the link.

dereuromark:
Most likely your tool cannot work with aliases maybe?
https://github.com/dereuromark/cakephp-tools/blob/master/src/Model/Table/Table.php#L7
is called by  https://github.com/dereuromark/cakephp-setup/blob/master/tests/test_app/src/Model/Table/UsersTable.php#L5

TomasVotruba:
I'll look on during weekend, thanks for provided sources. Very helpful :+1:

TomasVotruba:
I see it now: it just missed isset :)

https://github.com/rectorphp/rector/blob/master/src/Rector/MethodCall/MethodNameReplacerRector.php#L126

I'll fix it when I get to pc

TomasVotruba:
Code can't depends on missing classes. They should be included, even it should be in `require-dev`.

It works the same way as PHPStan, so there have to be either [autolaoded class or excluded file](https://github.com/rectorphp/rector#extra-autoloading). 


```yaml
parameters:
    exclude_paths:
        - '../somePath'
```

I'm just testing the package upgrade it and there actually is demo app code.
It only misses `App\Controller\AppController`.

I'll look on how to make these errors non-blocking.

TomasVotruba:
See: https://github.com/dereuromark/cakephp-setup/pull/15/files

There are no changes reported. Seem the package is up-to-date

I've noticed `config` → `getConfig` change. 
Is there something else that should be changed?

I've fixed few bugs thanks to this testing, please use latest Rector version.

dereuromark:
The deprecations are related to 3.6 upgrade
They should be Cake/Http/ instead of Cake/Network namespace.

TomasVotruba:
Could you share a commit where this namespace was changed?

[...]

PR #627, opened 2018-09-17
make error on process non-blocking

TomasVotruba:
Ref: https://github.com/rectorphp/rector/issues/619#issuecomment-421799657

## Before

before

<br><br><br>

## After

after

What do you think @dereuromark ?

### DIFF
--- a/src/Console/Command/ProcessCommand.php
+++ b/src/Console/Command/ProcessCommand.php
@@ -3,13 +3,14 @@
 namespace Rector\Console\Command;
 
 use Nette\Utils\FileSystem;
+use PHPStan\AnalysedCodeException;
+use Rector\Application\Error;
 use Rector\Application\FileProcessor;
 use Rector\Autoloading\AdditionalAutoloader;
 use Rector\Configuration\Option;
 use Rector\Console\ConsoleStyle;
 use Rector\Console\Output\ProcessCommandReporter;
 use Rector\ConsoleDiffer\DifferAndFormatter;
-use Rector\Exception\Command\FileProcessingException;
 use Rector\FileSystem\FilesFinder;
 use Rector\Guard\RectorGuard;
 use Rector\Reporting\FileDiff;
@@ -81,6 +82,11 @@ final class ProcessCommand extends Command
      * @var RectorGuard
      */
     private $rectorGuard;
+
+    /**
+     * @var Error[]
+     */
+    private $errors = [];
 
     public function __construct(
         FileProcessor $fileProcessor,
@@ -157,6 +163,11 @@ final class ProcessCommand extends Command
         $this->processCommandReporter->reportFileDiffs($this->fileDiffs);
         $this->processCommandReporter->reportChangedFiles($this->changedFiles);
 
+        if ($this->errors) {
+            $this->processCommandReporter->reportErrors($this->errors);
+            return 1;
+        }
+
         if ($input->getOption(Option::OPTION_WITH_STYLE)) {
             $command = sprintf('vendor/bin/ecs check %s --config ecs-after-rector.yml --fix', implode(' ', $source));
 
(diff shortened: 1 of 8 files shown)


dereuromark:
Awesome!

src/Console/Command/ProcessCommand.php
```diff
@@ -157,6 +163,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $this->processCommandReporter->reportFileDiffs($this->fileDiffs);
         $this->processCommandReporter->reportChangedFiles($this->changedFiles);
 
+        if ($this->errors) {
+            $this->processCommandReporter->reportErrors($this->errors);
+            return 1;
```
dereuromark: Instead of this we usually recommend using a speaking const, in cake core we have here

    public const CODE_ERROR = 1;

to make it clear why 1 is here not a succesful one.

src/Console/Output/ProcessCommandReporter.php
```diff
@@ -92,4 +93,25 @@ public function reportFileDiffs(array $fileDiffs): void
             $this->consoleStyle->newLine();
         }
     }
+
+    /**
+     * @param Error[] $errors
```
dereuromark: You dont use FQCN for doc blocks? This is recommended. We in spryker have a autofixing sniff, but I am sure other packages as well.

TomasVotruba: Why?

I prefer unique class names, that are short and clear without namespace.

Namespaces tend programmers to write names like "App\Product\Manager". Then you have 3 classes named "Manager" in no time. IDE autocomple suffers

src/Console/Command/ProcessCommand.php
TomasVotruba: I had it somewhere but dropped after while.
Good idea. Where do you place it in application with many commands?

src/Console/Output/ProcessCommandReporter.php
dereuromark: You need to use sniffing tools to assert it, as we do.
And no, doc blocks must never be linked to code (usage). they must be copy and pastable and still be valid, e.g. in docs and other resources.
And if you use proper tooling about use statement usage you will see the logical error in not using FQCN here. Trust me :)

src/Console/Command/ProcessCommand.php
dereuromark: Usually the base command, to use static::XYZ
I am surprised the framework doesnt offer this, as they should.

Symfony, CakePHP usually all do.

src/Console/Output/ProcessCommandReporter.php
TomasVotruba: I trust you it works for you. This works for me - I don't copy docs blocks, it's only for static analysis.

And of course I use sniff tools :) 

https://github.com/Symplify/Symplify/blob/8a613ea8e691beb543bc4c5060bfe042bbae64b0/packages/CodingStandard/config/symplify.yml#L6-L11

dereuromark: An example of the logical error in your current usage is
https://github.com/spryker/cart-extension/blob/master/src/Spryker/Zed/CartExtension/Dependency/Plugin/CartPreCheckPluginInterface.php
=> `@return \Generated\Shared\Transfer\CartPreCheckResponseTransfer` should not make it necessary in use statements only because of doc block.
Sure, you can maybe add return type hint (here not for BC reasons), but if you e.g. have Type[] collections as array return type it already starts to get really messy.
So again: Highly recommend to rethink that - especially since tooling here can do all the job for you.

TomasVotruba: I still don't get the issue you have.

But I see WTF in duplicated approach - FQN doc but imported namespaced. I'm ok with both, if they're consistent. They'll be moving to typehints (as scalars did) in the future, so future-compat-consistency is a good move from my POV.

dereuromark: different topic :) those are usually still needed for additional info or overwrite 
so it is saner to keep it consistent and keep then. again thanks to tooling no effort.

src/Console/Command/ProcessCommand.php
TomasVotruba: Ref https://github.com/cakephp/cakephp/blob/master/src/Console/Shell.php

dereuromark:
:tada:

PR #639, opened 2018-09-27
AnnotatedPropertyInjectToConstructorInjectionRector issues

bendavies:
Hey there,

First of all, thanks for the amazing project! It's been such a timesaver in upgrading some legacy code which doing manually would have been soul destroying.

I've been trying to use `AnnotatedPropertyInjectToConstructorInjectionRector` to clean up some uses of JMS DI in one of our projects. 
I've encountered a few issues which require manual cleanup after rector has run.
Hopefully these tests highlight the issues I'm facing, but if not, just shout.

I've not had time to look to see if I can fix these myself yet, but I'd imagine I wouldn't be able to do so quickly, so I just thought i get the ball rolling with some 'failing' tests.

Cheers!

### DIFF
--- a/tests/Rector/Architecture/DependencyInjection/AnnotatedPropertyInjectToConstructorInjectionRector/AnnotatedPropertyInjectToConstructorInjectionRectorTest.php
+++ b/tests/Rector/Architecture/DependencyInjection/AnnotatedPropertyInjectToConstructorInjectionRector/AnnotatedPropertyInjectToConstructorInjectionRectorTest.php
@@ -21,6 +21,11 @@ final class AnnotatedPropertyInjectToConstructorInjectionRectorTest extends Abst
     public function provideWrongToFixedFiles(): Iterator
     {
         yield [__DIR__ . '/Wrong/wrong.php.inc', __DIR__ . '/Correct/correct.php.inc'];
+        yield [__DIR__ . '/Wrong/wrong2.php.inc', __DIR__ . '/Correct/correct2.php.inc'];
+        yield [__DIR__ . '/Wrong/wrong3.php.inc', __DIR__ . '/Correct/correct3.php.inc'];
+        yield [__DIR__ . '/Wrong/wrong4.php.inc', __DIR__ . '/Correct/correct4.php.inc'];
+        yield [__DIR__ . '/Wrong/wrong5.php.inc', __DIR__ . '/Correct/correct5.php.inc'];
+        yield [__DIR__ . '/Wrong/wrong6.php.inc', __DIR__ . '/Correct/correct6.php.inc'];
     }
 
     protected function provideConfig(): string
--- a/tests/Rector/Architecture/DependencyInjection/AnnotatedPropertyInjectToConstructorInjectionRector/Correct/correct2.php.inc
+++ b/tests/Rector/Architecture/DependencyInjection/AnnotatedPropertyInjectToConstructorInjectionRector/Correct/correct2.php.inc
@@ -1,19 +1,19 @@
 <?php declare (strict_types=1);
 
-class ClassWithInjects
+class ClassWithInjects2
 {
     /**
      * @var stdClass
+     * @autowired
      */
-    private $property;
+    public $property;
 
     /**
      * @var DateTimeInterface
-     * @inject
      */
-    public $otherProperty;
-    public function __construct(\stdClass $property)
+    private $otherProperty;
+    public function __construct(\DateTimeInterface $otherProperty)
     {
-        $this->property = $property;
+        $this->otherProperty = $otherProperty;
     }
 }
--- a/tests/Rector/Architecture/DependencyInjection/AnnotatedPropertyInjectToConstructorInjectionRector/Correct/correct3.php.inc
+++ b/tests/Rector/Architecture/DependencyInjection/AnnotatedPropertyInjectToConstructorInjectionRector/Correct/correct3.php.inc
@@ -1,0 +1,16 @@
+<?php declare(strict_types=1);
+
+use Rector\Tests\Rector\Architecture\DependencyInjection\AnnotatedPropertyInjectToConstructorInjectionRector\Source\SomeProduct;
+
+class ClassWithInjects3
+{
+    /**
+     * @var SomeProduct
+     */
+    private $property;
+
+    public function __construct(SomeProduct $property)
+    {
+        $this->property = $property;
+    }
+}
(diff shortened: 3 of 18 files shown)


carusogabriel:
Thank you for this failing tests. I've requested @TomasVotruba review, he might know how to fix them

tests/Rector/Architecture/DependencyInjection/AnnotatedPropertyInjectToConstructorInjectionRector/Correct/correct2.php.inc
```diff
@@ -1,19 +1,19 @@
...
      */
-    public $otherProperty;
-    public function __construct(\stdClass $property)
+    private $otherProperty;
+    public function __construct(\DateTimeInterface $otherProperty)
     {
-        $this->property = $property;
+        $this->otherProperty = $otherProperty;
     }
 }
```
TomasVotruba: Why this needs a change?

tests/Rector/Architecture/DependencyInjection/AnnotatedPropertyInjectToConstructorInjectionRector/AnnotatedPropertyInjectToConstructorInjectionRectorTest.php
```diff
@@ -20,7 +20,12 @@ public function test(string $wrong, string $fixed): void
 
     public function provideWrongToFixedFiles(): Iterator
     {
-        yield [__DIR__ . '/Wrong/wrong.php.inc', __DIR__ . '/Correct/correct.php.inc'];
+        yield 'one' => [__DIR__ . '/Wrong/wrong.php.inc', __DIR__ . '/Correct/correct.php.inc'];
+        yield 'two' => [__DIR__ . '/Wrong/wrong2.php.inc', __DIR__ . '/Correct/correct2.php.inc'];
+        yield 'three' => [__DIR__ . '/Wrong/wrong3.php.inc', __DIR__ . '/Correct/correct3.php.inc'];
+        yield 'four' => [__DIR__ . '/Wrong/wrong4.php.inc', __DIR__ . '/Correct/correct4.php.inc'];
+        yield 'five' => [__DIR__ . '/Wrong/wrong5.php.inc', __DIR__ . '/Correct/correct5.php.inc'];
+        yield 'six' => [__DIR__ . '/Wrong/wrong6.php.inc', __DIR__ . '/Correct/correct6.php.inc'];
```
TomasVotruba: Why the key here? The filename is reported, so it's clear which it is.

tests/Rector/Architecture/DependencyInjection/AnnotatedPropertyInjectToConstructorInjectionRector/Correct/correct2.php.inc
bendavies: yeah, sorry, should have been split to a separate commit.

this was an old test file that wasn't being used any more. it tested a the `autowired ` annotation.
https://github.com/rectorphp/rector/commit/01652b1f9e4f6e2594400238dbc5995662a74131
I've corrected the expected result because the current test works on the `inject` annotation.

TomasVotruba:
Hi, thank you for the feedback. I'm glad it works for you :+1: 

Have you configured the `@autowire` annotation?

See docs: https://github.com/rectorphp/rector/blob/master/docs/AllRectorsOverview.md#dependencyinjection

Would expect some other behavior?

tests/Rector/Architecture/DependencyInjection/AnnotatedPropertyInjectToConstructorInjectionRector/AnnotatedPropertyInjectToConstructorInjectionRectorTest.php
bendavies: no reason, just personal preference. I was going to name the properly. happy to revert!

tests/Rector/Architecture/DependencyInjection/AnnotatedPropertyInjectToConstructorInjectionRector/Correct/correct2.php.inc
TomasVotruba: No need for commit, but rather completely new file.

tests/Rector/Architecture/DependencyInjection/AnnotatedPropertyInjectToConstructorInjectionRector/AnnotatedPropertyInjectToConstructorInjectionRectorTest.php
TomasVotruba: Please do so, file names are just fine to locate.

bendavies:
Hi @TomasVotruba, thanks for looking.
I've configured the service correctly.

To summarise, the issue is that the parameter type of the property in the constructor is wrong. It does not use the class of the autowired property, but the parent of the class, or an interface the class is implementing, or even a trait the class is implementing.
The tests should make it clear.

Oh, and also the class in the constructor is full quantified, rather than imported, but that is less of an issue.

tests/Rector/Architecture/DependencyInjection/AnnotatedPropertyInjectToConstructorInjectionRector/Correct/correct2.php.inc
bendavies: For clarity, that test file was not being used.
The test was deleted by the `php.inc` files were not.
https://github.com/rectorphp/rector/commit/8dcc357104af24ef4810045cfee53756b9748778

TomasVotruba: Ah, I see. My bad :)

TomasVotruba:
> It does not use the class of the autowired property, but the parent of the class, or an interface the class is implementing, or even a trait the class is using.

That's what I was missing, thanks I get it now :+1:

TomasVotruba:
Thanks for reporting, very nice tests btw!

I'll look on it :+1:

TomasVotruba:
Fixed via: https://github.com/rectorphp/rector/pull/640 + tagged :heavy_check_mark: 

As for imports, some people prefer FQN over imported ones. So Rector provides FQN by default and let the personal coding standard handle that. [See README for more](https://github.com/rectorphp/rector#how-to-apply-coding-standards).

If you have coding standard, use it.

**But if you want to apply basic coding standard that the most people prefer, just add  `--with-style` to the command.** This set will be applied: https://github.com/rectorphp/rector/blob/master/ecs-after-rector.yml (imports inclueded :))

PR #640, opened 2018-09-27
 Fix AnnotatedPropertyInject Rector type resolving

TomasVotruba:
Fixes #639


bendavies:
Thanks for turning this around so quickly @TomasVotruba ! 💥

PR #641, opened 2018-09-28
Add FileSystemProcessor with first PSR4 splitter Rector

### LINKED ISSUE #630: Split file to PSR-4 classes
E.g. `exceptions.php` to 10 classes named like the class, so it can be loaded

https://github.com/nette/utils/blob/master/src/Utils/exceptions.php

Is that what you've meant @hason? 
I might look on it if noone is volunteering ;)

TomasVotruba:
Closes #630


TomasVotruba:
Ping @hasonm, I think you mentioned need of this in some talk, so just FYI :)

TomasVotruba:
@hason, I think you mentioned need of this in some talk, so just FYI :)

Issue #638, opened 2018-09-27
Add PHP 5.3 to 7.4 upgrades


TomasVotruba:
Mission: develop PHP core faster without letting anyone behind.

# Resources

- [x] https://github.com/php/php-src/blob/PHP-7.2/UPGRADING

- [x] PHP 7.3 Regex regression - https://github.com/rectorphp/rector/pull/847

    ```diff
    -preg_match("#[\w-()]#", 'some text');
    +preg_match("#[\w\-()]#", 'some text');
    ```

- [x] **https://www.oreilly.com/ideas/upgrading-to-php7 - VERY GOOD**
- [x] https://auth0.com/blog/amp/migrating-a-php5-app-to-php7-part-two/

- [x] https://github.com/php/php-src/blob/PHP-7.0/UPGRADING
- [x] https://github.com/php/php-src/blob/PHP-7.1/UPGRADING

- [x] removed `create_function`, [e.g. PR](https://github.com/Automattic/jetpack/pull/8478/files#diff-e854d0045a643c0938d4cb9816afc4f3R273), [PR2](https://github.com/Automattic/jetpack/pull/8870/files)

- [x] Interesting source: https://www.exakat.io/get-ready-php-7-2/
- [x] PHP 7.1 argument count - https://www.reddit.com/r/PHP/comments/a1ie7g/is_there_a_linter_for_argumentcounterror_for_php/

- [x] https://wiki.php.net/rfc/null_coalesce_equal_operator `??=`
- [x] https://www.exakat.io/get-ready-php-7-2/  - function changes + `get_class(null);` - https://github.com/rectorphp/rector/pull/842
- [x] PHP 7.1 typehints to code, same as properties typehints - https://github.com/rectorphp/rector/pull/795
- [x] PHP 7.3 flexible here/now doc - https://github.com/rectorphp/rector/pull/707
- [x] case insensitive constants: 
https://wiki.php.net/rfc/case_insensitive_constant_deprecation, https://github.com/nikic/php-src/blob/fb60f60ad8b459ea635b1dab762bb1c7f8de3977/Zend/tests/case_insensitive_constant_deprecation.phpt  [PHP73]
- [x] removed functions and their alternatives - https://github.com/PHPCompatibility/PHPCompatibility/blob/master/PHPCompatibility/Sniffs/FunctionUse/RemovedFunctionsSniff.php 	
[...]

PR #652, opened 2018-10-04
[Php] Add RandomFunctionRector

TomasVotruba:
Ref #638


TomasVotruba:
Replaced by #653, where build succesfuly passed minutes ago and here it's stuck

PR #667, opened 2018-10-08
[WIP] [PHP] Add EregToPregMatchRector

TomasVotruba:
Ref #638


TomasVotruba:
http://php.net/reference.pcre.pattern.posix

TomasVotruba:
Add also split preg_split

TomasVotruba:
Where they differ in a big way is that the preg functions do not support the wordy character classes that ereg supports. So if you have any eregs that include codes like [:digit:] or [:alpha:] then you'll have more work to convert them. They can still be converted, though.'


Ref: https://stackoverflow.com/a/3769161/1348344

TomasVotruba:
https://gist.github.com/lifthrasiir/704754/7e486f43e62fd1c9d3669330c251f8ca4a59a3f8

Has it all :) 

Maybe test the conberter itself

PR #668, opened 2018-10-09
[Php] Add string array rector

TomasVotruba:
Ref #635

Inspiration: 

https://www.drupal.org/files/issues/2905980-create-l-function-class-array-before-element-assignment-edited.patch


config/level/php/php70.yml
```diff
@@ -11,3 +11,4 @@ services:
     Rector\Php\Rector\List_\ListSwapArrayOrderRector: ~
 
     Rector\Php\Rector\FuncCall\CallUserFuncRector: ~
+    Rector\Php\Rector\FuncCall\EregToPregMatchRector: ~
```
TomasVotruba: Add current rector too
Maybe validation that package rectors are in at one set?

PR #671, opened 2018-10-10
[PHP] Add Add ArrayKeyFirstLastRector

TomasVotruba:


TomasVotruba:
@todo later

- cover more cases with nested expressions
- cover out-of-function/namespace cases

PR #679, opened 2018-10-12
Create SimplifyConditionsRector

carusogabriel:
Ref: https://github.com/rectorphp/rector/issues/424#issuecomment-412317813

### DIFF
--- a/src/Rector/CodeQuality/SimplifyConditionsRector.php
+++ b/src/Rector/CodeQuality/SimplifyConditionsRector.php
@@ -1,0 +1,57 @@
+<?php declare(strict_types=1);
+
+namespace Rector\Rector\CodeQuality;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\BinaryOp;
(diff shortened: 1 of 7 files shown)


src/Rector/CodeQuality/SimplifyConditionsRector.php
```diff
@@ -0,0 +1,71 @@
...
+
+        $binaryOpType = get_class($notNode->expr);
+
+        if (! isset($this->binaryOpClassMap[$binaryOpType])) {
+            return $notNode;
+        }
+
+        $newBinaryOp = $this->binaryOpClassMap[$binaryOpType];
+
+        return new $newBinaryOp($notNode->expr->left, $notNode->expr->right);
```
TomasVotruba: So elegant! Just 9 lines :+1:

carusogabriel:
This rector actually improves performance: https://belineperspectives.com/2016/03/24/php-not-strict-equals-or-strict-not-equals/ :tada:

PR #681, opened 2018-10-13
[CodeQuality] Add SimplifyArraySearchRector

TomasVotruba:
Ref #424

### DIFF
--- a/src/NodeAnalyzer/FuncCallAnalyzer.php
+++ b/src/NodeAnalyzer/FuncCallAnalyzer.php
@@ -1,0 +1,30 @@
+<?php declare(strict_types=1);
+
+namespace Rector\NodeAnalyzer;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Name;
+
+/**
+ * Read-only utils for FuncCall Node:
+ * "someMethod()"
+ */
+final class FuncCallAnalyzer
+{
+    /**
+     * Checks "specificFunction()"
+     */
+    public function isName(Node $node, string $name): bool
+    {
+        if (! $node instanceof FuncCall) {
+            return false;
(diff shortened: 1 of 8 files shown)


src/NodeAnalyzer/FuncCallAnalyzer.php
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Read-only utils for FuncCall Node:
+ * "someMethod()"
+ */
+final class FuncCallAnalyzer
+{
+    /**
+     * Checks "specificFunction()"
+     */
+    public function isName(Node $node, string $name): bool
```
carusogabriel: This is awesome. I’ll provide some refactorings across our Rectors :)

tests/Rector/CodeQuality/SimplifyArraySearchRector/Wrong/wrong.php.inc
```diff
@@ -0,0 +1,17 @@
+<?php declare(strict_types=1);
+
+if (array_search('searching', $array) !== false) {
+    echo 'yes!';
+}
+
+if (array_search('searching', $array) === true) {
```
carusogabriel: This refactoring is actually wrong: `array_search` returns false if didn't find the **index**, otherwise, the index itself. Also, it has a third argument `$strict`. I'll refactor the implementation, as well handle this argument

TomasVotruba: Thanks

src/NodeAnalyzer/FuncCallAnalyzer.php
TomasVotruba: I tried to do that already, but there might be some cases I missed :)

Issue #646, opened 2018-10-02
PHP Recoverable fatal error:  Object of class PhpParser\Node\Expr\Variable could not be converted to string


vasilvestre:
Function called from a variable content don't work.
This is a real case, it's ugly but still a valid case.

```php
$functionName = "dummy$typeName";
$haveToUpdate = $this->$functionName($result);
```

TomasVotruba:
What set/class did cause this?

Full stack trace would be really helpful

vasilvestre:
Oh course !
```
PHP Recoverable fatal error:  Object of class PhpParser\Node\Expr\Variable could not be converted to string in /home/vsilvestre/.composer/vendor/rector/rector/packages/Symfony/src/Rector/HttpKernel/GetRequestRector.php on line 161
```

PR #684, opened 2018-10-13
[CodeQuality] Add SimplifyStrposLowerRector

TomasVotruba:
Ref #424

Also make sure #646 won't happen again


tests/Rector/CodeQuality/SimplifyStrposLowerRector/Wrong/wrong.php.inc
```diff
@@ -0,0 +1,6 @@
+<?php declare(strict_types=1);
+
+strpos(strtolower($string), 'find-me');
+
+//$funcName = 'strpos';
+//$funcName(strtolower($string), 'find-me');
```
TomasVotruba: @todo, try to break things

Related issue: https://github.com/rectorphp/rector/issues/646

Should be resolved in every `(string) FunCall->name` call, probably via `FuncCallAnalyzer`

PR #690, opened 2018-10-14
[CodeQuality] Add missing rectors

carusogabriel:
These new Rector were added in #684, #682, #681, #677 and #676


TomasVotruba:
Nice :+1:

Maybe they could be PSR-4 autodiscovered to avoid forgetting?

carusogabriel:
> Maybe they could be PSR-4 auto-discovered to avoid forgetting?

@TomasVotruba Maybe, but I don't think so. We can discuss a better way to handle it in phpCE week :)

TomasVotruba:
Why not?

PR #698, opened 2018-10-15
[CodeQuality] Add SimplifyFalseComparisonToNegateRector

TomasVotruba:
Ref #424


src/Rector/Architecture/RepositoryAsService/ServiceLocatorToDIRector.php
```diff
@@ -130,7 +130,7 @@ public function refactor(Node $node): ?Node
                 return null;
             }
         }
-        if (! Strings::endsWith($className, 'Repository') === false) {
+        if (Strings::endsWith($className, 'Repository')) {
```
TomasVotruba: These double negative are now not missed :+1:

PR #701, opened 2018-10-16
[CodeQuality] Add SimplifyBinarySwitchRector

TomasVotruba:
Ref #424

### DIFF
--- a/packages/CodeQuality/tests/Rector/Switch_/SimplifyBinarySwitchRector/Correct/correct.php.inc
+++ b/packages/CodeQuality/tests/Rector/Switch_/SimplifyBinarySwitchRector/Correct/correct.php.inc
@@ -1,0 +1,28 @@
+<?php
+
+if ($foo == 'my string') {
+    $result = 'ok';
+} else {
+    $result = 'not ok';
+}
+
+if ($foo == 'my string') {
+    $result = 'ok';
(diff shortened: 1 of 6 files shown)


packages/CodeQuality/tests/Rector/Switch_/SimplifyBinarySwitchRector/Correct/correct.php.inc
```diff
@@ -0,0 +1,28 @@
...
+    $result = 'not ok';
+}
+
+if ($foo == 'my string') {
+    $result = 'ok';
+}
+
+if ($foo == 'my string') {
+    $result = 'ok';
+} elseif ($foo == 'your string') {
```
carusogabriel: A switch to a else/if is totally no go :-1: 

My idea was just a switch with **1** case and **default**.

TomasVotruba: Why?

TomasVotruba: Anyhow, we shall make it at least configurable via ctor options.

To enable/disalbe 2nd case

Issue #700, opened 2018-10-15
AnnotatedPropertyInjectToConstructorInjectionRector cannot handle imported aliased annotations


bendavies:
Hi there,

Consider this example, which is directly pulled from the JMSDIExtra bundle docs:
```php
<?php

use JMS\DiExtraBundle\Annotation as DI;

class Controller
{
    /** @DI\Inject("doctrine.orm.entity_manager") */
    private $em;

    /** @DI\Inject("session") */
    private $session;
}
```

It seems `AnnotatedPropertyInjectToConstructorInjectionRector` can't handle this case. I would expect this to work:

```yaml
services:
  Rector\Rector\Architecture\DependencyInjection\AnnotatedPropertyInjectToConstructorInjectionRector:
    $annotation: JMS\DiExtraBundle\Annotation\Inject
```

Is there a way to make the above example work, without reworking all the injection annotations?

Thanks!

TomasVotruba:
Hi Ben, 

thanks for reporting your idea.

I never actually used this package. What is your expected output? 

"session" refers to a name of services in Symfony container?

bendavies:
hi @TomasVotruba 

Ah yes, they are service names from the symfony di container.
That wasn't the problem i was highlighting, although it would be awesome if it did!

Let me reduce the problem a bit:

```php
<?php

use Doctrine\ORM\EntityManagerInterface;
use JMS\DiExtraBundle\Annotation as DI;

class Controller
{
    /** 
	 * @var EntityManagerInterface
	 * @DI\Inject("doctrine.orm.entity_manager") 
	 */
    private $em;
}
```

So, that will work with me rector config above, as the anotation and processed is not detected.

Does that make sense?
Thanks!

TomasVotruba:
Hi, it makes sense,

thanks for explaining.

The question is - do wee need `("doctrine.orm.entity_manager")` or just `@var` is enough? 

The second would be much easier to fix.

bendavies:
My project has `@var` annotations, so that's fine!

In the long run, it would be great to offer a rector for removing   `JMSDiExtraBundle` from a project!

https://jmsyst.com/bundles/JMSDiExtraBundle/master/usage

TomasVotruba:
Got you, let me see what I can do.

PR #704, opened 2018-10-16
 Add class-based annotation support to AnnotatedPropertyInject... Rector

TomasVotruba:
Ref #700 

@bendavies Check this


bendavies:
that looks like the ticket! top work!

PR #705, opened 2018-10-16
AnnotatedPropertyInjectToConstructorInjectionRector: support inheritance

bendavies:
hi @TomasVotruba 

Having used this rector on a large project, these test should show a situation that I currently have to fix manually after the rector as run.

What do you think about supporting this case? 
Looks tricky, as it creates a dependency chain!

Thanks


TomasVotruba:
Hi Ben,

I see. The point is keeping the parent constructor, right?

That makes sense :+1: 

This is the method responsible for creating a new `__construct` method:
https://github.com/rectorphp/rector/blob/bcd20fa20e0b075a441c6e0e9077689c94861028/src/Builder/ConstructorMethodBuilder.php#L55-L58

To fix that, it should analyze a parent and if it has some dependencies itself, add it too. Is that right?

bendavies:
that sounds right, yeah! but that means the parents must always be processed first which is where i guess the complexity comes in.

TomasVotruba:
Indeed, I just wanted to make sure I get the  problem. 

Would you like to try to contribute this? I'll guide you

bendavies:
hey @TomasVotruba yes! i'd love to help, but i am on vacation from tomorrow. So it won't be for a while.

TomasVotruba:
I see, no troubles.

PR #707, opened 2018-10-17
[PHP] Add SensitiveHereNowDocRector - mbereg function renames

TomasVotruba:
Ref #638


paulinevos:
Hi @TomasVotruba, I ran into this rule at work today and got curious as to why having heredoc/nowdoc wrappers duplicated in the string content is actually considered unsafe... I can't find anything about it online and it's not described in the commit and/or PR description. Do you happen to remember what this was about?

TomasVotruba:
Hey, it's been couple of years, so I'm not 100 % sure. But I think the problem was with downgrades going to PHP 7.2. Rector is heavily used to downgrade packages/tools to make them available for older projects, and this was nasty bug. Of course it be better to include in downgrade rules only, but this is a compability for all solution that we landed.

https://3v4l.org/SXQ4P#veol

paulinevos:
Super fast reply, wow. Thanks for explaining.

PR #712, opened 2018-10-17
[JMS] Add JmsInjectAnnotationRector

### LINKED ISSUE #700: AnnotatedPropertyInjectToConstructorInjectionRector cannot handle imported aliased annotations

TomasVotruba:
Closes #700 

This should allow JMS Inject annotation to constructor, without `@var` type

@bendavies Could you review?


packages/Jms/src/Rector/Property/JmsInjectAnnotationRector.php
```diff
@@ -0,0 +1,128 @@
...
+                    ,
+                    <<<'CODE_SAMPLE'
+use JMS\DiExtraBundle\Annotation as DI;
+
+class SomeController
+{
+    /**
+     * @var EntityManager
+     */
+    private $entityManager;
```
bendavies: missing constructor

packages/Jms/src/Rector/Property/JmsInjectAnnotationRector.php
```diff
@@ -0,0 +1,128 @@
...
+                new CodeSample(
+                    <<<'CODE_SAMPLE'
+use JMS\DiExtraBundle\Annotation as DI;
+
+class SomeController
+{
+    /**
+     * @DI\Inject("entity.manager")
+     */
+    private $entityManager;
```
bendavies: public

packages/Jms/src/Rector/Property/JmsInjectAnnotationRector.php
```diff
@@ -0,0 +1,128 @@
...
+        /** @var PhpDocTagNode $injectTagNode */
+        $injectTagNode = $this->docBlockAnalyzer->getTagByName($node, self::INJECT_ANNOTATION);
+
+        $serviceName = $this->resolveServiceNameFromInjectTag($injectTagNode);
+        if ($serviceName === null) {
+            return null;
+        }
+
+        if (! $this->analyzedApplicationContainer->hasService($serviceName)) {
+            return null;
```
bendavies: can we log an error/warning to the user when this happens?

packages/Jms/tests/Rector/Property/JmsInjectAnnotationRector/JmsInjectAnnotationRectorTest.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Rector\Jms\Tests\Rector\Property\JmsInjectAnnotationRector;
+
+use Iterator;
+use Rector\Testing\PHPUnit\AbstractRectorTestCase;
+
+/**
+ * @covers \Rector\Jms\Rector\Property\JmsInjectAnnotationRector
+ */
+final class JmsInjectAnnotationRectorTest extends AbstractRectorTestCase
```
bendavies: there is `JmsRectorTest`
combine these?

packages/Jms/src/Rector/Property/JmsInjectAnnotationRector.php
TomasVotruba: In previous examples you've send it was private. What is the correct then?

Any docs link I could add to Rector to explain?

packages/Jms/src/Rector/Property/JmsInjectAnnotationRector.php
TomasVotruba: Good point.

It might be collected like this: https://github.com/rectorphp/rector/blob/master/src/Console/Command/ProcessCommand.php#L250

packages/Jms/tests/Rector/Property/JmsInjectAnnotationRector/JmsInjectAnnotationRectorTest.php
TomasVotruba: I think so, since only this one will be used.

Should be here fallback to @var annotation in case of missing named service?
Again, docs would help

packages/Jms/src/Rector/Property/JmsInjectAnnotationRector.php
TomasVotruba: Is it this one? 
https://jmsyst.com/bundles/JMSDiExtraBundle/master/annotations#inject

bendavies: yes, i meant to say, they can be **private**

packages/Jms/src/Rector/Property/JmsInjectAnnotationRector.php
```diff
@@ -0,0 +1,128 @@
...
+
+        /** @var PhpDocTagNode $injectTagNode */
+        $injectTagNode = $this->docBlockAnalyzer->getTagByName($node, self::INJECT_ANNOTATION);
+
+        $serviceName = $this->resolveServiceNameFromInjectTag($injectTagNode);
+        if ($serviceName === null) {
+            return null;
+        }
+
+        if (! $this->analyzedApplicationContainer->hasService($serviceName)) {
```
bendavies: need to handle parameters:
```
@Inject("%kernel.cache_dir%")
```

packages/Jms/src/Rector/Property/JmsInjectAnnotationRector.php
TomasVotruba: Kk, now are ale accepted. Thanks to you and your feedback. 
Before it was only limited to public.

packages/Jms/src/Rector/Property/JmsInjectAnnotationRector.php
TomasVotruba: Well, how would you refactor this? 

It requires modification of service in `services.yml` or changing to `$this->getParamter()`. Both depend on actualy context.

Maybe we should leave this on the user.

packages/Jms/tests/Rector/Property/JmsInjectAnnotationRector/JmsInjectAnnotationRectorTest.php
TomasVotruba: Added

TomasVotruba:
Btw, what happens for this case? I don't get how it works

```
/**
 * @Inject
 */
private $session;
```

bendavies:
> If you do not specify the service explicitly, we will try to guess it based on the name of the property or the parameter.

SO that would try find a service in the container called `session `

packages/Jms/src/Rector/Property/JmsInjectAnnotationRector.php
bendavies: 👍

composer.json
```diff
@@ -224,5 +227,7 @@
     "bin": ["bin/rector"],
     "config": {
         "sort-packages": true
-    }
+    },
+    "minimum-stability": "dev",
```
carusogabriel: This is something we should definitely discuss in person during phpCE :man_facepalming:

PR #715, opened 2018-10-19
Make ConstructorMethodBuilder respect parent constructor

TomasVotruba:


TomasVotruba:
@bendavies It respect parent/child ctors now. Thanks for initial PR :+1: 

Could you briefly review?

bendavies:
Amazing work! You are fast!

TomasVotruba:
> Here, I think you need to iterate up the parents until you find one with a constructor.

Good point, I'll add that!

> Amazing work! You are fast!

:+1:

PR #718, opened 2018-10-21
Fix duplicated property naming

### LINKED ISSUE #610: PHP can't have two same name parameter
Before : (note that the $request in param is useless)
```php
public function fooAction(Request $request) {
    $oRequest = $this->getRequest();
}
```

After : 
```php
public function fooAction(Request $request, Symfony\Component\HttpFoundation\Request $request) {
    $oRequest = $request;
}
```

PHP generate an error about same name in signature, should be check.
I still don't know what behaviour it should have..

TomasVotruba:
Closes #610


packages/Sylius/src/Rector/Review/ReplaceCreateMethodWithoutReviewerRector.php
```diff
@@ -60,15 +60,13 @@ public function refactor(Node $node): ?Node
...
-            && ! $this->methodArgumentAnalyzer->isMethodNthArgumentNull($node, 2)
-        ) {
+        if (isset($node->args[1]) && ! $this->isNull($node->args[1])) {
             return null;
         }
 
         $this->identifierRenamer->renameNode($node, 'createForSubject');
 
-        if ($this->methodArgumentAnalyzer->hasMethodNthArgument($node, 2)) {
+        if (isset($node->args[1])) {
```
TomasVotruba: To keep it consistent with `$node->params`, `$node->stmts`, `$node->args` where keys are form 0. The start with 1 is confusing.

PR #729, opened 2018-10-23
[WIP][CodeQuality] Create SimplifyEmptyArrayCheckRector

carusogabriel:
Ref #424 

Prove that the logic isn't changed: https://3v4l.org/nr0NG

### DIFF
--- a/packages/CodeQuality/src/Rector/FuncCall/SimplifyEmptyArrayCheckRector.php
+++ b/packages/CodeQuality/src/Rector/FuncCall/SimplifyEmptyArrayCheckRector.php
@@ -1,0 +1,88 @@
+<?php declare(strict_types=1);
+
+namespace Rector\CodeQuality\Rector\FuncCall;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Expr\BinaryOp\BooleanAnd;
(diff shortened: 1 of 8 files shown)


TomasVotruba:
The compared value [] should be on right, to be consistent with rest of code

packages/CodeQuality/src/Rector/FuncCall/SimplifyEmptyArrayCheckRector.php
```diff
@@ -0,0 +1,88 @@
...
+    public function isNoneOfPreferableNodes($node): bool
+    {
+        return ! $this->isName($node, 'is_array') && ! $node instanceof Empty_ && ! $node instanceof BooleanNot;
+    }
+
+    /**
+     * @param mixed $leftValue
+     * @param mixed $rightValue
+     */
+    private function isPreferableCheck($leftValue, $rightValue): bool
```
TomasVotruba: `mixed` is very weak typing. This is at least `Node` type, right?

```php
private function isPreferableCheck(Node $leftValueNode, Node $rightValueNode): bool
```

carusogabriel: @TomasVotruba Thanks for the tip, it can :)

Issue #745, opened 2018-10-30
CallUserFuncRector rector is missing


frazjp65:
With a fresh install of rector (v0.3.16), I get the error below when I run `rector process src/ --level php70 --dry-run`. 

>   Class "Rector\Php\Rector\FuncCall\CallUserFuncRector" was not found while loading "vendor/rector/rector/config/level/php/php70.yml" file. Are you sure it is correctly spelled?

frazjp65:
It looks like it broke with ecbf96a9ed9dbd41f62d21fd12b9f47946eea957. They call it "Func", but name it "Method".

TomasVotruba:
Closed by #746

PR #746, opened 2018-10-30
use correct rector name

frazjp65:
For issue #745 : 

The rector name is actually for a method, not a function.


TomasVotruba:
Thank you :+1:

PR #763, opened 2018-11-04
Revert "[CodeQuality] Create SimplifyEmptyArrayCheckRector"

carusogabriel:
This is an invalid refactoring: https://3v4l.org/egtb5 :cry:


carusogabriel:
Actually, some branches are true: https://3v4l.org/HiYXX

Gonna search witch one is invalid to drop

TomasVotruba:
Do not merge anything like this without my check.

Add new test cases instead.

PR #803, opened 2018-11-30
make ECS optional dependency to reduce amount of installed packages

TomasVotruba:
And allow own cs tools

As suggested by @carusogabriel


TomasVotruba:
### Before

```
composer show | wc -l
54
```

### After

```
composer show | wc -l
34
```

Issue #791, opened 2018-11-26
Please add gitattributes


dereuromark:
See https://github.com/cakephp/cakephp/blob/master/.gitattributes

This is useful for
- newline definitions

TomasVotruba:
The better place for that is `.editorconfig`: https://github.com/Symplify/Symplify/blob/master/.editorconfig

dereuromark:
No, spacing is editorconfig, this is about newlines in windows vs linux etc.
Very important to let git know about the type here in use.

dereuromark:
Ah true
but the gitattributes also works directly on checkout, and also envs where there is no IDE plugin for editorconfig available (yet).

TomasVotruba:
I see. Feel free to add it

dereuromark:
Sure, if no one is faster I can do it this week.

dereuromark:
PR open.

PR #836, opened 2018-12-12
Add gitattributes file

dereuromark:
Resolves https://github.com/rectorphp/rector/issues/791

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 /vendor
 /build
 composer.lock
+.idea/
 
 /demo
 rector-symfony.yml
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -5,6 +5,7 @@
 /vendor
 /build
 composer.lock
+.idea/
```
TomasVotruba: Why is this one needed?

dereuromark: As most people use phpstorm, this is just annoying to have not in this file.
Doesnt hurt anyone else so a recommended addition in general.

TomasVotruba: Ok, fine with me :+1:

TomasVotruba:
Thank you @dereuromark

PR #847, opened 2018-12-16
[PHP 7.3] Add RegexDashEscapeRector

TomasVotruba:
Ref #638 

```diff
-preg_match("#[\w-()]#", 'some text');
+preg_match("#[\w\-()]#", 'some text');
```


packages/Php/tests/Rector/FuncCall/RegexDashEscapeRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,33 @@
...
+    preg_match("#[\w()-]#", 'some text'); // ok
+    preg_match("#[-\w()]#", 'some text'); // ok
+    preg_match("#[\w\-()]#", 'some text'); // NOPE!
+    preg_match("#[\w(-)]#", 'some text'); // ok
+
+    preg_match('#[\w()-]#', 'some text'); // ok
+    preg_match('#[-\w()]#', 'some text'); // ok
+    preg_match('#[\w\-()]#', 'some text'); // NOPE!
+    preg_match('#[\w(-)]#', 'some text'); // ok
+}
```
TomasVotruba: @vrana Hi, we talked about regex on Zdroják.

Is this correct PHP 7.3 Preg regression replacement? Only 2 cases with `NOPE!` should be affected

It took me while you mean `-` as connecting 2 signs :)

---

This is the pattern I use:

```php
    /**
     * Matches:
     * [\w-\d]
     *
     * Skips:
     * [-
     * [\-
     * a-z
     * A-Z
     * 0-9
     * (-
     * -]
     * -)
     * @var string
     * @see https://regex101.com/r/6zvPjH/1/
     */
    private const PATTERN_DASH_NOT_AROUND_BRACKETS = '#(?<!\[|\(|A|a|0|\\\\)-(?!\]|\)|A|a|0)#';
```

vrana:
This is too complicated for me so I couldn't answer directly. Basically, `-` must be escaped if it's between character set (e.g. `\w` or `\d`) and something else. Otherwise, it doesn't need to be.

TomasVotruba:
@vrana Thanks, I'll try to simplify it. Where can I learn about this specific rule?

TomasVotruba:
This is the best I could think of.

```php
// left [\D-*]
'#(\\\\(w|s|d))-(?!\])#i';

// right [*-\w]
'#(?<!\[)-\\\\(w|s|d)#i';
```

PR #866, opened 2018-12-17
Adjust docker image name

egobude:
`rector:latest` is not available. You have to use `rector/rector:latest`


TomasVotruba:
Thanks :+1:

PR #898, opened 2018-12-27
Fix typo

Great-Antique:
Btw such changes can be incorrect if you forgot to define constant.


TomasVotruba:
Thanks for the fix :+1: 

I'm working with this example: https://3v4l.org/3qJMG

Could you share incorrect change?

PR #940, opened 2019-01-08
remove NullCoalescingOperatorRector - not part of PHP itself

TomasVotruba:
See

- https://github.com/php/php-src/pull/1795#issuecomment-425748467
- https://github.com/squizlabs/PHP_CodeSniffer/issues/1342#issuecomment-452382695


TomasVotruba:
This can be returned under PHP 7.4

Merged today few mins ago
https://github.com/php/php-src/pull/3747

PR #944, opened 2019-01-11
"bin/bootstrap.php" -> fix autoload

voku:


bin/bootstrap.php
```diff
@@ -23,7 +25,29 @@
...
+    if (!is_dir($vendor_path)) {
+        fwrite(STDERR, "Please check if 'composer.phar install' was run already (expected to find '$vendor_path')\n");
+        exit(1);
+    }
+
+    /** @noinspection PhpIncludeInspection */
+    require $vendor_path . '/autoload.php';
+    return;
+}
+
```
TomasVotruba: This might decopuled to `includeProjectsAutoload()` function, so it's clear what it does.

voku:
yep, this fixed my use-case

TomasVotruba:
Thank you

PR #949, opened 2019-01-12
fix for e.g. negative class const values in CompleteVarDocTypeConstantRector

voku:
... otherwise we get a empty "@var" comment.


TomasVotruba:
Thanks!

What does `[+]` stands for? 
Title of merge request is going to be part of git history, so it should be clear to understand.

TomasVotruba:
I see.

I use convention words like add/remove/fix etc. from [Keep a changelog](https://keepachangelog.com/en/1.0.0/)

Then [ChangelogLinker](https://github.com/symplify/changeloglinker#write-or-dry-run) can generate nice changelogs from Github :)

Thanks for the fix :+1:

PR #951, opened 2019-01-13
add: detect broken php

voku:
... and prevent of creating broken php files. 

-> this is a quick first try... we need better error output, so
   that the user see what is wrong (maybe without the full diff)

### DIFF
--- a/src/Application/RectorApplication.php
+++ b/src/Application/RectorApplication.php
@@ -151,7 +151,16 @@ final class RectorApplication
             if ($this->configuration->isDryRun()) {
                 $newContent = $this->fileProcessor->printToString($fileInfo);
             } else {
-                $newContent = $this->fileProcessor->printToFile($fileInfo);
+                // detect broken php after the "refactoring"
+                $parser = (new \PhpParser\ParserFactory())->create(\PhpParser\ParserFactory::PREFER_PHP7);
+                try {
+                    $parser->parse($this->fileProcessor->printToString($fileInfo));
+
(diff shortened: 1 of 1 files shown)


src/Application/RectorApplication.php
```diff
@@ -151,7 +151,16 @@ private function processFileInfo(SmartFileInfo $fileInfo): void
             if ($this->configuration->isDryRun()) {
                 $newContent = $this->fileProcessor->printToString($fileInfo);
             } else {
-                $newContent = $this->fileProcessor->printToFile($fileInfo);
+                // detect broken php after the "refactoring"
+                $parser = (new \PhpParser\ParserFactory())->create(\PhpParser\ParserFactory::PREFER_PHP7);
```
TomasVotruba: Here you can use `Parser` service.

src/Application/RectorApplication.php
```diff
@@ -151,7 +151,16 @@ private function processFileInfo(SmartFileInfo $fileInfo): void
...
+                try {
+                    $parser->parse($this->fileProcessor->printToString($fileInfo));
+
+                    $newContent = $this->fileProcessor->printToFile($fileInfo);
+                } catch (\PhpParser\Error $error) {
+                    fwrite(STDERR, "Parse error: " . $error->getMessage() . "\n");
```
TomasVotruba: Here exception should be thrown. This error is fatal and should not happen.

TomasVotruba:
Closing for no response

PR #973, opened 2019-01-21
[WIP] Improve docker usage

JanMikes:


TomasVotruba:
What needs to be done here?

JanMikes:
This one is dependent on prefixer and will not work as nicely as it could with it 😢

TomasVotruba:
Kk. 

The Docker (or anything) for prefixed version should be handled in https://github.com/rectorphp/compiler

PR #995, opened 2019-01-31
Add RouterListToControllerAnnotationsRector

TomasVotruba:
From @entrydo project


packages/NetteToSymfony/src/Rector/RouterListToControllerAnnotationsRector.php
```diff
@@ -0,0 +1,365 @@
...
+
+            $getFqnTypeNode = $staticReturnType->getFqnTypeNode();
+            if ($getFqnTypeNode === null) {
+                return null;
+            }
+
+            if ($this->isName($getFqnTypeNode, $this->routeListClass)) {
+                return null;
+            }
+        }
```
TomasVotruba: Type Resolver - add resolve return type method

JanMikes:
@TomasVotruba i found a bug:
```php
/**
 * @Symfony\Component\Routing\Annotation\Route(path="/events/<eventId>", methods={"GET"})
 */
```

`<eventId>`  should be `{eventId}`, symfony parameters format is like that 😄

TomasVotruba:
Let me reged it :)

TomasVotruba:
Issue here: https://github.com/rectorphp/rector/issues/1042

PR #1004, opened 2019-02-02
make MethodNameReplacerRector acccept fnmatch as well

TomasVotruba:
Closes #42

### DIFF
--- a/src/Rector/TypeAnalyzerTrait.php
+++ b/src/Rector/TypeAnalyzerTrait.php
@@ -2,6 +2,7 @@
 
 namespace Rector\Rector;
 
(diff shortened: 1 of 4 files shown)


src/Rector/TypeAnalyzerTrait.php
```diff
@@ -40,6 +41,18 @@ public function autowireTypeAnalyzerDependencies(
     protected function isType(Node $node, string $type): bool
```
JanMikes: So we are not renaming this method, to make it more clear what it does?

TomasVotruba: This is just extension of `$type`input

PR #1018, opened 2019-02-05
Add RenameEventNamesInEventSubscriberRector

TomasVotruba:


packages/NetteToSymfony/src/Rector/ClassMethod/RenameEventNamesInEventSubscriberRector.php
```diff
@@ -0,0 +1,197 @@
...
+        'nette.application.presenter.startup' => ['Symfony\Component\HttpKernel\KernelEvents', 'CONTROLLER'],
+        'nette.application.presenter.shutdown' => ['Symfony\Component\HttpKernel\KernelEvents', 'CONTROLLER'],
+        'nette.application.response' => ['Symfony\Component\HttpKernel\KernelEvents', 'RESPONSE'],
+        'nette.application.error' => ['Symfony\Component\HttpKernel\KernelEvents', 'EXCEPTION'],
+    ];
+
+    /**
+     * @var string[][]
+     */
+    private $netteClassConstToSymfonyClassConstMap = [
```
JanMikes: What about doing it `private static` property or `const`?

packages/NetteToSymfony/src/Rector/ClassMethod/RenameEventNamesInEventSubscriberRector.php
```diff
@@ -0,0 +1,197 @@
...
+     * @param ClassMethod $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        $classNode = $node->getAttribute(Attribute::CLASS_NODE);
+        if ($classNode === null) {
+            return null;
+        }
+
+        if (! $this->isType($classNode, 'Symfony\Component\EventDispatcher\EventSubscriberInterface')) {
```
JanMikes: Just to make sure, does this cover subtypes of this interface? `Interface XyzShit extends EventSubscriberInterface`?

packages/NetteToSymfony/src/Rector/ClassMethod/RenameEventNamesInEventSubscriberRector.php
```diff
@@ -0,0 +1,197 @@
...
+
+        /** @var Return_[] $returnNodes */
+        $returnNodes = $this->betterNodeFinder->findInstanceOf($node, Return_::class);
+
+        foreach ($returnNodes as $returnNode) {
+            if (! $returnNode->expr instanceof Array_) {
+                continue;
+            }
+
+            $this->renameArrayKeys($returnNode);
```
JanMikes: I smell a bug here, that should be covered with test.

There are 2 possible usecases:
```php
public static function getSubscribedEvents(): array
	{
		return [StartupEvent::NAME => 'addHeaders'];
	}
```
and
```php
public static function getSubscribedEvents(): array
	{
		return [StartupEvent::NAME];
	}
```

Both are valid event subscribers. In 1st case, this works, because you rename array keys. For 2nd usecase this will not work.

packages/NetteToSymfony/src/Rector/ClassMethod/RenameEventNamesInEventSubscriberRector.php
TomasVotruba: Yes yes, it does check type of the node. Not just strings

packages/NetteToSymfony/src/Rector/ClassMethod/RenameEventNamesInEventSubscriberRector.php
TomasVotruba: Resolved in person

JanMikes:
Big up 👍

PR #1027, opened 2019-02-09
Fixed Contributte events namespaces and added missing

JanMikes:


JanMikes:
```php
public static function getSubscribedEvents(): array
    {
        return [\Symfony\Component\HttpKernel\KernelEvents::CONTROLLER => 'validateRequestBody'];
    }

    public function validateRequestBody(PresenterEvent $event): void
    {
```

It might replace event as argument for given method as well, but I do not know how complicated it would be and it anyways always needs manual revision, because event has different parameters so I am not sure if it is worth it.

TomasVotruba:
> It might replace event as argument for given method as well, but I do not know how complicated it would be and it anyways always needs manual revision, because event has different parameters so I am not sure if it is worth it.

:+1: Agreed. 

It is worth it, just give me the diff :)

TomasVotruba:
Could you fix CI? Feel free to merge after it passes :+1:

TomasVotruba:
CI fail looks unrelated, I'll check it

JanMikes:
@TomasVotruba for inspiration here are some:

screenshot 2019-02-09 14 06 20

screenshot 2019-02-09 14 07 44

screenshot 2019-02-09 14 08 21

I still found it useful to do this one manually, it was pretty straightforward.

TomasVotruba:
Good. I need FQN names of those event objects and related event constant.

JanMikes:
It might use class replacer rector, right?

because in fact it is `someMethod(SomeOldClass $event)` -> `someMethod(SomeNewClass $event)`

Then it might be hard to get replacement for methods but as well for that we have rectors so basically it is only about writing config if i understand it?

JanMikes:
Is this you have been looking for: https://symfony.com/doc/current/components/http_kernel.html#creating-an-event-listener ?

TomasVotruba:
> Is this you have been looking for: [symfony.com/doc/current/components/http_kernel.html#creating-an-event-listener](https://symfony.com/doc/current/components/http_kernel.html#creating-an-event-listener) ?

Yes, and original Contribute object names.

TomasVotruba:
> It might use class replacer rector, right?
>
>because in fact it is `someMethod(SomeOldClass $event)` -> `someMethod(SomeNewClass $event)`
>
>Then it might be hard to get replacement for methods but as well for that we have >rectors so basically it is only about writing config if i understand it?

Yes :+1: then if method names are different, to change them as well. That might require custom Rector for now.

PR #1039, opened 2019-02-11
Fix extra commas in EventInfosFactory

JanMikes:


vasilvestre:
https://github.com/rectorphp/rector/issues/1040
Extra comma in function call is valid with 7.3 but composer requirement is 7.2

TomasVotruba:
> In addition, you could add no_trailing_comma_in_list_call at php cs fixer before 7.3 :)

Does is separate list and functions args?

vasilvestre:
no_trailing_comma_in_list_call
Remove trailing commas in list function calls.

I don't know much more than what doc is telling

TomasVotruba:
I just tried it on this case and it ignored it.

Issue #1040, opened 2019-02-11
[Docker] unexpected ')' in /rector/packages


vasilvestre:
Hello, here's an error that pop off after running this command `docker run -v $(pwd):/project rector/rector:latest`

```
Parse error: syntax error, unexpected ')' in /rector/packages/NetteToSymfony/src/Event/EventInfosFactory.php on line 24
```

TomasVotruba:
Resolved by #1039

PR #1041, opened 2019-02-11
Fix file by removing extra unallowed trailing

vasilvestre:
https://github.com/rectorphp/rector/issues/1040

Trailing comma in function call [is valid with PHP7.3](https://wiki.php.net/rfc/trailing-comma-function-calls)

Actual minimum php requirement is 7.2 so this is not consistent.


vasilvestre:
There's already one open...
I'm closing it.

PR #1086, opened 2019-02-21
SensitiveConstantNameRector: ignore existing contants

jkuchar:
case-sensitivity test of `defined()`: https://3v4l.org/DA23r


TomasVotruba:
Thank you :+1:

Issue #1127, opened 2019-02-27
TypeError in ValueResolver for class constants that have an int value


cgkkevinr:
Hello

Rector throws an exception when it has to inspect a constant with an integer value.

Should be reproducible with this class:

```php
<?php

namespace App\Command;

use Symfony\Component\Console\Command\Command;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;

class TestCommand extends Command
{
    const TEST = 1;
    const FOOBAR = 2;

    protected static $defaultName = 'TestCommand';

    protected function execute(InputInterface $input, OutputInterface $output)
    {
        echo $this->returnSomething('down');
    }

    public function returnSomething(string $compareMe)
    {
        if ($compareMe === 'foobar') {
            return self::TEST;
        } else {
            return self::FOOBAR;
        }
    }
}
```

Result:

```
root@36fa5d845905:/app# vendor/bin/rector process src/App/Command/TestCommand.php --level=php70 --dry-run
Rector v0.4.3

 3/3 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%


 [ERROR] Could not process "src/App/Command/TestCommand.php" file, due to:
         "Return value of Rector\PhpParser\Node\Value\ValueResolver::resolveClassConstFetch() must be of the type string, integer returned".
```

Locally overwriting this line https://github.com/rectorphp/rector/blob/master/src/PhpParser/Node/Value/ValueResolver.php#L123 to cast to string resolved the issue.

TomasVotruba:
Thanks for complete description of issue.

The `resolveClassConstFetch` return type should be mixed, in docblock. Could you send PR for that?

cgkkevinr:
Hi, sure. PR #1128 created.

PR #1128, opened 2019-02-27
Change type hint for resolveClassConstFetch to mixed

cgkkevinr:
Rector throws an exception when it has to inspect a constant with an integer value. See: #1127.

Signed-off-by: Kevin R <22906111+cgkkevinr@users.noreply.github.com>

### DIFF
--- a/src/PhpParser/Node/Value/ValueResolver.php
+++ b/src/PhpParser/Node/Value/ValueResolver.php
@@ -2,7 +2,6 @@
 
 namespace Rector\PhpParser\Node\Value;
 
-use PhpParser\ConstExprEvaluationException;
 use PhpParser\ConstExprEvaluator;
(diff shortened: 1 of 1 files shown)


src/PhpParser/Node/Value/ValueResolver.php
```diff
@@ -92,7 +93,12 @@ private function resolveFileConstant(File $file): string
         return $fileInfo->getPathname();
     }
 
-    private function resolveClassConstFetch(ClassConstFetch $classConstFetch): string
+    /**
+     * @return mixed
+     * @throws ShouldNotHappenException
+     * @throws ConstExprEvaluationException
```
TomasVotruba: We don't use `@throws` tags

TomasVotruba:
Thank you :clap:

PR #1142, opened 2019-03-03
[Laravel 5.8] Add MinutesToSecondsInCacheRector + misc

TomasVotruba:
Ref https://github.com/laravel/framework/pull/27276/files

### DIFF
--- a/packages/Laravel/src/Rector/StaticCall/MinutesToSecondsInCacheRector.php
+++ b/packages/Laravel/src/Rector/StaticCall/MinutesToSecondsInCacheRector.php
@@ -1,0 +1,115 @@
+<?php declare(strict_types=1);
+
+namespace Rector\Laravel\Rector\StaticCall;
+
+use PhpParser\Node;
+use PhpParser\Node\Arg;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\BinaryOp\Mul;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Scalar\LNumber;
+use Rector\Rector\AbstractRector;
+use Rector\RectorDefinition\CodeSample;
+use Rector\RectorDefinition\RectorDefinition;
+
+/**
+ * @see https://github.com/laravel/framework/pull/27276
+ */
+final class MinutesToSecondsInCacheRector extends AbstractRector
+{
+    /**
+     * @var int[][]
+     */
+    private $typesToMethods = [
+        'Illuminate\Support\Facades\Cache' => [
+            'put' => 2, // time argument position
+            'add' => 2,
+        ],
+        'Illuminate\Contracts\Cache\Store' => [
+            'put' => 2,
+            'putMany' => 1,
+        ],
+        'Illuminate\Cache\DynamoDbStore' => [
+            'add' => 2,
+        ],
+    ];
+
+    public function getDefinition(): RectorDefinition
+    {
+        return new RectorDefinition(
+            'Change minutes argument to seconds in Illuminate\Contracts\Cache\Store and Illuminate\Support\Facades\Cache',
+            [
+                new CodeSample(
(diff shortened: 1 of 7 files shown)


TomasVotruba:
cc @barryvdh I think this is what you needed :)

Could you give me code-review and complete more cases if needed?

barryvdh:
My preference would go to actually using the Datetime instance, but maybe that can be configurable? https://twitter.com/laravelshift/status/1097514254016696323

packages/Laravel/src/Rector/StaticCall/MinutesToSecondsInCacheRector.php
```diff
@@ -0,0 +1,110 @@
...
+        return new RectorDefinition(
+            'Change minutes argument to seconds in Illuminate\Contracts\Cache\Store and Illuminate\Support\Facades',
+            [
+                new CodeSample(
+                    <<<'CODE_SAMPLE'
+class SomeClass
+{
+    public function run()
+    {
+        Illuminate\Support\Facades::put('key', 'value', 60);
```
barryvdh: Missing 'Cache' here

packages/Laravel/src/Rector/StaticCall/MinutesToSecondsInCacheRector.php
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+CODE_SAMPLE
+                    ,
+                    <<<'CODE_SAMPLE'
+class SomeClass
+{
+    public function run()
+    {
+        Illuminate\Support\Facades::put('key', 'value', 3600);
```
barryvdh: You actually modifying it to 60*60, right?

packages/Laravel/src/Rector/StaticCall/MinutesToSecondsInCacheRector.php
```diff
@@ -0,0 +1,110 @@
...
+
+        return $node;
+    }
+
+    private function processArgumentPosition(StaticCall $staticCall, int $argumentPosition): StaticCall
+    {
+        if ($staticCall->args[$argumentPosition]->value instanceof LNumber) {
+            // number
+            $value = $this->getValue($staticCall->args[$argumentPosition]->value);
+            if ($value === 60) {
```
barryvdh: Why only 60?

barryvdh:
Although when you use * 60 it's pretty clear.

packages/Laravel/src/Rector/StaticCall/MinutesToSecondsInCacheRector.php
TomasVotruba: Thanks :+1:

packages/Laravel/src/Rector/StaticCall/MinutesToSecondsInCacheRector.php
TomasVotruba: Just first idea, no reason really :) fixed

packages/Laravel/src/Rector/StaticCall/MinutesToSecondsInCacheRector.php
TomasVotruba: Yes, later I conclude it's more readable. 

I forgot to fix it here, thanks :+1:

TomasVotruba:
> My preference would go to actually using the Datetime instance, but maybe that can be configurable? twitter.com/laravelshift/status/1097514254016696323

That should be handled by another rule, so it's consisten through whole code, like `laravel-coding-style`.

barryvdh:
Yeah perhaps you're right. The main issue was that it wasn't totally clear if it's converted or not, but with * 60 that isn't an issue. So if this is in the upgrade  set it's fine, and indeed a code style set would be better for 5.8+

TomasVotruba:
Ready to merge for me

barryvdh:
Looks good to me too :) Thanks!

Issue #1138, opened 2019-03-02
[Laravel DI] Create static to constructor injection level


TomasVotruba:
## Source

- https://medium.freecodecamp.org/moving-away-from-magic-or-why-i-dont-want-to-use-laravel-anymore-2ce098c979bd by @mrpixeldream
- https://laravel.com/docs/5.7/facades#facades-vs-dependency-injection
- https://gist.github.com/barryvdh/bb6ffc5d11e0a75dba67
- https://github.com/laravel/framework/blob/e32e92706fe3ba26aef496d23a452750a4653725/src/Illuminate/Foundation/helpers.php#L545

TomasVotruba:
@mrpixeldream Hi, thanks for awesome post! :+1: 

Could you please help me with `Request::*` to h$request->*` migration?
I didn't find `Reuqest::validate()` method anywhere in Laravel code (I don't use Laravel).

Diff with FQN names would be the best

mrpixeldream:
Thank you very much for using my post as an inspiration for your project. Nice work there!

It's registered as a "macro" in the `FoundationServiceProvider`: https://github.com/laravel/framework/blob/5.8/src/Illuminate/Foundation/Providers/FoundationServiceProvider.php#L50

Maybe that helps you already. If not or if you need anything else, let me know and I'll check tomorrow. Heading to bed now ✌️

barryvdh:
About the Example in PR #1140; DI can be used on both the constructor and the method itself. Wouldn't it make more sense to add the instances to the method itself (eg. `store`, `index`), because otherwise all dependencies need to be resolved for each method?

TomasVotruba:
@barryvdh I understand. That could be again resolved in standalone Rector.

TomasVotruba:
> It's registered as a "macro" in the FoundationServiceProvider: laravel/framework:src/Illuminate/Foundation/Providers/FoundationServiceProvider.php@5.8#L50

I would never figure that out, thank you :+1: 

> Thank you very much for using my post as an inspiration for your project. Nice work there!

It's easy with great input and carefully selected words, great job to you sir.

> Heading to bed now 

Of course :+1:

mrpixeldream:
> I would never figure that out, thank you 👍

That's basically why I wrote the article, haha. It makes it easy to use but as soon as you try to figure out what's going on, it screws you over.

> Wouldn't it make more sense to add the instances to the method itself (eg. store, index), because otherwise all dependencies need to be resolved for each method?

[...]

Issue #290, opened 2018-01-14
[Laravel] Add Laravel Rectors


carusogabriel:
Following [Symfony migration on Rector](https://github.com/rectorphp/rector/tree/master/src/config/level/symfony), I guess we can have that automation as well for [Laravel](https://laravel.com/). List of versions:
- [ ] [5.0 to 5.1](https://laravel.com/docs/5.1/upgrade#upgrade-5.1.0)
- [ ] [5.1 to 5.2](https://laravel.com/docs/5.2/upgrade#upgrade-5.2.0)
- [ ] [5.2 to 5.3](https://laravel.com/docs/5.3/upgrade#upgrade-5.3.0)
- [ ] [5.3 to 5.4](https://laravel.com/docs/5.4/upgrade#upgrade-5.4.0)
- [ ] [5.4 to 5.5](https://laravel.com/docs/5.5/upgrade#upgrade-5.5.0)

Also, make use of `CHANGELOG-5.x.md` files in [Framework's core](https://github.com/laravel/framework). Some useful and undocumented changes may be found there :man_technologist: 

---

EDIT by @TomasVotruba 

## Resources

- [x] https://laravel.com/docs/5.8/upgrade
- [ ] https://laravel.com/docs/5.7/upgrade
- [ ] https://laravel.com/docs/5.6/upgrade
- [ ] https://laravel.com/docs/5.5/upgrade
- [ ] https://laravel.com/docs/5.4/upgrade
- [ ] https://laravel.com/docs/5.3/upgrade
- [ ] https://laravel.com/docs/5.2/upgrade

carusogabriel:
Closing as Laravel migrations usually adds new folders struct, new helper methods, and bugfixes. Out of scope for Rector now.

TomasVotruba:
Also handled better by https://laravelshift.com

We can now focus in improving Symfony and architecture Rectors

TomasVotruba:
- maybe make a demo list, just to make this really free:

- also a post https://medium.com/@taylorotwell/laravel-blog-contest-may-mayhem-edition-1aa9f8ac25a9

TomasVotruba:
In the meantime, looks like the author of Laravel Shifts is taking care of the products: https://medium.com/@jasonmccreary/laravel-by-the-numbers-262a1cdd9c72

I really like it and price seems affordable. I also don't want decrease somebody's workin business just because I can.

Let's make focus on projects that do not have such a service - Symfony, Guzzle, Doctrine... first

[...]

PR #1143, opened 2019-03-04
[Laravel] Add RequestStaticValidateToInjectRector

TomasVotruba:
Ref #1138 ,
ref #290

```diff
  use Illuminate\Http\Request;
	  
  class SomeClass
  {
-     public function store()
+     public function store(\Illuminate\Http\Request $request)
      {
-         $validatedData = Request::validate(['some_attribute' => 'required']);
+         $validatedData = $request->validate(['some_attribute' => 'required']);
      }
  }
```

### DIFF
--- a/packages/Laravel/tests/Rector/StaticCall/RequestStaticValidateToInjectRector/Fixture/fixture.php.inc
+++ b/packages/Laravel/tests/Rector/StaticCall/RequestStaticValidateToInjectRector/Fixture/fixture.php.inc
@@ -1,0 +1,31 @@
+<?php
+
+namespace Rector\Laravel\Tests\Rector\StaticCall\RequestStaticValidateToInjectRector\Fixture;
+
+use Illuminate\Http\Request;
+
+class SomeClass
+{
+    public function store()
+    {
+        $validatedData = Request::validate(['some_attribute' => 'required']);
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\Laravel\Tests\Rector\StaticCall\RequestStaticValidateToInjectRector\Fixture;
+
+use Illuminate\Http\Request;
+
+class SomeClass
+{
+    public function store(\Illuminate\Http\Request $request)
+    {
+        $validatedData = $request->validate(['some_attribute' => 'required']);
+    }
+}
+
+?>
--- a/config/level/laravel/laravel-static-to-injection.yaml
+++ b/config/level/laravel/laravel-static-to-injection.yaml
@@ -1,2 +1,3 @@
 services:
     Rector\Laravel\Rector\StaticCall\FacadeStaticCallToConstructorInjectionRector: ~
+    Rector\Laravel\Rector\StaticCall\RequestStaticValidateToInjectRector: ~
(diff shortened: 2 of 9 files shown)


TomasVotruba:
@mrpixeldream Could I ask you for a review? I hope I got that right :sweat_smile:

barryvdh:
Why just the validate? Could also be for the other Requests calls, right?

TomasVotruba:
Sure. Again, I don't use Laravel so I have no idea about the logic here :).

If you provide list of these magic static methods, that can be changed here, we can update it

barryvdh:
Ah my bad, I guess usually it's the Request facade, not the Request class itself that is called static.
I usually prefer to inject the Request class like this ($request in the controller method), but that's already done by the other call (all facades to injection in __construct). But I would say that the Request object belongs to the current controller action usually, so it would belong there. But could be a matter of preference.

TomasVotruba:
I'm not sure I understand. It's in controller action.

See description of this PR:

image

barryvdh:
This one is, but not the Facade one (that is done in https://github.com/rectorphp/rector/pull/1140)

TomasVotruba:
Why do you talk about differnt PR here?

TomasVotruba:
Mergin for `validate` method as it is now

mrpixeldream:
Yeah, seems fine 👍   
If you want, you can check out https://github.com/laravel/framework/blob/5.8/src/Illuminate/Support/Facades/Request.php for a list of methods that are usable on the facade and hence also on the request object. The docblock shows all of them. `Request::get()` is frequently used as well for example. I am of course open to reviewing if you want that, just wait a little longer :P

barryvdh:
Because people usually call the Request facade, not the actual class. eg. `Request::get()` is usually an alias for `Illuminate\Support\Facades\Request`, not for `\Illuminate\Http\Request`. So those calls are already handled by the other PR (which handles all facades)

TomasVotruba:
The code-review would be best :+1;

packages/Laravel/tests/Rector/StaticCall/RequestStaticValidateToInjectRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,31 @@
...
+-----
+<?php
+
+namespace Rector\Laravel\Tests\Rector\StaticCall\RequestStaticValidateToInjectRector\Fixture;
+
+use Illuminate\Http\Request;
+
+class SomeClass
+{
+    public function store(\Illuminate\Http\Request $request)
```
mrpixeldream: You are `use`-ing the class above already, no need to fully specify it here.

TomasVotruba: This is job of coding standard. Rector is doing the hard work, ECS makes it nice again :)

See https://github.com/rectorphp/rector#how-to-apply-coding-standards

mrpixeldream: Ah yeah, alright. Everything's fine then I guess 👍

PR #1148, opened 2019-03-04
Add Shopware Rector config

shyim:
I am new to PHP-Parser. Feel free for ideas to improve 🙈

### DIFF
--- a/config/level/shopware/shopware55.yaml
+++ b/config/level/shopware/shopware55.yaml
@@ -1,3 +1,4 @@
+# See https://github.com/shopware/shopware/blob/5.5/UPGRADE-5.5.md
 services:
     Rector\Rector\MagicDisclosure\GetAndSetToMethodCallRector:
         Enlight_View_Default:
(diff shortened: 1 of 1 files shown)


config/level/shopware/shopware55.yaml
```diff
@@ -0,0 +1,18 @@
+services:
```
TomasVotruba: Could you link some source of upgrade path?

E.g. link to `UPGRADE.md`, `changelog.md`, post on blog...

```yaml
# see: https://....file.md
services:
   ...
```

That way it's easier to find out more about the project and complete the list.

TomasVotruba:
> I am new to PHP-Parser. Feel free for ideas to improve see_no_evil

You learn fast, it looks good :+1:

config/level/shopware/shopware55.yaml
shyim: I added the Upgrade.md for the developers. But the part with Magic Calls is currently not documented. We want to get rid of those calls in the future :)

PR #1149, opened 2019-03-04
Fix GetAndSetToMethodCallRector in class itself

shyim:
**Before the Fix**

Before
```php
class Enlight_View_Default
{
    protected $engine;
    public function __construct($engine)
    {
        $this->engine = $engine;
    }
}
```

After
```php
class Enlight_View_Default
{
    protected $engine;
    public function __construct($engine)
    {
        $this->addService('engine', $engine);
    }
}
```

### DIFF
--- a/tests/Rector/MagicDisclosure/GetAndSetToMethodCallRector/Fixture/fixture3.php.inc
+++ b/tests/Rector/MagicDisclosure/GetAndSetToMethodCallRector/Fixture/fixture3.php.inc
@@ -1,0 +1,25 @@
+<?php
+
+class Enlight_View_Default
+{
+    protected $engine;
+
+    public function __construct($engine)
+    {
+        $this->engine = $engine;
+    }
+}
+?>
+-----
+<?php
+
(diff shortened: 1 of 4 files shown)


TomasVotruba:
Great work, thanks for the test :+1:

tests/Rector/MagicDisclosure/GetAndSetToMethodCallRector/Fixture/fixture3.php.inc
```diff
@@ -0,0 +1,25 @@
...
+class Enlight_View_Default
+{
+    protected $engine;
+
+    public function __construct($engine)
+    {
+        $this->engine = $engine;
+    }
+}
+?>
```
TomasVotruba: If the code should remain the same after Rector run, just keep the first part. 

Remove the code after `-----`

shyim:
@TomasVotruba Now the tests are failing or i missed something?

TomasVotruba:
Thank you :+1:

PR #1153, opened 2019-03-05
Add annotation replacement to phpunit80

sebastianfeldmann:
The annotation replacement is part of the phpunit70 process but the
'expectedException*' annotations are an actual BC-break in PHPUnit 8.0.

I copied the annotation replacement configuration from `phpunit70.yml` over to `phpunit80.yml`


TomasVotruba:
This can be a solution. 

But what about duplication? What if some level will be fixed and the other not?

sebastianfeldmann:
So what you are saying is that phpunit70 should be executed anyway even if you are already using PHPUnit 7

TomasVotruba:
Yes, and also that levels are based on added new-features. Not when it's too late - fatal error, missing class.

The problems with PHPUnit deprecations are kinda random, e.g. https://thephp.cc/news/2016/02/questioning-phpunit-best-practices

Something was removed, something was re-added, somethign was only written about.

PR #1170, opened 2019-03-09
[PSR4] Add UniteFileAndClassNameRector

### LINKED ISSUE #1154: [FileNames] Renaming file name (or class name) according to PSR-4
Hi! Check following problem: `TestParallel.php` and class is named `TestParallelPresenter`:   
screenshot 2019-03-06 13 20 25

I found https://github.com/rectorphp/rector/issues/630, but I do not know if this can be used as well for this?

TomasVotruba:
Closes #1154

### DIFF
--- a/src/Rector/Psr4/UniteFileAndClassNameRector.php
+++ b/src/Rector/Psr4/UniteFileAndClassNameRector.php
@@ -1,0 +1,111 @@
+<?php declare(strict_types=1);
+
+namespace Rector\Rector\Psr4;
+
+use PhpParser\Node;
+use PhpParser\Node\Stmt\ClassLike;
+use Rector\FileSystemRector\Contract\FileSystemRectorInterface;
+use Rector\PhpParser\Node\BetterNodeFinder;
+use Rector\PhpParser\Node\Resolver\NameResolver;
+use Rector\PhpParser\Parser\Parser;
+use Rector\RectorDefinition\CodeSample;
+use Rector\RectorDefinition\RectorDefinition;
+use Symfony\Component\Filesystem\Filesystem;
+use Symplify\PackageBuilder\FileSystem\SmartFileInfo;
+
+final class UniteFileAndClassNameRector implements FileSystemRectorInterface
+{
+    /**
+     * @var BetterNodeFinder
+     */
+    private $betterNodeFinder;
+
+    /**
+     * @var Filesystem
+     */
+    private $filesystem;
+
+    /**
+     * @var Parser
+     */
+    private $parser;
+
+    /**
+     * @var NameResolver
+     */
+    private $nameResolver;
+
+    public function __construct(
+        BetterNodeFinder $betterNodeFinder,
+        Filesystem $filesystem,
+        Parser $parser,
+        NameResolver $nameResolver
+    ) {
+        $this->betterNodeFinder = $betterNodeFinder;
+        $this->filesystem = $filesystem;
+        $this->parser = $parser;
+        $this->nameResolver = $nameResolver;
+    }
+
+    public function getDefinition(): RectorDefinition
+    {
+        return new RectorDefinition(
+            'Corrects file name to be in sync with class name in it and autoloadable by PSR-4 classes.',
+            [
+                new CodeSample(
+                    <<<'CODE_SAMPLE'
+// file: SomeClass.php
+final class SomeCllass 
+{
+    
+}
+CODE_SAMPLE
+                    ,
+                    <<<'CODE_SAMPLE'
+// file: SomeClass.php
+final class SomeClass 
+{
+    
+}
+CODE_SAMPLE
+                ),
+            ]
+        );
+    }
+
+    public function refactor(SmartFileInfo $smartFileInfo): void
+    {
+        $nodes = $this->parser->parseFile($smartFileInfo->getRealPath());
+
+        // collect named class nodes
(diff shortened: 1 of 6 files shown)


TomasVotruba:
@JanMikes Ready for review :)

src/Rector/Psr4/UniteFileAndClassNameRector.php
```diff
@@ -0,0 +1,111 @@
...
+                    <<<'CODE_SAMPLE'
+// file: SomeClass.php
+final class SomeCllass 
+{
+    
+}
+CODE_SAMPLE
+                    ,
+                    <<<'CODE_SAMPLE'
+// file: SomeClass.php
```
mrpixeldream: According to the code and test case, it renames the file not the class. This example shows that the file name stays the same but the class inside gets renamed.

mrpixeldream: Maybe add an option to let the user decide in which direction they want to sync the name? From file to class or from class to file? I could imagine that this is a very hard decision though.

TomasVotruba: The idea is to automate this for user.  Any case for "asking" should be converted to algorithm or not handled by Rector at all but by a sniff.


I thought about this and there should be some function that detects which of file/class is typoed. Any idea for typo-detecting PHP package?

mrpixeldream: Even if we'd add a spell checker, the user would still have to configure which language he used for his project, which would then again be a case of "asking" I think.

TomasVotruba: Better done than perfect. Got any idea about the package?

mrpixeldream: I found this package: https://github.com/mekras/php-speller
It uses popular spell checkers like Hunspell (LibreOffice, OpenOffice, etc.) under the hood, seems to be well-tested and even provides suggestions what the right spelling would be. Looks like it's also actively being maintained.

TomasVotruba: Thanks. I know Hunspell.

People won't use anything that requires them install some binary file and setup path to it. The package must be *composer-only*.

mrpixeldream:
I can offer to write a simple implementation of a spell checker as a separate package if you want. It will not be as advanced as the professional tools like hunspell but I think it could be enough to start with and we can still evolve it as needed.

src/Rector/Psr4/UniteFileAndClassNameRector.php
mrpixeldream: I can offer to write a simple implementation of a spell checker as a separate package if you want. It will not be as advanced as the professional tools like hunspell but I think it could be enough to start with and we can still evolve it as needed.

TomasVotruba: If you think that's the best option, got for it :+1:

JanMikes: Hi guys! I agree that it would be awesome to let user choose, what way he wants to use it.

1) Fix filename  
2) Fix classname and all usages in app  

When i say *let user choose* there might be option/parameter in rector's config, maybe with default value? No prompts or ask dialogs :-).

mrpixeldream:
I think that would be the case for a different rector, right @TomasVotruba ? I would go only for renaming the namespace instead of renaming the folder in that case.

src/Rector/Psr4/UniteFileAndClassNameRector.php
TomasVotruba: Rector's idea is not to bother user at all. Anything requiring user's attention is better done in PHPStorm, where you prefer control with slow work.

Rector should handle typos for users, like `SomeeeClass` in `SomeClass.php`. Any user-opinionated moving is not Rector's job.

It would be better to add more real-life tests, so I know what you need this for :)

TomasVotruba:
> I think that would be the case for a different rector, right @TomasVotruba ? I would go only for renaming the namespace instead of renaming the folder in that case.

Yep, differnt issue.

src/Rector/Psr4/UniteFileAndClassNameRector.php
JanMikes: Real life example:
`SomeClassPresenter` + `SomeClass.php` - discussion with team, agreed to rename `SomeClass.php` -> `SomeClassPresenter.php` this will be most common use case for us now, but i can imagine in some cases we want to rename class instead of file. Maybe create 2 rectors? One for renaming files, different for renaming classes? 😄

TomasVotruba: I'd do these manually, since it needs team energy anyway.

TomasVotruba:
Closing as too vague to automate. This is better done manually, with human decission.

nschoellhorn:
Just to be sure: you don't want to follow the spell checker approach anymore? I've already started the development of a spell checking library that doesn't require additional binaries. If not needed here, I'd push back further development of it a little bit though.

TomasVotruba:
Nono, sorry if you invested some time it.

These opinionated filesystem changes would often definitely go wrong and annoy people. That's not feeling Rector should bring up.

PR #1175, opened 2019-03-10
[Laravel] Add request() to injection

TomasVotruba:
## Before

```php
<?php

namespace App\Http\Controllers;

class WelcomeController extends Controller
{
    public function index()
    {
        return view('welcome', [
            'foo' => request()->get('foo'),
        ]);
    }
}
```

## After

```php
<?php

namespace App\Http\Controllers;

use Illuminate\Http\Request;

class WelcomeController extends Controller
{
    public function index(Request $request)
    {
        return view('welcome', [
            'foo' => $request->get('foo'),
        ]);
    }
}
```

cc @azdanov :)

### DIFF
--- a/packages/Laravel/src/Rector/StaticCall/RequestStaticValidateToInjectRector.php
+++ b/packages/Laravel/src/Rector/StaticCall/RequestStaticValidateToInjectRector.php
@@ -3,9 +3,12 @@
 namespace Rector\Laravel\Rector\StaticCall;
 
 use PhpParser\Node;
+use PhpParser\Node\Expr\FuncCall;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Stmt\Class_;
+use Rector\NodeTypeResolver\Node\Attribute;
 use Rector\PhpParser\Node\Manipulator\ClassMethodManipulator;
 use Rector\Rector\AbstractRector;
 use Rector\RectorDefinition\CodeSample;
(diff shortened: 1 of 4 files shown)


azdanov:
In case of `request()->get('foo');` there are no issues.

But I get problems with `request('foo')`, it gets transformed to `$request;`

azdanov:
I think it would be better to use [`$request->input('foo');`](https://laravel.com/docs/5.8/requests#retrieving-input), here's an explanation https://laravel.com/api/5.8/Illuminate/Http/Request.html#method_get.

TomasVotruba:
Thank you for detailed review. 

Could you send cases  that are failing now as  PR test cases to this branch?

azdanov:
https://github.com/rectorphp/rector/pull/1177

TomasVotruba:
That's matter of PHPStan upgrade. This feature alone works. My local tests passes.

Thanks for failing test cases :+1:

mrpixeldream:
Are you sure it's related to PHPStan? How can PHPStan possibly let the test cases fail that worked before? Isn't PHPStan only relevant when being invoked manually?

TomasVotruba:
I'm sure. It worked on 0.11.2, it doesn't on 0.11.3

Rector uses part of PHPStan it to analyse types: https://github.com/rectorphp/rector/blob/master/composer.json#L14

azdanov:
Oh, now it makes sense. Usually, PHPStan is used as an outside tool.

mrpixeldream:
Ah, I see, didn't know that. Tested locally with a pinned version of 0.11.2 and it works. Maybe we should pin the version then? I can submit a PR if you want.

TomasVotruba:
Please do send the PR. I'm looking into the upgrade now, but it might take a while

PR #1180, opened 2019-03-10
Pin PHPStan version to 0.11.2

mrpixeldream:
This pins the PHPStan version to 0.11.2 since the newer version breaks tests.


TomasVotruba:
Thank you 

(The error is not related to this PR)

Issue #1185, opened 2019-03-11
PHPUnit MethodCall Rectors do not fix staticly called methods


bendavies:
Hey there.

All rectors here: https://github.com/rectorphp/rector/tree/master/packages/PHPUnit/src/Rector/MethodCall do not support methods called statically, e.g. `static::assertContains(...)`. PHPUnit assert methods are static so rector should support both static and non static. 

Cheers!

TomasVotruba:
Hi, 

I think it's tested on `self`.
Could you send failing test case to one of them?

bendavies:
hi @TomasVotruba 
I was wrong above.

my assert is actually not being converted because this check fails:
https://github.com/rectorphp/rector/blob/d2c925d366a7bd26e2b34e63256b554780fcd4cf/packages/PHPUnit/src/Rector/MethodCall/SpecificAssertContainsRector.php#L89

My assert is like this:

```php
    protected function assertHeader(Response $response, string $contentType): void
    {
        self::assertContains(
            $contentType,
            $response->headers->get('Content-Type')
        );
    }
```

Can anything be done?

TomasVotruba:
Can be somehow detected that `$response->headers->get('Content-Type')` return string?

bendavies:
well, it us a `UnionType`. string or array or null. we have no way of knowing if it's string or iterable i guess.

btw, i notice that `SpecificAssertContainsRector` docs say that it can fix `assertIterableContains`, but that does not seem to be implemented.

TomasVotruba:
> well, it is a UnionType. string or array or null. we have no way of knowing if it's string or iterable i guess.

In that case there is nothing we can do.

> btw, i notice that SpecificAssertContainsRector docs say that it can fix assertIterableContains, but that does not seem to be implemented.

Care to test that and fix that?

bendavies:
nothing to do here

TomasVotruba:
I just found out you were right. I'm sorry for false info.

There were 19 rules that only worked with `MethodCall` node.

The fix is ready: https://github.com/rectorphp/rector/pull/1197

PR #1187, opened 2019-03-11
remove incorrect docs

bendavies:
`assertIterableContains`/`assertIterableNotContains` were never implemented in phpunit.

https://github.com/rectorphp/rector/issues/1185#issuecomment-471768214


TomasVotruba:
Thank you

They were actually implemented as announced in PHPunit 8, but they were remove few moment before PHPunit 8 release from the core. So we had to remove them here as well.

PR #1192, opened 2019-03-12
Check minimum php version before running IsCountableRector and TernaryToSpaceshipRector

### LINKED ISSUE #1191: IsCountableRector is executed even if php_version_features is < 7.3
Using the following rector.yml with rector 0.4.4 it modifies the code using `is_countable` function which is [available only on PHP 7.3+](http://php.net/is_countable) (as expected by the IsCountableRector rule).

```
parameters:
  php_version_features: '7.2'
services:
  Rector\Php\Rector\BinaryOp\IsCountableRector: ~
```

Is this an expected behavior? Shouldn't IsCountableRector be bypassed due to php_version_features parameter?

endelwar:
fix #1191

### DIFF
--- a/packages/Php/src/Rector/BinaryOp/IsCountableRector.php
+++ b/packages/Php/src/Rector/BinaryOp/IsCountableRector.php
@@ -52,6 +52,10 @@ CODE_SAMPLE
      */
     public function refactor(Node $node): ?Node
     {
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Great job, thank you :+1:

PR #1197, opened 2019-03-13
[PHPUnit] make Rectors work with self and this as well

TomasVotruba:


bendavies:
any tests to update?

TomasVotruba:
I've extended few to verify functionality.

PR #1207, opened 2019-03-15
add failing test for RemoveUnusedPrivateMethodRector

bendavies:
Here's a bug I found when running `dead-code` level on my project


TomasVotruba:
Nice catch!

I think methods with name starting `__` should be all skipped, since there is no way to verify they're usage/non-usage. What do you think?

bendavies:
that sounds sensible to me

TomasVotruba:
Could you add it the same way `isAbstract()`?

Something like this:

```php
if ($this->matchName($node, '#^__*#')) {
    return null;
}
```

bendavies:
yep

TomasVotruba:
Great job :+1:

PR #1212, opened 2019-03-15
Failing tests

enumag:
Ref #1209

PR #1217, opened 2019-03-15
Fix underscore-to-namespace set

### LINKED ISSUE #1209: [Twig] underscore-to-namespace upgrade ignores annotations
I managed to run rector on my application by removing PHPStan completely from composer.json and then adding rector. This way it somewhat works with some hiccups:

### Rector outputs a lot of errors like this:

```
 [ERROR] Could not process "<filename>"   
         file, due to:                                                                                      
         "Name cannot be empty".
```

Which tells me nothing about what's wrong. But since some classes were updated anyway maybe it's not important.

### Annotations are ignored:

My classes ended up like this after the upgrade:

```
    /**
     * @var \Twig_Environment
     */
    private $twig;

    public function __construct(\Twig\Environment $twig)
    {
        $this->twig = $twig;
    }
```

As you can see while Rector did successfully upgrade the constructor typehint, it completely ignored the `@var` annotation.

Also I expected it to add a use-statement as well and change the contructor to `__construct(Environment $twig)` which did not happen either. Is there any way to enforce that?

TomasVotruba:
Make test from #1212 actually fail

Closes #1209

- one bug was caused by missing `Property` node on `RenameClassRector` :heavy_check_mark: 
- fix `PseudoNamespaceToNamespaceRector` for doclocks :heavy_check_mark:


enumag:
Wow that actually took quite a lot of work to fix. Thank you very much for doing it! I'll try it on my project this weekend and let you know about the result.

TomasVotruba:
If you find any missed cases, just put samples here. I'll add tests and fix them

PR #1218, opened 2019-03-15
Fix logging in phar

shyim:
In phar on boot of the kernel. The kernel tries to write inside the phar 🙈 

Btw got the phar version running 😊 

https://github.com/Symplify/Symplify/pull/1498

### DIFF
--- a/src/HttpKernel/RectorKernel.php
+++ b/src/HttpKernel/RectorKernel.php
@@ -34,6 +34,12 @@ final class RectorKernel extends Kernel implements ExtraConfigAwareKernelInterfa
         return sys_get_temp_dir() . '/_rector';
     }
 
+    public function getLogDir(): string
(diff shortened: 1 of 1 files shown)


src/HttpKernel/RectorKernel.php
```diff
@@ -34,6 +34,12 @@ public function getCacheDir(): string
         return sys_get_temp_dir() . '/_rector';
     }
 
+    public function getLogDir(): string
+    {
+        // manually configured, so it can be replaced in phar
+        return sys_get_temp_dir() . '/_rector';
```
TomasVotruba: Log dir should be different one than cache, to make finding logs easier. By convention

`_rector_log` should do

TomasVotruba:
Thank you :+1:

PR #1252, opened 2019-03-23
[CakePHP] Add Fixture importing name style rector(for 3.7)

o0h:
Since CakePHP 3.7,  declaring `TestCase::$fixtures`  in underscored has been deprecated.
Now, it should be declared in SnakeCase.

I think rector is the perfect solution for saving cakephp-er from such a boring work "check and renaming ALL `$fixtures = []` VARS in exists test file" 😄 

---

> Using underscored fixtures names in TestCase::$fixtures is deprecated. Use CamelCased names instead. For e.g. app.FooBar, plugin.MyPlugin.FooBar.
[(migration guide)](https://book.cakephp.org/3.0/en/appendices/3-7-migration-guide.html)

also see: https://github.com/cakephp/cakephp/pull/12560/files#diff-00c99fa549dd6a8e04203fff9bcc684aR196

### DIFF
--- a/config/level/cakephp/cakephp37.yaml
+++ b/config/level/cakephp/cakephp37.yaml
@@ -79,3 +79,5 @@ services:
             className:
                 set: 'setClassName'
                 get: 'getClassName'
+
+    Rector\CakePHP\Rector\Name\ChangeSnakedFixtureNameToCamel: ~
--- a/packages/CakePHP/src/Rector/Name/ChangeSnakedFixtureNameToCamelRector.php
+++ b/packages/CakePHP/src/Rector/Name/ChangeSnakedFixtureNameToCamelRector.php
@@ -10,6 +10,9 @@ use Rector\Rector\AbstractRector;
 use Rector\RectorDefinition\CodeSample;
 use Rector\RectorDefinition\RectorDefinition;
 
+/**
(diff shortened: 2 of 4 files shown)


o0h:
> Rector\PhpSpecToPHPUnit\Tests\Rector\Class_\PhpSpecToPHPUnitRector\PhpSpecToPHPUnitRectorTest::test

It seems to be occurred  in https://github.com/rectorphp/rector/commit/1fae638c39b11213debc90f1f08ad9657a343a1f .
Should I solve it in this branch before reviewing?

TomasVotruba:
> Should I solve it in this branch before asking code-review?

No no, I just fixed that on `master`. Just rebase and it should be fine

TomasVotruba:
Intresting change. I personaly autowire `under_score` to `CamelCase` with no problems.

Could you add link to the PR above the class? Like this: https://github.com/rectorphp/rector/blob/44810fb39fe3b608ec1dfcf2bd55e768664d67b2/packages/PHPUnit/src/Rector/MethodCall/SpecificAssertContainsRector.php#L14

That way there is reference to source of this change in case we want to investigate in the future.



Thank you :

o0h:
@TomasVotruba 
Thanks for your feedback!
I rebased master & force-pushed, and add commit to link for migration guid 🚀

TomasVotruba:
Thank you :+1:

PR #1320, opened 2019-04-14
[CakePHP]FIx rule for cakephp37

o0h:
Sorry, I made a mistake! 🙇 
In #1252, I added incorrect service name.
Lacking  of `~Rector` suffix, `--level  cakephp37` is  interrupted.


TomasVotruba:
Thanks for the fix

Issue #1319, opened 2019-04-13
Could not process error because of unrelated file


BernhardWebstudio:
Thank you very much for this utility! 
I am having a little issue with the `symfony-constructor-injection` level. 

**Rector Version**: v4.10 as well as dev-master@a65c7f2

**Command** I execute: 
`./bin/rector process ./src/ --autoload-file ./vendor/autoload.php --level symfony-constructor-injection --dry-run`
(I also tried without dry-run & without autoload-file, but I get the same error)

**Error** I get (multiple times, for different controllers but the same service & validator):
```
[ERROR] Could not process                                                                                              
         "/myDirectory/src/Acme/MyBundle/Controller/GeneralController.php" file, due to:                                                                                           
         "Cannot autowire service "Acme\MyOtherBundle\Validator\Constraints\AuthorizationValidator": argument "$security" of
         method "__construct()" references class "Symfony\Component\Security\Core\Security" but no such service         
         exists.".
...
```

While, when I check for the service with `bin/console debug:autowiring`, the service shows up fine (the app works fine the way it is, phpstan does not see any problems either):

 ```
Helper class for commonly-needed security tasks.
 Symfony\Component\Security\Core\Security (security.helper)
```

May I ask for some guidance on how to resolve this issue, what I am missing or how to debug rector's symfony autowiring?

TomasVotruba:
Hi, this is very hard to debug without full application.
I could give you random tips, but it would only lead us to comment ping-pong.

Could you provide minimal application on Github with as few packages and files as possible, that still gives you this error?

BernhardWebstudio:
Hi, thanks for your response. The easy fix for me was just to comment the classes requiring the security service, then it worked alright. 
Still curious (after all, when the problem is in context of security, it may be kind of relevant), I tried what you suggested to create a minimum viable application. Slowly deleting one file after another, I found that the error is related to the Symfony security configuration. My project had separate security configurations for production & debug. Apparently, rector may not be able to decide which environment it's run in. I guess this is a no-fix? Or a bug in an under-rector-lying library?

Anyway, if you want to reproduce, one may setup a Symfony app, create some service requiring the Security service and finally move `/config/security.yaml` to `/config/{prod,dev}/security.yaml` or use the following repo (might not work as an application, but gives the error): 

[https://github.com/BernhardWebstudio/rector-symfony-security-error](https://github.com/BernhardWebstudio/rector-symfony-security-error)

TomasVotruba:
[...]

PR #1331, opened 2019-04-21
Use `dev` as Symfony default env to fix issue #1319

BernhardWebstudio:
This PR includes what I was asked for in issue #1319 . Thank you very much for the assistance!


TomasVotruba:
Thanks for looking into this :+1:

PR #1304, opened 2019-04-08
[SOLID] Add PrivatizeLocalClassConstantRector

TomasVotruba:
Ref #151

PR #1333, opened 2019-04-23
Add PrivatizeLocalClassConstantRector to SOLID

mxr576:
Original PR: #1304

### DIFF
--- a/config/level/solid/solid.yaml
+++ b/config/level/solid/solid.yaml
@@ -1,2 +1,4 @@
 services:
     Rector\SOLID\Rector\Class_\FinalizeClassesWithoutChildrenRector: ~
+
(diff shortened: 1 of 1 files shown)


mxr576:
@TomasVotruba Although I am not sure what is the minimum required PHP version by the SOLID config set. I saw an "Add PHP 7.1 level min" task in the original PR and adding visibility to constants is only supported in >= PHP 7.1. 

https://www.php.net/manual/en/language.oop5.constants.php

TomasVotruba:
The PHP version is checked inside the rule:
https://github.com/rectorphp/rector/blob/de9eab49747662ae070b35ff1bcbc691488400bd/packages/SOLID/src/Rector/ClassConst/PrivatizeLocalClassConstantRector.php#L69-L71

TomasVotruba:
Thank you for PR!

PR #1367, opened 2019-04-30
[DeadCode] Anonymous class implementing an interface doesn't respect interface signature

pierredup:
I found this while running the dead code level on a code base. I don't have a fix for it yet, but thought I'd open a PR so long with a failing test case while I investigate further (and maybe get some pointers to the right solution)

When instantiating an anonymous class that implements an interface, and the methods in the class is empty, then all the parameters in the method signature is removed. The signature that comes from the interface isn't respected. When running the test case in the PR, I get the following failure:

```diff
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
         {
             $class = new class implements Foo
             {
-                public function bar($value1)
+                public function bar()
                 {
                 }
             };\n
```

but I expect the test to pass, since the function `bar` matches the signature in the interface it implements.

The strange thing, though, is that in the `RemoveUnusedParameterRector` class, the line `$classNode = $node->getAttribute(AttributeKey::CLASS_NODE);` returns the current class node, and not the anonymous class node. So I think the issue might lie when traversing the nodes, and not with the actual rector class. I'm still digging through everything, but any nudge in the right direction would be appreciated.

### DIFF
--- a/packages/DeadCode/tests/Rector/ClassMethod/RemoveUnusedParameterRector/Fixture/anonymous_classes.php.inc
+++ b/packages/DeadCode/tests/Rector/ClassMethod/RemoveUnusedParameterRector/Fixture/anonymous_classes.php.inc
@@ -1,0 +1,28 @@
+<?php
+
+namespace Rector\DeadCode\Tests\Rector\ClassMethod\RemoveUnusedParameterRector\FixtureInterface {
+
+    interface Foo
+    {
+        public function bar($value1);
+    }
+
+}
+
+namespace Rector\DeadCode\Tests\Rector\ClassMethod\RemoveUnusedParameterRector\Fixture {
+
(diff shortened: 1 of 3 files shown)


TomasVotruba:
There are many edge-cases in anonymous classes, that are hard to resolve right.
I think anonymous classes should be skipped by this rule.

As for the incorrect class node: each node is decorated by it's parent node.
The class is probably incorrectly resolved here: https://github.com/rectorphp/rector/blob/ea1a151897748065f6716a49abf371ab37f8480f/packages/NodeTypeResolver/src/NodeVisitor/ClassAndMethodNodeVisitor.php#L55

pierredup:
Thanks @TomasVotruba, I updated the PR with a patch for handling methods on an anonymous class.

TomasVotruba:
Thank you, looks good :+1:

PR #1377, opened 2019-05-02
Add function support to ImportFullyQualifiedNamesRector

### LINKED ISSUE #1376: [Bug] ImportFullyQualifiedNamesRector imports functions as classes
I'm using the library https://github.com/thecodingmachine/safe.

Until now, I've used their functions this way:

    \Safe\sprintf(...);

Now, with the new Rector `ImportFullyQualifiedNamesRector` imported it this way:

```php

use Safe\sprintf;

\Safe\sprintf(...);

```

This causes issues with static analysis and with PHPStorm.

The solution is to import the functions using `import function Safe\sprintf`.

See https://github.com/phan/phan/issues/2705

TomasVotruba:
Closes #1376

### DIFF
--- a/docs/AllRectorsOverview.md
+++ b/docs/AllRectorsOverview.md
@@ -2892,7 +2892,7 @@ Changes rand, srand and getrandmax by new md_* alternatives.
 
 - class: `Rector\Php\Rector\FuncCall\PregReplaceEModifierRector`
 
-The /e modifier is no longer supported, use preg_replace_callback instead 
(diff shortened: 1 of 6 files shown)


GrahamCampbell:
It would be useful to be able to turn this off, so one can only enable automatic importing of classes.

TomasVotruba:
There are already 2 options to skip docblocks and short classes: https://github.com/rectorphp/rector/pull/2428

The function one is not implemented yet

PR #1395, opened 2019-05-06
Preserve file permissions when updating a file

LeSuisse:
The change of permissions is not explicitely requested by the user
and is not shown in the dry-run. It makes Rector harder to use when
the sources of a project has some executable scripts.


TomasVotruba:
Thank you :+1:

PR #1410, opened 2019-05-10
[PHP 7.4] Add ClosureToArrowFunctionRector

### LINKED ISSUE #1405: [PHP 7.4] Add arrow functions
Merged in `php-parser` https://github.com/nikic/PHP-Parser/commit/f3b19c19ef147dc4e0adfce628d21031a7e20869

```diff
-return array_filter($meetups, function (Meetup $meetups) {
+return array_filter($meetups, fn(Meetup $meetup) => is_object($meetup));
-   return is_object($meetup);
-);

```

TomasVotruba:
Closes #1405

```diff
-return array_filter($meetups, function (Meetup $meetups) {
+return array_filter($meetups, fn(Meetup $meetup) => is_object($meetup));
-   return is_object($meetup);
-);
```

### DIFF
--- a/packages/Php/src/Rector/Closure/ClosureToArrowFunctionRector.php
+++ b/packages/Php/src/Rector/Closure/ClosureToArrowFunctionRector.php
@@ -1,0 +1,134 @@
+<?php declare(strict_types=1);
+
+namespace Rector\Php\Rector\Closure;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\ArrowFunction;
+use PhpParser\Node\Expr\Closure;
+use PhpParser\Node\Expr\ClosureUse;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Stmt\Return_;
+use Rector\Rector\AbstractRector;
+use Rector\RectorDefinition\CodeSample;
+use Rector\RectorDefinition\RectorDefinition;
+
+/**
+ * @see https://wiki.php.net/rfc/arrow_functions_v2
+ */
+final class ClosureToArrowFunctionRector extends AbstractRector
+{
+    public function getDefinition(): RectorDefinition
+    {
+        return new RectorDefinition('Change closure to arrow function', [
+            new CodeSample(
+                <<<'CODE_SAMPLE'
+class SomeClass
+{
+    public function run($meetups)
+    {
+        return array_filter($meetups, function (Meetup $meetup) {
+            return is_object($meetup);
+        });
+    }
+}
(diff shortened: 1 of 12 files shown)


packages/Php/src/Rector/Closure/ClosureToArrowFunctionRector.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $this->isAtLeastPhpVersion('7.4')) {
+            return null;
+        }
+
+        if (count((array) $node->stmts) !== 1) {
+            return null;
+        }
```
nikic: Shouldn't this check that there are no by-ref uses somewhere?

packages/Php/src/Rector/Closure/ClosureToArrowFunctionRector.php
```diff
@@ -0,0 +1,81 @@
...
+
+        $arrowFunction = new ArrowFunction();
+        $arrowFunction->params = $node->params;
+        $arrowFunction->returnType = $node->returnType;
+        $arrowFunction->byRef = $node->byRef;
+
+        $expr = $node->stmts[0];
+        if ($expr instanceof Return_) {
+            $expr = $expr->expr;
+        }
```
nikic: I don't think you should do this refactoring if there is no return. It may convert a void closure in a non-void one.

packages/Php/src/Rector/Closure/ClosureToArrowFunctionRector.php
TomasVotruba: Could you share example?

packages/Php/src/Rector/Closure/ClosureToArrowFunctionRector.php
TomasVotruba: Good point, thanks :+1:

packages/Php/src/Rector/Closure/ClosureToArrowFunctionRector.php
nikic: `function() use(&$i) { return ++$i; }` can't be converted to `fn() => ++$i`, because `$i` will be captured by-value instead of by-reference.

TomasVotruba: I see, thanks

TomasVotruba:
@nikic Thanks, I've included your notes

PR #1414, opened 2019-05-10
[PSR-4 Split] Split interfaces and traits as well

JanMikes:


JanMikes:
@TomasVotruba before this MR files containing interface and/or trait with class would fail splitting - it deleted non class nodes and did not move them to new files.

TomasVotruba:
Better create new test instead of deleting one

JanMikes:
Fixed 😉

JanMikes:
@TomasVotruba can we merge? 😉

TomasVotruba:
Sure, nice test btw, simple and clear :+1:

PR #1416, opened 2019-05-11
[PHP 7.4] Spread array

### LINKED ISSUE #1411: [PHP 7.4] Add array spread operator
- php-Parser PR: https://github.com/nikic/PHP-Parser/pull/609
- RFC: https://wiki.php.net/rfc/spread_operator_for_array
- array_merge of 2 items

TomasVotruba:
Closes #1411

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "jetbrains/phpstorm-stubs": "^2019.1",
         "nette/robot-loader": "^3.1",
         "nette/utils": "^2.5|^3.0",
-        "nikic/php-parser": "^4.2.1",
+        "nikic/php-parser": "dev-master",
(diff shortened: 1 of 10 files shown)


lyrixx:
Hi Tomas

Thanks for this PR. I used rector on a projet of mine and I saw this diff.
So I ran some benchmark to compare both version : https://github.com/lyrixx/php-bench/pull/5

The new version (spread array) is slower. And the new version is not necessary more readable.
So what do you think of making this rector optionnal in the php 7.4 set ?

TomasVotruba:
Hi,

how would people know about optional sets? It's like optional dependency.

Also, could you test with opcodes? I recall there was some difference about it in RFC.

TomasVotruba:
Thank you for performance testing :+1:

PR #1430, opened 2019-05-14
Add working directory option

ktomk:
New command-line utility option:

   -d, --working-dir=WORKING-DIR

If specified, use the given directory as working directory.

Follows same syntax as Composer (quick port from there).

### DIFF
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -6,6 +6,8 @@ use Jean85\PrettyVersions;
 use Rector\ContributorTools\Command\DumpNodesCommand;
 use Rector\ContributorTools\Command\DumpRectorsCommand;
 use Rector\ContributorTools\Exception\Command\ContributorCommandInterface;
+use Rector\Exception\Configuration\InvalidConfigurationException;
+use RuntimeException;
 use Symfony\Component\Console\Application as SymfonyApplication;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputDefinition;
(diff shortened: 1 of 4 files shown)


TomasVotruba:
What's the use case?

The right place for this is https://github.com/rectorphp/rector/blob/master/src/Console/Command/AbstractCommand.php

ktomk:
Changing the working directory on invocation. Compare w/ composer, tar, git etc. (for any Symfony console based CLI util, the composer variant is a known good from what I know). This helps with utility invocation, esp. with all other arguments relative to it. For example check with the docker example in the read me which different it could do.

As for the right place I must admit it's me now who does not fully understand. Why is that a better place?  Isn't the class chosen equivalent to [the one in composer](https://github.com/composer/composer/blob/master/src/Composer/Console/Application.php#L126)? I'm not so often using Symfony console, but do from time to time, so please give me a bit of your thoughts there.

TomasVotruba:
I understand the reason.

The location is due to tree architecture. Application is like router, it only delegates to specific command. It should not decide anything about options.

Symfony Console is pretty old and full of these nowadays antipattens of mixed responsibilities, but I still think it's better to keep them separated.

ktomk:
Oh router since just like the right place. It's an option for the application. E.g. for it to work it is important that it have it's effects before obtaining the path to the configuration file for example.

Also looked into the abstract class, must continue to admit that I don't get how that could work in there.

src/Console/Application.php
```diff
@@ -147,10 +157,34 @@ private function addCustomOptions(InputDefinition $inputDefinition): void
             InputOption::VALUE_NONE,
             'Enable debug verbosity (-vvv)'
         ));
+
+        $inputDefinition->addOption(new InputOption(
+            '--working-dir',
+            '-d',
+            InputOption::VALUE_REQUIRED,
+            'If specified, use the given directory as working directory.'
+        ));
```
TomasVotruba: I talk only about this section

ktomk: I can't get them displayed w/ --help when I put them in the abstract command. I assume this is b/c I want to set a global option which is an option of the *application* and a *command* only can configure non-global options.

TomasVotruba: Ah, that's problem of this coupling. It should be separated service.

Thanks for investigation. This is good enough

PR #1464, opened 2019-05-23
do not override printing files with previous stmts if possible

TomasVotruba:


TomasVotruba:
@janmikes I got the same error :) :party: 

error_tokens

JanMikes:
Hehe, should we celebrate that we have encounter same errors? 😄

TomasVotruba:
> Hehe, should we celebrate that we have encounter same errors? smile

Yes, because that's 50 % of the fix :)

TomasVotruba:
If found little bit hacky, but error-proof solution. It always uses previous formatting, if possible. But still works if not possible.

PR #1468, opened 2019-05-24
[MultipleClassFileToPsr4ClassesRector] Original file is deleted even if class matches filename

### LINKED ISSUE #1467: [MultipleClassFileToPsr4ClassesRector] Original file is deleted and it should not be if it contains class matching filename
The original file is being deleted completely, instead of just moving away duplicated classes.

I will investigate further and create failing test, i wonder if tests are passing now :-) 

Could be related to #1440 or #1360 ?

Screenshot 2019-05-24 11 21 37

JanMikes:
Covers #1467


TomasVotruba:
I changed "fixes" to "covers"

PR #1493, opened 2019-05-29
Fix documentation minor mistake

alterphp:


JanMikes:
@TomasVotruba isn't this file autogenerated? (It would mean typo in php code instead)

TomasVotruba:
@JanMikes Yes :+1: 

@alterphp Please fix it in this file: https://github.com/rectorphp/rector/blob/7bd2d79d2d9d2a6a9e66f01b3cd9378ca064b10c/packages/Symfony/src/Rector/HttpKernel/GetRequestRector.php#L49-L61

Then regenarated docs with:

```bash
composer docs
```

And push, thank you :+1:

alterphp:
@TomasVotruba As you see in https://github.com/rectorphp/rector/pull/1493/commits/18ca7fb5f7410e105f94fd731e6418ef15f19a16, there is a big diff in the generated doc files... That seems not good to me. WDYT ?

TomasVotruba:
@alterphp Why not? It's the way generated files work.

alterphp:
Ok, if it's only a sorting difference, why not. I used the Docker image from the provided Dockerfile to dump them and I work on a Linux environment.

TomasVotruba:
I see. 

Is is better when you try adding sort here? https://github.com/rectorphp/rector/blob/7bd2d79d2d9d2a6a9e66f01b3cd9378ca064b10c/packages/ContributorTools/src/Finder/RectorsFinder.php#L72

It should only change your line, there were no new rules since

TomasVotruba:
Thanks for the PR :+1:

alterphp:
> It's out of scope of this PR, I'm just curious. Is is better when you try adding sort here?

I tried many ways, still a big diff. Are you sure the generated doc in the `master` was up to date with source files ?

TomasVotruba:
Yes. The ordering is inconsistent, see: https://github.com/rectorphp/rector/pull/1493#issuecomment-497030391

I'll fix it, no troubles

Issue #1501, opened 2019-05-30
BinaryOpBetweenNumberAndStringRector should ignore concatenation "." but it doesn't


Aerendir:
```diff
    ---------- begin diff ----------
--- Original
+++ New
@@ -319,7 +319,7 @@
                 throw new MissingConnectionParameterException('You need to pass an ID to make a request for a Resource.');
             }

-            return $this->requestUrl . '/' . $this->resourceId;
+            return 0 . $this->resourceId;
         }

         return $this->requestUrl;
    ----------- end diff -----------

Applied rectors:

 * Rector\Php\Rector\BinaryOp\BinaryOpBetweenNumberAndStringRector
```

I expect that this concatenation is simply ignored and not transformed in this way as I need an URL in return, not what actually returns the transformed code.

TomasVotruba:
Btw, to highlight diff code here in Markdown, use *\`\`\`diff*

PR #1502, opened 2019-05-30
Test concatenation dot is ignored.

Aerendir:
Ensure that if a concatenation is made between a `string` and an `int`, Rector will not change anything, leaving the code as is.

https://github.com/rectorphp/rector/issues/1501


TomasVotruba:
Great, thanks :+1:

PR #1519, opened 2019-05-31
[Symfony] Add MakeCommandLazyRector

### LINKED ISSUE #1497: Feature Request: Symfony Rector to make commands lazy
See: https://symfony.com/doc/current/console/commands_as_services.html

TomasVotruba:
Closes #1497

What do you think @stloyd ?

### DIFF
--- a/packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php
+++ b/packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php
@@ -3,11 +3,13 @@
 namespace Rector\Symfony\Rector\Class_;
 
 use PhpParser\Node;
+use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Stmt\Class_;
 use PhpParser\Node\Stmt\ClassMethod;
 use PhpParser\Node\Stmt\Expression;
+use PHPStan\Type\StringType;
 use Rector\PhpParser\NodeTraverser\CallableNodeTraverser;
 use Rector\Rector\AbstractRector;
 use Rector\RectorDefinition\CodeSample;
@@ -108,11 +110,20 @@ CODE_SAMPLE
             &$commandName
         ) {
             if ($node instanceof MethodCall) {
+                if (! $this->isType($node->var, self::COMMAND_CLASS)) {
+                    return null;
+                }
+
                 if (! $this->isName($node, 'setName')) {
                     return null;
                 }
 
                 $commandName = $node->args[0]->value;
+                $commandNameStaticType = $this->getStaticType($commandName);
+                if (! $commandNameStaticType instanceof StringType) {
+                    return null;
+                }
+
                 $this->removeNode($node);
             }
 
@@ -124,12 +135,29 @@ CODE_SAMPLE
                 $onlyNode = $node->stmts[0];
                 if ($onlyNode instanceof Expression) {
                     $onlyNode = $onlyNode->expr;
-                    if (! $this->isName($onlyNode, '__construct')) {
+
+                    $commandName = $this->matchCommandNameNodeInConstruct($onlyNode);
+                    if ($commandName === null) {
                         return null;
                     }
 
-                    $commandName = $onlyNode->args[0]->value;
-                    if (! is_string($this->getValue($commandName))) {
+                    if (count($node->params) === 0) {
+                        $this->removeNode($node);
+                    }
+                }
+            }
+
+            if ($node instanceof ClassMethod && $this->isName($node, '__construct')) {
+                if (count((array) $node->stmts) !== 1) {
+                    return null;
+                }
+
+                $onlyNode = $node->stmts[0];
+                if ($onlyNode instanceof Expression) {
+                    $onlyNode = $onlyNode->expr;
+
+                    $commandName = $this->matchCommandNameNodeInConstruct($onlyNode);
+                    if ($commandName === null) {
                         return null;
                     }
 
@@ -140,16 +168,41 @@ CODE_SAMPLE
             }
 
             if ($node instanceof StaticCall) {
-                if (! $this->isName($node, '__construct')) {
+                if (! $this->isType($node, self::COMMAND_CLASS)) {
                     return null;
                 }
 
-                $commandName = $node->args[0]->value;
+                $commandName = $this->matchCommandNameNodeInConstruct($node);
+                if ($commandName === null) {
+                    return null;
+                }
 
                 array_shift($node->args);
             }
         });
 
         return $commandName;
+    }
+
+    private function matchCommandNameNodeInConstruct(Expr $expr): ?Node
+    {
+        if (! $expr instanceof MethodCall && ! $expr instanceof StaticCall) {
+            return null;
+        }
+
+        if (! $this->isName($expr, '__construct')) {
+            return null;
+        }
+
+        if (count($expr->args) < 1) {
+            return null;
+        }
+
+        $staticType = $this->getStaticType($expr->args[0]->value);
+        if (! $staticType instanceof StringType) {
+            return null;
+        }
+
+        return $expr->args[0]->value;
     }
 }
--- a/ecs.yaml
+++ b/ecs.yaml
@@ -101,7 +101,8 @@ parameters:
 
         Symplify\CodingStandard\Sniffs\CleanCode\CognitiveComplexitySniff:
             # tough logic
-            - 'packages/BetterPhpDocParser/src/'
+            - 'src/PhpParser/Node/Commander/NodeRemovingCommander.php'
+            - 'packages/BetterPhpDocParser/src/*'
             - 'packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php'
             - 'packages/Legacy/src/Rector/ClassMethod/ChangeSingletonToServiceRector.php'
             - 'src/Rector/Psr4/MultipleClassFileToPsr4ClassesRector.php'
@@ -150,4 +151,3 @@ parameters:
         Symplify\CodingStandard\Sniffs\DependencyInjection\NoClassInstantiationSniff:
             # 3rd party api
             - 'src/PhpParser/Node/Value/ValueResolver.php'
-
--- a/packages/Symfony/tests/Rector/Class_/MakeCommandLazyRector/Fixture/set_name_fluent.php.inc
+++ b/packages/Symfony/tests/Rector/Class_/MakeCommandLazyRector/Fixture/set_name_fluent.php.inc
@@ -1,0 +1,29 @@
+<?php
+
+namespace Rector\Symfony\Tests\Rector\Class_\MakeCommandLazyRector\Fixture;
+
+class SetNameFluent extends \Symfony\Component\Console\Command\Command
+{
+    public function configure()
+    {
+        $this->setName('some:command')
+            ->setDescription('Description');
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\Symfony\Tests\Rector\Class_\MakeCommandLazyRector\Fixture;
+
+class SetNameFluent extends \Symfony\Component\Console\Command\Command
+{
+    protected static $defaultName = 'some:command';
+    public function configure()
+    {
+        $this->setDescription('Description');
+    }
+}
+
+?>
--- a/packages/Symfony/tests/Rector/Class_/MakeCommandLazyRector/Fixture/skip_non_string_param_construct.php.inc
+++ b/packages/Symfony/tests/Rector/Class_/MakeCommandLazyRector/Fixture/skip_non_string_param_construct.php.inc
@@ -1,0 +1,16 @@
+<?php
+
+namespace Rector\Symfony\Tests\Rector\Class_\MakeCommandLazyRector\Fixture;
+
+use Symfony\Component\Console\Command\Command;
+
+final class SkipNonStringParamConstruct extends Command
+{
+    private $another;
+
+    public function __construct($someService, $another)
+    {
+        parent::__construct($someService);
+        $this->another = $another;
+    }
+}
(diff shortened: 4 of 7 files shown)


packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * @see https://symfony.com/doc/current/console/commands_as_services.html
+ */
+final class MakeCommandLazyRector extends AbstractRector
+{
+    /**
+     * @var string
+     */
+    private const COMMAND_CLASS = 'Symfony\Component\Console\Command\Command';
```
stloyd: You should also check the `ContainerAwareCommand` (it's deprecated but still used in legacy systems).

packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            if ($node instanceof Node\Expr\StaticCall) {
+                if (! $this->isName($node, '__construct')) {
+                    return null;
+                }
+
+                $commandName = $this->getValue($node->args[0]->value);
+
+                $this->removeNode($node);
```
stloyd: Call to parent should not be removed cause, it will have side effects.

TomasVotruba: So what should happen?

packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php
TomasVotruba: Doesn't that inherit from `Symfony\Component\Console\Command\Command`?

stloyd: If you check all parent classes then this okay  ;)

packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php
stloyd: You should leave empty call to parent:
```
parent::_construct();
```

packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php
TomasVotruba: It's type-based control, so yes :)

packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php
stloyd: If you have in mind simple: why leave empty parent call. Answer is that by having commands as services you inject those sevices in constructor  ;)

TomasVotruba:
I see. So just remove the argument?

stloyd:
If you find that call with parameter remove just the parameter and you will be safe  ;)

TomasVotruba:
(@stloyd Sorry, I wanted to comment to reply you, but on mobile it dropped me somewhere else)

packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    private function resolveCommandNameAndRemove(Class_ $class): ?Node
+    {
+        $commandName = null;
+        $this->callableNodeTraverser->traverseNodesWithCallable((array) $class->stmts, function (Node $node) use (
+            &$commandName
+        ) {
+            if ($node instanceof MethodCall) {
+                if (! $this->isName($node, 'setName')) {
```
stloyd: Done testing & this is too fragile, you must check it's call on `$this` (or fluent) cause I saw replacement:
```
+    protected static $defaultName = $something->name;
-                $object->setName($something->name);

File: packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php [hunk 4]
```diff
@@ -0,0 +1,155 @@
...
+                    }
+                }
+            }
+
+            if ($node instanceof StaticCall) {
+                if (! $this->isName($node, '__construct')) {
+                    return null;
+                }
+
+                $commandName = $node->args[0]->value;
```
stloyd: You need to check argument here too, cause I see:
```
+    protected static $defaultName = $entityManager;

-        parent::__construct($entityManager, $client);
+        parent::__construct($client);

Comment (stloyd) [reply packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php]: Also I see notice:
```
PHP Notice:  Trying to get property of non-object in /vendor/rector/rector/packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php on line 147
```
And error which may be related or not:
```
[ERROR] Could not process
         "/src/TestBundle/Command/SomeCommand.php" file, due to:
         "leaveNode() returned invalid value of type integer".
```

File: packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php (hunk 3 above)
Comment (TomasVotruba) [reply packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php]: How would you solve that?

File: packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php (hunk 4 above)
Comment (TomasVotruba) [reply packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php]: What code is causing it? Error alone is not much helpful

File: packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php (hunk 3 above)
Comment (stloyd) [reply packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php]: Can we check on what the method call is being done? If so I would validate its object called `$this` or of command instance.

File: packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php (hunk 4 above)
Comment (stloyd) [reply packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php]: Was not able yet to separate it, will check later today or tomorrow.

Comment (stloyd) [reply packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php]: Seems like such command:
```php
class SomeCommand extends ContainerAwareCommand
{
    protected function configure()
    {
        $this
            ->setName('some:command')
            ->setDescription('Description')
            ->addArgument('id', InputArgument::REQUIRED, 'integer')
            ->addArgument('string', InputArgument::REQUIRED, 'string')
        ;
    }

    protected function execute(InputInterface $input, OutputInterface $output)
    {
        return true
    }
}
```
Is able to generate that issue.

packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php
TomasVotruba: Yes, that's already done.

 
The problem is with "($something->name)". It cannot be move to static property.

packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php
TomasVotruba: Perfect, thanks

stloyd: Full trace:
```
In NodeTraverser.php line 168:

  [LogicException]
  leaveNode() returned invalid value of type integer


Exception trace:
 () at /vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:168
 PhpParser\NodeTraverser->traverseNode() at /vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:146
 PhpParser\NodeTraverser->traverseNode() at /vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:146
 PhpParser\NodeTraverser->traverseNode() at /vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:146
 PhpParser\NodeTraverser->traverseNode() at /vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:223
 PhpParser\NodeTraverser->traverseArray() at /vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:114
 PhpParser\NodeTraverser->traverseNode() at /vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:223
 PhpParser\NodeTraverser->traverseArray() at /vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:114
 PhpParser\NodeTraverser->traverseNode() at /vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:223
 PhpParser\NodeTraverser->traverseArray() at /vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:114
 PhpParser\NodeTraverser->traverseNode() at /vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:223
 PhpParser\NodeTraverser->traverseArray() at /vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:91
 PhpParser\NodeTraverser->traverse() at /vendor/rector/rector/src/PhpParser/Node/Commander/NodeRemovingCommander.php:53
 Rector\PhpParser\Node\Commander\NodeRemovingCommander->traverseNodes() at /vendor/rector/rector/src/Rector/AbstractRector.php:126
 Rector\Rector\AbstractRector->afterTraverse() at /vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:94
 PhpParser\NodeTraverser->traverse() at /vendor/rector/rector/src/PhpParser/NodeTraverser/RectorNodeTraverser.php:52
 Rector\PhpParser\NodeTraverser\RectorNodeTraverser->traverse() at /vendor/rector/rector/src/Application/FileProcessor.php:114
 Rector\Application\FileProcessor->refactor() at /vendor/rector/rector/src/Application/RectorApplication.php:108
 Rector\Application\RectorApplication->Rector\Application\{closure}() at /vendor/rector/rector/src/Application/RectorApplication.php:137
 Rector\Application\RectorApplication->tryCatchWrapper() at /vendor/rector/rector/src/Application/RectorApplication.php:109
 Rector\Application\RectorApplication->runOnFileInfos() at /vendor/rector/rector/src/Console/Command/ProcessCommand.php:154
 Rector\Console\Command\ProcessCommand->execute() at /vendor/symfony/console/Command/Command.php:255
 Symfony\Component\Console\Command\Command->run() at /vendor/rector/rector/src/Console/Command/AbstractCommand.php:32
 Rector\Console\Command\AbstractCommand->run() at /vendor/symfony/console/Application.php:926
 Symfony\Component\Console\Application->doRunCommand() at /vendor/symfony/console/Application.php:269
 Symfony\Component\Console\Application->doRun() at /vendor/rector/rector/src/Console/Application.php:92
 Rector\Console\Application->doRun() at /vendor/symfony/console/Application.php:145
 Symfony\Component\Console\Application->run() at /vendor/rector/rector/bin/rector:24

File: packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php (hunk 3 above)
Comment (stloyd) [reply packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php]: > The problem is with "($something->name)". It cannot be move to static property.

Yes, but this call is not on the command itself, it's on totally unrelated object (in this case it's value object class :) )

Comment (stloyd) [reply packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php]: Ok, test class will be much cleaner I guess :)
```php
class SomeCommand extends ContainerAwareCommand
{
    protected function configure()
    {
        $this->setDescription('Description');
    }

    protected function execute(InputInterface $input, OutputInterface $output)
    {
         // This will fail
         $input->setName('whatever you put here, object, string, you name it');
    }
}
```
Second:
```
class SomeCommand extends ContainerAwareCommand
{
    protected function configure()
    {
        $this->setName('some:command');
    }

    protected function execute(InputInterface $input, OutputInterface $output)
    {
         // This will fail
         $input->setName('whatever you put here, object, string, you name it');
    }
}

File: packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php (hunk 4 above)
Comment (TomasVotruba) [reply packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php]: Stack trace is very rarely needed. The best is to provide:

- src file that causes the error (the smaller the better, it's usually 1 line)
- error message
- Rector class that causes the error

**Comment by TomasVotruba** [comment]
New update :+1:

**Comment by stloyd** [comment]
@TomasVotruba Works almost perfect! Just randomly (like one on 20-30) classes is not getting correct property being set:
```diff
325) src/Command/TestCommand.php

    ---------- begin diff ----------
--- Original
+++ New
@@ -16,9 +16,7 @@
 {
     protected function configure()
     {
-        $this
-            ->setName('someclass')
-            ->setDescription('do something')
+        $this->setDescription('do something')
```
Other than this random thing it looks you covered all our legacy cases 👍 🏆

TomasVotruba:
@stloyd Thanks for deep testing! I think it will not be random. 

Could you get me the minimal anonymized code that has this error so I can fix it?

packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php
```diff
@@ -140,16 +168,37 @@ private function resolveCommandNameAndRemove(Class_ $class): ?Node
             }
 
             if ($node instanceof StaticCall) {
```
stloyd: @TomasVotruba Here is last bug :)

With class:
```php
class SomeCommand extends Command
{
    protected function configure()
    {
        $this->setName('some:command');
    }

    protected function execute(InputInterface $input, OutputInterface $output)
    {
        // This will fail
        SomeClass::staticMethodCall();
    }
}
```
It will fail cause that check is global but should be local, adding:
```diff
            if ($node instanceof StaticCall) {
+                if (! $this->isType($node, self::COMMAND_CLASS)) {
+                    return null;
+                }
```
Fixes the issue cause otherwise the `$commandName` was replaced with `null`.

TomasVotruba: Great examples + fix, just copy-pasted and it works :+1:

TomasVotruba: Btw, is this the "randomly (like one on 20/30) classes is not getting correct property being set" bug?

stloyd:
@TomasVotruba Just check latest version against our legacy codebase with over 400 (don't ask 🤓 ) commands being properly adjusted 😃 🚀

TomasVotruba:
The more code the bigger amount of work Rector saves  :+1: 

Thanks for testig and feedback. I wouldn't think about those edge cases myself, as I'm biased in single approach.

TomasVotruba:
It's :ship: ed

Issue #1539, opened 2019-06-03
StringClassNameToClassConstantRector transforms strings into namespaces when should not do it


Aerendir:
```diff
    ---------- begin diff ----------
--- Original
+++ New
@@ -188,7 +188,7 @@
         }

         if (null !== $exception && $event->isDebug()) {
-            $this->getLogger()->error('{message}', ['exception' => $exception, 'message' => $message]);
+            $this->getLogger()->error('{message}', [\exception::class => $exception, 'message' => $message]);
         }

         $this->getConsoleNotificationManager()->write('errorLine', $message);
    ----------- end diff -----------

    ---------- begin diff ----------
--- Original
+++ New
@@ -218,7 +218,7 @@
             $options = $reflected->invoke($s3Adapter, 'dummy');

             $bucket    = $options['Bucket'];
-            $directory = $options['directory'];
+            $directory = $options[\directory::class];
         }

         return 'S3://' . $bucket . '/' . $directory . '/';
    ----------- end diff -----------

Applied rectors:

 * Rector\Php\Rector\String_\StringClassNameToClassConstantRector
```

Assuming both classes exist (`\Exception` and `\Directory`), the rector anyway transforms them wrongly using all lowercase letters.

But the main problem here is that both `exception` and `directory` are keys of an array, so should not be transformed at all.

TomasVotruba:
Thanks for reporting.

There are many approaches with differnt -/+:

- 1. ignore upper/lower case of clases (current approach) - it will replace typoed classes, but also strings sometimes, becuase "directory" is usually a string, but "Directory" might be a class

- 2. skip simple words like "exception", "directory" and "error", because people tend to use them as a string

Maybe we could make it case sensitive and upper/lower case mistakes delegate to another fixer/rule. What do you think?

Aerendir:
I think the second one approach is the best one: highlight such misspellings should be delegated to IDEs like PHPStorm and to static code analyzers. And I think is also the simplest approach...

PR #1540, opened 2019-06-03
[test] String class name to class constant rector ignore array keys 1539

Aerendir:
Issue #1539.

### DIFF
--- a/packages/Php/tests/Rector/String_/StringClassNameToClassConstantRector/Fixture/ignore_array_keys.php.inc
+++ b/packages/Php/tests/Rector/String_/StringClassNameToClassConstantRector/Fixture/ignore_array_keys.php.inc
@@ -1,0 +1,13 @@
+<?php
+
+namespace Rector\Php\Tests\Rector\String_\StringClassNameToClassConstantRector\Fixture;
+
+class IgnoreArrayKeys
+{
+    public function run()
+    {
+        return ['exception' => new \Exception()];
+    }
+}
+
+?>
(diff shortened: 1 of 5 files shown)


TomasVotruba:
First commit says, that keys are ignored.

Would you add test with namespaces class that proves that?

Aerendir:
@TomasVotruba no, it is not clear my comment: it says that if I call `$var['Exception']`, then I get the `\Exception` object.

But instead, the key is transformed into `$var[\Exception::class]` and this is not correct: it should remain `$var['Exception']`, also if `Exception` is the name of an existent class.

TomasVotruba:
> But instead, the key is transformed into $var[\Exception::class]

This is expected behavior.

```diff
'SomeClass' => new SomeClass

Aerendir:
@TomasVotruba , in my case it is not intended as the name of a class but, instead, as a key by which I can get the Exception.

Maybe I need to change the name of my key to something like `ExceptionObj` 🤔

TomasVotruba:
Why is it upper class anyway? How do you use it?

Aerendir:
@TomasVotruba there is no particular reason: I simply wrote it this way 😅.

You are right: I can simply switch to lowercase, am I right?

TomasVotruba:
I see :) lowercase would be better in this particular case. We'll see how it's used in the wild

TomasVotruba:
Resolved by https://github.com/rectorphp/rector/pull/1565

Thanks for the test case

Issue #1551, opened 2019-06-04
CountOnNullRector should verify the variable can actually be null


Aerendir:
```diff
    ---------- begin diff ----------
--- Original
+++ New
@@ -59,7 +59,7 @@
             }
         }

-        return count($return) > 0 ? $return : null;
+        return (is_array($return) || $return instanceof \Countable ? count($return) : 0) > 0 ? $return : null;
     }

     /**
@@ -104,6 +104,6 @@
         // Flatten the array
         $results = array_merge(...$results);

-        return count($results) > 0 ? $results : null;
+        return (is_array($results) || $results instanceof \Countable ? count($results) : 0) > 0 ? $results : null;
     }
 }
    ----------- end diff -----------

Applied rectors:

 * Rector\Php\Rector\FuncCall\CountOnNullRector
```

The original code is this:

```php
    public function myMethod(string $string, array $regexes): ?array
    {
        $results = [];

        foreach ($regexes as $toFind) {
            if (false !== \Safe\preg_match_all($toFind, $string, $result)) {
                $results[] = $result[0];
            }
        }

        // Flatten the array
        $results = array_merge(...$results);

        return count($results) > 0 ? $results : null;
    }
```

The code changed by the rector causes also errors with PHPStan:

> Else branch is unreachable because ternary operator condition is always true.

There is also another error about the fact that the variable will never be `\Countable` as it is an `array` and not an object.

TomasVotruba:
Covered by #1563

PR #1552, opened 2019-06-04
Don't apply `CountOnNullRector` to variable that are `array` for sure.

Aerendir:
Issue #1551.

### DIFF
--- a/packages/Php/tests/Rector/FuncCall/CountOnNullRector/Fixture/local_variable.php.inc
+++ b/packages/Php/tests/Rector/FuncCall/CountOnNullRector/Fixture/local_variable.php.inc
@@ -1,0 +1,143 @@
+<?php
+
+namespace Rector\Php\Tests\Rector\FuncCall\CountOnNullRector\Fixture;
+
+use Rector\Php\Tests\Rector\FuncCall\CountOnNullRector\Source\CountableClass;
+
+final class LocalVariable
+{
+    public function methodWithLocalVariableInitializedAsArray(array $anArray): ?array
+    {
+        $results = [];
+
+        foreach ($anArray as $value) {
+            $results[] = $value;
+        }
+
+        return count($results) > 0 ? $results : null;
+    }
+
+    public function methodWithLocalVariableArrayAsResult(string $string, array $regexes): ?array
+    {
+        $results = [];
+
+        foreach ($regexes as $toFind) {
+            if (false !== preg_match_all($toFind, $string, $result)) {
+                $results[] = $result[0];
+            }
+        }
+
+        // Flatten the array
+        $results = array_merge(...$results);
+
+        return count($results) > 0 ? $results : null;
+    }
+
+    public function methodWithLocalVariableInitializedAsArrayAndTransformedIntoString(array $anArray): ?array
+    {
+        $results = [];
+
+        // This code may not have much sense,
+        // but in real world scenario it may
+        // appear very similar to this.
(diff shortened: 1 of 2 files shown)


Aerendir:
The tests here passes, as if it is not problem.

But as per Issue, problems actually exist.

@TomasVotruba , please, can you check?

packages/Php/tests/Rector/FuncCall/CountOnNullRector/Fixture/local_variable.php.inc
```diff
@@ -0,0 +1,143 @@
...
+        // appear very similar to this.
+        foreach ($anArray as $value) {
+            if ('condition' === $value) {
+                $results = 'transform into a string';
+            } elseif (is_array($results)) {
+                $results[] = $value;
+            }
+        }
+
+        return (is_array($results) || $results instanceof \Countable ? count($results) : 0) > 0 ? $results : null;
```
stloyd: What about using: `is_iterable($results)`?

packages/Php/tests/Rector/FuncCall/CountOnNullRector/Fixture/local_variable.php.inc
```diff
@@ -0,0 +1,143 @@
...
+    public function methodWithLocalVariableInitializedAsArrayAndTransformedIntoString(array $anArray): ?array
+    {
+        $results = [];
+
+        // This code may not have much sense,
+        // but in real world scenario it may
+        // appear very similar to this.
+        foreach ($anArray as $value) {
+            if ('condition' === $value) {
+                $results = 'transform into a string';
```
stloyd: TBH. I'm not sure this should be supported at all cause string is not countable in newer PHP versions, yet this probably will fail while checking code with phpstan.

TomasVotruba:
@Aerendir Could you maybe add just the minimal example that broke your code? 4-5 lines shold do.

It's too long to review and with no added value.

packages/Php/tests/Rector/FuncCall/CountOnNullRector/Fixture/local_variable.php.inc
Aerendir: This is not something I can decide: I'm simply following the way the rector works.

To change it, a change in the rector, too, is required.

ping @TomasVotruba

packages/Php/tests/Rector/FuncCall/CountOnNullRector/Fixture/local_variable.php.inc
Aerendir: Yes, you are right. In my opinion, it is something that should be avoided at all.

I added it for completeness, but there's nothing against removing it.

ping @TomasVotruba

Aerendir:
@TomasVotruba , I'll try, but the test I've submitted already contains the failing code: is the method `methodWithLocalVariableArrayAsResult()`: in my code it is changed while in this test it isn't...

packages/Php/tests/Rector/FuncCall/CountOnNullRector/Fixture/local_variable.php.inc
TomasVotruba: @stloyd Why not. Could you show me the compatibility with https://3v4l.org ?

TomasVotruba:
The smaller the code the easier to find why

TomasVotruba:
Closing in favor of #1563

PR #1558, opened 2019-06-04
Do not call parent constructor of AutowiredEventDispatcher unless it exists

cgkkevinr:
Signed-off-by: Kevin R <22906111+cgkkevinr@users.noreply.github.com>


TomasVotruba:
Thanks, works well!

PR #1559, opened 2019-06-04
Fix classname change for FilterControllerEvent

keulinho:
As described in https://github.com/symfony/symfony/blob/4.4/UPGRADE-4.3.md#httpkernel the new name for the FilterControllerEvent is ControllerEvent


TomasVotruba:
Thank you!

PR #1563, opened 2019-06-04
add array_merge to CountOnNullRector

TomasVotruba:


TomasVotruba:
@Aerendir I've added the `array_merge` case, locally it passes

Aerendir:
I have to check if I still have the broken code: if I have, I'll try to understand better which is the problem... Otherwise, we have to wait for the next time it will broke 😅

TomasVotruba:
No troubles, it's also possible it was accidentally fixed in PHPStan which used for static analysis here.

PR #1567, opened 2019-06-05
Fix solid.yaml

Great-Antique:
Currently solid.yaml references to not existed class. This PR fix this


TomasVotruba:
Thank you :+1:

PR #1582, opened 2019-06-08
Rename "level" directory to "set"

TomasVotruba:


TomasVotruba:
cc @jkuchar Final part of "level" → "set" renaming

PR #1605, opened 2019-06-10
Fix Open Collective link for FUNDING.yml

pxgamer:
This fixes the `FUNDING.yml` file to use the correct Open Collective id.

Currently the Sponsor button shows the following error notice:
> Unknown funding platform 'opencollective'.


TomasVotruba:
Thank you!

PR #1627, opened 2019-06-18
Symfony42: Remove invalid rectors

bendavies:
Ref: https://github.com/symfony/symfony/commit/f75fffa9974d98fb40f0e7365e99ec35e01e3509 

These deprecation warnings are for when inherited class calls parent method.

### DIFF
--- a/config/set/symfony/symfony42.yaml
+++ b/config/set/symfony/symfony42.yaml
@@ -14,57 +14,6 @@ services:
     Rector\Symfony\Rector\New_\StringToArrayArgumentProcessRector: ~
     Rector\Symfony\Rector\New_\RootNodeTreeBuilderRector: ~
 
-    Rector\Rector\Argument\ArgumentAdderRector:
-        # https://github.com/symfony/symfony/commit/fa2063efe43109aea093d6fbfc12d675dba82146
-        # https://github.com/symfony/symfony/commit/e3aa90f852f69040be19da3d8729cdf02d238ec7
-        Symfony\Component\BrowserKit\Client:
-            submit:
-                2:
-                    name: 'serverParameters'
-                    default_value: []
-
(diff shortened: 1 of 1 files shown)


TomasVotruba:
I don't understand. Could you explain bit more?

bendavies:
These deprecations warnings for for classes extending the symfony ones. for example.

This will trigger a deprecation warning.
```php
<?php

class ClassThatInheritClient extends Client
{
    public function submit(DomCrawlerForm $form, array $values []))
    {
        return parent::submit($form, $values);
    }
}
```

It should be fixed to:
```php
<?php

class ClassThatInheritClient extends Client
{
    public function submit(DomCrawlerForm $form, array $values [], array $serverParameters = [])))
    {
        return parent::submit($form, $values, $serverParameters);
    }
}
```

The current rectors are fixing calls to the current symfony classes, i.e.
Changing this:
```php
<?php


class MyService
{
    public function __invoke()
    {
        $client = new \Symfony\Component\HttpKernel\Client(...);
        $client->submit($form);
    }
}
```

to this
```php
<?php


class MyService
{
    public function __invoke()
    {
        $client = new \Symfony\Component\HttpKernel\Client(...);
        $client->submit($form, []);
    }
}
```

which is incorrect

TomasVotruba:
I see. In that way it should not be removed but fixed.

There should be new rule that has the same configuration, but adds it only to parent call + class method.

bendavies:
👍 so it's not possible right now?

TomasVotruba:
Not yet

bendavies:
Right. So shall we merge this for now? As this config is causing incorrect changes

TomasVotruba:
No, it needs to be fixed so it works as it should.
Merging would cause other troubles with missing migration

bendavies:
up to you, but at the moment keeping this causes an incorrect migration, which is better than no migration.

TomasVotruba:
I know. But finished work is better than correct double work.

PR #1631, opened 2019-06-19
Fix typo : rename jsm-decouple.yaml to jms-decouple.yaml

gnutix:
I've quickly looked for any usage in the code but found none. Maybe this should be documented though, as users might rely on it.


TomasVotruba:
Thank you

PR #1640, opened 2019-06-23
Add --rule option to process only single rule from set

TomasVotruba:
```bash
bin/rector process packages/Guzzle --set symfony43 -n -vvv --rule "Rector\Symfony\Rector\MethodCall\SimplifyWebTestCaseAssertionsRector"
```

Must be quoted, or "\\ " are trimmed

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -169,3 +169,5 @@ parameters:
 
         # future compat
         - '#Call to function method_exists\(\) with (.*?) will always evaluate to false#'
+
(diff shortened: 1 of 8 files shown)


TomasVotruba:
Ref #42

TomasVotruba:
This feature is causing some confusion in how to work with sets and rules, as it give 2 ways to do 1 thing:
https://github.com/rectorphp/rector/issues/4690

Also after switch from YAML to PHP configs: https://getrector.org/blog/2020/08/31/rector-is-moving-from-yaml-to-php-configs-what-changes-and-how-to-get-ready,
and dropping `--set` CLI option to make use of the PHP configs,
this feature is contradictory towards PHP autocomplete filosophy

Sayint that, it will be probably removed

PR #1674, opened 2019-07-02
Add Polyfil function support

### LINKED ISSUE #1650: [PHP] for new version code verify both version and polyfil
E.g. `is_iterable` can be added via symfony polyfil before PHP version is upgraded

Suggested by @stloyd

TomasVotruba:
Closes #1650

### DIFF
--- a/packages/Php/src/Rector/BinaryOp/IsCountableRector.php
+++ b/packages/Php/src/Rector/BinaryOp/IsCountableRector.php
@@ -4,7 +4,7 @@ namespace Rector\Php\Rector\BinaryOp;
 
 use PhpParser\Node;
 use PhpParser\Node\Expr\BinaryOp\BooleanOr;
-use Rector\Php\DualCheckToAble;
+use Rector\Php\IsArrayAndDualCheckToAble;
 use Rector\Rector\AbstractRector;
 use Rector\RectorDefinition\CodeSample;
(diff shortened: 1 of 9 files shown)


TomasVotruba:
Cc @@stloyd

packages/Php/src/Rector/BinaryOp/IsCountableRector.php
```diff
@@ -52,10 +52,19 @@ public function getNodeTypes(): array
...
             return null;
         }
 
-        return $this->dualCheckToAble->processBooleanOr($node, 'Countable', 'is_countable') ?: $node;
+        return $this->isArrayAndDualCheckToAble->processBooleanOr($node, 'Countable', 'is_countable') ?: $node;
+    }
+
+    private function shouldSkip(): bool
+    {
+        if ($this->isAtLeastPhpVersion('7.3')) {
```
stloyd: Shouldn't those check be opposite?

You can have polyfill with that method on even PHP 7.0, so first it should check for polyfill/method availability, then check PHP version, no?

TomasVotruba: You're right of course!

Care for you PR?

stloyd: https://github.com/rectorphp/rector/pull/1677

PR #1677, opened 2019-07-03
[Bugfix] `IsCountableRector` & `IsIterableRector` should first check method availability

stloyd:
User can have polyfill with that method on even PHP 7.0, so first it should check for polyfill/method availability, then check PHP version.


TomasVotruba:
CI is failing. Could you look at it?

packages/Php/tests/Rector/BinaryOp/IsCountableRector/IsCountableRectorTest.php
```diff
@@ -9,7 +9,11 @@ final class IsCountableRectorTest extends AbstractRectorTestCase
 {
     public function test(): void
     {
-        $this->doTestFiles([__DIR__ . '/Fixture/fixture.php.inc']);
+        if (PHP_VERSION_ID < 70300 && ! function_exists('is_countable')) {
```
stloyd: This is required cause some of Symfony dependencies force installation of polyfill on `dev` instalation of Rector.

packages/Php/tests/Rector/BinaryOp/IsCountableRector/Fixture/fixture71.php.inc
```diff
@@ -0,0 +1,17 @@
...
+?>
+-----
+<?php
+
+function isCountable()
+{
+    is_array($foo) || $foo instanceof Countable;
+}
+
+?>
```
TomasVotruba: If code are the same before/after (= no change), you can remove the part under `-----` (included)

```diff
 <?php
 function isCountable()
 {
     is_array($foo) || $foo instanceof Countable;
 }
 ?>
------
-<?php
-function isCountable()
-{
-    is_array($foo) || $foo instanceof Countable;
-}
-?>
```

packages/Php/tests/Rector/BinaryOp/IsCountableRector/IsCountableRectorTest.php
TomasVotruba: I see. The PHP version is taken from the config though, not real PHP version. So tests are consistent on various PHP versions.

See this: https://github.com/rectorphp/rector/pull/1674/files#diff-18a25bf46cd398e8a56a8ef9bf4f7172R26

You change PHP version to your desired one and this should be 2 tests then.

TomasVotruba: If it's not possible, add a comment you wrote there, so we know "why" later

packages/Php/tests/Rector/BinaryOp/IsIterableRector/Fixture/fixture70.php.inc
```diff
@@ -0,0 +1,37 @@
...
+
+    is_string($foo) || $foo instanceof Traversable;
+
+    is_array($foo2) || $foo instanceof Traversable;
+
+    $foo2 || $foo instanceof Traversable;
+}
+
+?>
+-----
```
TomasVotruba: Again no change, so this line and bellow can be removed

TomasVotruba:
Thanks for quick response!

There is one more issue on Travis with lowest dependencies: https://travis-ci.org/rectorphp/rector/jobs/553693071#L658

This is how you replicate on local machine:

```bash
composer update --prefer-lowest
vendor/bin/phpunit
```

TomasVotruba:
Thank you :+1:

PR #1679, opened 2019-07-03
[MakeCommandLazyRector] Remove duplicated check

stloyd:


stloyd:
https://github.com/rectorphp/rector/blob/master/packages/Symfony/src/Rector/Class_/MakeCommandLazyRector.php#L119-L137

TomasVotruba:
I see

TomasVotruba:
Thank you

PR #1687, opened 2019-07-04
[Symfony] Set few default common service names for Symfony App Analyzer

stloyd:
Allow to set custom ones via configuration:
```yaml
# rector.yaml
services:
  Rector\Symfony\Bridge\DefaultAnalyzedSymfonyApplicationContainer:
    autowire: true
    arguments:
      $commonNamesToTypes:
        'doctrine.orm.other_entity_manager': Doctrine\ORM\EntityManagerInterface
        'pseudo.buggy.service_name': App\Some\ServiceClass

### DIFF
--- a/packages/Symfony/src/Bridge/DefaultAnalyzedSymfonyApplicationContainer.php
+++ b/packages/Symfony/src/Bridge/DefaultAnalyzedSymfonyApplicationContainer.php
@@ -18,6 +18,8 @@ final class DefaultAnalyzedSymfonyApplicationContainer implements AnalyzedApplic
      */
     private $commonNamesToTypes = [
         'doctrine' => 'Doctrine\Bundle\DoctrineBundle\Registry',
+        'doctrine.orm.entity_manager' => 'Doctrine\ORM\EntityManagerInterface',
+        'doctrine.orm.default_entity_manager' => 'Doctrine\ORM\EntityManagerInterface',
     ];
 
     /**
@@ -35,14 +37,19 @@ final class DefaultAnalyzedSymfonyApplicationContainer implements AnalyzedApplic
      */
     private $containerFactory;
 
+    /**
+     * @param array<string, string> $commonNamesToTypes
+     */
     public function __construct(
         ParameterProvider $parameterProvider,
         SymfonyKernelParameterGuard $symfonyKernelParameterGuard,
-        ContainerFactory $containerFactory
+        ContainerFactory $containerFactory,
+        array $commonNamesToTypes = []
     ) {
         $this->parameterProvider = $parameterProvider;
         $this->symfonyKernelParameterGuard = $symfonyKernelParameterGuard;
         $this->containerFactory = $containerFactory;
+        $this->commonNamesToTypes = array_merge($this->commonNamesToTypes, $commonNamesToTypes);
     }
 
     public function getTypeForName(string $name): ?string

### CODE REVIEW

File: packages/Symfony/src/Bridge/DefaultAnalyzedSymfonyApplicationContainer.php [hunk 1]
```diff
@@ -35,14 +37,22 @@ final class DefaultAnalyzedSymfonyApplicationContainer implements AnalyzedApplic
      */
     private $containerFactory;
 
+    /**
+     * @param ParameterProvider           $parameterProvider
+     * @param SymfonyKernelParameterGuard $symfonyKernelParameterGuard
+     * @param ContainerFactory            $containerFactory
```
TomasVotruba: There should be no duplicated for PHP code info.
The sniff is disabled for this, because it was doing to much

packages/Symfony/src/Bridge/DefaultAnalyzedSymfonyApplicationContainer.php
```diff
@@ -35,14 +37,22 @@ final class DefaultAnalyzedSymfonyApplicationContainer implements AnalyzedApplic
...
         ParameterProvider $parameterProvider,
         SymfonyKernelParameterGuard $symfonyKernelParameterGuard,
-        ContainerFactory $containerFactory
+        ContainerFactory $containerFactory,
+        array $commonNamesToTypes = []
     ) {
         $this->parameterProvider = $parameterProvider;
         $this->symfonyKernelParameterGuard = $symfonyKernelParameterGuard;
         $this->containerFactory = $containerFactory;
+        $this->commonNamesToTypes = array_merge($this->commonNamesToTypes, $commonNamesToTypes);
```
TomasVotruba: What is this needed for? Default names should not depend on custom user input, but on default Symfony naming.

packages/Symfony/src/Bridge/DefaultAnalyzedSymfonyApplicationContainer.php
```diff
@@ -18,6 +18,8 @@ final class DefaultAnalyzedSymfonyApplicationContainer implements AnalyzedApplic
      */
     private $commonNamesToTypes = [
         'doctrine' => 'Doctrine\Bundle\DoctrineBundle\Registry',
+        'doctrine.orm.entity_manager' => 'Doctrine\ORM\EntityManagerInterface',
+        'doctrine.orm.default_entity_manager' => 'Doctrine\ORM\EntityManagerInterface',
```
TomasVotruba: :+1:

packages/Symfony/src/Bridge/DefaultAnalyzedSymfonyApplicationContainer.php
stloyd: Problem is for example that by default container says that: `doctrine.orm.entity_manager` is instance of `Doctrine\Common\Persistence\ObjectManager`, which in fact is not really true, cause it's instance of `Doctrine\ORM\EntityManagerInterface` (which is also used for autowire functionality).

This is just one example, if there is more edge cases (in big legacy like we have, we can occur them most likely), user is not able to have automatic fix for such cases and he/she will be forced to fix it manually.

With this you can have most common known cases for replacement, but also user is able to not wait for new release of Rector if he/she finds another edge case.

TomasVotruba: :+1: 

This is just one example, if there is more edge cases (in big legacy like we have, we can occur them most likely), user is not able to have automatic fix for such cases and he/she will be forced to fix it manually.

I see. It makes sense now :) I missed this description on PR. Could you add it next time, so I  can read the code with context of your intention? It would speed up the review

PR #1697, opened 2019-07-05
[DeadCode] Allow static constant call on RemoveUnusedPrivateConstantRector

ravanscafi:
Basically there were false positives regarding private constant uses when called with `static::` instead of `self::`

**Note**: `phpstan` is failing [on master too](https://github.com/rectorphp/rector/commits/master). I've fixed it on #1698


ravanscafi:
I'm not sure why `phpstan` is failing on totally unrelated code. Is it something that I did?

TomasVotruba:
Thank you :+1:

PR #1698, opened 2019-07-05
Fix NameTypeResolver@resolveFullyQualifiedName return type

ravanscafi:
It was broken on `master` branch.

PR #1699, opened 2019-07-05
Update composer scripts

ravanscafi:
There was a call to `@update-docs` but the script is called `@docs`.

**Note**: `phpstan` is failing [on master too](https://github.com/rectorphp/rector/commits/master). I've fixed it on #1698

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -139,7 +139,7 @@
             "@check-cs",
             "phpunit",
             "@phpstan",
-            "@update-docs"
+            "@docs"
         ],
         "check-cs": "vendor/bin/ecs check bin packages src tests utils --ansi",
         "fix-cs": [


composer.json
```diff
@@ -139,18 +139,18 @@
...
             "phpunit",
             "@phpstan",
-            "@update-docs"
+            "@docs"
         ],
-        "check-cs": "vendor/bin/ecs check bin packages src tests utils --ansi",
+        "check-cs": "ecs check bin packages src tests utils --ansi",
         "fix-cs": [
-            "vendor/bin/ecs check bin packages src tests utils --fix --ansi",
+            "ecs check bin packages src tests utils --fix --ansi",
```
TomasVotruba: Why this? It's problematic when copy-pasted to modify

ravanscafi: No problem, @TomasVotruba. I've updated the PR.

TomasVotruba:
Thank you

Issue #1692, opened 2019-07-05
[SOLID] Rector\DeadCode\Rector\ClassMethod\RemoveUnusedParameterRector : inverts "public" and "abstract", which breaks PSR-2 standard + removes parameters to abstract methods


gnutix:
I've noticed this strange change that happens when I add the `solid` level to my code base. It's an utility class that, currently, is not used in the code. But as it's an abstract method, I'm not sure the Rector should remove the parameters (it could be used from another bundle/project).

And the order of public / abstract should not be touched I guess, as PSR-2 states : 

> Visibility MUST be declared on all properties and methods; abstract and final MUST be declared before the visibility; static MUST be declared after the visibility.

```diff
$ vendor/bin/rector process --dry-run --level=solid src/SomeNamespace/SomeClass.php
Rector v0.5.7
Config file: rector.yaml

 3/3 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

1 file with changes
===================

1) src/SomeNamespace/SomeClass.php

    ---------- begin diff ----------
--- Original
+++ New
@@ -23,7 +23,7 @@
      *
      * @param array $options ...
      */
-    abstract public function buildHeaders(array $options = []): array;
+    public abstract function buildHeaders(): array;

     /**
      * ...
@@ -33,7 +33,7 @@
      *                       ...
      * @param array $options ...
      */
-    abstract public function formatData(array $data, array &$headers, array $options = []): array;
+    public abstract function formatData(): array;

     /**
      * Retourne un object de réponse au format csv
    ----------- end diff -----------

Applied rectors:

 * Rector\DeadCode\Rector\ClassMethod\RemoveUnusedParameterRector
```

TomasVotruba:
This should not happen.

TomasVotruba:
This looks like a duplicate of the other issue. What's different?

gnutix:
> This looks like a duplicate of your other issue. What's different?

Which issue are you talking about?

TomasVotruba:
> Which issue are you talking about?

There are much more issues now, so I'm not sure anymore, sorry.



On closer look this behaviors is in php-parser: https://github.com/nikic/PHP-Parser/blob/3f718ee2c3a22b656773476f19dd8d72a50e2413/lib/PhpParser/PrettyPrinterAbstract.php#L1042-L1049

If you need to change it, report the issue there

[...]

PR #1701, opened 2019-07-06
Make sure parameter is not removed when a child class does use the parameter

jeroensmit:


jeroensmit:
Fixes part of #1692

TomasVotruba:
Very nice work :clamp:

PR #1702, opened 2019-07-06
Fixed some issues for RemoveZeroAndOneBinaryRector

jeroensmit:
Fixed issue where '1 / x' is considered to be equal to 'x'
Fixed issue where '($var + 1)' inside another expression is considered to be equal to '1'
Expanded unit test for ValueResolver

### DIFF
--- a/packages/DeadCode/src/Rector/Plus/RemoveZeroAndOneBinaryRector.php
+++ b/packages/DeadCode/src/Rector/Plus/RemoveZeroAndOneBinaryRector.php
@@ -105,9 +105,11 @@ CODE_SAMPLE
      */
     private function processBinaryMulAndDiv(BinaryOp $binaryOp): ?Expr
     {
-        if ($this->isValue($binaryOp->left, 1)) {
-            if ($this->isNumberType($binaryOp->right)) {
-                return $binaryOp->right;
+        if ($binaryOp instanceof Mul) {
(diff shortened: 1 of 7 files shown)


packages/DeadCode/src/Rector/Plus/RemoveZeroAndOneBinaryRector.php
```diff
@@ -105,9 +105,11 @@ private function processBinaryPlusAndMinus(BinaryOp $binaryOp): ?Expr
      */
     private function processBinaryMulAndDiv(BinaryOp $binaryOp): ?Expr
     {
-        if ($this->isValue($binaryOp->left, 1)) {
-            if ($this->isNumberType($binaryOp->right)) {
-                return $binaryOp->right;
+        if ($binaryOp instanceof Mul) {
```
TomasVotruba: This will skip all the `Div`, right?

Or will it change:

```diff
-5/1
+5
```

jeroensmit: it will only skip the div for 1/5:
$binaryOp->left == 1 
$binaryOp->right == 5

TomasVotruba: Could you add test case for it, just to be sure?

TomasVotruba:
Thank you :+1:

PR #1707, opened 2019-07-08
Don't mess with lines between docblock comment and var type.

ravanscafi:
Previously, this code:

```php
/**
 * My comment.
 *
 * @var array
 */
public const CONST = [
    'string'
];
```

would become:

```php
/**
 * My comment.
 * @var string[]
 */
public const CONST = [
    'string'
];
```
messing with code standards/style (removing the line between comment and `@var`).

Now it stays as is:

```php
/**
 * My comment.
 *
 * @var string[]
 */
public const CONST = [
    'string'
];
```

(without breaking anything else)


TomasVotruba:
Thanks

PR #1709, opened 2019-07-08
[FuncCall] Don't add $result to parse_str if second parameter is already set

ravanscafi:
Before this fix, `parse_str($query, $data);` would turn into `parse_str($query, $result);` by no reason, breaking valid code.

### DIFF
--- a/packages/Php/tests/Rector/FuncCall/ParseStrWithResultArgumentRector/Fixture/already_set.php.inc
+++ b/packages/Php/tests/Rector/FuncCall/ParseStrWithResultArgumentRector/Fixture/already_set.php.inc
@@ -1,0 +1,29 @@
+<?php
+
+namespace Rector\Php\Tests\Rector\FuncCall\ParseStrWithResultArgumentRector\Fixture;
+
+class AlreadySet
+{
+    public function run()
+    {
+        $query = 'bla';
+        parse_str($query, $data);
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\Php\Tests\Rector\FuncCall\ParseStrWithResultArgumentRector\Fixture;
+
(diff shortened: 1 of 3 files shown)


packages/Php/tests/Rector/FuncCall/ParseStrWithResultArgumentRector/Fixture/already_set.php.inc
```diff
@@ -0,0 +1,29 @@
...
+class AlreadySet
+{
+    public function run()
+    {
+        $query = 'bla';
+        parse_str($query, $data);
+    }
+}
+
+?>
```
TomasVotruba: If the code should have no change, you can keep only the first part (above `----`):

```diff
<?php

namespace Rector\Php\Tests\Rector\FuncCall\ParseStrWithResultArgumentRector\Fixture;

class AlreadySet
{
    public function run()
    {
        $query = 'bla';
        parse_str($query, $data);
    }
}
?>
------
-<?php
-
-namespace Rector\Php\Tests\Rector\FuncCall\ParseStrWithResultArgumentRector\Fixture;
-
-class AlreadySet
-{
-    public function run()
-    {
-        $query = 'bla';
-        parse_str($query, $data);
-    }
-}
-?>
```

TomasVotruba: Also, the standard here is  to prefix these test cases with `skip_`.
That way we can group them visually by *fix* and *skip* groups

Here `skip_already_set.php.inc`.

TomasVotruba:
On second though, these details are not related to the feature and can be cleared later in chunk, so thanks for the PR! Mregin

Issue #1400, opened 2019-05-07
Allow other namespaces for @Inject in JmsInjectAnnotationRector


holtkamp:
I was checking whether I could use rector to change my [PHP-DI @Inject annotation-based](http://php-di.org/doc/annotations.html#inject) dependency injections to constructor based injections.

To my big surprise I found out about an already existing rector:

https://github.com/rectorphp/rector/blob/master/docs/AllRectorsOverview.md#jmsinjectannotationrector

The only difference seems to be that the PHP-DI `@Inject` annotation does not have/use a namespace. That made me wonder, instead of making a new rector, would it be a (good) idea to change https://github.com/rectorphp/rector/blob/fac1503915aea843a3c123a5c622ecf0708bda62/packages/Jms/src/Rector/Property/JmsInjectAnnotationRector.php in such a way that the namespace of the annotation can be configured / left empty, probably defaulting to https://github.com/rectorphp/rector/blob/fac1503915aea843a3c123a5c622ecf0708bda62/packages/Jms/src/Rector/Property/JmsInjectAnnotationRector.php#L28

TomasVotruba:
Hi,

so it's basically the very same change just with different annotation?

What will happen when you change it in `JmsInjectAnnotationRector` manually?

If that's what helps, we can make the rule configurable.

holtkamp:
>What will happen when you change it in JmsInjectAnnotationRector manually?

Will give it a shot and report back here, this week is quite busy, so might take some time.

holtkamp:
I finally found some time to try this out.

Using rector `0.5.5` and the following `rector.yaml` file:

```yaml
# rector.yaml
services:
  Rector\Rector\Property\InjectAnnotationClassRector:
    $annotationClass: 'Inject'
```

This results in an error:

>Make sure "kernel_class" parameters is set in rector.yaml in "parameters:" section

Which seems to indicate this functionality is only available for Symfony applications?

When the property type indicates that it is a FQCN (using `class_exists()`?), this type could probably be used directly instead of having the [`AnalyzedApplicationContainer` determine the type](https://github.com/rectorphp/rector/blob/2883e39869fc527651124ad78f9363c8751196e4/src/Rector/Property/InjectAnnotationClassRector.php#L153). This way it should also work for "non-Symfony" applications.

Not sure this is possible, just thinking out loud 😄 


```php
class Example {

  /**
   * @Inject
   * @var \Fully\Qualified\Class\Name\To\Dependency
   */
[...]

PR #1711, opened 2019-07-09
Do not mark injected properties as private when moved to constructor

holtkamp:
As suggested in https://github.com/rectorphp/rector/issues/1400#issuecomment-508336071

### DIFF
--- a/src/Rector/Property/InjectAnnotationClassRector.php
+++ b/src/Rector/Property/InjectAnnotationClassRector.php
@@ -195,9 +195,6 @@ CODE_SAMPLE
 
         $this->docBlockManipulator->removeTagFromNode($property, $annotationClass);
 
-        // set to private
-        $property->flags = Class_::MODIFIER_PRIVATE;
-
         $classNode = $property->getAttribute(AttributeKey::CLASS_NODE);
         if (! $classNode instanceof Class_) {
             throw new ShouldNotHappenException();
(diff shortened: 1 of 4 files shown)


TomasVotruba:
What about classes with no children?

holtkamp:
As mentioned here https://github.com/rectorphp/rector/issues/1400#issuecomment-508675676, I do not think "detecting" this is the responsibility of this rector...? Better to have a lot of small and focussed rectors and let them work together, right?

TomasVotruba:
Great approach and usually I'd agree. 

I try to look at this from users perspective - what would they change in code at one go?
Would they not care about visibility?

holtkamp:
Well, from my user-perspective. In case I have marked my `property` as public or `protected`, this has a reason...

Of course, in case rector can determine properly that the visibility can be set to `private`, this would be great. However: I do not know how to determine that reliable, so I choose the easy way 😄 

In case it is possible for rector to determine whether:
- a `public` property is used by by other code
- a `protected` property is used by specializing classes

=> then this could be added. But I "assumed" this would be possibly get quite complex?

TomasVotruba:
I see. Thanks for explaning your point of view. It helped me to improve my pespective :)

Let's try your idea and see how it goes.

TomasVotruba:
Thanks for the PR!

PR #1713, opened 2019-07-09
Remove `--with-style` in favour of mentioning ECS

### LINKED ISSUE #1683: [README] Add "big" note about why there are both FCQN in inject & imports
As per comment: https://github.com/rectorphp/rector/issues/1564#issuecomment-499146022

> The best ways to solve this is using coding standard to handle your imports.
Using FQN by default is to allow them import or keep FQN - some people prefer first, some second. And it's harder to FQNize than import in coding standard rules.
>
> tl;dr; AST sucks in coding standard, and always will. It doesn't make sense to rewrite EasyCodingStandard, PHP CS Fixer or CodeSnffier. Just use it.
>
> See this comment for deeper answer: #1456 (comment)
>
> I recommend using option --with-style: https://github.com/rectorphp/rector#how-to-apply-coding-standards

This IMHO should be hardly noted in `README` file, so there is clear point why it's this way & why it will not be fixed.

stloyd:
List reasons why Rector doesn't work well with coding standards instead.

Closes #1683.


README.md
```diff
@@ -59,6 +59,10 @@ Rector **instantly upgrades and instantly refactors PHP code of your application
 
 ...**look at overview of [all available Rectors](/docs/AllRectorsOverview.md)** with before/after diffs and configuration examples. You can use them to build your own sets.
 
+## How to Apply Coding Standards?
+
+AST libraries that Rector use, doesn't work well with coding standards, so it's better to let coding standard tools do that. Your project doesn't have one? Consider usage of i.e. [EasyCodingStandard](https://github.com/Symplify/EasyCodingStandard), that covers namespaces import, 1 empty line between class elements & many more _boring_ stuff that cs-fixer tool should do for you.
```
TomasVotruba:
```suggestion
AST libraries that Rector use, doesn't work well with coding standards, so it's better to let coding standard tools do that. Your project doesn't have one? Consider usage of i.e. [EasyCodingStandard](https://github.com/Symplify/EasyCodingStandard) or your custom set.
```

TomasVotruba: The list actually referects to local `ecs-after-rector.yaml` set, not the tool.

TomasVotruba:
Very nice job with the removal :+1:

Issue #1724, opened 2019-07-10
is_float() replaced with is_real()


holtkamp:
| Subject        | Details                                                      |
| :------------- | :----------------------------------------------------------- |
| PHP version    | PHP 7.3.7                              |
| Full Command   | `./vendor/bin/rector process src --level php74 --debug`  | 

## Current Behaviour 

`is_float()` is replaced with `is_real()`

```
    ---------- begin diff ----------

-            'fixed-base-price' => is_float($this->fixedBasePrice) ? (string) $this->fixedBasePrice : $this->fixedBasePrice,
+            'fixed-base-price' => is_real($this->fixedBasePrice) ? (string) $this->fixedBasePrice : $this->fixedBasePrice,

    ----------- end diff -----------

Applied rectors:

 * Rector\Rector\Function_\RenameFunctionRector

```

## Minimal PHP Code Causing Issue

Just use a `is_float()`

## Expected Behaviour

No replacements should be done for `is_float()`. 

`is_real()` SHOULD be replaced with `is_float()`

Probably caused by: https://github.com/rectorphp/rector/pull/905/files#r302064547
Also see: https://wiki.php.net/rfc/deprecations_php_7_4

holtkamp:
Resolved in https://github.com/rectorphp/rector/pull/1728

TomasVotruba:
Tip for closing keywords in PR description:
https://help.github.com/en/articles/closing-issues-using-keywords

holtkamp:
Ah, thanks for the feedback!

PR #1728, opened 2019-07-11
is_real() is deprecated instead of is_float()

holtkamp:
As reported  in https://github.com/rectorphp/rector/issues/1724


TomasVotruba:
Thank you :+1:

Issue #1734, opened 2019-07-13
Support reference symbol in phpdoc


tigitz:
It seems that putting an `&` in the docblock of a parameter fails the `Rector\TypeDeclaration\Rector\FunctionLike\ParamTypeDeclarationRector`

I guess it would be a new feature to support this.

However even if this `&` is common in popular codebases it's not supported by phpdoc nor the PSR-5

I let you decide.

Some food for thoughts: https://github.com/phpstan/phpstan/issues/699
## Diff 

```diff
<?php

    /**
    * @param array &$array
    */
-    function test(&$array)
+    function test(array &$array)
    {

    };
```

TomasVotruba:
Why not! Send PR with failing test case to `ParamTypeDeclarationRectorTest`

PR #1735, opened 2019-07-13
Consider reference symbol in docblock for param type declaration rector

tigitz:
Linked to #1734 

The idea here would be to ignore the `&` during parsing. Unfortunately in this case, rector depends on phpstan doc parser which maintainers decided not to support this type of documenting: https://github.com/phpstan/phpstan/issues/699

I've dug a little and found that adding this line:
```php
$tokens->tryConsumeTokenType(Lexer::TOKEN_REFERENCE);
```

After this line: 
https://github.com/phpstan/phpdoc-parser/blob/master/src/Parser/PhpDocParser.php#L166

Would effectively ignore it and make the test green.

I'm wondering if there could be any workaround that can be put in rector code directly.


TomasVotruba:
I'll look at it, thanks for failing test and propsed solution :+1:

PR #1744, opened 2019-07-15
add reference support to ParamTypeDeclarationRector

### LINKED ISSUE #1734: Support reference symbol in phpdoc

TomasVotruba:
Close #1734

### DIFF
--- a/packages/BetterPhpDocParser/src/Attributes/Ast/AttributeAwareNodeFactory.php
+++ b/packages/BetterPhpDocParser/src/Attributes/Ast/AttributeAwareNodeFactory.php
@@ -96,6 +96,10 @@ final class AttributeAwareNodeFactory
             return $this->createFromPhpDocValueNode($node);
         }
 
+        if ($node instanceof TypeNode) {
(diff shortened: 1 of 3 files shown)


TomasVotruba:
@tigitz What do you think about this?

tigitz:
@TomasVotruba awesome! Thanks, I almost feel bad requiring that amount of complexity to handle such a specific use case :sweat_smile:

TomasVotruba:
If it becomes a performance issue, we can always improve it :)

That's the price of open source. Some people don't care about problem, so you have to take a way around it.

PR #1747, opened 2019-07-16
Move annotation-based property dependency injections to constant array

holtkamp:
As briefly introduced in https://github.com/rectorphp/rector/issues/1400#issuecomment-510798349

This rector allows to collect all properties that use an `@Inject` annotation in an associative array assigned to a constant. Constants are used to ensure the arrays can be *merged* statically, which would not be possible with properties:

```php
protected const DEPENDENCIES = parent::DEPENDENCIES + ['childProperty' => \Some\Dependencies::class];
```

This array can be used during instantiation of the class to actually inject the dependencies. This is particularly useful in situations where the signature of the constructor can not be changed and therefore does not allow constructor-based dependency injection, for example for Zend Framework 1.

The actual dependency injection is not part of this rector but could look something like:

```php
class Controller
{
   protected const DEPENDENCIES = [
      'someProperty' = \Some\Class\To\Dependency::class,
   ];

   /**
    * @var \Some\Class\To\Dependency
    */
   private $someProperty;

   public function __construct(\Some\Dependency\Required\By\Framework $dependency)
   {
      parent::__construct($dependency);
      $this->injectPropertyDependencies();
   }

   private function injectPropertyDependencies() : void
   {
      foreach(self::DEPENDENCIES as $propertyName => $propertyType){
         $this->$propertyName = $this->container->get($propertyType);
      }
   }
}
```

### DIFF
--- a/src/Rector/Property/InjectAnnotationsToConstantRector.php
+++ b/src/Rector/Property/InjectAnnotationsToConstantRector.php
@@ -1,0 +1,297 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Rector\Property;
+
+use function array_unshift;
+use Nette\Utils\Strings;
+use PhpParser\Node;
+use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\Property;
+use PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocTagNode;
+use Rector\Application\ErrorAndDiffCollector;
+use Rector\Bridge\Contract\AnalyzedApplicationContainerInterface;
+use Rector\Exception\ShouldNotHappenException;
+use Rector\NodeTypeResolver\Node\AttributeKey;
+use Rector\NodeTypeResolver\PhpDoc\NodeAnalyzer\DocBlockManipulator;
+use Rector\Rector\AbstractRector;
+use Rector\RectorDefinition\ConfiguredCodeSample;
+use Rector\RectorDefinition\RectorDefinition;
+use Symplify\PackageBuilder\FileSystem\SmartFileInfo;
+use Throwable;
+
+final class InjectAnnotationsToConstantRector extends AbstractRector
+{
+    /**
+     * @var string
+     */
+    private $constantNameThatContainsDependencyMap;
+
+    /**
+     * @var DocBlockManipulator
+     */
+    private $docBlockManipulator;
+
+    /**
+     * @var AnalyzedApplicationContainerInterface
+     */
+    private $analyzedApplicationContainer;
+
+    /**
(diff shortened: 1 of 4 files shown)


TomasVotruba:
Hi, how do you find this useful to general public?

holtkamp:
Well, I don't think the *general* public will find it useful.

This will be useful for people trying to (gradually) migrate away from:
- a framework which involves a FrontController with a constructor with a specific signature
  - ZendFramework 1
  - Symfony 2
  - ... ?
- using property-based dependency injection using `@Inject` annotations

So I guess people who used this:

- http://php-di.org/doc/best-practices.html#writing-controllers
- http://php-di.org/doc/frameworks/zf1.html
- https://github.com/PHP-DI/Zend-Framework-Bridge

Feel free to close this PR in case it is not appropriate. But maybe you can use it as inspiration to see how "tedious" it is to "simply" add a constant to a class 😄 , some helper functions in Rector might be useful here.

TomasVotruba:
I'm not sure I understand the change of code. How would minimal diff code, that would show your idea, would look like?

holtkamp:
Mm, there is a code sample:

https://github.com/rectorphp/rector/pull/1747/files#diff-eb420a4f5867f37714234df3080065e7R79

And a test:

https://github.com/rectorphp/rector/pull/1747/files#diff-cd49a700fc66ca6ddeffad024bf4e7b4R21

Is that not clear enough?

Class-properties with an (`@Inject`) annotation will be collected in an array.

TomasVotruba:
Ah, so it remove `@inject` and creates some random constant?

holtkamp:
The used constant name is configurable, just as the name of the annotation.

TomasVotruba:
Then what happens with this constant?

holtkamp:
Then it can be used by some mechanism to actually perform the dependency injection, see the last section of https://github.com/rectorphp/rector/pull/1747#issue-298163293

TomasVotruba:
I see. Still not sure how anyone can use this. It looks like breaking code, while moving from one service locator to anothre.

holtkamp:
I used it to get rid of @Inject annotations in a legacy ZF1 framework application in a framework/technology-agnostic way, so I thought it would be nice to share. But feel free to close this pull request 😉

TomasVotruba:
I migrate Zend 1 projects directly to Symfony 4.4 constructor injection, that's why I can't related to this.

If you think it might be useful to other for this case, feel free o create own repo with rules related to Zend 1 upgrade. As Drupal 8 has: https://github.com/drupal8-rector/drupal8-rector. If it will be appreciated by community, I'll be happy to add it here. 

Closing for now

PR #1757, opened 2019-07-21
make SymfonyContainer factory configurable with "kernel_environment" parameter in rector.yaml

TomasVotruba:
```yaml
# rector.yaml
paramters:
    # configuratioon for analyzed Symfony application (not Rector itself)
    kernel_environment: "test"
```

### DIFF
--- a/src/Configuration/Option.php
+++ b/src/Configuration/Option.php
@@ -27,12 +27,12 @@ final class Option
     /**
      * @var string
      */
-    public const OPTION_LEVEL = 'level';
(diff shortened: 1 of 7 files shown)


src/Configuration/Option.php
```diff
@@ -27,12 +27,12 @@ final class Option
     /**
      * @var string
      */
-    public const OPTION_LEVEL = 'level';
```
TomasVotruba: Never used

PR #1764, opened 2019-07-25
[Symfony] Use Symfony bridge interface for `doctrine` service

stloyd:
Autowire is using `Symfony\Bridge\Doctrine\RegistryInterface` not the Doctrine one.


TomasVotruba:
Thanks

PR #1773, opened 2019-07-28
add original doc types to AbstractTypeInfo

TomasVotruba:


TomasVotruba:
Not needed in the end

PR #1793, opened 2019-08-03
Break class name in @var when relation is defined in same namespace

snapshotpl:
Doctrine allow to define relation in same namespace without FQDN. However after rector process namespace is broken.

### DIFF
--- a/packages/TypeDeclaration/tests/Rector/Property/PropertyTypeDeclarationRector/Fixture/doctrine_relation.php.inc
+++ b/packages/TypeDeclaration/tests/Rector/Property/PropertyTypeDeclarationRector/Fixture/doctrine_relation.php.inc
@@ -59,7 +59,7 @@ class DoctrineRelation
 
     /**
      * @ORM\OneToOne(targetEntity="DoctrineRelated")
-     * @var DoctrineRelated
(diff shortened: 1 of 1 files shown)


packages/TypeDeclaration/tests/Rector/Property/PropertyTypeDeclarationRector/Fixture/doctrine_relation.php.inc
```diff
@@ -44,6 +57,12 @@ class DoctrineRelation
      */
     private $products;
 
+    /**
+     * @ORM\OneToOne(targetEntity="DoctrineRelated")
+     * @var DoctrineRelated
```
TomasVotruba:
```suggestion
     * @var \Rector\TypeDeclaration\Tests\Rector\FunctionLike\PropertyTypeDeclarationRector\Fixture\DoctrineRelated
```

This should be FQN by default, that's convention for all changed imports here to have a consistent output

snapshotpl:
I provide falling test and feel free to find solution how to fix it @TomasVotruba :rescue_worker_helmet:

TomasVotruba:
Thanks!

PR #1810, opened 2019-08-05
add failing test for doctrine decimal type

bendavies:


TomasVotruba:
Could you send fix as well?

Issue #1746, opened 2019-07-16
What can I do, if 3rd party function uses func_get_args, but RemoveExtraParametersRectors want's to remove them.


przemo-hemmersbach:
## Question
Accordingly to this [php variadics rfc](https://wiki.php.net/rfc/variadic) `::isVariadic` may not properly recognize variadic functions if they are using old style `func_get_args`.
`RemoveExtraParametersRector` will remove those arguments at call sites.

What workaround could I use to whitelist particular functions in 3rd party libs, while still benefit from this rector in functions I have a control over.

PS Question, since that's a "bug" in php reflection implementation if anything. Can create full bug report if advised to do so.

TomasVotruba:
Hi, add PR with failing test case and I'll look at it.
It should be fixable

TomasVotruba:
Closing for missing faling test, see https://github.com/rectorphp/rector/pull/1811#issuecomment-538743867

PR #1811, opened 2019-08-05
added failing `func_get_arg()` testcase for RemoveExtraParametersRector

staabm:
adding a failing testcase for https://github.com/rectorphp/rector/issues/1746 since we are also experiencing the mentioned problem.

hopefully I did everything right - my first PR to this project.
In case I missing something I am willing to improve the testcase with you guidance.

### DIFF
--- a/packages/Php/tests/Rector/FuncCall/RemoveExtraParametersRector/Fixture/func_get_arg.php.inc
+++ b/packages/Php/tests/Rector/FuncCall/RemoveExtraParametersRector/Fixture/func_get_arg.php.inc
@@ -1,0 +1,35 @@
+<?php
+
+namespace Rector\Php\Tests\Rector\FuncCall\RemoveExtraParametersRector\Fixture;
+
+class FuncGetArg
+{
+    public function __construct() {
+        self::("aa", "bb", "cc");
+    }
+    
+    static public function dumpArgs() {
+        $arg_list = func_get_args();
+        var_dump($arg_list);
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\Php\Tests\Rector\FuncCall\RemoveExtraParametersRector\Fixture;
+
+class FuncGetArg
+{
+    public function __construct() {
+        self::("aa", "bb", "cc");
+    }
+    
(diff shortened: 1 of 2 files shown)


packages/Php/tests/Rector/FuncCall/RemoveExtraParametersRector/Fixture/func_get_arg.php.inc
```diff
@@ -0,0 +1,35 @@
...
+?>
+-----
+<?php
+
+namespace Rector\Php\Tests\Rector\FuncCall\RemoveExtraParametersRector\Fixture;
+
+class FuncGetArg
+{
+    public function __construct() {
+        self::("aa", "bb", "cc");
```
staabm: I expect rector not to strip this call-site arguments, because the method beeing called uses some `func_get_args` magic to handle them

staabm:
hmm very interessting.. I can see rector stripping arguments for my function throughout my project... but its not reproducible as a simple single file test

TomasVotruba:
Why closing?
Resolved?

staabm:
I was not able to build a small repro. It still happens on my projects codebass

TomasVotruba:
> but its not reproducible as a simple single file test

Oh, I see it now. So the test case was passing.

No need to build small repo. If you create local directory with:

```php
composer require rector/rector
touch src/SomeCode.php # the failing code
```

withc

```yaml
# rector.yaml
services:
    Rector\Php\Rector\FuncCall\RemoveExtraParametersRector: ~
```

and run Rector that fails:

```
vendor/bin/rector process src
```

Just paste the `src/SomeCode.php` here. I'll run it myself locally and fix it

staabm:
ok thx. so in https://github.com/rectorphp/rector/pull/1811#issuecomment-518189309 is all you need from me?

gnutix:
@TomasVotruba I guess this one should be re-opened? Otherwise we'll lose track of it.

TomasVotruba:
This report needs failing test. Without it there is nothing anyone can do of it

PR #1817, opened 2019-08-06
Fix analysed files for PHPStan scope resolver

TomasVotruba:


mglaman:
What if we want to persist fully qualified class names in our parameters or return types? Rector is now changing code we don't want it to.

Issue #1809, opened 2019-08-05
Rector errors with anon class


bendavies:
| Subject        | Details                                                      |
| :------------- | :----------------------------------------------------------- |
| PHP version    | e.g. PHP 7.3                            |
| Full Command   | e.g. vendor/bin/rector process Foo.php  | 

## Current Behaviour 

```
 [ERROR] Could not process "Foo.php" file, due to:
         "Call to a member function getMethod() on null".

```

## Minimal PHP Code Causing Issue

```yaml
# rector.yaml
services:
    Rector\TypeDeclaration\Rector\Property\PropertyTypeDeclarationRector: ~
```

```php
class Foo
{
    private function makeClass(int $bar = 1)
    {
        $class =  new class() {
            public $bar;
        };
        $class->bar = $bar;
    }
}

```

## Expected Behaviour

no error, and:

```php
class Foo
{
    private function makeClass(int $bar = 1)
    {
        $class =  new class() {
            /** @var int */
            public $bar;
        };
        $class->bar = $bar;
    }
}

```

TomasVotruba:
Send failing test case

TomasVotruba:
Anonymous classes are very hard to analyse reliable, moreover when it comes to public scope, like public property assign. For that, be better skip it

PR #1820, opened 2019-08-06
add failing testcase for anon classes

bendavies:
https://github.com/rectorphp/rector/issues/1809#issuecomment-518270057


TomasVotruba:
Full feature is needed, I don't have free time for resolving these anymore

Issue #1816, opened 2019-08-05
Resolving constant fetches is unsafe


scheb:
| Subject        | Details                                                      |
| :------------- | :----------------------------------------------------------- |
| PHP version    | PHP 7.3.7                               |
| Full Command   | vendor/composer/bin/rector process src test -vvv | 

```
parameters:
    php_version_features: '7.2'
    autoload_paths:
        - src/autoload.php
services:
    Rector\SOLID\Rector\ClassConst\PrivatizeLocalClassConstantRector: ~
    Rector\Php\Rector\ClassConst\PublicConstantVisibilityRector: ~
```

## Current Behaviour 

Some of my constants are wrongly assigned with a `private` visibility, while there was clear public usages in the code. Recognized that the parsing order actually matters. If the PHP file that declares the constant is parsed first, everything is fine. If the file(s) with the usage come first and then the declaring file at the end, the usages won't be recognized.

## Minimal PHP Code Causing Issue

The problem seems to be located in `ParsedNodesByType::collect()`, which resolves the usages on-the-fly, but at that point the file with the constant declaration hasn't necessarily been parsed yet.

Call stack (line numbers might be off, was playing around with the code):
```
#0  Rector\NodeContainer\ParsedNodesByType->matchClassTypeThatContainsConstant() called at [project/vendor/composer/rector/rector/src/NodeContainer/ParsedNodesByType.php:514]
#1  Rector\NodeContainer\ParsedNodesByType->addClassConstantFetch() called at [project/vendor/composer/rector/rector/src/NodeContainer/ParsedNodesByType.php:392]
#2  Rector\NodeContainer\ParsedNodesByType->collect() called at [project/vendor/composer/rector/rector/packages/NodeTypeResolver/src/NodeVisitor/NodeCollectorNodeVisitor.php:27]
#3  Rector\NodeTypeResolver\NodeVisitor\NodeCollectorNodeVisitor->enterNode() called at [project/vendor/composer/nikic/php-parser/lib/PhpParser/NodeTraverser.php:123]
#4  PhpParser\NodeTraverser->traverseNode() called at [project/vendor/composer/nikic/php-parser/lib/PhpParser/NodeTraverser.php:146]
```

You end up in the continue here:
https://github.com/rectorphp/rector/blob/b31b99a3aea1ba86467aa6c26ca5e971e6a1b0c0/src/NodeContainer/ParsedNodesByType.php#L446-L449
[...]

PR #1822, opened 2019-08-07
Add test for #1816: Resolving constant fetches is unsafe 

scheb:
Test case to demonstrate issue #1816.

1) The constant usages must be parsed before the class declaring the constant.
2) The class declaring the constant must be a subclass.

rector is then forced find the declaring class from the class hierarchy, which it can't when the declaring class and its superclasses haven't been parsed yet.

### DIFF
--- a/src/NodeContainer/ParsedNodesByType.php
+++ b/src/NodeContainer/ParsedNodesByType.php
@@ -70,11 +70,6 @@ final class ParsedNodesByType
     private $classConstantFetches = [];
 
     /**
-     * @var string[][][]
-     */
-    private $classConstantFetchByClassAndName = [];
-
-    /**
      * @var NodeTypeResolver
      */
     private $nodeTypeResolver;
@@ -298,13 +293,11 @@ final class ParsedNodesByType
     }
 
     /**
-     * @return string[]|null
+     * @return ClassConstFetch[]
      */
-    public function findClassConstantFetches(string $className, string $constantName): ?array
+    public function getClassConstantFetches(): array
     {
-        $this->processClassConstantFetches();
-
-        return $this->classConstantFetchByClassAndName[$className][$constantName] ?? null;
+        return $this->classConstantFetches;
     }
 
     public function findFunction(string $name): ?Function_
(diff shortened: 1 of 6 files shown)


scheb:
Pushed a suggestion how to fix it. Not sure if it's the way to go, but it solves the issue.

Idea is to collect all `ClassConstFetch` first (without processing them) and only process them when `findClassConstantFetches()` is called, which should be after all classes have been collected.

src/NodeContainer/ParsedNodesByType.php
```diff
@@ -484,6 +491,21 @@ private function addClassConstant(ClassConst $classConst): void
...
+    {
+        if ($this->classConstantFetches) {
+            foreach ($this->classConstantFetches as $rawClassConstantFetch) {
+                $this->processClassConstantFetch($rawClassConstantFetch);
+            }
+            $this->classConstantFetches = [];
+        }
+    }
+
+    private function processClassConstantFetch(ClassConstFetch $classConstFetch): void
```
TomasVotruba: What does "process" stand for? Better naming would help to understand

src/NodeContainer/ParsedNodesByType.php
```diff
@@ -484,6 +491,21 @@ private function addClassConstant(ClassConst $classConst): void
...
+        $this->classConstantFetches[] = $classConstFetch;
+    }
+
+    private function processClassConstantFetches(): void
+    {
+        if ($this->classConstantFetches) {
+            foreach ($this->classConstantFetches as $rawClassConstantFetch) {
+                $this->processClassConstantFetch($rawClassConstantFetch);
+            }
+            $this->classConstantFetches = [];
```
TomasVotruba: Why is the proprety resetted?

src/NodeContainer/ParsedNodesByType.php
scheb: It's processing the `ClassConstFetch` items that have been collected during `collect`, resolving them to their respective classes. Everything that was previously done in the `addClassConstantFetch` call. Maybe "resolve" or "evaluate" is the better name?

src/NodeContainer/ParsedNodesByType.php
scheb: So that this is only executed once on the first call to `findClassConstantFetches`. Basically lazy-loading the data into `$classConstantFetchByClassAndName`. Once this is done we don't need to do it again, that's why `$this->classConstantFetches` is reset to an empty value.

TomasVotruba: Could you think of better to add class constant fetches to prevent these time-dependency?

scheb: I think the only proper way to fix it would be to introduce another step between "parsing" and "refactoring". In that step all the work for resolving dependencies would be done, once parsing has completed and therefore all information about classes/methods/properties/constants has been collected. The current approach is doing this during `collect` and implictly relies on parsing order.

TomasVotruba: I'm trying to say that we have all we need now.

Look at: https://github.com/rectorphp/rector/blob/master/packages/DeadCode/src/Analyzer/SetterOnlyMethodAnalyzer.php#L68

It gets all the setter only methods just once before any refactoring, but after the parsing.

scheb: Ah, I see. So this is the more reliable and cleaner approach to process information after parsing. I think I can refactor it into following that implementation pattern. I'll have a look at it this evening. Thanks for the hint!

scheb:
Pushed update following the same implementation pattern as SetterOnlyMethodAnalyzer.

TomasVotruba:
Looks good, thanks :+1:

PR #1840, opened 2019-08-11
Constants declared in interfaces have to be public

scheb:
Constants declared in an interface have to be `public`, having `private` or `protected` constants in interfaces is not supported by the language. PHP will error with `PHP Fatal error:  Access type for interface constant SomeInterface::CONSTANT must be public`.

So whenever a constant is defined in an interface, we can just make it `public`.


TomasVotruba:
Good to know. Do you have any source about that language feature?

scheb:
Couldn't find anything in the official docs, but it's in the PHP source code: https://github.com/php/php-src/blob/master/Zend/zend_API.c#L3786

And it makes sense, since an interface is meant to declare a public contract to be implemented by a class, so allowing private/protected constants there doesn't make much sense.

TomasVotruba:
To bad it's so hidden.
Thanks for such a research :+1:

TomasVotruba:
Thank you for contribution with fix :+1:

PR #1842, opened 2019-08-11
Overriding constants require at least the parent's visibility

scheb:
When a constant is overridden in a subclass, it must have at least the access level of the superclass' constant. E.g. when the parent declares the constant `public`, the subclass _must_ also declare it as `public`, regardless if it's used or not.

It is possible to weaken the access level of a constant in a subclass, e.g. the superclass declares it as `private`, then the subclass can declare it `protected` or `public`. It's not allowed to have a more restrictive access level in a subclass tough.

If the access level of an overriding constant is not allowed, the PHP parser errors with:
`PHP Fatal error:  Access level to Subclass::OVERRIDDEN_CONSTANT must be [private|protected] (as in class Superclass::OVERRIDDEN_CONSTANT) or weaker in ... on line ...`

So what does that mean for rector:

Whenever it recognizes that a constant of the same name is declared in a superclass (not sure how to do that), it must take that constant's visibility into account (also not sure how to do that, if it's not explicitly declared, rector would first need to do the visibility evaluation for the superclass constant).

This is the same rules that also apply to class properties btw., so there could be a similar issue with property visibility rectors and however the solution to this issue may look like, it could be adopted.

(This is the last constant-related issue I've found with my codebase. Sorry for bothering you about all of these, I blieve if we get them fixed it makes it better for everyone who's dealing with a 5.x style codebase. Keep up the good work and let me know if I can help :) )

### DIFF
--- a/packages/SOLID/src/Rector/ClassConst/PrivatizeLocalClassConstantRector.php
+++ b/packages/SOLID/src/Rector/ClassConst/PrivatizeLocalClassConstantRector.php
@@ -103,50 +103,30 @@ CODE_SAMPLE
 
         /** @var string $constant */
         $constant = $this->getName($node);
-        $classNode = $this->parsedNodesByType->findClass($class);
-        $mustBeAtLeastProtected = false;
+        $parentConstIsProtected = false;
 
-        if ($classNode !== null && $classNode->hasAttribute(AttributeKey::PARENT_CLASS_NAME)) {
-            /** @var string $parentClassName */
-            $parentClassName = $classNode->getAttribute(AttributeKey::PARENT_CLASS_NAME);
-            if ($parentClassName) {
-                $parentClassConstant = $this->parsedNodesByType->findClassConstant($parentClassName, $constant);
-                if ($parentClassConstant) {
-                    $this->refactor($parentClassConstant);
-
-                    // The parent's constant is public, so this one must become public too
-                    if ($parentClassConstant->isPublic()) {
-                        $this->makePublic($node);
-                        return $node;
-                    }
-
-                    // The parent's constant is protected, so this one must become protected or weaker
-                    if ($parentClassConstant->isProtected()) {
-                        $mustBeAtLeastProtected = true;
-                    }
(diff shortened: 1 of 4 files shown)


scheb:
Pushed a fix to this issue. Not pretty yet, but it demonstrates how to approach the issue. Wanted to share this to get some feedback.

Please note this is rebased on top of the changes from #1822 because I needed them to fix this one.

TomasVotruba:
#1822 is merged now, please rebase so we can see better the changes

scheb:
Now rebased on master. Looking forward for some feedback.

packages/SOLID/src/Rector/ClassConst/PrivatizeLocalClassConstantRector.php
```diff
@@ -92,17 +101,50 @@ public function refactor(Node $node): ?Node
...
+                        return $node;
+                    }
+
+                    // The parent's constant is protected, so this one must become protected or weaker
+                    if ($parentClassConstant->isProtected()) {
+                        $mustBeAtLeastProtected = true;
+                    }
+                }
+            }
+        }
```
TomasVotruba: This should be decouple to own method, so code is better readable e.g. `resolveMinimalParentAllowedVisbilityLevel()`, returning `public`, `protected` or `private` string

scheb: Refactored for better readability.

TomasVotruba: Thanks

TomasVotruba:
Thank you for the PR, beautiful work

scheb:
Thank you for maintaining this project. Really impressed how you drive it forward :)

TomasVotruba:
Well, without contributors and patient users who discover bugs and fix them - like you - it would be just a small pet project. I'm happy you find it useful in your own code bases :+1:

PR #1848, opened 2019-08-14
Add a Laravel rector to change all the array_* helpers to Arr::* calls.

tvbeek:
The array_* helpers are deprecated in Laravel 5.8 (See: https://laravel.com/docs/5.8/upgrade#string-and-array-helpers )

The only part that I'm not sure about is that the composer complete-check from https://github.com/rectorphp/rector#how-to-contribute changed more on the docs than about the Rector I just created. If wanted I can reverse that part.

### DIFF
--- a/packages/Laravel/src/Rector/FuncCall/ArrayHelperToIlluminateSupportClassRector.php
+++ b/packages/Laravel/src/Rector/FuncCall/ArrayHelperToIlluminateSupportClassRector.php
@@ -1,0 +1,162 @@
+<?php declare(strict_types=1);
+
+namespace Rector\Laravel\Rector\FuncCall;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Name;
+use Rector\Rector\AbstractRector;
+use Rector\RectorDefinition\CodeSample;
+use Rector\RectorDefinition\RectorDefinition;
+
+/**
+ * @see https://laravel.com/docs/5.8/upgrade#string-and-array-helpers
+ */
+
+/**
+ * @see \Rector\Laravel\Tests\Rector\FuncCall\ArrayHelperToIlluminateSupportClassRector\ArrayHelperToIlluminateSupportClassRectorTest
+ */
+final class ArrayHelperToIlluminateSupportClassRector extends AbstractRector
+{
+    private $helpersToFunction = [
+        'array_add' => 'add',
+        'array_collapse' => 'collapse',
+        'array_divide' => 'divide',
+        'array_dot' => 'dot',
+        'array_except' => 'except',
+        'array_first' => 'first',
+        'array_flatten' => 'flatten',
+        'array_forget' => 'forget',
+        'array_get' => 'get',
+        'array_has' => 'has',
+        'array_last' => 'last',
+        'array_only' => 'only',
+        'array_pluck' => 'pluck',
+        'array_prepend' => 'prepend',
+        'array_pull' => 'pull',
+        'array_random' => 'random',
+        'array_set' => 'set',
+        'array_sort' => 'sort',
+        'array_sort_recursive' => 'sortRecursive',
+        'array_where' => 'where',
+        'array_wrap' => 'wrap',
+    ];
+
+    private $supportClass = '\Illuminate\Support\Arr';
+
+    public function getDefinition(): RectorDefinition
+    {
+        return new RectorDefinition('Use the Illuminate Support Arr class instead of the array helper functions', [
+            new CodeSample(
+                <<<'CODE_SAMPLE'
+class SomeClass
+{
+    public function run($data)
+    {
+        array_add($data, 'field', 'value');
+        array_collapse($data);
+        array_divide($data);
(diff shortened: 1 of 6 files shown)


packages/Laravel/src/Rector/FuncCall/ArrayHelperToIlluminateSupportClassRector.php
```diff
@@ -0,0 +1,162 @@
...
+    public function refactor(Node $node): ?Node
+    {
+        foreach ($this->helpersToFunction as $helperFunction => $classFunction) {
+            if (! $this->isName($node, $helperFunction)) {
+                continue;
+            }
+
+            return new StaticCall(new Name($this->supportClass), new Identifier($classFunction), $node->args);
+        }
+        return $node;
```
sashabeton: If the node hasn't been modified, `null` should be returned (screen from README)
image

packages/Laravel/src/Rector/FuncCall/ArrayHelperToIlluminateSupportClassRector.php
```diff
@@ -0,0 +1,162 @@
...
+    {
+        return [FuncCall::class];
+    }
+
+    /**
+     * @param FuncCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        foreach ($this->helpersToFunction as $helperFunction => $classFunction) {
```
sashabeton: Maybe it would be better to check using `array_key_exists($this->getName($node), $this->helpersToFunction)`?

packages/Laravel/src/Rector/FuncCall/ArrayHelperToIlluminateSupportClassRector.php
tvbeek: Thanks for the feedback, I have changed the function to return null.

packages/Laravel/src/Rector/FuncCall/ArrayHelperToIlluminateSupportClassRector.php
tvbeek: Thanks for the feedback, good point, I have changed the function.

sashabeton:
It seems `getName()` can return `null`.. So condition `($name === null || !array_key_exists(...))` will make it workable!
I'm pleased you paid an attention to my suggestions :)
image

packages/Laravel/src/Rector/FuncCall/ArrayHelperToIlluminateSupportClassRector.php
```diff
@@ -0,0 +1,159 @@
...
+        return [FuncCall::class];
+    }
+
+    /**
+     * @param FuncCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        $name = $this->getName($node);
+        if (!array_key_exists($name, $this->helpersToFunction)) {
```
KartaviK: What if use ternary operator?

```php
$name = $this->getName($node);

return $name && isset($this->helpersToFunction[$name])
    ? new StaticCall(new Name($this->supportClass), new Identifier($this->helpersToFunction[$name]), $node->args)
    : \null
```

sashabeton: Great idea, but it can be enhanced by removing first condition (`isset` supports null index)
The final version can be
```php
$name = $this->getName($node);

return isset($this->helpersToFunction[$name])
    ? new StaticCall(new Name($this->supportClass), new Identifier($this->helpersToFunction[$name]), $node->args)
    : \null
```

KartaviK: Yeah, my variant have a few redundant condition)

tvbeek: The final version noted by @sashabeton is to long, that is why I didn't choose the ternary operator.

Else you get:
```php
return isset($this->helpersToFunction[$name])
    ? new StaticCall(new Name($this->supportClass), new Identifier(
        $this->helpersToFunction[$name]
    ), $node->args)
    : null;
```

I have changed the code to use isset to prevent an error with null and working faster.

TomasVotruba:
Hi, 
thanks for your PR!

How is that different to https://github.com/rectorphp/rector/blob/ae2d85239118b43a10c4e18a9c50519ff792d2bd/config/set/laravel/static/array-str-functions-to-static-call.yaml#L2-L26 ?

tvbeek:
> 
> 
> Hi,
> thanks for your PR!
> 
> How is that different to
> 
> https://github.com/rectorphp/rector/blob/ae2d85239118b43a10c4e18a9c50519ff792d2bd/config/set/laravel/static/array-str-functions-to-static-call.yaml#L2-L26
> ?

Not, I totally missed it 🤦‍♂ 
Sorry

TomasVotruba:
Ah, no troubles.

Well, maybe it should be included in `laravel58.yaml`? What do you think?

tvbeek:
> 
> Well, maybe the config should be included in `laravel58.yaml`? What do you think?

It was already in the import.

TomasVotruba:
All good then :sweat_smile:

PR #1854, opened 2019-08-15
Ignores resource also when type is nullable

tigitz:
Follow up of https://github.com/rectorphp/rector/commit/34e1a8d46588bcaba90b419214f2318b367ff471 that also ignores resource when it's nullable


packages/TypeDeclaration/tests/Rector/FunctionLike/ParamTypeDeclarationRector/ParamTypeDeclarationRectorTest.php
```diff
@@ -16,6 +16,7 @@ public function test(): void
             __DIR__ . '/Fixture/variadic.php.inc',
             __DIR__ . '/Fixture/mixed.php.inc',
             __DIR__ . '/Fixture/resource.php.inc',
+            __DIR__ . '/Fixture/nullable_resource.php.inc',
```
TomasVotruba: Could you name the file `skip_*` → `skip_nullable_resource.php.inc`?

It's more clear from the name, that nothing should be changed

tigitz: Sure

TomasVotruba:
Merci beaucoup!

PR #1869, opened 2019-08-19
Refactoring order creates incompatible return types

scheb:
This test case shows how the order in which files are parsed & refactored can lead to incompatible method signatures on the return type. The rector does `populateChildren`, but the implementation implicity relies on the superclass being processed before its subclasses to ensure return type compatibility. This in combination with the subtype logic leads to this result.

(Note: Actually I have to say I don't really understand why this subtype logic is there, since [return type variance is only supported from PHP 7.4 on](https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters). Earlier versions support [type variance for the "iterable" type](https://www.php.net/manual/de/language.types.iterable.php#language.types.iterable.variance) which is there as the test case, but I have the impression that's the one single case when PHP <7.4 supports return type variance).

I'd guess an approach similar to #1842 would be the solution. Doing it the other way round: instead of refactoring children after refactoring the node itself, the parents should be refactored before the node is refactored. This would ensure the return type decision for the parent class was made by the rector and we can make a correct decision for the node itself.

Bonus objective would be to support refactoring with return type variance when PHP7.4 is set as the language level.

I'll have a look at a potential fix as soon as I find time to have a deeper look.

### DIFF
--- a/packages/TypeDeclaration/tests/Rector/FunctionLike/ReturnTypeDeclarationRector/CovarianceTest.php
+++ b/packages/TypeDeclaration/tests/Rector/FunctionLike/ReturnTypeDeclarationRector/CovarianceTest.php
@@ -21,6 +21,7 @@ final class CovarianceTest extends AbstractRectorTestCase
     public function provideDataForTest(): iterable
     {
         yield [__DIR__ . '/Fixture/nikic/inheritance_covariance.php.inc'];
+        yield [__DIR__ . '/Fixture/nikic/inheritance_covariance_order.php.inc'];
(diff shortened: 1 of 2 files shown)


TomasVotruba:
There's been major refactoring of types from string to PHPStan value object types.
Please rebase and I'll give this priority

scheb:
Added a new test case, which is technically the same as the `inheritance_covariance.php.inc` test but with different parsing order.

TomasVotruba:
Thanks :+1: 

I'll look at it locally

PR #1875, opened 2019-08-20
Remove default excluded file patterns (closes #1815)

### LINKED ISSUE #1815: Certain file paths ignored per default
| PHP version    | PHP 7.3.7                               |
| Full Command   | vendor/composer/bin/rector process src test -vvv | 

## Current Behaviour 

Hello there, I tried to upgrade my project with rector (great tool btw!) and was wondering why certain parts of my codebase aren't upgraded, although I haven't excluded these files and I've seen the files being auto-loaded during the parsing process. Running rector with `-vvv` revealed that they're never listed during `[parsing]`.

Recognized that it's only ignoring directories with name "fixtures", which is where I store classes to load fixture data and prepare state. They're part of the test suite and they're code like everything else following the same coding standards and autoloading conventions, so I want to include them in the upgrade process of course.

Turned out that rector has some hardcoded filters to ignore certain file paths per default (including "fixtures") and these rules cannot be deactivated/overwritten. Imo that's a bad idea, because it is not obvious to the user and there's no way to contol this behavior.

https://github.com/rectorphp/rector/blob/548d13276e71db195c33819ec80aa23cca324849/src/FileSystem/FilesFinder.php#L100-L102

## Expected Behaviour

Rector should not make assumptions that certain paths can be excluded per default. Exclusion patterns should be required to be configured explicitly.

scheb:
Remove the default excluded file patterns, change for #1815

### DIFF
--- a/src/FileSystem/FilesFinder.php
+++ b/src/FileSystem/FilesFinder.php
@@ -97,10 +97,6 @@ final class FilesFinder
             ->files()
             ->in($absoluteDirectories)
             ->name($suffixesPattern)
-            ->exclude(
-                ['examples', 'Examples', 'stubs', 'Stubs', 'fixtures', 'Fixtures', 'polyfill', 'Polyfill', 'vendor']
-            )
-            ->notName('*polyfill*')
             ->sortByName();
 
         $this->addFilterWithExcludedPaths($finder);


TomasVotruba:
Thank you.

Could you refer the issue id with closing keyword?

TomasVotruba:
It could be done just in PR description, but this is also ok.

Thanks

PR #1870, opened 2019-08-19
[DeadCode] Remove null value from property

jacekll:
Closes #1870

PR #1880, opened 2019-08-21
[DeadCode] Add RemoveNullPropertyInitializationRector to dead-code set

TomasVotruba:
(before `RemoveDeadInitializationRector`)


TomasVotruba:
Ref #1870

UFTimmy:
Maybe this is the wrong place to ask this, but why are nulls on property declarations considered bad? I know I can remove the rector from applying if I don't like, but I was hoping to get a better understanding of why it's considered bad first.

I don't typically use them, but with PHP 7.4 there is a difference between not initialized and null, so for some properties I do initialize them to null so you don't get the accessing an uninitialized field error.

TomasVotruba:
They're not bad at all. 

They're as useful as first line here:

```php
$value = [];
$value = [
  $this->getSome();
];
```

TomasVotruba:
For typed properties it's a different case, see: https://github.com/rectorphp/rector/blob/master/rules/php74/src/Rector/Property/TypedPropertyRector.php

If they're in conflict and this rule breaks PHP 7.4 code, please open an issue. That's a bug.

PR #1884, opened 2019-08-21
Parent constant visibility when it is declared in a super-superclass

scheb:
After my recent improvements on the `PrivatizeLocalClassConstantRector`, I've executed it on our big PHP project. Found out that the parent constant visibility is not detected correctly, when the constant is declared in the parent's parent or futher up the class hierarchy. This PR fixes the issue.

Worth mentioning: I've executed it on a project with 50k PHP files, did an almost perfect job. This issue was only a handful of cases. So :thumbsup: for rector!

### DIFF
--- a/packages/SOLID/src/Rector/ClassConst/PrivatizeLocalClassConstantRector.php
+++ b/packages/SOLID/src/Rector/ClassConst/PrivatizeLocalClassConstantRector.php
@@ -137,6 +137,9 @@ CODE_SAMPLE
                     $this->refactor($parentClassConstant);
 
                     return $parentClassConstant;
+                } else {
+                    // If the constant isn't declared in the parent, it might be declared in the parent's parent
+                    return $this->findParentClassConstant($parentClassName, $constant);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


TomasVotruba:
> I've executed it on a project with 50k PHP files, did an almost perfect job.

50 K php file? Not 50 k lines? Wow, how much loc is it? :D 

Thanks for trying it out, this is very important on real huge projects. So I'm happy Rector did good job - also thanks to your contributoins :+1:

test.php
```diff
@@ -0,0 +1,25 @@
+<?php
```
TomasVotruba: This file was just for fast local testing and could be remove, right?

scheb: Yea, my bad, that one slipped on. Removed.

scheb:
> Wow, how much loc is it? :D
```
Non-Comment Lines of Code (NCLOC)            1661434
Class Constants                                20129
```
:see_no_evil:

TomasVotruba:
That's huge :open_mouth:

PR #1898, opened 2019-08-26
Start CakePHP 4.0 rectors

markstory:
* Add removed method renames
* Add a handful of return typehints

This isn't a complete list but it is a start with more to come as I upgrade do an upgrade on a small application I have.


TomasVotruba:
Thanks :+1: 

I look forward to next PRs!

PR #1921, opened 2019-08-29
Add uuid only to entities with id

TomasVotruba:


packages/Doctrine/src/Rector/Class_/AddUuidToEntityWhereMissingRector.php
```diff
@@ -70,6 +70,11 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
+        // id is needed
+        if (! $this->classManipulator->getProperty($node, 'id')) {
```
stloyd: Maybe it's good idea to make this configurable via constructor parameter? I.e. I saw in legacy few times where that field was named `identifier` not `id`.

TomasVotruba: Good idea! Will do

TomasVotruba: https://github.com/rectorphp/rector/pull/1921/commits/02237369489d00d6cef297dd24afaab354be2040 :)

PR #1935, opened 2019-08-31
remove getDoctrine*() methods from PhpDocInfo, use getByType() instead

TomasVotruba:


.phpstorm.meta.php
```diff
@@ -6,6 +6,9 @@
 // $container->get(Type::class) → instance of "Type"
 override(\Psr\Container\ContainerInterface::get(0), type(0));
 
+// $propertyPhpDocInfo->getByType(Type::class) → instance of "Type"|null - @todo how to make this nullable?
+override(\Rector\BetterPhpDocParser\PhpDocInfo\PhpDocInfo::getByType(0), type(0));
```
TomasVotruba: @nikic Do you have any tip how to teach PHPStorm nullable type?

I tried variations of ↓... but no success

```diff
-type(0)
+type(0)|null
```

nikic: Unfortunately this isn't supported right now. PhpStorm itself doesn't do much in terms of nullability analysis, so the missing null type probably won't result in false positive warnings at least.

TomasVotruba: I see, feature then, thank you :+1:

PR #1957, opened 2019-09-04
Move from string types to PHPStan types

TomasVotruba:
This will merge writing rules for both Rector and PHPStan to 1 skill :)


packages/DeadCode/src/Rector/Class_/RemoveSetterOnlyPropertyAndMethodCallRector.php
```diff
@@ -107,22 +108,28 @@ public function refactor(Node $node): ?Node
 
     /**
      * @param Property|ClassMethod|MethodCall|Assign $node
-     * @return string[][]][]|null
+     * @return string[][] ][]|null
```
OskarStark: Hmm 🤔

packages/CodeQuality/src/Rector/FuncCall/SimplifyStrposLowerRector.php
```diff
@@ -31,7 +31,7 @@ public function getNodeTypes(): array
     }
 
     /**
-     * @param FuncCall $node;
+     * @param FuncCall $node ;
```
OskarStark: Is this correct?

TomasVotruba: I don't thinks so. PR?

packages/DeadCode/src/Rector/Class_/RemoveSetterOnlyPropertyAndMethodCallRector.php
TomasVotruba: PR?

Issue #1591, opened 2019-06-10
Ability to exclude specific file per rule


Levivb:
Hey,

I'm currently in the process of implementing Rector in one of our projects as testcase and am running into a few issues.

At the moment there doesn't seem to be a way to exclude a specific class for a specific rule. Some rules tend to crash on some code and it would be nice to be able to use the rule but exclude some classes.

There's the possibility to exclude certain paths in the parameters key:
```yml
parameters:
  exclude_paths:
    - 'Assets/*'
    - 'ci/*'
    - 'Database/Migrations/*'
```
But that's an all or nothing exclusion

Would it be possible to support configuration per rule? Similar to how the PhpCS supports it.

```yml
Rector\DeadCode\Rector\ClassMethod\RemoveUnusedParameterRector:
  exclude_paths:
    - 'Some/Exceptional/Directory/*'
    - 'Some/Other/Class.php
```

As icing of the cake having the ability to suppress a rule on a line by line basis, or code blocks would also be pretty cool.

Code sniffer example:
```
// some code

// phpcs:disable Some.Rule.Not.ApplicableHere
// violating but allowed code

// more violating but allowed code
// phpcs:enable

// some more code

// phpcs:ignore Some.Rule.Not.ApplicableHere
// a single line allowed violating code

// final code
```

Which could be supported in rector like this:
```
// some code

// rector:disable Rector\DeadCode\Rector\ClassMethod\RemoveUnusedParameterRector
// violating but allowed code

// more violating but allowed code
// rector:enable

// some more code

// rector:ignore Rector\DeadCode\Rector\ClassMethod\RemoveUnusedParameterRector
// a single line allowed violating code

// final code
```

Let me know what you think!

TomasVotruba:
Hi,

I wonder if that's bandaid to problem and desired feature.

> Some rules tend to crash on some code and it would be nice to be able to use the rule but exclude some classes.

That should not happen and it should be fixed. What exactly is crashing?

JanMikes:
Imho nice example of how it could work is https://github.com/Symplify/EasyCodingStandard#ignore-what-you-cant-fix:

```yaml
parameters:
    skip:
        SlevomatCodingStandard\Sniffs\TypeHints\TypeHintDeclarationSniff:
            - 'packages/EasyCodingStandard/packages/SniffRunner/src/File/File.php'
```

Levivb:
I just submitted the actual crash: https://github.com/rectorphp/rector/issues/1592

[...]

PR #1987, opened 2019-09-17
Added possibility to ignore rectors per statement

jeroensmit:
This would be a solution for issue #1591 

In my experience it can be useful to exclude a single statement. 

I took the liberty of writing the feature first and discuss it afterwards. 

So I look to know what you thing of this approach.

### DIFF
--- a/src/Rector/AbstractRector.php
+++ b/src/Rector/AbstractRector.php
@@ -156,8 +156,9 @@ abstract class AbstractRector extends NodeVisitorAbstract implements PhpRectorIn
         }
 
         // recurse up until a Stmt node is found since it might contain a noRector
-        if (! $node instanceof Stmt && $node->getAttribute(AttributeKey::PARENT_NODE)) {
-            return $this->shouldIgnoreRectorForNode($node->getAttribute(AttributeKey::PARENT_NODE));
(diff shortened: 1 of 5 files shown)


TomasVotruba:
Hi, I'm usually against these custom doc blocks in code. 
But in combination with other tool ignores, it could make sense.

To continue I need from you:

- Numbers on performance drop on checking every node's annotation
- Twitter or any other pool feedback from community (I'll retweet on @rectorphp)
- More tests, e.g. on FileSystemRector as well

src/Rector/AbstractRector.php
```diff
@@ -174,6 +196,18 @@ protected function isAtLeastPhpVersion(string $version): bool
...
+    {
+        if (preg_match_all('#@noRector\s*(?<rectorName>\S+)#i', $doc->getText(), $matches)) {
+            foreach ($matches['rectorName'] as $ignoreSpec) {
+                if (static::class === ltrim($ignoreSpec, '\\')) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
```
TomasVotruba: The logic should be decoupled to own service

jeroensmit:
Came up with a different solution, which I think is better than what I did before...
Coming up soon ;)

TomasVotruba:
Replaced by #2087

PR #1988, opened 2019-09-17
Added option to use ignore annotations from other tools

jeroensmit:
This relates to issue #1591 and extends pull request #1987 

It helps with letting work rector out of the box, because it recognises situation that would deliberately ignored for other tools. 

I can imagine the downside of this approach would be that maintaining mappings between inspections from different tools can cause a lot work.

I like to know that you think.


src/Rector/AbstractRector.php
```diff
@@ -179,6 +210,41 @@ protected function isAtLeastPhpVersion(string $version): bool
...
+                        return true;
+                    }
+
+                    break;
+                default:
+                    throw new UnknownToolException('Unknown tool: ' . $tool);
+            }
+        }
+        return false;
+    }
```
TomasVotruba: Looks good! Just extract the logic to external service and add via `@require` to keep `AbstractRector` prevent from being a God class

TomasVotruba:
Replaced by #2087 

@jeroenherczeg Please close previous PRs if they're replaced by another one. It will keep PRs clean with only active ones and easier to maintain/merge. Thanks :heart:

jeroenherczeg:
Wasn't me ¯\_(ツ)_/¯

TomasVotruba:
Sorry, wrong name :D

TomasVotruba:
Sorry, wrong name :D 

cc @jeroensmit

PR #2010, opened 2019-09-20
[Doctrine] Step #3 - id to UUID values

TomasVotruba:
This handles `setId()`

Next: handle `getId()` - instead of int, now it return `UuidInterface`


TomasVotruba:
Replaced by #2011

PR #2018, opened 2019-09-23
fix renaming class to existing one [closes #1438]

### LINKED ISSUE #1438: [RenameClassRector] Edge case - rename to class that already exists
Hey, it happened that i did not notice there is already class existing and renamed some other class that new name matched it as well:

Example:
```yaml
services:
    Rector\Rector\Class_\RenameClassRector:
        GlamiImage: Glami\Image
```

But there is already class `Glami\Image` existing and autoloaded via composer.

I am curious if we should cover this edge case and add check if renaming class, whether other class already exists but i am afraid that rector is not aware of existing classes, what do you think?

TomasVotruba:


packages/Renaming/src/Rector/Class_/RenameClassRector.php
```diff
@@ -323,4 +327,17 @@ private function refactorPhpDoc(Node $node): void
...
+    private function ensureClassWillNotBeDuplicate(string $newName, string $oldName): void
+    {
+        if (! ClassExistenceStaticHelper::doesClassLikeExist($newName)) {
+            return;
+        }
+
+        throw new InvalidPhpCodeException(sprintf(
+            'Renaming class "%s" to "%s" would create a duplicated class/interface/trait and cause PHP code to be invalid',
+            $oldName,
+            $newName
```
TomasVotruba: Todo: show exact location of the existing class file, so people can investigate better

PR #2027, opened 2019-09-25
Fix printing of tab-indented files

### LINKED ISSUE #1964: Whitespace indentation breaking
| PHP version    | e.g. PHP 7.3 (invoke `php -v`)                               |
| Full Command   | e.g. vendor/bin/rector process /src --set=cakephp34  | 

Using latest dev-master branch.

## Current Behavior 

Dry-Run shows what will happen
```
                /** @var \Cake\Console\ConsoleInputSubcommand[] $subcommands */
-               $subcommands = $parser->subcommands();
+  $subcommands = $parser->subcommands();
```
and
```
        /**
-        * Get the option parser
-        *
-        * @return \Cake\Console\ConsoleOptionParser
-        */
-       public function getOptionParser() {
+  * Get the option parser
+  *
+  * @return \Cake\Console\ConsoleOptionParser
+  */
+ public function getOptionParser() {
```

Once applied it actually destroys the indentation (2 tabs => 2 spaces).

It looks like the docblock before could cause this.

## Expected Behavior

Current indentation/whitespace should not be touched/modified.
Especially, since in this case there is not even a change here done.

TomasVotruba:
Closes #1964

### DIFF
--- a/src/PhpParser/Printer/BetterStandardPrinter.php
+++ b/src/PhpParser/Printer/BetterStandardPrinter.php
@@ -10,6 +10,7 @@ use PhpParser\Node\Expr\Yield_;
 use PhpParser\Node\Scalar\DNumber;
 use PhpParser\Node\Scalar\EncapsedStringPart;
 use PhpParser\Node\Scalar\String_;
+use PhpParser\Node\Stmt;
 use PhpParser\Node\Stmt\Class_;
 use PhpParser\Node\Stmt\ClassMethod;
 use PhpParser\Node\Stmt\Declare_;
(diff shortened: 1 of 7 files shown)


abz/SomeFile.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace App;
+
+class Test {
```
dereuromark: We could call it TabIndentationTest to be more speaking.

dereuromark:
I would also keep an line comment inside a function to be sure that one is also covered.

TomasVotruba:
One step at a time

src/PhpParser/Printer/BetterStandardPrinter.php
```diff
@@ -38,6 +46,14 @@ public function __construct(array $options = [])
         $this->insertionMap['Expr_Closure->returnType'] = [')', false, ': ', null];
     }
 
+    public function printFormatPreserving(array $stmts, array $origStmts, array $origTokens): string
+    {
+        // detect per print
+        $this->detectTabOrSpaceIndentCharacter($stmts);
```
TomasVotruba: @dereuromark Here is the important change. PhpParser prints everything with spaces-only by default. So there was actually 1 indent in the code, just not 1 tab, but 1 space.

Detecting spaces per line would be slow and not very useful, so 1st newlinew indent kind is taken into account - see `detectTabOrSpaceIndentCharacter()` bellow.

What do you think?

dereuromark: Sounds reasonable to me :)

TomasVotruba: Ok :), let's go with then. If anything breaks in the future, we know wher to look for.

dereuromark:
Wow, awesome :tada: Thanks a lot!

PR #2029, opened 2019-09-25
Update AllRectorsOverview.md

zeemil:
Hello,
I've discovered the following error while using the EregToPregMatchRector

`Fatal error: Uncaught Rector\Exception\DependencyInjection\ClassNotFoundException: Class "Rector\Php72\Rector\Each\WhileEachToForeachRector" was not found while loading "rector.yaml" file`

After checking Rector's source code, it appears that the documentation contains an error. Here is the correction.

Thanks for your work guys,

### DIFF
--- a/docs/AllRectorsOverview.md
+++ b/docs/AllRectorsOverview.md
@@ -4330,7 +4330,7 @@ Removes (unset) cast
 
 ### `WhileEachToForeachRector`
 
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Be sure to check you're using the same version of package as in README.
The class exists, it's just in `dev-master`:
https://github.com/rectorphp/rector/blob/master/packages/Php72/src/Rector/Each/WhileEachToForeachRector.php

PR #2043, opened 2019-09-26
[CI] run all sets check

### LINKED ISSUE #2042: [DX] Improve error message about invalid config - indent path
On `dev-master`, I get : 

```
Fatal error: Uncaught TypeError: Argument 1 passed to Rector\DependencyInjection\Loader\RectorServiceParametersShifter::resolveRectorConfiguration() must be of the type array, string given, called in /home/gnutix/dev/iomedia/bundles/aio-cms-bundle/vendor/rector/rector/src/DependencyInjection/Loader/RectorServiceParametersShifter.php on line 68 and defined in /home/gnutix/dev/iomedia/bundles/aio-cms-bundle/vendor/rector/rector/src/DependencyInjection/Loader/RectorServiceParametersShifter.php:109
```

_Originally posted by @gnutix in https://github.com/rectorphp/rector/issues/2041#issuecomment-535438506_

TomasVotruba:
Will prevent issues like: #2041

Closes #2042

### DIFF
--- a/config/set/celebrity/celebrity.yaml
+++ b/config/set/celebrity/celebrity.yaml
@@ -32,8 +32,7 @@ services:
     Rector\Celebrity\Rector\FuncCall\SetTypeToCastRector: ~
 
     # class { var $value; } → class { public $value; }
-    Rector\Php52\Rector\Property\VarToPublicPropertyRector: ~pu
-
+    Rector\Php52\Rector\Property\VarToPublicPropertyRector: ~
 
     #  false or true → false || true
     #  false and true → false && true
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,7 @@ script:
           bin/rector process src --set symfony40 --dry-run
           composer docs
           php bin/check_services_in_yaml_configs.php
+          php bin/run_all_sets.php
       fi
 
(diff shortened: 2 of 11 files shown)


config/set/celebrity/celebrity.yaml
```diff
@@ -32,8 +32,7 @@ services:
     Rector\Celebrity\Rector\FuncCall\SetTypeToCastRector: ~
 
     # class { var $value; } → class { public $value; }
-    Rector\Php52\Rector\Property\VarToPublicPropertyRector: ~pu
-
+    Rector\Php52\Rector\Property\VarToPublicPropertyRector: ~
```
TomasVotruba: @gnutix This was the issue :D slow PHPStorm + my hasty hands

gnutix: Ahahah that's a nice one xD Too bad it's still "valid YAML"... can't catch this with a linter.

TomasVotruba: Indeed :D 

But now it's not valid for the configuration - it will throw exception here now: https://github.com/rectorphp/rector/pull/2043/files#diff-4b356c6550ab7f23c4b4d17595032977R69

PR #2065, opened 2019-09-30
[CodingStyle] Allow private ctor override for static factory

### LINKED ISSUE #2063: [CodeQuality] Allow private construtor for named constructors
I guess you could look for "return new self/static" in static methods + protected/private constructor ? Not sure how hard it is to figure out such combination.

And yeah, the public parent comes from Symfony's HttpFoundation ParameterBag class.

_Originally posted by @gnutix in https://github.com/rectorphp/rector/issues/2057#issuecomment-536229922_

TomasVotruba:
Closes #2063


TomasVotruba:
@gnutix I've added you as a collaborator so I could ask you for a review here

PR #2070, opened 2019-10-01
Fixed the third argument in VarDumperTestTrait.

adrozdek:
In VarDumperTestTrait's asserts the third argument is called `$filter` and by default it should be equal to 0.
Right now it's called `$context` or `$format` and is set to null. Also `$data` argument is called `$format` sometimes.

`public function assertDumpEquals($expected, $data, $filter = 0, $message = '')`
`public function assertDumpMatchesFormat($expected, $data, $filter = 0, $message = '')` - 
https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/VarDumper/Test/VarDumperTestTrait.php

### DIFF
--- a/docs/AllRectorsOverview.md
+++ b/docs/AllRectorsOverview.md
@@ -5721,16 +5721,16 @@ Changes Process string argument to an array
 
 - class: `Rector\Symfony\Rector\VarDumper\VarDumperTestTraitMethodArgsRector`
 
-Adds new `$format` argument in `VarDumperTestTrait->assertDumpEquals()` in Validator in Symfony.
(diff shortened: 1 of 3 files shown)


TomasVotruba:
This rule is in Symfony 4.0 set. You address Symfony 4.4

When was this changed? Link to UPGRADE.md or PR in Symfony is needed

adrozdek:
Sorry for giving 4.4 version but this parameter is the same for 4.0: https://github.com/symfony/symfony/blob/4.0/src/Symfony/Component/VarDumper/Test/VarDumperTestTrait.php
The UPGRADE-4.0.md for VarDumper is wrong. I created PR for changing that: https://github.com/symfony/symfony/pull/33821

The description is wrong but the example taken from code is correct:
Zrzut ekranu z 2019-10-03 20-40-57

TomasVotruba:
I see, thanks for clarification

PR #2072, opened 2019-10-02
[CodingStyle] Skip common annotation aliases in RemoveUnusedAliasRector

### LINKED ISSUE #1116: Level coding-style breaks Doctrine annotations
Before applying:
`use Doctrine\ORM\Mapping as ORM;`

After:
`use Doctrine\ORM\Mapping;`

However the actual annotations are not changed, resulting in broken code:
```
/**
 * @ORM\Id
 * @ORM\Column(type="integer")
 * @ORM\GeneratedValue(strategy="AUTO")
 */
protected $id;
```

Command:
`docker run -v (pwd):/project rector/rector:latest bin/rector process /project/src --level coding-style --autoload-file /project/vendor/autoload.php`

TomasVotruba:
Closes #1116

### DIFF
--- a/packages/CodingStyle/src/Rector/Use_/RemoveUnusedAliasRector.php
+++ b/packages/CodingStyle/src/Rector/Use_/RemoveUnusedAliasRector.php
@@ -57,25 +57,28 @@ final class RemoveUnusedAliasRector extends AbstractRector
 
     public function getDefinition(): RectorDefinition
     {
-        return new RectorDefinition('Removes unused use aliases', [
-            new CodeSample(
-                <<<'PHP'
+        return new RectorDefinition(
+            'Removes unused use aliases. Keep annotation aliases like "Doctrine\ORM\Mapping as ORM" to keep convention format',
+            [
+                new CodeSample(
+                    <<<'PHP'
 use Symfony\Kernel as BaseKernel;
 
 class SomeClass extends BaseKernel
 {
 }
 PHP
-                ,
-                <<<'PHP'
+                    ,
+                    <<<'PHP'
 use Symfony\Kernel;
 
 class SomeClass extends Kernel
 {
 }
 PHP
-            ),
-        ]);
+                ),
+            ]
+        );
     }
 
     /**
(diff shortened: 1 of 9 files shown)


gnutix:
It feels fragile to me to have a static map of aliases to skip ; I know from another project of ours that people sometimes change them (it's `use Symfony\Component\Validator\Constraints as Validator;` in that project).

So I don't think it really "fixes" it, it just avoids it.

TomasVotruba:
That's just adding another layer of control. All used annotations are kept + those that people want to keep.

```diff
-use Doctrine\ORM\Mapping as ORM;
+use Doctrine\ORM\Mapping;

-@ORM\Table
+@Mapping\Table
```

gnutix:
In my project's example, that was not working : 

```diff
1 file with changes
===================

1) src/Entity/Media/Format.php

    ---------- begin diff ----------
--- Original
+++ New
@@ -22,7 +22,7 @@
 use Symfony\Component\HttpFoundation\File\File;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Component\Validator\Constraints as Assert;
-use Vich\UploaderBundle\Mapping\Annotation as Vich;
+use Vich\UploaderBundle\Mapping\Annotation;

 /**
  * @ORM\Table("Media_Format")
    ----------- end diff -----------

Applied rectors:

 * Rector\CodingStyle\Rector\Use_\RemoveUnusedAliasRector
```

It only removed the alias, it didn't change the annotations themselves (therefore they were broken).

TomasVotruba:
I see, I'll check that case too

packages/CodingStyle/src/Rector/Use_/RemoveUnusedAliasRector.php
```diff
@@ -299,17 +292,55 @@ private function resolveTraitUseNames(Node $searchNode): void
...
-        $this->traverseNodesWithCallable($searchNode, function (Node $node): void {
+        $possibleDocAliases = [];
+
+        $this->traverseNodesWithCallable($searchNode, function (Node $node) use (&$possibleDocAliases): void {
             if ($node->getDocComment() === null) {
                 return;
             }
 
-            $matches = Strings::matchAll($node->getDocComment()->getText(), '#\@(?<possible_alias>.*?)\\\\#s');
+            $matches = Strings::matchAll($node->getDocComment()->getText(), '#\@(?<possible_alias>\w+)(\\\\)?#s');
```
TomasVotruba: This line was the bugfix. Appearanbly, annotations without following `\\` broke this:

```php
* @UniqueEntity()
* @ThisWasSkipped
```

gnutix: Good catch! :)

Issue #1911, opened 2019-08-28
[CodingStyle] \Rector\CodingStyle\Rector\Namespace_\ImportFullyQualifiedNamesRector : add an option to skip "root namespace" classes (like \Throwable or \DateTime)


gnutix:
## Diff

Currently, when running the coding-style level, you will get changes like this : 

```diff
--- Original
+++ New
@@ -4,6 +4,8 @@

 namespace ...;

+use Throwable;
@@ -280,7 +282,7 @@
             }

             return new Response(...);
-        } catch (\Throwable $throwable) {
+        } catch (Throwable $throwable) {
```

It would be nice to have an option to disable this.

TomasVotruba:
Sure, free free to send PR to add optional disable via constructor

TomasVotruba:
Closing as answered

PR #2077, opened 2019-10-02
Add an option to skip importing root namespace classes (like \DateTime)

gnutix:
Trying to fix https://github.com/rectorphp/rector/issues/1911#issuecomment-525642537

```yaml
services:
    Rector\CodingStyle\Rector\Namespace_\ImportFullyQualifiedNamesRector:
        $shouldImportRootNamespaceClasses: false
```

I still need to prevent changing DocBlock annotations like `@var \DateTime`... which seems way harder, looking at the code.

### DIFF
--- a/packages/CodingStyle/src/Rector/Namespace_/ImportFullyQualifiedNamesRector.php
+++ b/packages/CodingStyle/src/Rector/Namespace_/ImportFullyQualifiedNamesRector.php
@@ -2,7 +2,6 @@
 
 namespace Rector\CodingStyle\Rector\Namespace_;
 
-use Nette\Utils\Strings;
 use PhpParser\Node;
 use PhpParser\Node\Name;
 use Rector\CodingStyle\Node\NameImporter;
@@ -83,9 +82,9 @@ PHP
     public function refactor(Node $node): ?Node
     {
         // Importing root namespace classes (like \DateTime) is optional
-        if (! $this->shouldImportRootNamespaceClasses) {
+        if (! $this->shouldImportRootNamespaceClasses && $node instanceof Name) {
             $name = $this->getName($node);
-            if (Strings::startsWith($name, '\\') && substr_count($name, '\\') === 1) {
+            if ($name !== null && substr_count($name, '\\') === 0) {
                 return null;
             }
         }
--- a/packages/CodingStyle/src/Rector/Namespace_/ImportFullyQualifiedNamesRector.php
+++ b/packages/CodingStyle/src/Rector/Namespace_/ImportFullyQualifiedNamesRector.php
@@ -20,14 +20,23 @@ final class ImportFullyQualifiedNamesRector extends AbstractRector
     private $shouldImportDocBlocks = true;
 
     /**
+     * @var bool
+     */
+    private $shouldImportRootNamespaceClasses = true;
+
+    /**
      * @var NameImporter
      */
     private $nameImporter;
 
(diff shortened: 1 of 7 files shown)


TomasVotruba:
Great pick!

Definitely start with a test with optional parameter to be on

gnutix:
Test added. But some help to get the condition `$node is NonRootNamespace` right would be more than welcome. Currently my condition broke the tests in `packages/CodingStyle/tests/Rector/Namespace_/ImportFullyQualifiedNamesRector/NonNamespacedTest.php`. And I've had a look at the DocBlock issue, and the code is quite complex in there (`DocBlockManipulator`, `DocBlockNameImporter`, `PhpDocNodeTraverser`, etc); I wouldn't know how/where to pass that new option around and where to use it exactly so I wouldn't break half the project.

TomasVotruba:
> But some help to get the condition $node is NonRootNamespace right would be more than welcome.

From top of my head, I'd detect `\DateTime` like:

```
$name = $this->getName($node);
if (Strings::startsWith($name, '\\')) {
   if (substr_count($name, '\\') === 1) {
        // skip
	    return null;
   }
}
```

gnutix:
Here's a working implementation. I've not handled DocBlock though, so the test is still failing.

packages/CodingStyle/src/Rector/Namespace_/ImportFullyQualifiedNamesRector.php
```diff
@@ -72,6 +81,14 @@ public function getNodeTypes(): array
...
+        if (! $this->shouldImportRootNamespaceClasses && $node instanceof Name) {
+            $name = $this->getName($node);
+            if ($name !== null && substr_count($name, '\\') === 0) {
+                return null;
+            }
+        }
+
         $this->useAddingCommander->analyseFileInfoUseStatements($node);
 
         if ($node instanceof Name) {
```
TomasVotruba: Why not add it here, where `Name` check is already added?

TomasVotruba:
As for doblocks, this might be place to check it:
https://github.com/rectorphp/rector/blob/master/packages/NodeTypeResolver/src/PhpDoc/NodeAnalyzer/DocBlockNameImporter.php#L125

gnutix:
I've finished the implementation. :+1:

TomasVotruba:
Feature looks ready, good job :+1: 

> Would there be some documentation to write for this new option ?

The documentation is generated from code samples, so adding one `ConfiguredCodeSample()` in `getDefinition()` will be enough

TomasVotruba:
> I've finished the implementation. +1 Would there be some documentation to write for this new option ?

Feel free to send that in next PR. I'm merging this, since feature is complete

gnutix:
Not sure how I can describe the new parameter within `new CodeSample(...)` as this only provides before/after code examples - but not the constructor parameters involved to achieve such results.

TomasVotruba:
`ConfiguredCodeSample`

gnutix:
Sorry, I should have read more attentively! :roll_eyes:

TomasVotruba:
No troubles :+1:

gnutix:
@TomasVotruba I can't find this option in `master` anymore. Was it removed purposefully ? :thinking:

TomasVotruba:
Could you add that to README as well?

Aerendir:
@TomasVotruba I documented it

PR #2081, opened 2019-10-02
Add failing test: StringifyStrNeedlesRector adds (string) to a method call that returns a string anyway

gnutix:


packages/Php73/tests/Rector/FuncCall/StringifyStrNeedlesRector/Fixture/skip_already_string.php.inc
```diff
@@ -0,0 +1,13 @@
...
+
+final class SkipAlreadyString
+{
+    public function getNeedle(): string
+    {
+        return 'needle';
+    }
+}
+
+strpos('needle', (new SkipAlreadyString)->getNeedle());
```
TomasVotruba: This better be wrapped in a method in the class, since fixtures are included (run), e.g.

```php
public function run(): string
{
    strpos('needle', (new self())->getNeedle());
}
```

TomasVotruba:
Thank you :+1:

PR #2082, opened 2019-10-02
Fixing 'return new static()' not being covered by MakeInheritedMethodVisibilitySameAsParentRector

### LINKED ISSUE #2057: Protected/private constructors are turned into public
| PHP version    | `PHP 7.3.9-1+ubuntu18.04.1+deb.sury.org+1 (cli) (built: Sep  2 2019 12:54:24) ( NTS )` |
| Full Command   | `vendor/bin/rector --ansi --dry-run --debug process src` | 

Running `0.5.13` on our entire codebase applies the following unexpected changes : 

```diff
2) src/.../CropParameters.php

    ---------- begin diff ----------
--- Original
+++ New
@@ -9,7 +9,7 @@

 final class CropParameters extends ParameterBag
 {
-    private function __construct(array $params)
+    public function __construct(array $params)
     {
         // Ensure mandatory parameters are present
         foreach (['width', 'height'] as $key) {
    ----------- end diff -----------

Applied rectors:

 * Rector\CodingStyle\Rector\ClassMethod\MakeInheritedMethodVisibilitySameAsParentRector
 * Rector\Rector\Argument\ArgumentRemoverRector
 * Rector\PHPUnit\Rector\ExceptionAnnotationRector
 * Rector\Renaming\Rector\Annotation\RenameAnnotationRector
 * Rector\CodingStyle\Rector\ClassMethod\ReturnArrayClassMethodToYieldRector
 * Rector\Rector\Argument\ArgumentDefaultValueReplacerRector
 * Rector\Rector\Argument\ArgumentAdderRector
 * Rector\Rector\ClassMethod\WrapReturnRector

5) src/.../SystemFormats.php

    ---------- begin diff ----------
--- Original
+++ New
@@ -33,7 +33,7 @@
-    protected function __construct(array $elements = [])
+    public function __construct(array $elements = [])
     {
         parent::__construct($elements);
     }
    ----------- end diff -----------

Applied rectors:

 * Rector\CodingStyle\Rector\ClassMethod\MakeInheritedMethodVisibilitySameAsParentRector
 * Rector\Rector\Argument\ArgumentRemoverRector
 * Rector\PHPUnit\Rector\ExceptionAnnotationRector
 * Rector\Renaming\Rector\Annotation\RenameAnnotationRector
 * Rector\CodingStyle\Rector\ClassMethod\ReturnArrayClassMethodToYieldRector
 * Rector\Rector\Argument\ArgumentDefaultValueReplacerRector
 * Rector\Rector\Argument\ArgumentAdderRector
 * Rector\Rector\ClassMethod\WrapReturnRector
```

gnutix:
`return new self();` was correctly handled, while `return new static();` was not. Now both are.

Fixes #2057

### DIFF
--- a/packages/CodingStyle/src/Rector/ClassMethod/MakeInheritedMethodVisibilitySameAsParentRector.php
+++ b/packages/CodingStyle/src/Rector/ClassMethod/MakeInheritedMethodVisibilitySameAsParentRector.php
@@ -214,7 +214,15 @@ PHP
                 return false;
             }
 
-            return $this->isName($node->expr->class, 'self') || $this->isName($node->expr->class, 'static');
+            if ($this->isName($node->expr->class, 'self')) {
+                return true;
+            }
+
+            if ($this->isName($node->expr->class, 'static')) {
+                return true;
+            }
+
+            return false;
         });
     }
 }
--- a/packages/CodingStyle/src/Rector/ClassMethod/MakeInheritedMethodVisibilitySameAsParentRector.php
+++ b/packages/CodingStyle/src/Rector/ClassMethod/MakeInheritedMethodVisibilitySameAsParentRector.php
@@ -177,7 +177,7 @@ PHP
                 continue;
             }
 
-            $isStaticSelfFactory = $this->isStaticSelfFactory($iteratedClassMethod);
+            $isStaticSelfFactory = $this->isStaticNamedConstructor($iteratedClassMethod);
 
             if ($isStaticSelfFactory === false) {
                 continue;
@@ -193,7 +193,7 @@ PHP
      * Looks for:
      * public static someMethod() { return new self(); }
      */
-    private function isStaticSelfFactory(ClassMethod $classMethod): bool
+    private function isStaticNamedConstructor(ClassMethod $classMethod): bool
     {
         if (! $classMethod->isPublic()) {
             return false;
@@ -212,7 +212,7 @@ PHP
                 return false;
             }
 
-            return $this->isName($node->expr->class, 'self');
+            return $this->isName($node->expr->class, 'self') || $this->isName($node->expr->class, 'static');
         });
     }
 }
--- a/packages/CodingStyle/src/Rector/ClassMethod/MakeInheritedMethodVisibilitySameAsParentRector.php
+++ b/packages/CodingStyle/src/Rector/ClassMethod/MakeInheritedMethodVisibilitySameAsParentRector.php
@@ -192,6 +192,8 @@ PHP
     /**
      * Looks for:
      * public static someMethod() { return new self(); }
+     * or
+     * public static someMethod() { return new static(); }
      */
     private function isStaticNamedConstructor(ClassMethod $classMethod): bool
     {
--- a/packages/CodingStyle/tests/Rector/ClassMethod/MakeInheritedMethodVisibilitySameAsParentRector/Fixture/skip_static_ctor.php.inc
+++ b/packages/CodingStyle/tests/Rector/ClassMethod/MakeInheritedMethodVisibilitySameAsParentRector/Fixture/skip_static_ctor.php.inc
@@ -2,19 +2,6 @@
 
 namespace Rector\CodingStyle\Tests\Rector\ClassMethod\MakeInheritedMethodVisibilitySameAsParentRector\Fixture;
 
-final class SkipSelfCtor extends ParentWithPublicConstructor
-{
-    private function __construct()
-    {
-        // do something
-    }
-
-    public static function create()
-    {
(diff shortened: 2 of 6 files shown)


TomasVotruba:
Please always create a new fixture file, so the fix is more clear and tests as small & readable as possible

Could you make fix with it?

gnutix:
Fixed.

gnutix:
And if you don't mind, I felt like fixing the implemention, while I was at it. :)

packages/CodingStyle/src/Rector/ClassMethod/MakeInheritedMethodVisibilitySameAsParentRector.php
```diff
@@ -212,7 +214,7 @@ private function isStaticSelfFactory(ClassMethod $classMethod): bool
                 return false;
             }
 
-            return $this->isName($node->expr->class, 'self');
+            return $this->isName($node->expr->class, 'self') || $this->isName($node->expr->class, 'static');
```
TomasVotruba: Better use early return for readability and consistence with the rest of the code

```php
if ($this->isName($node->expr->class, 'self')) {
    return true;
}

if ($this->isName($node->expr->class, 'static')) {
    return true;
}

return false;
```

packages/CodingStyle/tests/Rector/ClassMethod/MakeInheritedMethodVisibilitySameAsParentRector/Fixture/skip_static_ctor.php.inc
```diff
@@ -2,21 +2,25 @@
...
-class ParentWithPublicConstructor
+class InheritedSkipCtor extends SkipStaticCtor
 {
-    public function __construct()
+    protected function __construct()
     {
+        parent::__construct();
+
+        // do something more
     }
```
TomasVotruba: Why this change? It refers some `ParentWithPublicConstructor` class that is now missing here

gnutix: I've tried to make a difference between the Self and Static classes. Self extends just one class (ParentWithPublicConstructor), is final, private constructor, while Static is abstract, extended by another class (InheritedSkipCtor), protected constructor.

That way it matches "real code examples" better, IMHO.

packages/CodingStyle/src/Rector/ClassMethod/MakeInheritedMethodVisibilitySameAsParentRector.php
gnutix: Will do next time I sit behind a laptop.

packages/CodingStyle/tests/Rector/ClassMethod/MakeInheritedMethodVisibilitySameAsParentRector/Fixture/skip_static_ctor.php.inc
TomasVotruba: Better add new case fixture to cover this with standalone isolated classes.
In future one of test fixture might change it would break things

gnutix: Sorry, but I don't get the issue here. I have two features (covering "return new self();" and "return new static();"). Therefore I have two fixtures/test cases files.

What more would you like to "isolate" ?

TomasVotruba: Keep original fixture untouched + create a new file for your case.

The reason is to make debugging simpler in the future.

TomasVotruba: Oh, I see it now. You use one class in 2 tests.
In that case it should be extracted to `Source` directory

gnutix: I'll extract it then.

TomasVotruba: Thank you

TomasVotruba:
Thank you :+1:

Issue #1846, opened 2019-08-12
Rectors goes so far as to remove comments (which can contain valuable information)


gnutix:
| Subject        | Details                                                      |
| :------------- | :----------------------------------------------------------- |
| PHP version    | PHP 7.3.8-1+ubuntu18.04.1+deb.sury.org+1 (cli) (built: Aug  7 2019 09:52:12) ( NTS )                               |
| Full Command   | vendor/bin/rector --ansi process var/rector_test.php --level=code-quality  | 

## Current Behaviour 

```diff
Rector v0.5.9
Config file: rector.yaml

 3/3 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

1 file with changes
===================

1) var/rector_test.php

    ---------- begin diff ----------
--- Original
+++ New
@@ -5,12 +5,6 @@
     if ('A' === $input) {
         return false;
     }
-    
     // Some condition B we know it can't handle
-    if ('B' === $input) {
-        return false;
-    }
-    
-    // Yeah, everything else it should handle just fine
-    return true;
+    return (bool) ('B' !== $input);
 }
    ----------- end diff -----------

Applied rectors:

 * Rector\CodeQuality\Rector\If_\SimplifyIfReturnBoolRector


                                                                                                                        
 [OK] Rector is done! 1 changed files
```

## Minimal PHP Code Causing Issue

```
<?php
function canHandle($input)
{
    // Some condition A we know it can't handle
    if ('A' === $input) {
        return false;
    }
    
    // Some condition B we know it can't handle
    if ('B' === $input) {
        return false;
    }
    
    // Yeah, everything else it should handle just fine
    return true;
}
```

Which then turns into : 

```
<?php
function canHandle($input)
{
    // Some condition A we know it can't handle
    if ('A' === $input) {
        return false;
    }
    // Some condition B we know it can't handle
    return (bool) ('B' !== $input);
}
```

## Expected Behaviour

I don't think Rector should remove comments, as they can contain precious information. The Rector should skip this file change IMO.

gnutix:
By the way, my code example was voluntarily simplified. The conditions are more complex and the comments help understanding them better.

> I wonder how you decide the comment is useful?

[...]

PR #2083, opened 2019-10-02
Do not apply SimplifyIfReturnBoolRector when there are comments in between the if statements

gnutix:
See https://github.com/rectorphp/rector/issues/1846#issuecomment-536216506.


TomasVotruba:
Could you fix it at one go?

Collect comments of precious nodes and add it via $node->setComment()

gnutix:
I don't think we should apply the Rector if there's a comment in there. Because, as in my fixture, the comment might apply to the last return statement, and not the last two ones. Which could mess things up. So I would go for checking in there's a comment in between the ifs that the Rector is trying to simplify, and if yes, abort the operation.

TomasVotruba:
Agreed. Please send PR

gnutix:
Ready for review. :)

packages/CodeQuality/tests/Rector/If_/SimplifyIfReturnBoolRector/Fixture/should_keep_comments.php.inc
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public function both($var): bool
+    {
+        // Imagine this is a very important comment here that should not be removed.
+        // Imagine this is another very important comment here that should not be removed.
+        return is_string($var);
+    }
+}
+?>
```
TomasVotruba: If the code should remain the same, the part under `-----` can be fully removed here

packages/CodeQuality/tests/Rector/If_/SimplifyIfReturnBoolRector/Fixture/should_keep_doc_blocks.php.inc
```diff
@@ -0,0 +1,81 @@
...
+
+        /**
+         * @todo Imagine this is another very important comment here that should not be removed.
+         */
+        return true;
+    }
+}
+?>
+-----
+<?php
```
TomasVotruba: Same here

packages/CodeQuality/tests/Rector/If_/SimplifyIfReturnBoolRector/Fixture/should_keep_comments.php.inc
gnutix: It's not the same : the Rector changed the if conditions. Only the comments are the same, but I can't use a "one part" fixture file then.

packages/CodeQuality/tests/Rector/If_/SimplifyIfReturnBoolRector/Fixture/should_keep_doc_blocks.php.inc
gnutix: Same answer ;)

packages/CodeQuality/tests/Rector/If_/SimplifyIfReturnBoolRector/Fixture/should_keep_comments.php.inc
TomasVotruba: Ah, I see.

PR #2084, opened 2019-10-02
Add failing test case for ReturnArrayClassMethodToYieldRector (removing comments)

gnutix:
See https://github.com/rectorphp/rector/issues/1846#issuecomment-537026517


gnutix:
By the way, I was wondering if it wouldn't be better that ReturnArrayClassMethodToYieldRector uses : iterable instead of : \Iterator as the return type ? what do you think ?

EDIT: looking at PhpParser, I'm not sure there's a way to do that though...

TomasVotruba:
`Iterator` is more precise here, as it narrows it to `yield`

`iterable` would allow lead to non-sense code like:

```php
yield 5
return 5
```

TomasVotruba:
Care for fix again?

gnutix:
Not in the next 2 weeks, I've used all the hours I could spare. Need to focus on our project 'til I can.

TomasVotruba:
No troubles :+1:

PR #2087, opened 2019-10-03
Added way to exclude rectors

jeroensmit:
This replaces #1987 and #1988 

It's extracted to a different service. 

The ExcludeByDockblock check is default disabled so it is up to the user whether he wants to use it.

Custom Exclusion checks can be added later easily. For example to not run certain rector on classes that implement certain interfaces.

Also the mapping with ignore check from other tools can be easily added as separate checks, either in the rector repo, or in a separate rector-othertool-bridge repo, either maintained by rectorphp or third parties.


src/Contract/Exclusion/ExclusionCheckInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+declare(strict_types=1);
+
+namespace Rector\Contract\Exclusion;
+
+use PhpParser\Node;
+use Rector\Contract\Rector\PhpRectorInterface;
+
+interface ExclusionCheckInterface
```
TomasVotruba: Why interface?

src/Exclusion/Check/ExcludeByDocBlock.php
```diff
@@ -0,0 +1,36 @@
...
+use PhpParser\Comment\Doc;
+use PhpParser\Node;
+use PhpParser\Node\Stmt;
+use Rector\Contract\Exclusion\ExclusionCheckInterface;
+use Rector\Contract\Rector\PhpRectorInterface;
+use Rector\NodeTypeResolver\Node\AttributeKey;
+
+final class ExcludeByDocBlock implements ExclusionCheckInterface
+{
+    public function shouldExcludeRector(PhpRectorInterface $phpRector, Node $node): bool
```
TomasVotruba: Better name would be covering both arguments, `shouldExcludeRectorAndNode()`, or better "isNodeSkippedByRector()".

*Exclude* is rather "remove completelly" and is already used, see README: https://github.com/rectorphp/rector#exclude-paths-and-rectors

*Skip* is better - "just this once"

src/Exclusion/ExclusionManager.php
```diff
@@ -0,0 +1,34 @@
...
+    public function shouldExcludeRector(PhpRectorInterface $phpRector, Node $onNode): bool
+    {
+        foreach ($this->exclusionChecks as $check) {
+            if ($check->shouldExcludeRector($phpRector, $onNode)) {
+                return true;
+            }
+        }
+        return false;
+    }
+}
```
TomasVotruba: :clap:

tests/Exclusion/Check/ExcludeByDocBlockTest.php
```diff
@@ -0,0 +1,45 @@
...
+        $this->doTestFile($file);
+    }
+
+    public function provideDataForTest(): Iterator
+    {
+        yield [__DIR__ . '/Fixture/fixture.php.inc'];
+        yield [__DIR__ . '/Fixture/fixture-norector.php.inc'];
+        yield [__DIR__ . '/Fixture/fixture2.php.inc'];
+        yield [__DIR__ . '/Fixture/fixture2-norector.php.inc'];
+        yield [__DIR__ . '/Fixture/fixture3.php.inc'];
```
TomasVotruba: Eplicit fixture naming would be better for further maintenance, e.g. "exclude_with_rule_name.php.inc" etc.

tests/Exclusion/ExclusionManagerTest.php
```diff
@@ -0,0 +1,70 @@
...
+     * @return ExclusionCheckInterface
+     */
+    private function getCheckMock(bool $return, bool $shouldBeCalled): ExclusionCheckInterface
+    {
+        $exclusionCheck = Mockery::mock(ExclusionCheckInterface::class);
+        $exclusionCheck->shouldReceive('shouldExcludeRector')
+            ->times($shouldBeCalled ? 1 : 0)
+            ->andReturn($return);
+        return $exclusionCheck;
+    }
```
TomasVotruba: Convention for tests are get service from container to prevent reconstructing container with mocks.

See other non-Rector tests for something like this:

```php
protected function setUp()
{
     parent::setUp();
     $this->exclusionManager = self::$container->get(ExclusionManager::class);
}
```

config/services.yaml
```diff
@@ -5,7 +5,7 @@ services:
 
     Rector\:
         resource: '../src'
-        exclude: '../src/{Node/Attribute.php,Rector/**/*Rector.php,Reporting/FileDiff.php,Testing/PHPUnit,RectorDefinition,PhpParser/Node/*Info.php,Exception/*,DependencyInjection/CompilerPass/*,DependencyInjection/Loader/*,Application/Error.php,HttpKernel/RectorKernel.php,ValueObject/*}'
+        exclude: '../src/{Node/Attribute.php,Rector/**/*Rector.php,Reporting/FileDiff.php,Testing/PHPUnit,RectorDefinition,PhpParser/Node/*Info.php,Exception/*,DependencyInjection/CompilerPass/*,DependencyInjection/Loader/*,Application/Error.php,HttpKernel/RectorKernel.php,ValueObject/*,Exclusion/Check/*}'
```
TomasVotruba: The system one should be autolaoded by default, this should be removed

src/Contract/Exclusion/ExclusionCheckInterface.php
jeroensmit: To be implemented by anyone wanting to write their own exclusion check, for example the one checking also checking for other skip docblocks from other tools

tests/Exclusion/ExclusionManagerTest.php
jeroensmit: I'm not sure that this works given the fact that this test checks the ExclusionManager with different constructor params

config/services.yaml
jeroensmit: Nice

tests/Exclusion/Check/ExcludeByDocBlockTest.php
jeroensmit: fixed

src/Exclusion/Check/ExcludeByDocBlock.php
jeroensmit: fixed

src/Contract/Exclusion/ExclusionCheckInterface.php
TomasVotruba: That's what `ExclusionCheckInterface` does

TomasVotruba: Ah, my bad. I thought it's interface for `ExclusionManager`

Allright :+1:

src/Exclusion/Check/ExcludeByDocBlock.php
```diff
@@ -0,0 +1,36 @@
...
+namespace Rector\Exclusion\Check;
+
+use PhpParser\Comment\Doc;
+use PhpParser\Node;
+use PhpParser\Node\Stmt;
+use Rector\Contract\Exclusion\ExclusionCheckInterface;
+use Rector\Contract\Rector\PhpRectorInterface;
+use Rector\NodeTypeResolver\Node\AttributeKey;
+
+final class ExcludeByDocBlock implements ExclusionCheckInterface
```
TomasVotruba: The class name seems like value object.
It's convention to respect the interface name here ↓

`ExcludeByDocBlockExclusionCheck`

Also better naming as role to prevent confusion with value object:

- value vs role
- Check vs Checker
- object vs service

tests/Exclusion/ExclusionManagerTest.php
TomasVotruba: > ...with different constructor params

Those should be defined in the local config.yaml then, so it's 1:1 how people use Rector

tests/Exclusion/ExclusionManagerTest.php
```diff
@@ -0,0 +1,70 @@
...
+namespace Rector\Tests\Exclusion;
+
+use Mockery;
+use PhpParser\Node;
+use PHPUnit\Framework\TestCase;
+use Rector\Contract\Exclusion\ExclusionCheckInterface;
+use Rector\Exclusion\ExclusionManager;
+use Rector\Rector\AbstractRector;
+
+final class ExclusionManagerTest extends TestCase
```
TomasVotruba: Looking at this test again... the functionality is already covered well by: https://github.com/rectorphp/rector/pull/2087/files#diff-b4a2ec27faf137b39d3dbd3e8565d3e0

It can be removed completely.

TomasVotruba: It's ready to merge then, great job :+1:

jeroensmit: Technically not everything is covered in in the other test, like having multiple exclusionChecks or none at all. So removing this would result in losing some test-coverage.

src/Exclusion/Check/ExcludeByDocBlock.php
jeroensmit: I renamed it to ExcludeByDocBlockExclusionCheck as suggested

tests/Exclusion/ExclusionManagerTest.php
TomasVotruba: Of course, but the code coverage isn't the goal of tests.

Usable and maintainble tests are. In respect of Occam's razor, this adds more complexity than value. Thus should be removed

src/Exclusion/Check/ExcludeByDocBlock.php
TomasVotruba: Thanks :+1:

tests/Exclusion/ExclusionManagerTest.php
TomasVotruba: I'm mergin it as it is, so we can test the functionality.
If needed, I'll do change later.

TomasVotruba:
Thanks for great work :+1:

TomasVotruba:
Solves also #1591

gnutix:
Would be great to see a README update with this one :)

TomasVotruba:
Agreed :)

PR #2091, opened 2019-10-04
Add UseIncrementAssignRector

keulinho:
Fixes https://github.com/rectorphp/rector/issues/2004


packages/CodingStyle/tests/Rector/Assign/UseIncrementAssignRector/Fixture/increment-fixture.php.inc
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Rector\CodingStyle\Tests\Rector\Assign\UseIncrementAssignRector\Fixture;
+
+class SomeClass
```
TomasVotruba: The class name must be unique per namespace. Anything will fail in the future eventualy.

You can add all these changes to one file/method, since it's very simple change

packages/CodingStyle/tests/Rector/Assign/UseIncrementAssignRector/UseIncrementAssignRectorTest.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        $this->doTestFile($file);
+    }
+
+    public function provideDataForTest(): Iterator
+    {
+        yield [
+            __DIR__ . '/Fixture/increment-fixture.php.inc',
+            __DIR__ . '/Fixture/decrement-fixture.php.inc'
+        ];
```
TomasVotruba: This is incorrect. It must be 1 line = 1 test

```php
yield [__DIR__ . '/...'];
yield [__DIR__ . '/...'];
```

packages/CodingStyle/src/Rector/Assign/UseIncrementAssignRector.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * @param AssignOp $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if ((! $node->expr instanceof LNumber) || $node->expr->value !== 1) {
+            return null;
+        }
```
TomasVotruba: Better split into 2 early-return conditions, to prevent `||` and `&&` complexity

E.g.

```php
if (! $node->expr instanceof LNumber) {
    return null;
}

if ($node->expr->value !== 1) {
    return null;
}
````

File: packages/CodingStyle/src/Rector/Assign/UseIncrementAssignRector.php [hunk 4]
```diff
@@ -0,0 +1,79 @@
...
+    public function getNodeTypes(): array
+    {
+        return [
+            Plus::class,
+            Minus::class
+        ];
+    }
+
+    /**
+     * @param AssignOp $node
```
TomasVotruba:
```suggestion
     * @param Plus|Minus $node
```

Better use known more specific values

packages/CodingStyle/tests/Rector/Assign/UseIncrementAssignRector/Fixture/increment-fixture.php.inc
keulinho: I hope i understood you, i've merged the fixtures for increment and decrement in one fixture file

TomasVotruba: Yes, exactly :+1:

TomasVotruba:
Thanks :+1:

PR #2094, opened 2019-10-04
Add ShortenElseIfRector

keulinho:
Fixes https://github.com/rectorphp/rector/issues/1945


packages/CodeQuality/tests/Rector/If_/ShortenElseIfRector/ShortenElseIfRectorTest.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        $this->doTestFile($file);
+    }
+
+    public function provideDataForTest(): Iterator
+    {
+        yield [__DIR__ . '/Fixture/simple.php.inc'];
+        yield [__DIR__ . '/Fixture/nested-else.php.inc'];
+        yield [__DIR__ . '/Fixture/nested-elseif.php.inc'];
+        yield [__DIR__ . '/Fixture/recursive.php.inc'];
```
TomasVotruba: Please add also 1-2 cases that should be skipped

packages/CodeQuality/src/Rector/If_/ShortenElseIfRector.php
```diff
@@ -0,0 +1,100 @@
...
+        if (count($else->stmts) > 1) {
+            return null;
+        }
+
+        $if = $else->stmts[0];
+        if (! $if instanceof If_) {
+            return null;
+        }
+
+        $if = $this->refactor($if) ?? $if;
```
TomasVotruba: This is pretty difficult to understand recursion.
Could you think of more readable syntax?

keulinho: Removed the coalesc and made it more verbose, 
hopefully that makes it clearer

Do you have any better ideas in mind how to clarify this?

packages/CodeQuality/src/Rector/If_/ShortenElseIfRector.php
```diff
@@ -84,7 +84,11 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
-        $if = $this->refactor($if) ?? $if;
+        $refactored = $this->refactor($if);
+
+        if ($refactored) {
+            $if = $refactored;
+        }
```
TomasVotruba: Could you maybe add some explanatory comment?

packages/CodeQuality/tests/Rector/If_/ShortenElseIfRector/Fixture/multiple-stmts.php.inc
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    public function doSomething()
+    {
+
+    }
+}
+
+?>
```
TomasVotruba: For non-changed fixtures, this whole part can be removed

packages/CodeQuality/tests/Rector/If_/ShortenElseIfRector/ShortenElseIfRectorTest.php
```diff
@@ -22,6 +22,8 @@ public function provideDataForTest(): Iterator
         yield [__DIR__ . '/Fixture/nested-else.php.inc'];
         yield [__DIR__ . '/Fixture/nested-elseif.php.inc'];
         yield [__DIR__ . '/Fixture/recursive.php.inc'];
+        yield [__DIR__ . '/Fixture/no-else-if.php.inc'];
+        yield [__DIR__ . '/Fixture/multiple-stmts.php.inc'];
```
TomasVotruba: Naming convention with `skip-*` is more clear about skipped fixture

packages/CodeQuality/src/Rector/If_/ShortenElseIfRector.php
TomasVotruba: I'm not sure why have you chosen this approach, so I don't even understand the idea to describe it

keulinho: Ok the reason behind the recursion is that we get rid of the original nested `If_` node, but this rector runs only on `If_` nodes.
This means after refactoring the top-level `If_` we won't detect any more else/if condition that we could refactor inside that if statement, because now it is an `Elseif_` Node.
Consider the example from the `recursive` fixture:
```php
    public function run()
    {
        if ($this->cond1) {
            $this->doSomething();
        } else {
            if ($this->cond2) {
                $this->doSomething();
            } else {
                if ($this->cond3) {
                    $this->doSomething();
                }
            }
        }
    }
```
In that case we would move the `if ($this->cond2)` statement up and combine it to the `elseif ($this->cond2)`. By this refactoring our rector doesn't run for the nested `if ($this->cond3)`, it would get refactored if we run the rector a second time.
Therefore we have to run the refactoring on the inner `if ($this->cond3)` before applying the refactoring on the outer `if ($this->cond3)`.

I hope this clarifies the need for recursion, i'll try to express this inside the code

TomasVotruba: I see 

> I hope this clarifies the need for recursion, i'll try to express this inside the code

That would be perfect, so this comment is not needed to understand the mechanism

packages/CodeQuality/src/Rector/If_/ShortenElseIfRector.php
```diff
@@ -0,0 +1,110 @@
...
+    {
+        return [If_::class];
+    }
+
+    /**
+     * @param If_ $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        return $this->shortenElseIf($node);
```
TomasVotruba: :+1:  Thank you

PR #2098, opened 2019-10-04
Add option to keep declared constant visibility

scheb:
This PR adds an option for `PrivatizeLocalClassConstantRector` to keep the constant visibility when it has been explicitly declared, instead of calculating it from usages in the codebase and potentially changing it.

### DIFF
--- a/docs/AllRectorsOverview.md
+++ b/docs/AllRectorsOverview.md
@@ -5716,7 +5716,7 @@ Classes that have no children nor are used, should have abstract
 
 - class: `Rector\SOLID\Rector\ClassConst\PrivatizeLocalClassConstantRector`
 
-Finalize every class constant that is used only locally
+Privatize every class constant that is used only locally
 
 ```diff
  class ClassWithConstantUsedOnlyHere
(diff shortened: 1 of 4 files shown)


TomasVotruba:
This has same effect as removing the rule completely.

What's the use case?

scheb:
It's only skipping constants that already have the constant visibility explicitly set.

The use case is a pre-7.1 codebase that has constants without any visibility declaration and some constants already have the visibility set. I don't want to change the later ones, only get the visibility set one the ones without any visibility declaration.

Added some documentation to explain what the setting does.

TomasVotruba:
I see.

Why do you want to skip exisiting visibilities, that might be incorrect?

scheb:
Because I trust the developers that they have made a reasonable choice on the visibility ;)

Other reasons:
- Reduce risk, existing visibilities are know to be working
- Constants for values of enum-like types, that should be public even when they're not explicitly referenced in the code
- Some edge cases that static code analysis cannot fully understand and I'm working around those by manually declaring the visibility

TomasVotruba:
Such option would suggest that every rule might be unreliable and optionally partially disabled. E.g. autompleting annotation only where missing etc. 

Similar suggestion were made before, but it only work-around the real bug in particular rule.

> Constants for values of enum-like types, that should be public even when they're not explicitly referenced in the code

This should be reflected in this rule without any configuration. Do you have such code?

> Some edge cases that static code analysis cannot fully understand and I'm working around those by manually declaring the visibility

This should be included too. Share a code

Another solution is rather generic, see: https://github.com/rectorphp/rector/pull/2087/files

scheb:
> Such option would suggest that every rule might be unreliable and optionally partially disabled. E.g. autompleting annotation only where missing etc.
>
> Similar suggestion were made before, but it only work-around the real bug in particular rule.

Oke, fair enough, I see your point that you don't want to weaken rectors. I'll get this somehow working for me. The `@noRector` PR in fact looks interesting, thanks for the hint.

If any of these edge cases could be added to the rector, be sure I'd send you a PR ;)

TomasVotruba:
Thanks for understanding :bowing_man:

PR #2101, opened 2019-10-04
Fix virtual property

### LINKED ISSUE #2086: [BetterPhpDocParser] SerializerTypePhpDocNodeFactory doesn't count with virtual props
No need for lengthy description.

This line https://github.com/rectorphp/rector/blob/598f228c6d12b74913f8915375abc8c222dddbd7/packages/BetterPhpDocParser/src/PhpDocNodeFactory/JMS/SerializerTypePhpDocNodeFactory.php#L24 is triggered for following code

```php
class Foo {
    /**
     * @Serializer\Type("boolean")
     * @Serializer\VirtualProperty
     */
    public function isBar(): bool
    {
        return true;
    }
}
```

Assumption that this annotation works only for properties is not correct.

TomasVotruba:
Closes #2086

### DIFF
--- a/packages/Php55/src/Rector/FuncCall/PregReplaceEModifierRector.php
+++ b/packages/Php55/src/Rector/FuncCall/PregReplaceEModifierRector.php
@@ -41,7 +41,7 @@ final class PregReplaceEModifierRector extends AbstractRector
 
     public function getDefinition(): RectorDefinition
     {
-        return new RectorDefinition('The /e modifier is no longer supported, use preg_replace_callback instead ', [
+        return new RectorDefinition('The /e modifier is no longer supported, use preg_replace_callback instead', [
             new CodeSample(
                 <<<'PHP'
 class SomeClass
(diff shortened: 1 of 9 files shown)


packages/Php55/src/Rector/FuncCall/PregReplaceEModifierRector.php
```diff
@@ -41,7 +41,7 @@ public function __construct(Parser $parser)
 
     public function getDefinition(): RectorDefinition
     {
-        return new RectorDefinition('The /e modifier is no longer supported, use preg_replace_callback instead ', [
+        return new RectorDefinition('The /e modifier is no longer supported, use preg_replace_callback instead', [
```
TomasVotruba: This finally fixes these repreated spaces in docs dump

utils/DocumentationGenerator/src/Command/DumpNodesCommand.php
```diff
@@ -402,7 +402,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
                 $this->nodeInfoResult->addNodeInfo($category, new NodeInfo(
                     $nodeClass,
-                    $this->betterStandardPrinter->print($node),
+                    trim($this->betterStandardPrinter->print($node)),
```
TomasVotruba: This finally fixes these repreated spaces in docs dump

Issue #2119, opened 2019-10-06
[Symfony 4.4] Command must return int


TomasVotruba:
See https://github.com/symfony/symfony/pull/33775/files

PR #2132, opened 2019-10-10
[Symfony] Add ConsoleExecuteReturnIntRector

keulinho:
Fixes: https://github.com/rectorphp/rector/issues/2119


packages/Symfony/src/Rector/Console/ConsoleExecuteReturnIntRector.php
```diff
@@ -0,0 +1,125 @@
...
+        if (! $return->expr ) {
+            $return->expr = new LNumber(0);
+        }
+
+        if ($return->expr instanceof ConstFetch && $this->getName($return->expr) === 'null') {
+            $return->expr = new LNumber(0);
+        }
+
+        if ($return->expr instanceof MethodCall || $return->expr instanceof StaticCall) {
+            // how to get the Node of teh called method?!
```
keulinho: @TomasVotruba i guess i need help here, am i able to the get the nodes that are called?!

or do you see another way to handle the case from the `return-function-call` testcase?

TomasVotruba: From top of my head:

```php
$returnStaticType = $this->getStaticType($return->expr);
if ($returnStaticType instanceof IntegerType) {
    return null;
}

$return->expr = new LNumber(0);
```

TomasVotruba:
Do you plan to handle no `return ` in `execute()` as well?

E.g.

```php
public function execute($input, $output)
{
    $output->write('hi');
}
```

packages/Symfony/src/Rector/Console/ConsoleExecuteReturnIntRector.php
keulinho: I got that covered, what I have trouble with is returning the value of a method call, e.g.
```
return $this->doRun();
```
My first idea was to apply the rector to the `doRun()` method too, but I have no idea how to get the node of the called method.

keulinho:
That should already be covered.

packages/Symfony/src/Rector/Console/ConsoleExecuteReturnIntRector.php
TomasVotruba: Good!

What happens when you try the code I wrote above?

keulinho: That leads to:
```diff
    public function execute(InputInterface $input, OutputInterface $output)
    {
-       return $this->doSomething();
+      return 0;
    }

    private function doSomething()
    {
        return null;
    }
```

but this is want i want to accomplish:
```diff
    public function execute(InputInterface $input, OutputInterface $output)
    {
        return $this->doSomething();
    }

    private function doSomething()
    {
-       return null;
+      return 0;
    }
```

I need to be able to somehow get the node of the called function, in this example the `doSomething()` method, but i don't know how as it seems that the `MethodCall`-Node just has a name of the called method but no reference to the node of that method.

keulinho: nevermind i got now, guess i just needed a bit of rubber ducking :laughing:

packages/Symfony/src/Rector/Console/ConsoleExecuteReturnIntRector.php
```diff
@@ -84,42 +99,119 @@ public function refactor(Node $node): ?Node
         return $this->addReturn0ToMethod($node);
     }
 
-    private function setReturnTo0InsteadOfNull(Return_ $return)
+    public function beforeTraverse(array $nodes): void
```
TomasVotruba: This method should never be used in Rector rules as it is rather internal part (I know it's allowed, my bad, I gotta make it `final`).

What do you need to achieve here?

packages/Symfony/src/Rector/Console/ConsoleExecuteReturnIntRector.php
```diff
@@ -84,42 +99,119 @@ public function refactor(Node $node): ?Node
...
         }
     }
 
-    private function addReturn0ToMethod(FunctionLike $node): FunctionLike
+    private function addReturn0ToMethod(FunctionLike $functionLike): FunctionLike
     {
         $hasReturn = false;
-        $this->traverseNodesWithCallable($node->getStmts(), function (Node &$stmt) use ($node, &$hasReturn) {
-            if (!$stmt instanceof Return_) {
+        $this->traverseNodesWithCallable($functionLike->getStmts(), function (Node &$stmt) use (
```
TomasVotruba: As Node is an object, no need to refernce it. It's done by default by PHP

```diff
-Node &$stmt
+Node $stmt
```

packages/Symfony/tests/Rector/Console/ConsoleExecuteReturnIntRector/Fixture/return-external-function-call.php.inc
```diff
@@ -0,0 +1,57 @@
...
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class ReturnExternalFunctionCall extends Command
+{
+    /** @var ExternalClass */
+    private $external;
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        return $this->external->doSomething();
```
TomasVotruba: How does Command work now with `null` value?

Maybe better would be retype to integer, e.g. `(int) $this->external->doSomething`

Chanding value in other method might break code elsewhere, where `null` is expected

```php
// app/SomeController.php

class SomeController
{
    public function go()
    {
        // this would not pass after the change
        return $this->external->doSomething() === null;
    }
}
```

keulinho: Yep i didn't thought about just casting the expressions, that makes the solution alos much cleaner and no need to for the `beforeTraverse()` anymore

packages/Symfony/src/Rector/Console/ConsoleExecuteReturnIntRector.php
keulinho: as already said not necessary anymore ;)

TomasVotruba: Understood :)

packages/Symfony/src/Rector/Console/ConsoleExecuteReturnIntRector.php
```diff
@@ -90,7 +75,7 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
-        if (! $class->extends || $class->extends->toCodeString() !== self::COMMAND_CLASS) {
+        if (! $class->extends || $class->extends->toCodeString() !== '\\' . Command::class) {
```
TomasVotruba: Better:

```php
if (! $this->isName($class->extends, Command::class)) {
```

TomasVotruba:
Related Symfony PR: https://github.com/symfony/symfony/pull/33775

TomasVotruba:
Btw, only numberic values are converted to int:
https://github.com/symfony/symfony/pull/33775/files#diff-8e80595663798837db1b033187835535R262

It seems like anything else will return `0`:

```diff
-return 'hey';
+return 0;
```

```diff
-return '13';
+return 13;
```


return `hey` would

keulinho:
So what do you suggest?
instead of always casting the return value we could use something like:

```diff
- return $this->doSomething();
+ $result = $this->doSomething();
+ is_numeric($result) ? return (int) $result : return 0;
```

TomasVotruba:
I suggest to change the value only if know what is returned, but that's just an detail.
 
But moreover, we should merge this PR as it is now, as there is lot of opened work. So next step is just small iteration

TomasVotruba:
Could you just make CI pass? It's ready then

keulinho:
:+1: 
CI did pass, merging is blocked because you've requested changes ;)

TomasVotruba:
Few moments I saw if fail, probably miss click.
All right then :)

PR #2128, opened 2019-10-09
EncapsedStringsToSprintfRector fails when using class properties

gnutix:
```
vendor/bin/phpunit packages/CodingStyle/tests/Rector/Encapsed/EncapsedStringsToSprintfRector/EncapsedStringsToSprintfRectorTest.php
PHPUnit 7.5.0 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.3.9-1+ubuntu18.04.1+deb.sury.org+1
Configuration: /home/gnutix/dev/oss/rectorphp/rector/phpunit.xml

.F.                                                                 3 / 3 (100%)

Time: 1.49 seconds, Memory: 56.50MB

There was 1 failure:

1) Rector\CodingStyle\Tests\Rector\Encapsed\EncapsedStringsToSprintfRector\EncapsedStringsToSprintfRectorTest::test with data set #1 ('/home/gnutix/dev/oss/rectorph...hp.inc')
Caused by /home/gnutix/dev/oss/rectorphp/rector/packages/CodingStyle/tests/Rector/Encapsed/EncapsedStringsToSprintfRector/Fixture/fixture2.php.inc
Failed asserting that string matches format description.
--- Expected
+++ Actual
@@ @@
 
     public function run(string $format)
     {
-        return sprintf('Unsupported format %s', $this->format);
+        return sprintf('Unsupported format ');
     }
 }

/home/gnutix/dev/oss/rectorphp/rector/src/Testing/PHPUnit/AbstractRectorTestCase.php:182
/home/gnutix/dev/oss/rectorphp/rector/src/Testing/PHPUnit/AbstractRectorTestCase.php:136
/home/gnutix/dev/oss/rectorphp/rector/packages/CodingStyle/tests/Rector/Encapsed/EncapsedStringsToSprintfRector/EncapsedStringsToSprintfRectorTest.php:16
```

PR #2135, opened 2019-10-10
Fix EncapsedStringsToSprintfRector for non-var exprs

TomasVotruba:
Closes #2128


gnutix:
Well, that was a simple fix! ;)

TomasVotruba:
Indeed :) lucky one

PR #2159, opened 2019-10-14
Screen file command improvements

TomasVotruba:
```
vendor/bin/rector screen-file SomeFile.php
```

## Input file

```php
<?php

declare(strict_types=1);

namespace App;

final class ScreenSample
{
    public function run($value)
    {
        $value = is_string($value) ? 5 : $value;
        return $value;
    }
}
```

## Output helper

```php
<?php
// node_type: "Declare_"
declare(strict_types=1);
// node_type: "Namespace_"
// name: "App"
namespace App;

// node_type: "Class_"
// class_name: "App\ScreenSample"
final class ScreenSample
{
    // node_type: "ClassMethod"
    public function run($value)
    {
        // node_type: "Assign"
        // assign_var: 
        //   * node_type: "Variable"
        //   * variable_name: "value"
        //   * variable_types: "mixed"
        
        // assign_expr: 
        //   * node_type: "Ternary"
        $value = is_string(
            // node_type: "Arg"
            $value
        ) ? 5 : $value;
        // node_type: "Return_"
        // returned_node: 
        //   * node_type: "Variable"
        //   * variable_name: "value"
        //   * variable_types: "mixed"
        return $value;
    }
}
```


TomasVotruba:
cc @DaveLiddament @gnutix What is missing to be useful here?

gnutix:
Are comments / docblocks "explained" too ?

TomasVotruba:
Could you show me some example?

gnutix:
Hmm.. maybe I spoke too soon. Regarding the type of information that this command provides, I'm not sure it could/should say anything about comments.

Is the purpose of this command solely to help people code/contribute Rectors ? If so, I think it might help if the command returned code examples on how to check for these types or fetch some information (like how to get data for a Node, or how to get a comment/docblock, stuff like that). But maybe that's too much and not the scope of this PR (and I guess most of this is related to Nikic PHP Parser right ? so maybe a link to https://github.com/nikic/PHP-Parser#documentation might "just work" as well ?

TomasVotruba:
> Is the purpose of this command solely to help people code/contribute Rectors ?

Yes! We used it at workshop, so people don't have to learn/remember how php-parser, PHPStan and Rector works (way too much info for anyone).

You just run the command, look at the line you want to change and see it's `MethodCall` of type `A` - half job done.

> how to get data for a Node

At the moment simple `dump($node)` in Rector shows the most important stuff. Docs is also one of the sources, but again too complicated. One programmre wants to change method, another extra function etc.

gnutix:
This is definitely a very nice first step indeed. I guess we'll see down the road if there are more improvements that could be done. Dumping a $node for example provides quite a huuuuuuge tree of information, most of them probably never useful to write a simple check or change. But let's start with that already. :)

TomasVotruba:
> This is definitely a very nice first step indeed. I guess we'll see down the road if there are more improvements that could be done.

Agreed, I'll put it out on trainings and start to collect feedback what is relevant and what not.

> I guess we'll see down the road if there are more improvements that could be done. Dumping a $node for example provides quite a huuuuuuge tree of information, most of them probably never useful to write a simple check or change. But let's start with that already. :)

Agreed, thinking of that, there might be useful function `dump_node()` which would dump only what is useful.


Thanks for feedback

PR #2182, opened 2019-10-18
[PHP 74] Add ChangeReflectionTypeToStringToGetNameRector

TomasVotruba:


packages/NodeTypeResolver/src/DependencyInjection/PHPStanServicesFactory.php
```diff
@@ -31,6 +31,11 @@ public function __construct()
 
         $additionalConfigFiles[] = __DIR__ . '/../../config/phpstan/type-extensions.neon';
 
+        $currentProjectConfigFile = __DIR__ . '/../../../../phpstan.neon';
+        if (file_exists($currentProjectConfigFile)) {
+            $additionalConfigFiles[] = $currentProjectConfigFile;
+        }
```
TomasVotruba: @sabbelasichon This is what you needed when we called? :)

TomasVotruba: Oh, `getcwd()` would be better I guess

sabbelasichon: @TomasVotruba Yes, i have tested it with the getcwd() function, and it works perfectly. I think this is very important for extensibility. You are incredible.

sabbelasichon: I already created the phpstan.neon for TYPO3 https://github.com/sabbelasichon/typo3-rector/blob/master/phpstan.neon

TomasVotruba: >  You are incredible 

:heart: 

This will take more days to finish. Could I ask you to cherry pick this feature into own PR so we can merge it?

PR #2196, opened 2019-10-25
Update grammar in the README for clarity

sbine:
Hey there, this PR makes some (opinionated) changes to the grammar in the README, and standardizes casing in the section headings using traditional title case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,4 +321,4 @@ docker run -v $(pwd):/project rector/rector:latest process /project/app --config
 Do you use Rector to upgrade your code? Add it here:
 
 - [drupal8-rector/drupal8-rector](https://github.com/drupal8-rector/drupal8-rector) by [@mxr576](https://github.com/mxr576) for [Drupal](https://www.drupal.org/)
-- [sabbelasichon/typo3-rector](https://github.com/sabbelasichon/typo3-rector] for [TYPO3](https://typo3.org/)
+- [sabbelasichon/typo3-rector](https://github.com/sabbelasichon/typo3-rector) for [TYPO3](https://typo3.org/)
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
-# Rector - Upgrade your Legacy App to Modern Codebase
+# Rector - Upgrade Your Legacy App to a Modern Codebase
 
 Rector is a **rec**onstruc**tor** tool - it does **instant upgrades** and **instant refactoring** of your code.
-Why doing it manually if 80% Rector can handle for you?
+Why refactor manually if Rector can handle it for you?
 
 [![Build Status](https://img.shields.io/travis/rectorphp/rector/master.svg?style=flat-square)](https://travis-ci.org/rectorphp/rector)
 [![Coverage Status](https://img.shields.io/coveralls/rectorphp/rector/master.svg?style=flat-square)](https://coveralls.io/github/rectorphp/rector?branch=master)
@@ -25,7 +25,7 @@ Thank you:
 
 ## Open-Source First
 
-Rector **instantly upgrades and instantly refactors the PHP code of your application**. It covers many open-source projects and PHP changes itself:
+Rector **instantly upgrades and instantly refactors the PHP code of your application**. It keeps up to date with the latest PHP changes, and supports many open-source projects:
 
 <br>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -81,7 +79,7 @@ Rector relies on project and autoloading of its classes. To specify your own aut
 vendor/bin/rector process ../project --autoload-file ../project/vendor/autoload.php
 ```
 
-Or make use of `rector.yaml` config:
+Or use a `rector.yaml` config file:
```
TomasVotruba: :+1:

README.md
```diff
@@ -102,7 +100,7 @@ parameters:
         - '*/src/*/Tests/*'
 ```
 
-Do you want to use a whole set, except that one rule? Exclude it:
+You can use a whole ruleset, except one rule:
```
TomasVotruba: :+1:

README.md
```diff
@@ -111,29 +109,27 @@ parameters:
...
 
 ```yaml
 # rector.yaml
 parameters:
-    php_version_features: '7.2' # your version 7.3
+    php_version_features: '7.2' # your version is 7.3
 ```
 
-Do you prefer importing FQN class names by default?
+You can import FQN class names automatically every time Rector performs a change, so you don't have to do it manually/after each run:
```
TomasVotruba: :+1:

README.md
```diff
@@ -281,9 +277,9 @@ That's it!
 
 ## More Detailed Documentation
 
-- **[All Rectors Overview](/docs/AllRectorsOverview.md)**
-- [How Rector Works?](/docs/HowItWorks.md)
-- [Nodes Overview](/docs/NodesOverview.md)
+- **[All Rectors: Overview](/docs/AllRectorsOverview.md)**
+- [How Does Rector Work?](/docs/HowItWorks.md)
```
TomasVotruba: :+1:

README.md
```diff
@@ -45,33 +45,31 @@ Rector **instantly upgrades and instantly refactors the PHP code of your applica
...
 ## Install
 
 ```bash
 composer require rector/rector --dev
 ```
 
-**Do you have conflicts on `composer require` or on run?**
-
-- Use [Docker image](#run-rector-in-docker)!
+**Did you have conflicts during `composer require` or on run? Use the [Docker image](#run-rector-in-docker).**
```
TomasVotruba: Better keep this separated on each line. 

- **bold** = question 
- under it = answer

README.md
```diff
@@ -169,37 +165,37 @@ imports:
...
     ```
 
-## 3 Steps to Create Own Rector
+## 3 Steps to Create Your Own Rector
 
-First, make sure it's not covered by [any existing Rectors yet](/docs/AllRectorsOverview.md).
+First, make sure it's not covered by [any existing Rectors](/docs/AllRectorsOverview.md).
 
-Let's say we want to **change method calls from `set*` to `change*`**.
+Let's say we want to **change method prefixes from `set*` to `change*`**.
```
TomasVotruba: *method calls* here is technical term from abstract syntax tree nodes, so better keep it

README.md
```diff
@@ -169,37 +165,37 @@ imports:
...
  $user = new User();
 -$user->setPassword('123456');
 +$user->changePassword('123456');
 ```
 
-### 1. Create New Rector and Implement Methods
+### 1. Create a New Rector and Implement Methods
 
-Create a class that extends [`Rector\Rector\AbstractRector`](/src/Rector/AbstractRector.php). It has useful methods like checking node type and name. Just run `$this->` and let PHPStorm show you all possible methods.
+Create a class that extends [`Rector\Rector\AbstractRector`](/src/Rector/AbstractRector.php). It will inherit useful methods e.g. to check node type and name. See the source (or type `$this->` in an IDE) for a list of available methods.
```
TomasVotruba: Nice one :+1:

README.md
```diff
@@ -281,9 +277,9 @@ That's it!
 
 ## More Detailed Documentation
 
-- **[All Rectors Overview](/docs/AllRectorsOverview.md)**
-- [How Rector Works?](/docs/HowItWorks.md)
-- [Nodes Overview](/docs/NodesOverview.md)
+- **[All Rectors: Overview](/docs/AllRectorsOverview.md)**
+- [How Does Rector Work?](/docs/HowItWorks.md)
+- [Nodes: Overview](/docs/NodesOverview.md)
```
TomasVotruba: It's really "Nodes Overview" aka "Overview of Nodes", better keep it as there is just this one page about nodes.

README.md
```diff
@@ -303,16 +299,15 @@ Just follow 3 rules:
...
 ## Run Rector in Docker
-
-With this command, you can process your project with Rector from docker:
+You can run Rector on your project using Docker:
 
 ```bash
 docker run -v $(pwd):/project rector/rector:latest process /project/src --set symfony40 --dry-run
 
-# Note that a volume is mounted from `pwd` into `/project` which can be accessed later.
+# Note that a volume is mounted from the current directory into `/project` which can be accessed later.
```
TomasVotruba: Better combine both, as one is more technical and other amateur (so both groups understand)

README.md
sbine: What do you think about quote-formatting this section? The single bullet point looked a little out of place to me.
e.g.
> **Did you have conflicts during `composer require` or on run?**  
> Use the [Docker image](#run-rector-in-docker).

TomasVotruba: Personally I don't read quotes in README. I'd expect them in posts

TomasVotruba:
Thank you

PR #2197, opened 2019-10-25
Add test to ConsoleExecuteReturnIntRector for a not command class

franmomu:
When using `ConsoleExecuteReturnIntRector` on a class with an `execute` method that not extends from `Command` it adds the `return 0` and adds `: int`.


TomasVotruba:
Thanks!

TomasVotruba:
@keulinho Could you look at it?

keulinho:
@TomasVotruba yep will provide an fix

TomasVotruba:
Thank you

PR #2200, opened 2019-10-25
Fix ConsoleExecuteReturnIntRector for non console commands

keulinho:


TomasVotruba:
What about parent class that extends Command class?

PR #2202, opened 2019-10-25
Fix ConsoleExecuteReturnIntRector if target class not directly extends Command

keulinho:
Fixes: https://github.com/rectorphp/rector/pull/2200#issuecomment-546327783

### DIFF
--- a/packages/Symfony/src/Rector/Console/ConsoleExecuteReturnIntRector.php
+++ b/packages/Symfony/src/Rector/Console/ConsoleExecuteReturnIntRector.php
@@ -76,7 +76,8 @@ PHP
             return null;
         }
 
-        if (! ($class->extends && $this->isName($class->extends, Command::class))) {
+        $reflection = new \ReflectionClass($this->getName($class));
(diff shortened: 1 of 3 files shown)


packages/Symfony/src/Rector/Console/ConsoleExecuteReturnIntRector.php
```diff
@@ -76,7 +76,8 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
-        if (! ($class->extends && $this->isName($class->extends, Command::class))) {
+        $reflection = new \ReflectionClass($this->getName($class));
+        if (! $reflection->isSubclassOf(Command::class)) {
```
TomasVotruba: What about:

```php
if (!$this->isObjectType($class, Command::class)) {
    return null;
}
```

Something like this should be in other rules like this

TomasVotruba:
My comment is marginal.
Merging as it is to fix the issue.

TomasVotruba:
Thank you for quick fix!

PR #2207, opened 2019-10-27
Previous statement rewrite

jeroensmit:
Current/Previous expression did not work as expected. 

It only took \PhpParser\Node\Stmt\Expression as boundary leading to issues when other instances of \PhpParser\Node\Stmt where previous. 

It also turned out that the current implementation of correctPregMatchType was working by incident. But would break if any random statement was put on packages/Php71/tests/Rector/FuncCall/CountOnNullRector/Fixture/preg_match_array.php.inc:9


TomasVotruba:
I look forward to review this! Let me know when CI passes

TomasVotruba:
Looks good, thank you :+1:

PR #2214, opened 2019-10-28
travis dogfood replay

ktomk:
fails at home.

expected execution on travis php 7.3 (dogfood).

#2205

PR #2218, opened 2019-10-29
Fix empty -c/--config value

### LINKED ISSUE #2205: Config File-Path Injection by Not Passing a Config File Option Argument
| Rector version | 0.5.x-dev@5757e86 (master this morning)          |
| PHP version    | PHP 7.3.10                               |
| Full Command   | `bin/rector -c`  | 

## Current Behaviour 

When invoking rector with an empty `-c` option, the rector utility is looking for the config file in the path "`--ansi`" (and fatals w/ an uncaught exception).

For this case Tom suspects a flaw of the utilities self-configuration that is adding command line options at wrong places.

TomasVotruba:
Closes #2205

Solves #2214

Inspired by: https://github.com/phpstan/phpstan/commit/d437e70f929bf508569e3c0f55ecfee7f977bd01

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,12 +51,6 @@ script:
           php bin/run_all_sets.php
       fi
 
-    # try #2205 w/ xdebug
-    - |
-      if [[ $COVERAGE == true ]]; then
-        bin/rector -c
-      fi
-
     # Eat your own dog food
     - |
       if [[ $DOG_FOOD == true ]]; then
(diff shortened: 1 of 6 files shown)


TomasVotruba:
Looks good now :+1: 

image

@ktomk Thank you for investigating this, it's really nasty bug and hard to find. 

The same problem happened here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3901

ktomk:
Thanks for the feedback. Happy to help, was just two quick pushes.

PR #2223, opened 2019-10-30
Add AbstractController as base class if no one exists

### LINKED ISSUE #2215: Edge Cases with TemplateAnnotationRector
## Question
Thanks for fixing my issues with the TemplateAnnotationRector before.

Now, I've found two more edge cases where I don't know if it is a case for Rector to solve it, so I just thought opening a Support Ticket:

In one case I've got a Controller which doesn't extend any class (no AbstractController or Controller). While the Template annotation works, the result ($this->render(...)) won't work, because the render method is a helper method from AbstractController (ControllerTrait to be more specific).

So I think the rector shouldn't do anything in this case or what do you think?

The other case is that some of our actions call another private method like in the following example:

```php
    /**
     * @Route("/create", name="survey_create")
     * @Template("PAPPSurveyBundle:Survey:create.html.twig")
     *
     * @return array
     */
    public function createAction(Request $request, Party $currentParty)
    {
        $survey = $this->surveyHelper->createSurvey($currentParty);

        return $this->processForm($request, $currentParty, $survey, $this->generateUrl('survey_save'));
    }

    private function processForm(Request $request, Party $party, Survey $survey, $url)
    {
        ...

        return [
            'form' => $form->createView(),
        ];
    }
```

In this case the Template annotation will be removed, but nothing more which breaks the application.

In both cases, I am not sure what the rector should do. If you want me to open issues for one or both of these cases, just let me know, I could add some failing tests.

stedekay:
This is a first draft to fix #2215.

It seems to work, but I get an error from PHPStan which I don't understand.
Maybe you can help me with that?

### DIFF
--- a/packages/Sensio/src/Rector/FrameworkExtraBundle/TemplateAnnotationRector.php
+++ b/packages/Sensio/src/Rector/FrameworkExtraBundle/TemplateAnnotationRector.php
@@ -8,6 +8,8 @@ use PhpParser\Node;
 use PhpParser\Node\Arg;
 use PhpParser\Node\Expr\Array_;
 use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Name\FullyQualified;
+use PhpParser\Node\Stmt\Class_;
 use PhpParser\Node\Stmt\ClassMethod;
 use PhpParser\Node\Stmt\Return_;
 use Rector\BetterPhpDocParser\PhpDocNode\Sensio\SensioTemplateTagValueNode;
@@ -15,6 +17,7 @@ use Rector\Rector\AbstractRector;
 use Rector\RectorDefinition\CodeSample;
 use Rector\RectorDefinition\RectorDefinition;
 use Rector\Sensio\Helper\TemplateGuesser;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 
 final class TemplateAnnotationRector extends AbstractRector
 {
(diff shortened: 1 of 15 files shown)


packages/Sensio/src/Rector/FrameworkExtraBundle/TemplateAnnotationRector.php
```diff
@@ -106,6 +113,17 @@ public function refactor(Node $node): ?Node
         return $node;
     }
 
+    private function addBaseClassIfMissing(Node $node): ?Node
+    {
+        if ($node->extends !== null) {
+            return null;
+        }
+
```
TomasVotruba: Not all classes need to be extended.
This would add parent controller to every single class, that has no parent yet.

To fix that, there must be check that at least one of class' method has the template annotation.
If not, return null here

stedekay: Good point. I updated the PR.

stedekay:
There is one more thing I thought about. What if the controller is extending a class which is not based on AbstractController, like:

```php
class MyController extends DefaultController {
}

class DefaultController {
}
```

In this case, the render method would be added, but still failing as its missing.

packages/Sensio/src/Rector/FrameworkExtraBundle/TemplateAnnotationRector.php
TomasVotruba: Very nice job!

TomasVotruba:
Agreed. Should we skip such case?
U wonder how real it is and if it's worth the attention

stedekay:
Exactly, I think it is something which can be skipped for now. Can be fixed when its needed :)

TomasVotruba:
Thank you

PR #2233, opened 2019-11-01
Also remove assignment if the value of the assignment is different

jeroensmit:


TomasVotruba:
CI needs fix.

Tip for better development: run Rector with the rule you're changing on itself:

```
bin/rector p src packages --set dead-code --dry-run
```

It would quickly show false positives of your implementation

TomasVotruba:
I've merge your other PR, please rebase on master

TomasVotruba:
Thank you :+1:

PR #2239, opened 2019-11-03
Better anonymous class handling

jeroensmit:
Also added test for nodeVisitors

I got a bit carried away while fixing a couple of issues that were caused by incorrect handling of anonymous classes.

### DIFF
--- a/packages/NodeTypeResolver/tests/NodeVisitor/FunctionMethodAndClassNodeVisitor/fixtures/anonymousClass.php.inc
+++ b/packages/NodeTypeResolver/tests/NodeVisitor/FunctionMethodAndClassNodeVisitor/fixtures/anonymousClass.php.inc
@@ -1,0 +1,148 @@
+<?php
+
+namespace Rector\NodeTypeResolver\Tests\NodeVisitor\FunctionMethodAndClassNodeVisitorTest\RandomProperty;
+
+
+class foo
+{
+    public function bar()
+    {
+        new class
+        {
+            public function baz()
+            {
+            }
+        };
+        moreCode();
+    }
+}
+?>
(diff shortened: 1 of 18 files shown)


packages/NodeTypeResolver/tests/NodeVisitor/FunctionMethodAndClassNodeVisitor/fixtures/anonymousClass.php.inc
```diff
@@ -0,0 +1,148 @@
...
+                )
+            )
+        )
+        attributes: array(
+            classNode: null
+            className: null
+            methodName: null
+        )
+    )
+)
```
TomasVotruba: This would be hell to maintain if php-parser or Rector changes any attribute/property.

Better check this with ->getAttribute() on particular node

jeroensmit: That was my original plan, but it turned out that it was hard to determine which node was what and how it related to other nodes. 

Since the attributes you want rendered are whitelisted only attributes that are relevant to the visitor you want to test should be shown. I consider any changes to these attributes relevant to the test as well.

As a solution about the maintainability (and because no sane man would type this dumps themselves) you can run this command:

`UPDATE_FIXTURE=1 vendor/bin/phpunit`

It will replace the dump.

TomasVotruba: I see. Let's give it a try then

PR #2250, opened 2019-11-04
Fixed edge cases of RemoveDeadStmtRector

jeroensmit:
Also covered all nodes that could be considered dead statement.

! important !
I changed the way it handles comments. Previously it didn't remove to preserve comments. Now the node is changed with a NOP preserving the command but removing the dead code.


TomasVotruba:
Great work :+1:

PR #2270, opened 2019-11-06
ParamTypeDeclarationRector with [optional] tag behind variable name

orklah:
Hi,

I saw an issue while using Rector in a project.

It seems that when there is a [optional] tag behind the variable name, the type is not added in the signature.

I couldn't launch the tests on my computer(I'm on windows and all my paths are broken).

So, I'm trying to add a failing test to show the issue on github. I will modify this PR if it doesn't fail the tests.

Thanks

### DIFF
--- a/packages/TypeDeclaration/tests/Rector/FunctionLike/ParamTypeDeclarationRector/Fixture/optional_tag_behind_variable_name.php.inc
+++ b/packages/TypeDeclaration/tests/Rector/FunctionLike/ParamTypeDeclarationRector/Fixture/optional_tag_behind_variable_name.php.inc
@@ -1,0 +1,26 @@
+<?php
+
+namespace Rector\TypeDeclaration\Tests\Rector\FunctionLike\ParamTypeDeclarationRector\Fixture;
+
+/**
+ * @var string $string[optional]
+ */
+function test($string = null){
+
+}
+
+?>
+-----
+<?php
+
+namespace Rector\TypeDeclaration\Tests\Rector\FunctionLike\ParamTypeDeclarationRector\Fixture;
+
+/**
+ * @var string $string[optional]
+ */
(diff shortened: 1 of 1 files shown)


TomasVotruba:
> I couldn't launch the tests on my computer(I'm on windows and all my paths are broken).

TomasVotruba:
Hey, thanks for the report!

It seems there is `@var` instead of `@param`

> I couldn't launch the tests on my computer(I'm on windows and all my paths are broken).

We could fix that with your help. It would be better I think

orklah:
> Hey, thanks for the report!
> 
> It seems there is `@var` instead of `@param`

Good catch, I fixed the phpdoc tag.

> > I couldn't launch the tests on my computer(I'm on windows and all my paths are broken).
> 
> We could fix that with your help. It would be better I think

We could try but I'm not very competent with command lines. Not sure I'm the best suited for the job!

TomasVotruba:
> We could try but I'm not very competent with command lines. Not sure I'm the best suited for the job!

Well, just create an issue with error message your get. Using `--debug` would be nice.
I think it would be mostly related to slashes: `\` `/`

orklah:
I couldn't reproduce, when I went to check my issue again, I find out that the type and variable were inverted in phpdoc (@param $username string[optional]) It was not due to [optional] tag at all. I'll close this issue. I still have one more issue to report.

Ok for the --debug. I'll create the issue

TomasVotruba:
Thanks! We might handle the doc issues as well. Anything that bothers developer and is on many places, should be automated

```diff
-@param $value string
+@param string $value 
```

There have been similar issues in the past, that can be automated.

```diff
-/*
+/**
```

Could you create an issue for your problem?

orklah:
Yes, I will

PR #2275, opened 2019-11-06
fix no-regular naming

### LINKED ISSUE #2187: Rector\CodingStyle\Rector\ClassMethod\ReturnArrayClassMethodToYieldRector should only apply to methods defined in a `@dataProvider` annotation.
I've added `config/set/phpunit/phpunit-yield-data-provider.yaml` to my `imports`, which does the following : 

```yaml
services:
    Rector\CodingStyle\Rector\ClassMethod\ReturnArrayClassMethodToYieldRector:
        PHPUnit\Framework\TestCase:
            - '#(provide|dataProvider)*#'
```

I would have assumed that this would apply only on methods targeted by a dataProvider, but it's not. I had a `private function getSomeData(): array;` method in a test that got changed by the Rector (furthermore, data providers in PHPUnit are **always** `public`!).

Is there something I misunderstood about the intended behavior of this config file ? Or should I try to add a broken test ?

TomasVotruba:
Closes #2187


TomasVotruba:
@gnutix :man_facepalming:  I just realized that naming support only `*` fnmatch and FQN types

gnutix:
@TomasVotruba Thanks! Missed [updating the config set](https://github.com/rectorphp/rector/pull/2277) though, so I made another PR :)

PR #2277, opened 2019-11-06
Update rule with proper syntax of ReturnArrayClassMethodToYieldRector

gnutix:


gnutix:
Also I was thinking : what about supporting methods referenced by the `@dataProvider [methodName]` PHPUnit annotation ? Do you think it should be implemented ?

Example of a failing test case `packages/CodingStyle/tests/Rector/ClassMethod/ReturnArrayClassMethodToYieldRector/Fixture/data_provider_via_annotation_only.php.inc` : 

```php
<?php

namespace Rector\CodingStyle\Tests\Rector\ClassMethod\ReturnArrayClassMethodToYieldRector\Fixture;

use Rector\CodingStyle\Tests\Rector\ClassMethod\ReturnArrayClassMethodToYieldRector\Source\ParentTestCase;

class DataProviderViaAnnotationOnly extends ParentTestCase
{
    /**
     * @dataProvider someMethod
     */
    public function testSomething($arg1, $arg2, $arg3): void
    {
    }

    public function someMethod(): array
    {
        return [['some', 'test', 'arguments']];
    }
}

?>
-----
<?php

namespace Rector\CodingStyle\Tests\Rector\ClassMethod\ReturnArrayClassMethodToYieldRector\Fixture;

use Rector\CodingStyle\Tests\Rector\ClassMethod\ReturnArrayClassMethodToYieldRector\Source\ParentTestCase;

class DataProviderViaAnnotationOnly extends ParentTestCase
{
    /**
     * @dataProvider someMethod
     */
    public function testSomething($arg1, $arg2, $arg3): void
    {
    }

    public function someMethod(): \Iterator
    {
        yield ['some', 'test', 'arguments'];
    }
}

?>
```

TomasVotruba:
Well, this rule is generic and configurable to any class/pattern, not just data providers (it's just an example).

If you want to add such a rule for data providers, you can re-use this in a new rule, that would collect all method names from `@dataProvider` annotation and then apply the logic on there class methdos.

PR #2281, opened 2019-11-07
Fix ECS for windows

orklah:
I added a line feed parameter to ecs.yaml to enforce \n as line feed. It fixes the LineEndingFixer issue on windows.


TomasVotruba:
Thanks!

PR #2293, opened 2019-11-08
RemoveSetterOnlyPropertyAndMethodRector and UnusedPrivatePropertyRector fixes

jeroensmit:
Once again got a little out of hand.

Bugs fixed:

- static property accessing using $this was not recognized
- removing assigns also removed logic from the left side
- variable property names are only relevant when used in read context
- rectors couldn't handle multiple properties in 1 declaration

Both rectors now share the same logic.

The old way RemoveSetterOnlyPropertyAndMethodRector turned out to be rather slow on our codebase (30x slower than all other rectors). This in now fixed.


TomasVotruba:
Thank you! :+1: 

New test fixtures makes this quite easy to review

PR #2300, opened 2019-11-09
Give testset a name

jeroensmit:


TomasVotruba:
What's the reason behind this?

jeroensmit:
Is easier to spot which example file is causing exceptions

PHPunit output:

before change:
Rector\DeadCode\Tests\Rector\Class_\RemoveSetterOnlyPropertyAndMethodCallRector\RemoveSetterOnlyPropertyAndMethodCallRectorTest::test with data set #0 

after change:
Rector\DeadCode\Tests\Rector\Class_\RemoveSetterOnlyPropertyAndMethodCallRector\RemoveSetterOnlyPropertyAndMethodCallRectorTest::test with data set "deal_with_property_fetches.php.inc"

TomasVotruba:
How would you re-run only the file?

jeroensmit:
the complete name filter filters on looks something like:
Rector\Architecture\Tests\Rector\Class_\ConstructorInjectionToActionInjectionRector\ConstructorInjectionToActionInjectionRectorTest::test with data set "duplicate.php.inc"

The filter option works as an regex so you could do:

```bash
vendor/bin/phpunit --filter ConstructorInjectionToAction.*duplicate
```
or

```bash
vendor/bin/phpunit --filter ConstructorInjectionToAction --filter duplicate
```

a part of the file names suffices (doesn't need to be from te start)

TomasVotruba:
How would you compare it to `--filter test#1`

TomasVotruba:
Having both options would be something I'd merge:

```bash
Rector\DeadCode\Tests\Rector\Class_\RemoveSetterOnlyPropertyAndMethodCallRector\RemoveSetterOnlyPropertyAndMethodCallRectorTest::test with data set "deal_with_property_fetches.php.inc", #0   
```

jeroensmit:
Just what I was thinking.

Sadly the best thing I can reach is this:

`vendor/bin/phpunit --filter RemoveSetter@1.*`

If you find that sufficient I'll update the pullrequest. 
Otherwise decline. I can live with that ;-)

TomasVotruba:
Yes, that's what I meant :) go for it

src/Testing/StaticFixtureProvider.php
```diff
@@ -14,8 +14,10 @@ public static function yieldFilesFromDirectory(string $directory, string $suffix
     {
         $fileInfos = self::findFilesInDirectory($directory, $suffix);
 
+        $key = 0;
         foreach ($fileInfos as $fileInfo) {
-            yield $fileInfo->getFilename() => [$fileInfo->getPathName()];
+            yield "{$key} " . $fileInfo->getFilename() => [$fileInfo->getPathName()];
+            $key++;
```
TomasVotruba: Why not use key of this foreach?

jeroensmit: key contains the filepath

TomasVotruba: Ah, I used `FinderSanitizer` to remove this WTF

Ok

TomasVotruba:
Thank you

TomasVotruba:
I'm finally testing it an it breaks filtering by provider name `test#1`

```bash
vendor/bin/phpunit packages/PHPUnit/tests/Rector/MethodCall/ReplaceAssertArraySubsetRector/ReplaceAssertArraySubsetRectorTest.php --filter test#1 
```

jeroensmit:
Sorry for taking a while.

Hopefully you managed by now, if not the solution was using @ instead of #
test@1.*

PR #2308, opened 2019-11-11
Fix removing 0 when on left side of Minus

jeroensmit:


packages/DeadCode/tests/Rector/Plus/RemoveDeadZeroAndOneOperationRector/Fixture/zero_minus.php.inc
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace Rector\DeadCode\Tests\Rector\Plus\RemoveDeadZeroAndOneOperationRector\ZeroMinus;
+
+class Assigns
+{
+    public function run(int $value)
+    {
+        $value = 0 - $value;
```
TomasVotruba: Shouldn't this be:

```php
$value = -$value;
```

?

jeroensmit: You're right, PR updated

TomasVotruba: Thank you!

PR #2318, opened 2019-11-13
Fixed issue with sed command:

jeroensmit:
on macOs the -E option was taken as the argument of the -i option causing backups to be made with the addional extension -E.
also the backslash of the \s needed additional escaping

### DIFF
--- a/ci/clean_trailing_spaces.sh
+++ b/ci/clean_trailing_spaces.sh
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 # trailing whitespaces
-sed -i -E 's#\s+$##g' config/set/*/*.yaml docs/*.md README.md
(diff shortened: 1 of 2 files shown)


ci/clean_trailing_spaces.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 # trailing whitespaces
-sed -i -E 's#\s+$##g' config/set/*/*.yaml docs/*.md README.md
+sed -i '' -E 's#\\s+$##g' config/set/*/*.yaml docs/*.md README.md
```
TomasVotruba: This now keeps failing on Linux.

Travis didn't fail, because the check was not added there. I've added it:
https://github.com/rectorphp/rector/pull/2322

jeroensmit: Fix for unix and macOs both is #2323

PR #2319, opened 2019-11-13
Test exclude_paths weird behaviour

JanMikes:


JanMikes:
In this case it is weird that `'Exception.php'` is in excluded paths but even SomeException.php is ignored.

TomasVotruba:
Rebased on master...

TomasVotruba:
CI build passes: https://travis-ci.org/rectorphp/rector/builds/612473224?utm_source=github_status&utm_medium=notification


If it still fails on your machine, could try same build on Travis just with Mac.

TomasVotruba:
image

PR #2321, opened 2019-11-13
[Nette 3.0]  Add Nette 2.x to Nette 3 upgrade set

TomasVotruba:


ci/clean_trailing_spaces.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 # trailing whitespaces
-sed -i '' -E 's#\\s+$##g' config/set/*/*.yaml docs/*.md README.md
+sed -i -E 's#\s+$##g' config/set/*/*.yaml docs/*.md README.md
```
TomasVotruba: Reverted because of #2322

PR #2337, opened 2019-11-19
Add more refactorings for CakePHP 4

markstory:
These help add more return type and fix parameter errors during the upgrade from cake3 to cake4.


TomasVotruba:
Thank you

PR #2353, opened 2019-11-24
Fixed error on classConstFetch outside class

jeroensmit:


src/NodeContainer/ParsedNodesByType.php
```diff
@@ -677,6 +677,10 @@ private function isThisVariable(Node $node): bool
             /** @var string $className */
```
TomasVotruba:
```suggestion
            /** @var string|null $className */
```

Issue #2354, opened 2019-11-25
CakePHP 4 rectors no longer applying


markstory:
| Subject        | Details                                                      |
| :------------- | :----------------------------------------------------------- |
| Rector version | Rector 0.6.x-dev@0eceb8a          |
| PHP version    | e.g. PHP 7.3.6                               |
| Full Command   | see below  | 

## Current Behaviour 

While rector runs fine it isn't modifying code that it should. The minimal reproduction below does not change the file :cry:

I ran a git bisect using the repro below and ended up on bb7125361ec97eaab1425a9aebc5e77694faa6f1 as the 'bad' commit. I don't have enough context on this project to know what is actually going on with this change, but the NodeTypeResolver changes seem related.

## Minimal PHP Code Causing Issue

You'll need a small CakePHP application. You can get one with `composer create-project cakephp/app:4.0.0-RC1`, the test script assumes you've put it in `~/cake-app`. After you have that add the following file.

```php
<?php 
// in ~/cake-app/rector-test.php
require 'vendor/autoload.php';
require 'config/bootstrap.php';

use Cake\Console\ConsoleOutput;

function thing() {
    $s = new ConsoleOutput();
    $s->styles();
    $s->styles('error');
}
```

Then run `bin/rector process -d ~/cake-app/ -a ~/cake-app/vendor/autoload.php -s cakephp40 --dry-run ~/cake-app/rector-test.php`

## Expected Behaviour

The renaming rectors for modal methods should continue to apply.

TomasVotruba:
Could add failing test case to cake PHP modal set?

markstory:
@TomasVotruba I can try to reproduce the type inference issue in the test suite.

markstory:
Thanks for the quick fix :+1:

PR #2356, opened 2019-11-26
Add test case for #2354

markstory:


packages/CakePHP/tests/Rector/MethodCall/ModalToGetSetRector/Source/SomeModelType.php
```diff
@@ -4,7 +4,13 @@
 
 namespace Rector\CakePHP\Tests\Rector\MethodCall\ModalToGetSetRector\Source;
 
-final class SomeModelType
+class Entity
 {
+}
 
+final class SomeModelType
```
TomasVotruba: These should be 2 separated files, as that's `/Source/`

markstory: Fixed.

PR #2366, opened 2019-11-29
Fix typo in AllRectorsOverview.md

berezuev:


TomasVotruba:
This needs to be fixed at the rule class, as this file is generated

PR #2370, opened 2019-11-30
Update expectations for PHP74 TypedPropertyRector...

mallardduck:
...per #2367 setting proper expectations as the rector doesn't touch DocBlocks.

PR #2371, opened 2019-11-30
Update FilesFinder.php to fix builds...

mallardduck:
...the IdenticalFalseToBooleanNotRector file in testing is causing this line to fail builds.


mallardduck:
Once this is merged in, then I believe that #2370 can also be merged in.

PR #2389, opened 2019-12-04
composer: lock phpstan to 0.11.19 due to breaking changes [closes #2385]

### LINKED ISSUE #2385: Doesn't work with phpstan 0.12
A new version of phpstan has been published and rector doesn't work with it.

TomasVotruba:


ondrejmirtes:
This doesn't solve anything. ^0.11.19 does not include 0.12.0.

PR #2407, opened 2019-12-07
Added missing methods

ruudboon:
Did a small scan on the current Phalcon40 project. And noticed these missing changes


TomasVotruba:
Nice spot!

PR #2373, opened 2019-11-30
Another attempt to add Compiler + upgrade to PHPStan 0.12

TomasVotruba:
- Reference #177 - still needs lot of testing in the wild
- Closes #2385 

### Inspiration

- https://github.com/phpstan/phpstan-src/commit/339819882edeccf87db133b6856b9f4c707c5489

PR #2410, opened 2019-12-09
Added default to prevent rector from breaking

jeroensmit:
when rector.yaml does not contain rector_recipe


TomasVotruba:
I fixed it in #2373, but this is faster

PR #2420, opened 2019-12-10
Fix NewToStaticCallRector documentation

RusiPapazov:
Fix arguments documentation for `Rector\Rector\New_\NewToStaticCallRector`

### DIFF
--- a/src/Rector/New_/NewToStaticCallRector.php
+++ b/src/Rector/New_/NewToStaticCallRector.php
@@ -53,7 +53,7 @@ class SomeClass
 PHP
                 ,
                 [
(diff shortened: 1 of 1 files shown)


TomasVotruba:
This file is generated, so it has to be changed directly in `NewToStaticCallRector` class instead

TomasVotruba:
Thank you :+1:

RusiPapazov:
@TomasVotruba The question is: when is it generated? I updated the .md again but didn't push the commit, decided it's not needed as you said `instead`. But I still don't see the changes on the master branch's docs.

TomasVotruba:
Manually now occasionally

PR #2423, opened 2019-12-11
Add test case for removing paranthesis in TernaryToElvisRector

JoyceBabu:
Add test case for removing paranthesis when replacing ternary operator with elvis operator


TomasVotruba:
Thank you :+1:

PR #2456, opened 2019-12-19
Added failing test case to the RenameMethodRectorTest

lulco:


packages/Renaming/tests/Rector/MethodCall/RenameMethodRector/Fixture/fixture7.php.inc
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace Rector\Renaming\Tests\Rector\MethodCall\RenameMethodRector\Fixture;
+
+use Rector\Renaming\Tests\Rector\MethodCall\RenameMethodRector\Source\AbstractType;
+
+class SomeClass8
```
TomasVotruba: @lulco I mean this class. Is it in the test case configuration?

lulco: Yes, this test is executed, I tried it on my local machine. I just changed output to:
```
public function doSomething()
{
    $this->someOtherClass->hmhmhm();
}
```
and the test failed:
```
There was 1 failure:

1) Rector\Renaming\Tests\Rector\MethodCall\RenameMethodRector\RenameMethodRectorTest::test with data set #0 ('/var/www/libs/lulco/rector/pa...hp.inc')
Caused by /var/www/libs/lulco/rector/packages/Renaming/tests/Rector/MethodCall/RenameMethodRector/Fixture/fixture7.php.inc
--- Expected
+++ Actual
@@ @@
 
     public function doSomething()
     {
-        $this->someOtherClass->hmhmhm();
+        $this->someOtherClass->configureOptions();
     }
 }
```

TomasVotruba: It's missing here:
https://github.com/rectorphp/rector/blob/master/packages/Renaming/tests/Rector/MethodCall/RenameMethodRector/RenameMethodRectorTest.php#L35

lulco: I used AbstractType renaming rule setDefaultOptions -> configureOptions... so I don't think it's missing

TomasVotruba: I mean in this test case. It's passing because `SomeClass8` is not set anywhere. You need to add it:

```diff
RenameMethodRector::class => [
                '$oldToNewMethodsByClass' => [
+                    SomeClass8::class => [
```

lulco: hm MyCustomType from fixture.php.inc, SomeClass from fixture3.php.inc aren't there also and it works or am I missing something?

I was thinking in this array should be only mapping of renaming methods. SomeClass8 has no method to be renamed. As I mention before, I used AbstractType renaming rule setDefaultOptions -> configureOptions

TomasVotruba: Ah, sorry. I got lost in the code. The `AbstractType`is correct.

So if the test passes, what's the issue?

lulco: Issue is that it is not working on my application (repository created https://github.com/lulco/rector-nette-api)

On this line should be replaced method getApiHandler to getApi: https://github.com/lulco/rector-nette-api/blob/master/nette-api-app/app/Presenters/ApiPresenter.php#L48

I can't find the difference

TomasVotruba: Ah, sorry. I got lost in the code. The AbstractTypeis correct.

So if the test passes, what's the issue?

TomasVotruba: I'll check the repo :+1: 

Not sure when, I'm quite busy next couple of days

lulco: Thank you :)

PR #2476, opened 2019-12-24
Add ScanFatalErrors command

### LINKED ISSUE #2388: [NFR] Scan for fatal errors
When an interface changes the extended class should be adjusted.
Normally PHP will let you know that by generating a fatal error. It would be nice if you could scan your codebase for these kind of errors. Since it's a fatal error you need something to make sure the script continues.
I created a prototype to show a way how to solve this. It's a prototype just to show the idea. Code needs some improvement ;)

Our fatal error (test.php)
```php
<?php
    class test {
        public function add(string $message="") {

        }
    }

    class extendTest extends test {

        public function add($message) {

        }
    }
```
And the scanner
```php
<?php

class interfaceErrorDetector {
    public $errors=[];

    public function __construct()
    {
        set_error_handler([$this, 'logError']);
    }

    public function logError ( $num, $str, $file, $line, $context = null) {
        $this->errors[$file] = $str;
    }

    public function scanFile($file) {
        $content = str_replace("<?php", "", file_get_contents($file));
        eval($content);
    }

    public function __destruct()
    {
        restore_error_handler();
    }
}

$handler = new interfaceErrorDetector();
$handler->scanFile("test.php");
print_r($handler->errors);
```

TomasVotruba:
Closes #2388

Useful for creating param/return type completion based on your specific code

```bash
vendor/bin/rector scan-fatal-errors src
```

Then run Rector to complete param/return types:

```bash
vendor/bin/rector process src --config rector-types.yaml
```


TomasVotruba:
Now as a part of the build:

https://travis-ci.org/rectorphp/rector/jobs/629125978#L406

PR #2492, opened 2019-12-26
[docs] Fix codesample in TypedPropertyRector

ruudk:
This rector does not remove the PHPDoc.


TomasVotruba:
This part of the docs is actually generated with:

```bash
composer docs
```

So to update it, you need to change the `Rector` rule and then re-run this command

TomasVotruba:
Mergin as it reflects the rule content in parallel PR

PR #2500, opened 2019-12-26
[PHP] add PHP version feature checks

fsok:
This PR adds more PHP version checks to the PHP rectors.


TomasVotruba:
Very nice job :+1:

Issue #2506, opened 2019-12-27
[TypedPropertyRector] Compile Error: Type of Child::$name must not be defined (as in class VendorParent)


ruudk:
| Subject        | Details                                                      |
| :------------- | :----------------------------------------------------------- |
| Rector version | v0.6.3          |
| PHP version    | PHP 7.4                               |
| Full Command   | vendor/bin/rector process --dry-run file.php  | 

## Problem

There is a problem with TypedPropertyRector. When `Child` extends `VendorParent` (package coming from Composer) and overwrites a property with a default value, Rector will try to apply the `string` property type. 

When running the code, this fails:
```
 Compile Error: Type of Child::$name must not be defined (as in class VendorParent)
```

The reason is that the parent is not changed, so the parent has no type, but the child introduces a type. That's not allowed in PHP 7.4.

## Failing test scenario:
vendor/rector/rector/packages/Php74/tests/Rector/Property/TypedPropertyRector/Fixture/parent_has_untyped_property.php.inc:
```
<?php

namespace Rector\Php74\Tests\Rector\Property\TypedPropertyRector\Fixture;

class VendorParent
{
    /**
     * @var string
     */
    protected $name;
}

final class Child extends VendorParent
{
    /**
     * @var string
     */
    protected $name = 'child';
}

?>
-----
<?php

namespace Rector\Php74\Tests\Rector\Property\TypedPropertyRector\Fixture;

class VendorParent
{
    /**
     * @var string
     */
    protected $name;
}

final class Child extends VendorParent
{
    /**
     * @var string
     */
    protected $name = 'child';
}

?>

```

result:
```
Failed asserting that string matches format description.
--- Expected
+++ Actual
@@ @@
     /**
      * @var string
      */
-    protected $name;
+    protected string $name;
 }
 
 final class Child extends VendorParent
@@ @@
     /**
      * @var string
      */
-    protected $name = 'child';
+    protected string $name = 'child';
 }
 
 ?>
```

(it also tries to update the VendorParent, but that's because I didn't exclude this one. Should propably be done in the test somewhere, not sure where).

TomasVotruba:
I see.

So the expecting behavior is not to update vendor-locked cases? = skip yours?

ruudk:
Not sure if it's about "vendor" only, I think that the fixer should look at the parent. And only fix when the parent has that type. But that can be a race condition, what if the fixer first fixes the child and then the parent, it needs to come back to the child and fix that too. 

[...]

PR #2509, opened 2019-12-27
Do not suggest typed property when defined in vendored parent

ruudk:
When the property is defined in vendored code, it doesn't make sense to suggest typed properties on classes that extend from it.

Fixes https://github.com/rectorphp/rector/issues/2506

### DIFF
--- a/packages/Php74/src/Rector/Property/TypedPropertyRector.php
+++ b/packages/Php74/src/Rector/Property/TypedPropertyRector.php
@@ -11,6 +11,7 @@ use Rector\Rector\AbstractRector;
 use Rector\RectorDefinition\CodeSample;
 use Rector\RectorDefinition\RectorDefinition;
 use Rector\TypeDeclaration\TypeInferer\PropertyTypeInferer;
+use Rector\TypeDeclaration\VendorLock\VendorLockResolver;
 use Rector\ValueObject\PhpVersionFeature;
 
 /**
@@ -25,9 +26,15 @@ final class TypedPropertyRector extends AbstractRector
      */
     private $propertyTypeInferer;
 
-    public function __construct(PropertyTypeInferer $propertyTypeInferer)
+    /**
+     * @var VendorLockResolver
+     */
+    private $vendorLockResolver;
+
+    public function __construct(PropertyTypeInferer $propertyTypeInferer, VendorLockResolver $vendorLockResolver)
     {
         $this->propertyTypeInferer = $propertyTypeInferer;
+        $this->vendorLockResolver = $vendorLockResolver;
     }
 
     public function getDefinition(): RectorDefinition
@@ -89,6 +96,10 @@ PHP
 
         $propertyTypeNode = $this->staticTypeMapper->mapPHPStanTypeToPhpParserNode($varType, 'property');
         if ($propertyTypeNode === null) {
+            return null;
+        }
+
+        if ($this->vendorLockResolver->isPropertyChangeVendorLockedIn($node)) {
             return null;
         }
 
--- a/packages/Php74/tests/Rector/Property/TypedPropertyRector/Fixture/parent_of_parent_has_untyped_property.php.inc
+++ b/packages/Php74/tests/Rector/Property/TypedPropertyRector/Fixture/parent_of_parent_has_untyped_property.php.inc
@@ -1,0 +1,24 @@
+<?php
+
+namespace Rector\Php74\Tests\Rector\Property\TypedPropertyRector\Fixture;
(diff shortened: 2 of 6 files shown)


ruudk:
How come the PHP74 tests are executed on PHP 7.2? Shouldn't they be excluded? 

https://travis-ci.org/rectorphp/rector/jobs/630042621?utm_medium=notification&utm_source=github_status

tests/Source/SomeParent.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+// class that is used all over the code
+// autoloaded manually via composer to keep the rest of code clean
+namespace SomeNamespace;
+
+abstract class SomeParent
```
TomasVotruba: This class should be located next to the test, here create a new directory `Source`:

https://github.com/rectorphp/rector/tree/master/packages/Php74/tests/Rector/Property/TypedPropertyRector

And make class respect PSR-4 naming, so it's autolaoded by composer by default

packages/TypeDeclaration/src/VendorLock/VendorLockResolver.php
```diff
@@ -160,13 +174,66 @@ public function isReturnChangeVendorLockedIn(ClassMethod $classMethod): bool
...
+                return true;
+
+                // if not, look for it's parent parent - @todo recursion
+            }
+        }
+
+        return false;
+    }
+
+    // Until we have getProperty (https://github.com/nikic/PHP-Parser/pull/646)
```
TomasVotruba: :clap:  nice

packages/Php74/tests/Rector/Property/TypedPropertyRector/Fixture/parent_of_parent_has_untyped_property.php.inc
```diff
@@ -0,0 +1,45 @@
...
+    protected $name = 'child';
+
+    /**
+     * @var string
+     */
+    protected string $typedName = 'child';
+}
+
+?>
+-----
```
TomasVotruba: If the file should not be changed at all, you can remove all content from here to the end of the file.

TomasVotruba:
> How come the PHP74 tests are executed on PHP 7.2? Shouldn't they be excluded?

There is no limit on the class, so `test()` method runs and fixtures are autoloaded => thus the error.

To skip the test class even running, you need to add this above it:

```php
/**
 * @requires PHP >= 7.4
 */
```

tests/Source/SomeParent.php
ruudk: Done.

packages/Php74/tests/Rector/Property/TypedPropertyRector/Fixture/parent_of_parent_has_untyped_property.php.inc
ruudk: Removed.

ruudk:
Applied all feedback and added `@requires PHP >= 7.4` to the test.

composer.json
```diff
@@ -171,6 +171,7 @@
             "packages/Symfony/tests/Rector/FrameworkBundle/ContainerGetToConstructorInjectionRector/Source",
             "packages/NodeTypeResolver/tests/PerNodeTypeResolver/ParamTypeResolver/Source",
             "packages/NodeTypeResolver/tests/PerNodeTypeResolver/PropertyTypeResolver/Source",
+            "packages/Php74/tests/Rector/Property/TypedPropertyRector/Source",
```
TomasVotruba: This can be removed, because it's already covered in `psr-4` here

packages/Php74/src/Rector/Property/TypedPropertyRector.php
```diff
@@ -92,6 +99,10 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
+        if ($this->vendorLockResolver->isPropertyChangeVendorLockedIn($node)) {
+            return null;
+        }
```
TomasVotruba: Beautiful work :+1:

composer.json
ruudk: Oops, fixed.

TomasVotruba:
Thank you :+1:

ruudk:
@TomasVotruba Could you please tag a new release including this fix? 🙏

TomasVotruba:
:+1: 

There you go 
https://github.com/rectorphp/rector/blob/master/CHANGELOG.md#v065---2020-01-03

PR #2518, opened 2019-12-29
[TypeDeclaration] Fix static property type resolution

### LINKED ISSUE #2516: PropertyTypeDeclarationRector reports wrong type
| Rector version | 0.6.4          |
| PHP version    | PHP 7.3                           |
| Full Command   | `rector process redaxo/src/core/lib --autoload-file phpstan-bootstrap.php`  | 

## Current Behaviour 

running `rector process redaxo/src/core/lib --autoload-file phpstan-bootstrap.php` thru 
https://github.com/redaxo/redaxo suggests that the property `$cacheFile`[1] should have a phpdoc of `null`

image

when looking thru the class this property is writted from a method which returns a `string` on https://github.com/redaxo/redaxo/blob/735266da7cefc0d7caa259f7e39c9ac924e2acb4/redaxo/src/core/lib/autoload.php#L51

also the file is passed into `file_get_contents` which also could indicate a `string`:
https://github.com/redaxo/redaxo/blob/735266da7cefc0d7caa259f7e39c9ac924e2acb4/redaxo/src/core/lib/autoload.php#L134

another call site, where the property is passed and is expected as a `string`
https://github.com/redaxo/redaxo/blob/735266da7cefc0d7caa259f7e39c9ac924e2acb4/redaxo/src/core/lib/autoload.php#L164

(even more call-sites like that)

is there a reason why rector is not able to indentify the `string`  type?

[1] https://github.com/redaxo/redaxo/blob/735266da7cefc0d7caa259f7e39c9ac924e2acb4/redaxo/src/core/lib/autoload.php#L21

using rector.yaml:
```
# rector.yaml
services:
    Rector\TypeDeclaration\Rector\Property\PropertyTypeDeclarationRector: ~
```

TomasVotruba:
Closes #2516


packages/TypeDeclaration/tests/Rector/Property/PropertyTypeDeclarationRector/Fixture/typed_redax.php.inc
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace Rector\TypeDeclaration\Tests\Rector\FunctionLike\PropertyTypeDeclarationRector\Fixture;
+
+class TypedRedax
```
staabm: Nit.. the tool is named „redaxo“
```suggestion
class TypedRedaxo
```

TomasVotruba: I didn't notice that. It was just work in progress, untill I found the problem.
Thank for pointing it out, I give me more reason to think about better naming.

Got descriptive name now :)

PR #2532, opened 2019-12-30
[PHP 7.4] Add @var removal to TypedPropertyRector

### LINKED ISSUE #2367: TypedPropertyRector does not remove docblocks
rector.yaml:

```yaml
parameters:
    php_version_features: '7.4'

services:
    Rector\Php74\Rector\Closure\ClosureToArrowFunctionRector: ~
    Rector\Php74\Rector\Property\TypedPropertyRector: ~
```

```php
    /** @var string */
    public $name;
```

becomes

```php
    /** @var string */
    public string $name;
```

I expect it to be

```php
    public string $name;
```

TomasVotruba:
Closes #2367 propperly

## Before

```diff
 /**
  * @var int
  */
-public $count;
+public int $count;
```

## Now

```diff
-/**
- * @var int
- */
-public $count;
+public int $count;
```

/cc @brendt @enumag @collapso @mallardduck @ruudk

### DIFF
--- a/packages/NodeTypeResolver/src/StaticTypeMapper.php
+++ b/packages/NodeTypeResolver/src/StaticTypeMapper.php
@@ -554,6 +554,13 @@ final class StaticTypeMapper
         throw new NotImplementedException(sprintf('%s for "%s"', __METHOD__, $type));
     }
 
+    public function mapPHPStanPhpDocTypeNodeToPhpDocString(TypeNode $typeNode, Node $node): string
+    {
+        $phpStanType = $this->mapPHPStanPhpDocTypeNodeToPHPStanType($typeNode, $node);
+
+        return $this->mapPHPStanTypeToDocString($phpStanType);
+    }
+
     public function mapPHPStanPhpDocTypeNodeToPHPStanType(TypeNode $typeNode, Node $node): Type
     {
         if ($typeNode instanceof IdentifierTypeNode) {
--- a/packages/NodeTypeResolver/src/StaticTypeMapper.php
+++ b/packages/NodeTypeResolver/src/StaticTypeMapper.php
@@ -662,6 +662,10 @@ final class StaticTypeMapper
                         return new ArrayType(new MixedType(), $genericType);
                     }
 
+                    if ($typeName === 'Traversable') {
+                        return new ObjectType('Traversable');
+                    }
+
                     return new IterableType(new MixedType(), $genericType);
                 }
             }
--- a/packages/NodeTypeResolver/src/StaticTypeMapper.php
+++ b/packages/NodeTypeResolver/src/StaticTypeMapper.php
@@ -31,6 +31,7 @@ use PHPStan\Type\CallableType;
 use PHPStan\Type\ClosureType;
 use PHPStan\Type\ConstantType;
 use PHPStan\Type\FloatType;
+use PHPStan\Type\Generic\GenericObjectType;
 use PHPStan\Type\IntegerType;
 use PHPStan\Type\IntersectionType;
 use PHPStan\Type\IterableType;
@@ -645,7 +646,11 @@ final class StaticTypeMapper
         if ($typeNode instanceof GenericTypeNode) {
             if ($typeNode->type instanceof IdentifierTypeNode) {
                 $typeName = $typeNode->type->name;
-                if (in_array($typeName, ['array', 'iterable'], true)) {
+
+                // remove extra prefix
+                $typeName = ltrim($typeName, '\\');
+
+                if (in_array($typeName, ['array', 'iterable', 'Traversable'], true)) {
                     $genericTypes = [];
                     foreach ($typeNode->genericTypes as $genericTypeNode) {
                         $genericTypes[] = $this->mapPHPStanPhpDocTypeNodeToPHPStanType($genericTypeNode, $node);
@@ -660,6 +665,20 @@ final class StaticTypeMapper
                     return new IterableType(new MixedType(), $genericType);
                 }
             }
+
+            $mainType = $this->mapPHPStanPhpDocTypeNodeToPHPStanType($typeNode->type, $node);
+            if ($mainType instanceof TypeWithClassName) {
+                $className = $mainType->getClassName();
+            } else {
+                throw new NotImplementedException();
+            }
+
+            $genericTypes = [];
+            foreach ($typeNode->genericTypes as $genericType) {
+                $genericTypes[] = $this->mapPHPStanPhpDocTypeNodeToPHPStanType($genericType, $node);
+            }
+
+            return new GenericObjectType($className, $genericTypes);
         }
 
         throw new NotImplementedException(__METHOD__ . ' for ' . get_class($typeNode));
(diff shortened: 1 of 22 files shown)


ruudk:
Too bad that I did the work already in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4713... Any idea how I can get that merged?

TomasVotruba:
@ruudk That's ok, both tools have their user scope. I know many people who use CS only and many more who use Rector only.

Also, you give me next piece of motivation to make this happen. In older version of Rector this would be really complicatd. But with new API around php doc it was quite easy. So thank you as well :+1:

TomasVotruba:
@ruudk Btw, next time you can contribute here first. I do merge almost any feature improvements.

ruudk:
It was more that you said in the other issue it belongs in CS fixer, that’s why I did the work there , and that’s why I was surprised by this PR. But in the end, it does not matter, great to have this automatically fixable by either CS fixer or Rector ☺️

TomasVotruba:
At first yes, because it seemed as extra work not worth doing for me. Working PR with test is hard to reject :)

packages/Php74/tests/Rector/Property/TypedPropertyRector/FixtureUnionTypes/two_types.php.inc
```diff
@@ -18,9 +18,6 @@ namespace Rector\Php74\Tests\Rector\Property\TypedPropertyRector\FixtureUnionTyp
 
 final class TwoTypes
 {
-    /**
-     * @var bool|string
-     */
     private bool|string $unionValue;
```
enumag: This test seems to be in a wrong namespace. Union types are part of PHP 8.0, not 7.4.

packages/Php74/src/Rector/Property/TypedPropertyRector.php
```diff
@@ -103,8 +101,32 @@ public function refactor(Node $node): ?Node
...
+        if ($varTagValueNode === null) {
+            return;
+        }
+
+        // has description? keep it
+        if ($varTagValueNode->description !== '') {
+            return;
+        }
+
+        $propertyPhpDocInfo->removeByType(VarTagValueNode::class);
```
enumag: I don't see a condition that would preserve useful annotations such as

```
/**
 * @var array<int, string>
 */
private array $foo;
```

Is there a test that annotations with generic types are NOT removed?

packages/Php74/tests/Rector/Property/TypedPropertyRector/FixtureUnionTypes/two_types.php.inc
TomasVotruba: This tests property type, not union type

packages/Php74/src/Rector/Property/TypedPropertyRector.php
TomasVotruba: I need to add it

packages/Php74/src/Rector/Property/TypedPropertyRector.php
```diff
@@ -126,7 +129,44 @@ private function removeVarPhpTagValueNodeIfNotComment(Property $property): void
             return;
         }
 
+        // keep string[] etc.
+        if ($this->isNonBasicArrayType($property, $varTagValueNode)) {
```
enumag: What if it is non-array generic node?

```
/**
 * @var Traversable<int, string>
 */
private Traversable $foo;
```

or even non-iterable one?

```
/**
 * @var Promise<Foo>
 */
private Promise $foo;
```

Maybe the code already covers it but the method names shouldn't contain the word "array" imo.

TomasVotruba: I'll look on it, thanks

TomasVotruba: Added :+1:

packages/Php74/src/Rector/Property/TypedPropertyRector.php
TomasVotruba: Added

packages/NodeTypeResolver/src/StaticTypeMapper.php
```diff
@@ -645,7 +646,11 @@ public function mapPHPStanPhpDocTypeNodeToPHPStanType(TypeNode $typeNode, Node $
         if ($typeNode instanceof GenericTypeNode) {
             if ($typeNode->type instanceof IdentifierTypeNode) {
                 $typeName = $typeNode->type->name;
-                if (in_array($typeName, ['array', 'iterable'], true)) {
+
+                // remove extra prefix
+                $typeName = ltrim($typeName, '\\');
+
+                if (in_array($typeName, ['array', 'iterable', 'Traversable'], true)) {
```
enumag: What exactly is this condition for? There are more types like these - Iterator, Generator, ext-ds types, Doctrine collections, some SPL types... trying to white-list them for some special handling doesn't make much sense to me.

TomasVotruba: These all need to be listed there then. For phpdoc-parser it's just a string.

enumag: There is no way to whitelist everything. What is the purpose of this?

TomasVotruba: Only those that have type in object fo PHPStan.

The `mapPHPStanPhpDocTypeNodeToPHPStanType()` method converts doc type to phpstan type

TomasVotruba: If you're able to make it simpler and more robust, that would be great

enumag: > Only those that have type in object fo PHPStan.

Can you rephrase that? I don't understand...

Would something like `=== array || is_subclass_of(Traversable)` work?

TomasVotruba: > Can you rephrase that? I don't understand...

Sure, this is phpdoc-parsre doc type:

```php
@var int
```

To use in the code, we need to convert it to PHPStan type:

```php
$integerType = new IntegerType();
```

That's all :)


> Would something like === array || === iterable || is_subclass_of(Traversable) work?

Hard to tell, testing needs to be done, as this is 3rd party code we convert

TomasVotruba:
Thanks for the feedback! 

Feel free to add more failing cases in new issues. Merging as it is, so it can be used out in the wild

packages/NodeTypeResolver/src/StaticTypeMapper.php
TomasVotruba: Let's address in new PR

ruudk:
I just pulled in master to test this out, but this rector only removes the superfluous PHPDoc when the property does not have a type but does have a phpdoc:
```diff
-    /**
-     * @var Currency
-     */
-    public $fromCurrency;
+    public Currency $fromCurrency;
```

but it ignores this code: 
```php
    /**
     * @var Currency
     */
    public Currency $toCurrency;
```

Since I already ran the TypedPropertyRector before on my code, all my properties now have types, with phpdocs. Any idea how to solve that with Rector?

enumag:
@ruudk I'd recommend solving that with php-cs-fixer (with your patch).

TomasVotruba:
If you send PR to add this feature, I'd be happy to merge it

PR #2534, opened 2019-12-30
improve generic type conversion

TomasVotruba:
@enumag Apperently, only edge-case is generic `array`, that needs to be converted to array. It's treated as object without that, e.g. FQN by default `\array` and other problems.

Maybe there is more space for improvement. What do you think?


enumag:
Looks good! Just perhaps `iterable` might need special handling since it's not a class.

TomasVotruba:
@enumag Good point, updated + test included

Issue #1547, opened 2019-06-04
Problem with classes from global namespace like \Redis or \Memcached


pawel-brzezinski:
Hi,
I have a problem with classes from global namespace like \Redis or \Memcached. I run Rector by docker
```
docker run -v $(pwd):/project rector/rector:latest process /project/src --level symfony28 --autoload-file /project/vendor/autoload.php --dry-run
```
and I've got errors like this one:
```
"Analyze error: "Class Redis not found and could not be autoloaded.".
```
I suppose that should I point some more things than only vendor autoload?

Thanks for help!

JanMikes:
Hi @pawel-brzezinski !

First, I will explain what the problem is - if you check [Rector's Dockerfile](https://github.com/rectorphp/rector/blob/master/Dockerfile) it is pretty simple one, without much php extensions, but default ones. That means, in Rector's docker container (`rector/rector`) there is no Redis or Memcached extensions - so class \Redis or \Memcached does really not exist inside of it.

I have solved it on other project, where we had missing like 10 php extensions and i wanted to use rector (and it was failing for example on missing \Imagick class) this way - i created locally Dockerfile which extended rector/rector image, installing required extensions inside and then running this image instead of official one.

I think example will be worth more than 1000 words:

Dockerfile:
```Dockerfile
FROM rector/rector:latest

RUN apt update && apt install -y \
    gcc \
    git \
    unzip \
    gettext \
    libicu-dev \
    libmagickwand-dev \
    libfreetype6-dev libjpeg62-turbo-dev libmcrypt-dev libpng-dev \
    libxml2-dev \
    libxslt-dev \
    libgearman-dev \
    zlib1g-dev \
    libc-client-dev libkrb5-dev \
    libzip-dev \
    libmemcached-dev \
    --no-install-recommends

RUN pecl install imagick && docker-php-ext-enable imagick
RUN pecl install igbinary && docker-php-ext-enable igbinary
RUN pecl install memcached && docker-php-ext-enable memcached
RUN docker-php-ext-install gettext
RUN docker-php-ext-install gd
RUN docker-php-ext-install pdo_mysql
RUN docker-php-ext-install soap
RUN docker-php-ext-install sockets
RUN docker-php-ext-install bz2
RUN docker-php-ext-install calendar
RUN docker-php-ext-install mysqli
RUN docker-php-ext-install xsl
RUN docker-php-ext-install zip
[...]

PR #2535, opened 2019-12-30
Use non-root user for docker image

schrapel:
When running Rector inside a Docker container my project throws an error that we are running the code as a root user. It's specific issue with my project but I think it's a sensible change not to run rector under a root process inside the container regardless

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,14 @@ RUN  composer install --no-dev --optimize-autoloader --prefer-dist
 FROM php:7.4-cli
 WORKDIR /rector
 
-COPY . .
-COPY --from=composer /app .
+RUN groupadd -g 1000 rector
+RUN useradd -u 1000 -ms /bin/bash -g rector rector
+
+COPY . /rector
+COPY --from=composer /app /rector
+
+COPY --chown=rector:rector . /rector
+
+USER rector
 
 ENTRYPOINT [ "bin/rector" ]


JanMikes:
Hi, i am looking more into it, i always worked with root user in docker so i need to educate myself a little bit, will let you know asap (no later than tomorrow).


> my project throws an error 
Anyways how is this possible? Rector should not be executing your code, could you please share more details regards the issue, before we merge this PR? There might be possibly other solution :-)

schrapel:
I will share more details today or tomorrow. Little busy on other things today

schrapel:
@JanMikes I looked into it. Some of the spl autoloaded class files have side effects like `require_once(__DIR__.'/../file.php');` and that in turns has a function that runs to check if the script is being run by a root user or not. So this issue is more to do with this codebase I have inherited rather than Rector itself but I do still think that you shouldn't run as a root user in Docker personally. 

https://americanexpress.io/do-not-run-dockerized-applications-as-root/

Totally understand if you do not want to merge this though to keep the Dockerfile simple.

JanMikes:
I am already reading and studying 😄 as long as it is improvement (etc. security), there should be no reason to not merge this.

JanMikes:
@schrapel after reading a lot i have decided that this is not way we want to go with rector official docker image.

Situation when you want to run container as root is:

> This is really useful for development purposes, but can expose you to high risk once you put your containers into a production environment.

Rector should definitely not be used in production environment, but in dev as tool to modify your code and then review it.

> To modify the container system, allowing the user to do things like edit the host filesystem, install system packages at runtime, etc.

- This image is exactly for the purpose, to edit host filesystem (your application). 
- There are some images depending on official image, installing php extensions with apt-get which would now fail on `permission denied`, etc #1547 this is the main reason we can not merge this as this would cause big BC break with probably no way to extend the image with custom php extensions.

Now with new things i learned i suggest you to solve your issue by extending official docker image exactly the same way as in #1547 and use it instead, it should be really simple.

PR #2546, opened 2020-01-03
[Php70] add a rector to pass only variables as arguments by reference

Lctrs:
Still WIP.

Creating it to gather first reviews.

### DIFF
--- a/packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
+++ b/packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
@@ -1,0 +1,177 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Php70\Rector\FuncCall;
+
+use function lcfirst;
+use PhpParser\Node;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\ArrayDimFetch;
+use PhpParser\Node\Expr\Assign;
+use PhpParser\Node\Expr\AssignOp;
+use PhpParser\Node\Expr\AssignRef;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\New_;
+use PhpParser\Node\Expr\PropertyFetch;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Expr\StaticPropertyFetch;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Name;
+use PhpParser\Node\Stmt\Return_;
+use PHPStan\Analyser\MutatingScope;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\ParameterReflection;
+use Rector\CodingStyle\Naming\ClassNaming;
+use Rector\NodeTypeResolver\Node\AttributeKey;
+use Rector\Php70\Reflection\CallReflection;
+use Rector\Php70\ValueObject\VariableAssignPair;
+use Rector\Rector\AbstractRector;
+use Rector\RectorDefinition\CodeSample;
+use Rector\RectorDefinition\RectorDefinition;
+
+/**
+ * @see https://www.php.net/manual/en/migration70.incompatible.php
+ *
+ * @see \Rector\Php70\Tests\Rector\FuncCall\NonVariableToVariableOnFunctionCallRector\NonVariableToVariableOnFunctionCallRectorTest
+ */
+final class NonVariableToVariableOnFunctionCallRector extends AbstractRector
+{
+    /**
+     * @var string
+     */
+    private const DEFAULT_VARIABLE_NAME = 'tmp';
+
+    /**
+     * @var CallReflection
+     */
+    private $callReflection;
+
+    /**
+     * @var ClassNaming
+     */
+    private $classNaming;
+
+    public function __construct(CallReflection $callReflection, ClassNaming $classNaming)
+    {
+        $this->callReflection = $callReflection;
+        $this->classNaming = $classNaming;
+    }
+
+    public function getDefinition(): RectorDefinition
+    {
+        return new RectorDefinition(
+            'Transform non variable like arguments to variable where a function or method expects an argument passed by reference',
+            [new CodeSample('reset(a());', '$a = a(); reset($a);')]
+        );
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getNodeTypes(): array
+    {
+        return [FuncCall::class, MethodCall::class, StaticCall::class];
+    }
+
+    /**
+     * @param FuncCall|MethodCall|StaticCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        $scope = $node->getAttribute(AttributeKey::SCOPE);
+        if (! $scope instanceof MutatingScope) {
+            return null;
+        }
+
+        $arguments = $this->getNonVariableArguments($node, $scope);
+
+        if ($arguments === []) {
+            return null;
+        }
+
+        foreach ($arguments as $key => $argument) {
+            $replacements = $this->getReplacementsFor($argument, $scope);
+
+            $current = $node->getAttribute(AttributeKey::CURRENT_STATEMENT);
+            $this->addNodeBeforeNode($replacements->assign(), $current instanceof Return_ ? $current : $node);
+            $node->args[$key]->value = $replacements->variable();
+
+            $scope = $scope->assignExpression($replacements->variable(), $scope->getType($replacements->variable()));
+        }
+
+        $node->setAttribute(AttributeKey::SCOPE, $scope);
+
+        return $node;
+    }
+
+    /**
+     * @param FuncCall|MethodCall|StaticCall $node
+     *
+     * @return array<int, Expr>
+     */
+    private function getNonVariableArguments(Node $node, Scope $scope): array
+    {
+        $arguments = [];
+
+        /** @var ParameterReflection $parameter */
+        foreach ($this->callReflection->getParameterReflections($node, $scope) as $key => $parameter) {
+            // omitted optional parameter
+            if (! isset($node->args[$key])) {
+                continue;
+            }
+
+            if ($parameter->passedByReference()->no()) {
+                continue;
+            }
+
+            $argument = $node->args[$key]->value;
+
+            if ($this->isVariableLikeNode($argument)) {
+                continue;
+            }
+
+            $arguments[$key] = $argument;
+        }
+
+        return $arguments;
+    }
+
+    private function getReplacementsFor(Expr $expr, Scope $scope): VariableAssignPair
+    {
+        if (
+            (
+                $expr instanceof Assign
+                || $expr instanceof AssignRef
+                || $expr instanceof AssignOp
+            )
+            && $this->isVariableLikeNode($expr->var)
+        ) {
+            return new VariableAssignPair($expr->var, $expr);
+        }
+
+        $variable = new Variable($this->getVariableNameFor($expr, $scope));
+
+        return new VariableAssignPair($variable, new Assign($variable, $expr));
+    }
+
+    private function isVariableLikeNode(Node $node): bool
+    {
+        return $node instanceof Variable
+            || $node instanceof ArrayDimFetch
+            || $node instanceof PropertyFetch
+            || $node instanceof StaticPropertyFetch;
+    }
+
+    private function getVariableNameFor(Expr $expr, Scope $scope): string
+    {
+        if ($expr instanceof New_ && $expr->class instanceof Name) {
+            $name = $this->classNaming->getShortName($expr->class);
+        } else {
+            $name = $this->getName($expr);
+        }
+
+        return lcfirst($this->createCountedValueName($name ?? self::DEFAULT_VARIABLE_NAME, $scope));
+    }
+}
--- a/packages/Php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/assignment.php.inc
+++ b/packages/Php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/assignment.php.inc
@@ -1,0 +1,31 @@
+<?php
+
+namespace Rector\Php70\Tests\Rector\FuncCall\NonVariableToVariableOnFunctionCallRector\Fixture;
+
+function assignment()
+{
+    reset($var = []);
+    reset($var = [0][0]);
+    reset($var = (new \stdClass())->bar);
+    reset($var = \stdClass::$bar);
+}
+
+?>
+-----
+<?php
+
+namespace Rector\Php70\Tests\Rector\FuncCall\NonVariableToVariableOnFunctionCallRector\Fixture;
+
+function assignment()
+{
+    $var = [];
+    reset($var);
+    $var = [0][0];
+    reset($var);
+    $var = (new \stdClass())->bar;
+    reset($var);
+    $var = \stdClass::$bar;
+    reset($var);
+}
+
+?>
--- a/packages/Php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/func_calls.php.inc
+++ b/packages/Php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/func_calls.php.inc
@@ -1,0 +1,58 @@
+<?php
+
+namespace Rector\Php70\Tests\Rector\FuncCall\NonVariableToVariableOnFunctionCallRector\Fixture;
+
+function byRef($bar, &$baz) {}
+
+function allByRef(&$bar, &$baz) {}
+
+function funcCalls()
+{
+    reset(bar());
+    byRef(bar(), baz());
+    allByRef(bar(), baz());
+    allByRef(1, 2);
+
+    $anonymousFunction = function (&$bar) {};
+    $staticAnonymousFunction = static function (&$bar) {};
+    $anonymousFunction(bar());
+    $staticAnonymousFunction(bar());
+
+    return byRef(1, bar());
+}
+
+?>
+-----
+<?php
+
+namespace Rector\Php70\Tests\Rector\FuncCall\NonVariableToVariableOnFunctionCallRector\Fixture;
+
+function byRef($bar, &$baz) {}
+
+function allByRef(&$bar, &$baz) {}
+
+function funcCalls()
+{
+    $bar = bar();
+    reset($bar);
+    $baz = baz();
+    byRef(bar(), $baz);
+    $bar = bar();
+    $baz = baz();
+    allByRef($bar, $baz);
+    $tmp = 1;
+    $tmp2 = 2;
+    allByRef($tmp, $tmp2);
+
+    $anonymousFunction = function (&$bar) {};
+    $staticAnonymousFunction = static function (&$bar) {};
+    $bar = bar();
+    $anonymousFunction($bar);
+    $bar = bar();
+    $staticAnonymousFunction($bar);
+    $bar = bar();
+
+    return byRef(1, $bar);
+}
+
+?>
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -256,3 +256,6 @@ parameters:
         - '#Method (.*?) returns bool type, so the name should start with is/has/was#'
 
         - '#Property Rector\\BetterPhpDocParser\\PhpDocNode\\Gedmo\\BlameableTagValueNode\:\:\$value has no typehint specified#'
+
+        # known value
+        - "#^Parameter \\#1 \\$variable of class Rector\\\\Php70\\\\ValueObject\\\\VariableAssignPair constructor expects PhpParser\\\\Node\\\\Expr\\\\ArrayDimFetch\\|PhpParser\\\\Node\\\\Expr\\\\PropertyFetch\\|PhpParser\\\\Node\\\\Expr\\\\StaticPropertyFetch\\|PhpParser\\\\Node\\\\Expr\\\\Variable, PhpParser\\\\Node\\\\Expr given\\.$#"
--- a/config/set/php/php70.yaml
+++ b/config/set/php/php70.yaml
@@ -23,3 +23,4 @@ services:
 
     Rector\Php70\Rector\Break_\BreakNotInLoopOrSwitchToReturnRector: null
     Rector\Php70\Rector\FuncCall\RenameMktimeWithoutArgsToTimeRector: null
+    Rector\Php70\Rector\FuncCall\NonVariableToVariableOnFunctionCallRector: null
--- a/docs/AllRectorsOverview.md
+++ b/docs/AllRectorsOverview.md
@@ -1,4 +1,4 @@
-# All 425 Rectors Overview
+# All 426 Rectors Overview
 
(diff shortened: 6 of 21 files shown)


TomasVotruba:
Hi, is this ready for review or do you plan more work first?

Lctrs:
I'm planning to handle a few more cases and simplify code a little bit, but main structure/flow is here.

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
```diff
@@ -0,0 +1,204 @@
...
+            if (
+                $argument instanceof Node\Expr\New_
+            ) {
+                if ($argument->class instanceof Node\Name) {
+                    $variable = new Variable(lcfirst($argument->class->getLast()));
+                } elseif ($argument->class instanceof Node\Stmt\Class_ && $argument->class->name !== null) {
+                    $variable = new Variable(lcfirst($argument->class->name->toString()));
+                } else {
+                    continue;
+                }
```
TomasVotruba: Here you can use:

```php
$this->getName($argument->class)
```

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
```diff
@@ -0,0 +1,204 @@
...
+
+        return $node;
+    }
+
+    /**
+     * @param FuncCall|Node\Expr\MethodCall|Node\Expr\StaticCall $node
+     *
+     * @return array<int, Node\Expr>
+     */
+    private function getNonVariableArguments($node): array
```
TomasVotruba: At least some type is needed

```suggestion
    private function getNonVariableArguments(Node $node): array
```

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
```diff
@@ -0,0 +1,204 @@
...
+
+            if (
+                $argument instanceof Variable
+                || $argument instanceof ArrayDimFetch
+                || $argument instanceof PropertyFetch
+                || $argument instanceof StaticPropertyFetch
+            ) {
+                continue;
+            }
+
```
TomasVotruba: What is this for?
Extracted method with clear name would tell

packages/Php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,128 @@
...
+    $dummy = new Dummy();
+    $dummy->baz(baz());
+    $dummy->child()->bar(bar());
+
+    $anonymousClass = new class {
+        public function bar(&$baz) {}
+    };
+    $anonymousClass->bar(baz());
+}
+
```
TomasVotruba: It's best practise here to use more smaller test fixtures, instead of only 1 with all the use cases.

If one case breaks here in the future, it's very hard to find the bug. Having 10 files would make that easier

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
```diff
@@ -0,0 +1,204 @@
...
+                if ($argument->class instanceof Node\Name) {
+                    $variable = new Variable(lcfirst($argument->class->getLast()));
+                } elseif ($argument->class instanceof Node\Stmt\Class_ && $argument->class->name !== null) {
+                    $variable = new Variable(lcfirst($argument->class->name->toString()));
+                } else {
+                    continue;
+                }
+
+                $assign = new Node\Expr\Assign($variable, $argument);
+            }
```
TomasVotruba: I'm not sure what this section does.

It could be extrated into `processArgument()` or something similar, to improve readability.

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    /**
+     * @return array<int, ParameterReflection>
+     */
+    private function getParametersForName(Node\Name $name, Scope $scope): array
+    {
+        try {
+            return ParametersAcceptorSelector::selectSingle(
+                $this->broker->getFunction($name, $scope)->getVariants()
```
TomasVotruba: Broker is very low-level class here. It should not be part of Rector rules.

What is your goal here of this method?

Lctrs: Probably a lack of knowledge of phpstan internals on my side here.

Given a FuncCall identified by a Name, I want to return all Parameters.

Lctrs:
Yes I have used the create command.

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
TomasVotruba:
```php
$node->args
```

Lctrs: No, not the args. The parameters defined in the signature of the function.

TomasVotruba: In that case you can do simple `FunctionReflection`

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
Lctrs: In case of a namespaced class, I'll get the FQCN of the class (for example `App\Dummy`), which is not suitable for a variable name. What should I do in this case ?

TomasVotruba: You can use `Rector\CodingStyle\Naming\ClassNaming` service and `getShortName()` method.

It's conventional to process any FQN class/interface/type etc. names

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
Lctrs: I must admit I'm a little lost.

TomasVotruba: Keep you way than. This is just syntax suggar now

Lctrs:
@TomasVotruba Ready for review. Build failure is unrelated I guess.

Lctrs:
@TomasVotruba Finally still WIP. Tried it on a real project. Got errors.

TomasVotruba:
No troubles. Just ping me when you need me

TomasVotruba:
CI should pass completely now. Just fixed the master

Lctrs:
@TomasVotruba Any idea what `leaveNode() may only return an array if the parent structure is an array` means ?

TomasVotruba:
That happens when you get node, but return an array of nodes.

Lctrs:
How/Where can this happen ? I don't return an array in my rector.

TomasVotruba:
There can be like dozens of cases where this can happen.

I'd have to see the test case.

Lctrs:
@TomasVotruba It was caused by a return statement. When I add something like `return reset(bar());` for example, the node was inserted between the `return` and the `reset`, hence the error. I handled this case on line `100-101` in my rector, but maybe this can be generalized ?

Otherwise, it's finally ready for review. It worked well in one of my project. I tried to be as extensive as possible on test.

Lctrs:
Only the PHAR compilation fails ATM. Will take a look at it.

Lctrs:
@TomasVotruba All good :tada:

packages/Php70/src/Dto/VariableAssignPair.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Php70\Dto;
```
TomasVotruba: We have `ValueObject` namespace convention for any non-services object, to keep structure simple

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
```diff
@@ -0,0 +1,259 @@
...
+
+        return $arguments;
+    }
+
+    /**
+     * @param FuncCall|MethodCall|StaticCall $node
+     *
+     * @return array<int, ParameterReflection>
+     */
+    private function getParameters(Node $node, Scope $scope): array
```
TomasVotruba: As this works with reflection, it should be extracted in some service in `Reflection` namespace. Rector rules should contain as few business logic as possible, imagine it like a Controller.

TomasVotruba:
```suggestion
    private function getParameterReflections(Node $node, Scope $scope): array
```

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
```diff
@@ -0,0 +1,259 @@
...
+                $scope,
+                $args,
+                $this->broker->getFunction($name, $scope)->getVariants()
+            )->getParameters();
+        } catch (FunctionNotFoundException $functionNotFoundException) {
+            return [];
+        }
+    }
+
+    private function processArgument(Expr $expr, Scope $scope): VariableAssignPair
```
TomasVotruba: This looks like `VariableAssignPairFactory`.

*process/refactor* method should change nodes in some way

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
```diff
@@ -0,0 +1,259 @@
...
+        $i = 2;
+        $countedValueNamePart = $countedValueName;
+        while ($scope->hasVariableType($countedValueName)->yes()) {
+            $countedValueName = $countedValueNamePart . $i;
+            ++$i;
+        }
+
+        return $countedValueName;
+    }
+}
```
TomasVotruba: This would be great if abstracted in `AbstractRector`, as this logic is used in multiple places now

packages/Php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/array_dim_fetch.php.inc
```diff
@@ -0,0 +1,21 @@
...
+<?php
+
+namespace Rector\Php70\Tests\Rector\FuncCall\NonVariableToVariableOnFunctionCallRector\Fixture;
+
+function arrayDimFetch()
+{
+    reset([1][0]);
+}
+
+?>
```
TomasVotruba: If nothing changes, you can remove this whole part. 

Also, the fixture file should start with `skip_*`/`keep_*` to make it clear for future debugging.

packages/Php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/assignment.php.inc
```diff
@@ -0,0 +1,31 @@
...
+{
+    $var = [];
+    reset($var);
+    $var = [0][0];
+    reset($var);
+    $var = (new \stdClass())->bar;
+    reset($var);
+    $var = \stdClass::$bar;
+    reset($var);
+}
```
TomasVotruba: Nice :+1:

packages/Php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/optional_parameter.php.inc
```diff
@@ -0,0 +1,25 @@
...
+namespace Rector\Php70\Tests\Rector\FuncCall\NonVariableToVariableOnFunctionCallRector\Fixture;
+
+function withOptionalParameter(&$a = null) {}
+
+function optionalParameter()
+{
+    baz();
+}
+
+?>
```
TomasVotruba: This can be removed, as nothing changes

packages/Php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/static_property_fetch.php.inc
```diff
@@ -0,0 +1,21 @@
...
+<?php
+
+namespace Rector\Php70\Tests\Rector\FuncCall\NonVariableToVariableOnFunctionCallRector\Fixture;
+
+function staticPropertyFetch()
+{
+    reset(\stdClass::$dummy);
+}
+
+?>
```
TomasVotruba: This can be removed, as nothing changes

packages/Php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/property_fetch.php.inc
```diff
@@ -0,0 +1,21 @@
...
+<?php
+
+namespace Rector\Php70\Tests\Rector\FuncCall\NonVariableToVariableOnFunctionCallRector\Fixture;
+
+function propertyFetch()
+{
+    reset((new \stdClass())->dummy);
+}
+
+?>
```
TomasVotruba: This can be removed, as nothing changes

phpstan.neon
```diff
@@ -256,3 +256,9 @@ parameters:
         - '#Method (.*?) returns bool type, so the name should start with is/has/was#'
 
         - '#Property Rector\\BetterPhpDocParser\\PhpDocNode\\Gedmo\\BlameableTagValueNode\:\:\$value has no typehint specified#'
+
+        # known value
+        -
+            message: "#^Parameter \\#1 \\$variable of class Rector\\\\Php70\\\\Dto\\\\VariableAssignPair constructor expects PhpParser\\\\Node\\\\Expr\\\\ArrayDimFetch\\|PhpParser\\\\Node\\\\Expr\\\\PropertyFetch\\|PhpParser\\\\Node\\\\Expr\\\\StaticPropertyFetch\\|PhpParser\\\\Node\\\\Expr\\\\Variable, PhpParser\\\\Node\\\\Expr given\\.$#"
+            count: 1
+            path: packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
```
TomasVotruba: No need for such detailed ignore, as it's hard to maintain in case of renaming file.
Message is much better, as in the line above

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
```diff
@@ -0,0 +1,259 @@
...
+        return [FuncCall::class, MethodCall::class, StaticCall::class];
+    }
+
+    /**
+     * @param FuncCall|MethodCall|StaticCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        $scope = $node->getAttribute(AttributeKey::SCOPE);
+        if (! $scope instanceof MutatingScope) {
```
TomasVotruba: `Scope` is good enough as interface

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
```diff
@@ -0,0 +1,259 @@
...
+        }
+
+        foreach ($arguments as $key => $argument) {
+            $replacements = $this->processArgument($argument, $scope);
+
+            $current = $node->getAttribute(AttributeKey::CURRENT_STATEMENT);
+            $this->addNodeBeforeNode($replacements->assign(), $current instanceof Return_ ? $current : $node);
+            $node->args[$key]->value = $replacements->variable();
+
+            $scope = $scope->assignExpression($replacements->variable(), $scope->getType($replacements->variable()));
```
TomasVotruba: I never used this before. What is it for? When should we use it?

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
TomasVotruba: The service could be named `Rector\Php70\Reflection\CallReflection`

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
Lctrs: `Scope` doesn't contains the method `assignExpression` sadly. See (upcoming) explanation below.

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
Lctrs: It returns a new Scope from the previous Scope + the assigned expression (for example, if you pass a variable to it, the new scope will be aware of this variable).

I did this so that Scope is aware of newly created variables, so that `createCountedValueName()` returns a truly unique variable name.

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
TomasVotruba: Ah, I see. I'm affraid that this will go in mess like "which one to use when", since `instanceof Scope` is used everywhere now.

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
TomasVotruba: I see, thanks!

Maybe it would be more consistent to add type here:

```php
/** @var MutatingScope $scope */
```

packages/Php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/func_calls.php.inc
```diff
@@ -0,0 +1,58 @@
...
+{
+    $bar = bar();
+    reset($bar);
+    $baz = baz();
+    byRef(bar(), $baz);
+    $bar = bar();
+    $baz = baz();
+    allByRef($bar, $baz);
+    $tmp = 1;
+    $tmp2 = 2;
```
TomasVotruba: Nice job :clap

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
Lctrs: Will see if I can do something with the `ScopeFactory` instead.

Lctrs: But a simple rule could be if you need to mutate scope, go with `MutatingScope`, otherwise use `Scope`.

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
Lctrs: `assignExpression()` always returns a MutatingScope. No need to add a phpDoc IMHO.

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
TomasVotruba: It might, but classes with more public method than interfaces they implement always create crappy code. It's PHPStan design flaw we cannot cover here I'm affraid.

Let's keep it your way for now, not that important.

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
TomasVotruba: I meant it to have `instanceof Scope` above and here make the annotation for variable on the right:

```php
/** @var MutatingScope $scope */
$whatever = **$scope**->assignExpression(...);
```

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
Lctrs: Furthermore, `MutatingScope` seems to be the only available implementation of `Scope`.

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
Lctrs: @TomasVotruba How would you handle things like suffix ? (`ForRepeatedCountToOwnVariableRector` uses a suffix)

Lctrs:
@TomasVotruba New version with most of your comments adressed. Only things missing is extracting `createCountedValueName()` to `AbstractRector`. I don't know yet how to handle properly things such as suffix, etc. If you have any idea...

Lctrs:
@TomasVotruba Should be good.

Lctrs:
@TomasVotruba FYI, I have a working version locally that uses the core reflection API (well, and also `roave/better-reflection` for one case). So no more `Broker` use. Just need to split it into more functions, Should be up tomorrow.

TomasVotruba:
All rigth. Maybe look at PHPStan static reflection there is since 0.12.4

Ping me when ready for review

Lctrs:
Should I understand that you prefer to use the Reflection API from PHPStan ?

TomasVotruba:
Some of these commits: https://github.com/phpstan/phpstan-src/search?q=static+reflection&type=Commits

Lctrs:
Yeah but if we use a reflection provider from PHPStan, we'll get objects implementing the reflection API from PHPStan, not from the core. Or maybe I'm missing something ?

TomasVotruba:
It's just for inspiration, I don't personally use it

Lctrs:
@TomasVotruba WDYT ? Yeah I know that `roave/better-reflection` is required as `dev-master`, but due to version conflict, I had no other choice.

packages/Php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector.php
TomasVotruba: What do you mean? Suffix of what?

I'd just move this to protected method to abstract class, so it can be re-used in multiple Rector rules.

TomasVotruba:
> Yeah I know that roave/better-reflection is required as dev-master, but due to version conflict, I had no other choice.

That would disable `composer require rector/rector`

I think we should first finish the rule, then add more dependencies. This PR is already mixing many various problems. One step at a time.

Lctrs:
Okay, I reverted my last commit for now, I think I covered all your remarks. Maybe I'm missing something ?

TomasVotruba:
I'm looking at it

TomasVotruba:
All looks good!

Thanks for this aweseome feature. I think you're the most skilled *first contritbutor* Rector ever had :bowing_man:

PR #2553, opened 2020-01-03
Improve 3rd party property type resolution

### LINKED ISSUE #2474: [PHP] CountOnNull on properties with PHPDoc shouldn't apply
I've had the case in my code base that the following code change was triggered : 

```diff
/** @var string[] $array */
$array = $someObject->someMethod();

- if (count($array) > 0) {
+ if (is_countable($array) ? count($array) > 0 : 0) {

}
```

I was quite surprised by this proposition, as IMO a `string[]` doesn't need to be surrounded by `is_countable` (if the developer is misusing type hints/doc blocks, PHPStan should warn him).

I digged into it and even noticed there's a test Fixture that seems to affirm this behavior : 

`packages/Php71/tests/Rector/FuncCall/CountOnNullRector/Fixture/property_with_doc.php.inc`

Isn't that weird ?

TomasVotruba:
Closes #2474

### DIFF
--- a/packages/NodeTypeResolver/src/NodeTypeResolver.php
+++ b/packages/NodeTypeResolver/src/NodeTypeResolver.php
@@ -30,11 +30,13 @@ use PhpParser\Node\Stmt\ClassConst;
 use PhpParser\Node\Stmt\ClassLike;
 use PhpParser\Node\Stmt\ClassMethod;
 use PhpParser\Node\Stmt\Interface_;
+use PhpParser\Node\Stmt\Nop;
 use PhpParser\Node\Stmt\PropertyProperty;
 use PhpParser\Node\Stmt\Trait_;
 use PhpParser\NodeTraverser;
 use PHPStan\Analyser\Scope;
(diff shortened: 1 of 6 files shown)


packages/NodeTypeResolver/src/NodeTypeResolver.php
```diff
@@ -795,4 +840,46 @@ private function getClassNodeProperty(ClassLike $classLike, string $name): ?Prop
...
+            return null;
+        }
+
+        $typeNode = $varTagValues[0]->type;
+        if (! $typeNode instanceof TypeNode) {
+            return null;
+        }
+
+        return $this->staticTypeMapper->mapPHPStanPhpDocTypeNodeToPHPStanType($typeNode, new Nop());
+    }
```
TomasVotruba: @gnutix It's not perfect solution. But it's the only one without parsing the vendor file (which is performance and missing-classes trap).

Here we deduce type from the doc of property, thanks to php reflection.

PR #2558, opened 2020-01-03
[MinimalScope] Add ChangeLocalPropertyToVariableRector

### LINKED ISSUE #1900: [DeadCode] Local-only property to variable
```diff
-private $value;

private $usedVariable;

public function run()
{
-    $this->value = 5;
+    $value = 5;
-    $this->value += 31;
+    $value += 31;

    $this->usedVariable = 1515;

-    return $this->value;
+    return $value;
}

public function useMe()
{    
    $this->usedvVariable -= 100;
}
```

TomasVotruba:
Closes #1900

### DIFF
--- a/packages/MinimalScope/tests/Rector/Class_/ChangeLocalPropertyToVariableRector/Fixture/fixture.php.inc
+++ b/packages/MinimalScope/tests/Rector/Class_/ChangeLocalPropertyToVariableRector/Fixture/fixture.php.inc
@@ -1,0 +1,30 @@
+<?php
+
+namespace Rector\MinimalScope\Tests\Rector\Class_\ChangeLocalPropertyToVariableRector\Fixture;
+
+class SomeClass
+{
+    private $count;
+    public function run()
+    {
+        $this->count = 5;
+        return $this->count;
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\MinimalScope\Tests\Rector\Class_\ChangeLocalPropertyToVariableRector\Fixture;
+
+class SomeClass
+{
+    public function run()
+    {
+        $count = 5;
+        return $count;
+    }
+}
+
+?>
--- a/composer.json
+++ b/composer.json
@@ -104,7 +104,8 @@
             "Rector\\DynamicTypeAnalysis\\": "packages/DynamicTypeAnalysis/src",
             "Rector\\PhpDeglobalize\\": "packages/PhpDeglobalize/src",
             "Rector\\Phalcon\\": "packages/Phalcon/src",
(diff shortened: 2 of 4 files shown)


packages/MinimalScope/tests/Rector/Class_/ChangeLocalPropertyToVariableRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Rector\MinimalScope\Tests\Rector\Class_\ChangeLocalPropertyToVariableRector\Fixture;
+
+class SomeClass
+{
+    private $count;
+    public function run()
```
staabm: Wouldnt the rector fail when the method e.g. looks like

```
public function run()
     {
         If ($this->count == 5) {
           $this->count = 10;
           return $this->count;
         }
         $this->count = 5;
         return $this->count;
     }
```

TomasVotruba: Why should it fail?

staabm: I guess the logic will change because of the rector.

In case it gets transformed into
```

public function run()
     {
         If ($count == 5) {
           $count = 10;
           return $count;
         }
         $count = 5;
         return $count;
     }
```

The method will return different values when invoked 2 times in a row

TomasVotruba: > The method will return different values when invoked 2 times in a row

I see. That's what I was missing.

How would you approach it?

staabm: I guess it could work when you made sure that there is only a single property-write

TomasVotruba: That sound clear, I'll add fixture for that. Thanks for spotting this :+1:

TomasVotruba: https://github.com/rectorphp/rector/pull/2563#merged-event

TomasVotruba: It might needs full new service to analyze scope in the future. Let's see how it handles reality now

PR #2569, opened 2020-01-05
[Polyfill] Add UnwrapFutureCompatibleIfRector

### LINKED ISSUE #2540: Detect feature-detecting code as dead code
Its a good practice to use feature detects instead of hardcoding certain version-checks or try to sniff the environment.

This can lead to dead code, e.g.

```php
        // prevent session locking trough other addons
        if (function_exists('session_abort')) {
            session_abort();
        } else {
            session_write_close();
        }
```

This code is written like that because depending on the php-version the function in question might exists or not. In case it is missing we have implemented a less optimal alternative. 

When upgrading the min-php version of the project this code might easily be dead, because the function will always exist.

Maybe there is some way to detect this kind of codes and report it as dead

TomasVotruba:
Closes #2540

### DIFF
--- a/packages/Polyfill/src/FeatureSupport/FunctionSupportResolver.php
+++ b/packages/Polyfill/src/FeatureSupport/FunctionSupportResolver.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Polyfill\FeatureSupport;
+
+use Rector\Php\PhpVersionProvider;
+
+final class FunctionSupportResolver
+{
+    /**
+     * @var string[][]
+     */
+    private const FUNCTIONS_BY_VERSION = [
+        '5.6' => ['session_abort'],
+    ];
+
+    /**
+     * @var PhpVersionProvider
+     */
+    private $phpVersionProvider;
+
+    public function __construct(PhpVersionProvider $phpVersionProvider)
+    {
+        $this->phpVersionProvider = $phpVersionProvider;
+    }
+
+    public function isFunctionSupported(string $desiredFunction): bool
+    {
+        foreach (self::FUNCTIONS_BY_VERSION as $version => $functions) {
+            foreach ($functions as $function) {
+                if ($desiredFunction !== $function) {
+                    continue;
+                }
+
+                if (! $this->phpVersionProvider->isAtLeast($version)) {
+                    continue;
+                }
+
+                return true;
+            }
+        }
+
+        return false;
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -105,7 +105,8 @@
             "Rector\\PhpDeglobalize\\": "packages/PhpDeglobalize/src",
             "Rector\\Phalcon\\": "packages/Phalcon/src",
             "Rector\\DoctrineGedmoToKnplabs\\": "packages/DoctrineGedmoToKnplabs/src",
-            "Rector\\MinimalScope\\": "packages/MinimalScope/src"
+            "Rector\\MinimalScope\\": "packages/MinimalScope/src",
+            "Rector\\Polyfill\\": "packages/Polyfill/src"
         }
     },
     "autoload-dev": {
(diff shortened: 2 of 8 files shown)


packages/Polyfill/src/FeatureSupport/FunctionSupportResolver.php
```diff
@@ -0,0 +1,46 @@
...
+use Rector\Php\PhpVersionProvider;
+
+final class FunctionSupportResolver
+{
+    /**
+     * @var string[][]
+     */
+    private const FUNCTIONS_BY_VERSION = [
+        '5.6' => ['session_abort'],
+    ];
```
TomasVotruba: @staabm Here is just the code example you showed me. I need your help to complete list as far as possible.

staabm: I need to think about how to compile such method most easily

packages/Polyfill/src/FeatureSupport/FunctionSupportResolver.php
```diff
@@ -0,0 +1,46 @@
...
+
+    public function isFunctionSupported(string $desiredFunction): bool
+    {
+        foreach (self::FUNCTIONS_BY_VERSION as $version => $functions) {
+            foreach ($functions as $function) {
+                if ($desiredFunction !== $function) {
+                    continue;
+                }
+
+                if (! $this->phpVersionProvider->isAtLeast($version)) {
```
staabm: I guess it would also be helpfull to remove code when explicit php version checks are used (might be a separTe rector then?)

In codes like https://github.com/redaxo/redaxo/blob/591146a1dc60e8aacefd58dc9b7e9c307c0983b9/redaxo/src/addons/tests/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCodeGenerator.php#L85

TomasVotruba: Sure! I though of that too.

Could you help by making a new issue with minimal code (3-lines) in diff a format?

packages/Polyfill/src/FeatureSupport/FunctionSupportResolver.php
staabm: After thinking about it I feel we could compile this list from exisiting polyfil libs, e.g.
https://github.com/symfony/polyfill-php56
https://github.com/symfony/polyfill-php70
https://github.com/symfony/polyfill-php71
https://github.com/symfony/polyfill-php72
https://github.com/symfony/polyfill-php73
https://github.com/symfony/polyfill-php74

packages/Polyfill/src/FeatureSupport/FunctionSupportResolver.php
staabm: See https://github.com/rectorphp/rector/issues/2573 https://github.com/rectorphp/rector/issues/2574

packages/Polyfill/src/FeatureSupport/FunctionSupportResolver.php
staabm: Another thing I checked via github code-search
https://github.com/search?l=PHP&q=function_exists&type=Code

There we can see most examples are like
```
if( ! function_exists( 'supernova_comment' ) ){
function supernova_comment($comment, $args, $depth) {
 // impl
}
```

In these cases we could eleminate the code when we are sure that the function beeing checked is one from the php standard library

TomasVotruba: > After thinking about it I feel we could compile this list from exisiting polyfil libs, e.g.

Great!

> Another thing I checked via github code-search

Try to keep issues separated, so we don't get stuck with complexity. This is another use case.

staabm: > This is another use case.

Absolutely. Just wanted to let you know which things I tried

TomasVotruba: I see

TomasVotruba: There are many new PRs merged every day so there are new conflicts to rebase on.
I'll merge this now to prevent that.

You can send the function list as a new PR targeting this class :+1:

TomasVotruba: @staabm Any update on this? The rule is pretty lonely with just 1 function now :D

staabm: Was on vacation till yesterday.. will need a few more days to catch up

staabm: here we go https://github.com/rectorphp/rector/pull/2673

PR #2572, opened 2020-01-05
Don't return void when function contains empty return statement

snapshotpl:


packages/CodeQuality/tests/Rector/Array_/CallableThisArrayToAnonymousFunctionRector/Fixture/no_return_for_void.php.inc
```diff
@@ -13,6 +13,7 @@ class HasVoidMethod
 
     private function thisReturnsVoid(): void
     {
+        return;
     }
 }
```
TomasVotruba: New method with only `return;` is better. Now test for empty method was removed.

Please add a new function with only `return;` so we have both covered.

Thanks!

PR #2575, opened 2020-01-06
PHPStan 0.12.4 compatibility updates

ondrejmirtes:


ondrejmirtes:
There are still some failures which I don't what caused them.

TomasVotruba:
Wow, thanks a lot Ondřej!
You just save me lot of work, since I'd have to go through all the new commits on PHPStan.

I'll look at it after lunch.

TomasVotruba:
The ECS is fail is related to coding standard. It can be fixed with:

```bash
composer fix-cs
```

Test are passing :heavy_check_mark: 
I'll merge it as it is and resolve it locally. Thank you!

TomasVotruba:
I made it work except compiling of phar.

https://travis-ci.org/rectorphp/rector/jobs/633198099#L569

image


Any idea what is causing it?

ondrejmirtes:
This class is prefixed in PHPStan's PHAR. Not sure where you're referring it?

TomasVotruba:
It is not. PHAR cannot be used inside PHAR, so phpstan-src is used instead.

Seems like prefixing of PHPStan config.neon didn't work for this services. Which is weird, because it worked for 0.12.3

TomasVotruba:
Any idea what has changed since 0.12.3 that might have influence this?

ondrejmirtes:
IMHO you should take `phpstan.phar`, extract it using `box extract` and use the contents to build your PHAR. Otherwise you'd have to reimplement the same prefixing logic to have truly scoped PHAR.

This is exactly what I'm talking about - at some point you probably took my file https://github.com/phpstan/phpstan-src/blob/master/compiler/build/scoper.inc.php and reused it. But I kept changing it, see the history: https://github.com/phpstan/phpstan-src/commits/master/compiler/build/scoper.inc.php

TomasVotruba:
That's it, thanks!

> IMHO you should take phpstan.phar, extract it using box extract and use the contents to build your PHAR.

That might work. But it would add extra step of rewriting whole PHPStan code in Rector with current phpstan prefix. I can't imagine how I'd do that in saint way. 
Also it's pretty WTF behavior for further debugging, pack, unpack, reuse some code, pack again...


It would be much easier to re-use PHPStan's `scope.inc.php` content.
I'll copy it for now and if that happens again, I might autoamte it.

ondrejmirtes:
> That might work. But it would add extra step of rewriting whole PHPStan code in Rector with current phpstan prefix.

`PHPStan\\` namespace itself isn't prefixed, and you shouldn't touch PHPStan's dependencies (you shoud make them your dependencies), so I think it should be fine.

TomasVotruba:
Ah, I forgot that :man_facepalming: 

I can't imagine where would I extract the PHPStan code and how I'd merge it with Rector code. Somewhere in vendor?

ondrejmirtes:
I don’t know, you’d have to experiment on that 😊 To have everything under vendor/phpstan/phpstan (including PHPStan’s dependencies) might work.

TomasVotruba:
So I'd have to include `vendor/phpstan/phsptan/vendor/autoload.php`? :D 
That might go to another universe very quickly :smile: 

I won't have time to play with this, so I'll wait for future bugs to give me the motivation

PR #2576, opened 2020-01-06
travis: try phpdbg

TomasVotruba:
Inspired by https://kizu514.com/blog/phpdbg-is-much-faster-than-xdebug-for-code-coverage/

## Before - xdebug - 37 minutes

image

## After - phpdbg - 3 minutes

image

PR #2581, opened 2020-01-06
Try pcov for code coverage

staabm:
Similar to https://github.com/rectorphp/rector/pull/2576

Related article from the same source.. lets try

https://kizu514.com/blog/pcov-is-better-than-phpdbg-and-xdebug-for-code-coverage/
https://github.com/krakjoe/pcov

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -136,7 +136,11 @@ jobs:
             php: 7.3
             name: Test Coverage
             script:
-                - pecl install pcov; composer require pcov/clobber; vendor/bin/pcov clobber; vendor/bin/phpunit --configuration phpunit.xml --coverage-clover coverage.xml 
+                # https://kizu514.com/blog/pcov-is-better-than-phpdbg-and-xdebug-for-code-coverage/
+                - pecl install pcov
+                - composer require pcov/clobber
+                - vendor/bin/pcov clobber
+                - vendor/bin/phpunit --configuration phpunit.xml --coverage-clover coverage.xml
                 - wget https://github.com/php-coveralls/php-coveralls/releases/download/v2.1.0/php-coveralls.phar
                 - php php-coveralls.phar --verbose
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -138,8 +138,6 @@ jobs:
             script:
                 # https://kizu514.com/blog/pcov-is-better-than-phpdbg-and-xdebug-for-code-coverage/
                 - pecl install pcov
-                - composer require pcov/clobber
-                - vendor/bin/pcov clobber
                 - vendor/bin/phpunit --configuration phpunit.xml --coverage-clover coverage.xml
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -136,7 +136,7 @@ jobs:
             php: 7.3
             name: Test Coverage
             script:
-                - phpdbg -qrr -d memory_limit=-1 vendor/bin/phpunit --coverage-clover coverage.xml
+                - pecl install pcov; composer require pcov/clobber; vendor/bin/pcov clobber; vendor/bin/phpunit --configuration phpunit.xml --coverage-clover coverage.xml
```
TomasVotruba: should be in standalone lines instead of `x; y;`

TomasVotruba: Also reference to the post would be great, as I've never heard of it and it might be difficult to debug it in a year

staabm: Will wait for the actual perf number and adjust in case its faster

staabm:
What do you think about moving the simple checks to GithubAction so overall build-time can be reduced?

.travis.yml
staabm: Coverage is down to ~1min30

TomasVotruba: Nice, go for it then :+1:

TomasVotruba: Better remove fatal errors though: https://travis-ci.org/rectorphp/rector/jobs/633259038#L547

TomasVotruba:
> What do you think about moving the simple checks to GithubAction so overall build-time can be reduced? (Less sequential tasks)

PR prototype would help

.travis.yml
```diff
@@ -136,7 +136,9 @@ jobs:
             php: 7.3
             name: Test Coverage
             script:
-                - phpdbg -qrr -d memory_limit=-1 vendor/bin/phpunit --coverage-clover coverage.xml
+                # https://kizu514.com/blog/pcov-is-better-than-phpdbg-and-xdebug-for-code-coverage/
+                - pecl install pcov
+                - vendor/bin/phpunit --configuration phpunit.xml --coverage-clover coverage.xml
```
TomasVotruba: Btw, why `--configuration phpunit.xml`? It's default value

staabm: Copied from the article :)

staabm:
Remaining failures seem unrelated

TomasVotruba:
Indeed. Thanks :+1:

TomasVotruba:
The coverage just dropped by 4 %

https://coveralls.io/github/rectorphp/rector?branch=master

image

That's way too much for 1 min improvements on parallel Travis run.

We'll need to revert

staabm:
Imo the coverage did not actually change but pcov treats some details differenz and therefore leads to a different number

TomasVotruba:
The number changed for much worse. If that would be ~0,5 % than it's ok.
But this might be seriously missleading. I'd rather stay closer to original xdebug coverage, as we'll probably get back to it in 2.9

Thanks for you work though. I wish I noticed the number before

PR #2586, opened 2020-01-06
fix box autoload Neon class in scoper.php.inc by --no-parallel

TomasVotruba:


compiler/build/scoper.inc.php
```diff
@@ -105,11 +105,15 @@ function (string $filePath, string $prefix, string $content): string {
                 if (strpos($class, 'PhpParser\\') === 0) {
                     return $class;
                 }
+                if (strpos($class, 'Rector\\') === 0) {
+                    return $class;
```
TomasVotruba: If this is missing, it breaks https://github.com/rectorphp/rector/blob/a323fff8d4c85e9e7fd76ad28ed8af98ff3912e1/packages/NodeTypeResolver/config/phpstan/type-extensions.neon

PR #2587, opened 2020-01-06
[ForeachItemsAssignToEmptyArrayToAssignRector] apply on code it should not

gnutix:


TomasVotruba:
Could add fix as well?

gnutix:
Again, I might be able to provided some context. What is this rector rule really about ? What are all the edge cases that might be broken, except the two I had in my code base ? What's the vision for it and how to implement it ?

I know nothing of it right now, and as I only have a small amount of time to dedicate to OSS, the better the explanation the greater the chances I can contribute efficiently. :)

TomasVotruba:
Allright :)  I'll try my best. Ask whatever you need to know.

The rule is about removing foreach of array, that does basically assign of array:
https://github.com/rectorphp/rector/blob/master/docs/AllRectorsOverview.md#foreachitemsassigntoemptyarraytoassignrector

```diff
 class SomeClass
 {
     public function run($items)
     {
         $items2 = [];
-        foreach ($items as $item) {
-             $items2[] = $item;
-        }
+        $items2 = $items;
     }
 }
```

Everything else should be skipped - those edge-cases you've detected. I'll put comment there

packages/CodeQuality/tests/Rector/Foreach_/ForeachItemsAssignToEmptyArrayToAssignRector/Fixture/skip_false_positives.php.inc
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Rector\CodeQuality\Tests\Rector\Foreach_\ForeachItemsAssignToEmptyArrayToAssignRector\Fixture;
+
+class SkipFalsePositives
+{
+    public function indexBy(array $array, callable $key): array
+    {
+        $indexed = [];
+        foreach ($array as $item) {
+            $indexed[$key($item)] = $item;
```
TomasVotruba: There have to be assign with brackets: `$values[]`

Currently any key is accepted (`$values[$anything]`), see this line in the rule: https://github.com/rectorphp/rector/blob/master/packages/CodeQuality/src/Rector/Foreach_/ForeachItemsAssignToEmptyArrayToAssignRector.php#L107

It is missing check for `$arrayDimFetch->dim === null`, https://github.com/nikic/PHP-Parser/blob/88f3a669c12e576887dfded483e971d6c333adbe/lib/PhpParser/Node/Expr/ArrayDimFetch.php#L12

If the `$arrayDimFetch->dim` value is not null, stop and `return null;`

This should fix this case

TomasVotruba: After this is handled, we'll look on following cases

gnutix:
All good! :) Thanks for the help.

packages/CodeQuality/tests/Rector/Foreach_/ForeachItemsAssignToEmptyArrayToAssignRector/Fixture/skip_false_positive_1.php.inc
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Rector\CodeQuality\Tests\Rector\Foreach_\ForeachItemsAssignToEmptyArrayToAssignRector\Fixture;
+
+class SkipFalsePositive1
```
TomasVotruba: Explicit name would be better, when someone will try to debug this rule in the future.

E.g. `skip_non_empty_dim_key.php.inc`

gnutix: @TomasVotruba Done!

TomasVotruba:
Great job, thanks :+1:

PR #2589, opened 2020-01-06
Transition some travis checks to GithubActions

staabm:
parallize more jobs to speedup the overall CI build time.

In GithubActions we can get up to 8 parallel workers, while on travis its limited to 4.
current overall build time is ~17-18 min. distribution of the jobs should make it faster.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,17 @@
+name: PHP Checks
+
+on:
+  push:
+
+jobs:
+  ecs:
+      name: ECS
+      runs-on: ubuntu-latest
+      steps:
+          - uses: actions/checkout@v2
+          - name: Setup PHP
+            uses: shivammathur/setup-php@verbose
+            with:
+                php-version: 7.2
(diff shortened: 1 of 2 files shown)


staabm:
@TomasVotruba could you activate GithubActions for this repo?

TomasVotruba:
I added some file to master. Is that the activation?

staabm:
hmm I guess I cannot create new githubaction workflows because I am not a repo owner.

just reverted the changes to travis.yml so we could merge this as is and would iterate on the result in a new PR. I guess I can modify a then existing workflow, but creating a new one, as I did here is not allowed as a external collaborateur.

staabm:
(Alternatively you could take my changes and create a PR yourself)

TomasVotruba:
I've invited you as a collaborator, so you can can create a branch in here instead of work.

Could you re-check?

staabm:
Thx for the invitation. It seems I dont have permissions to create new branches in this repo

TomasVotruba:
Let's merge it and see

PR #2591, opened 2020-01-06
Remove travis-CI jobs which were moved to GithubAction

staabm:
You should configure the new checks to be required in case those were required before (in github branch protection).

In case it is essentiell that the „compile phar“ travis job only runs after all (including those beeing removed with this PR) previous jobs went green/successfull we should think about the job design.. (I guess travis ci does not know about status of single github actions and vice versa)

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,8 +1,10 @@
 name: PHP Checks
 
 on:
-  - push
-  - pull_request
+  pull_request:
+  push:
+    branches:
+      - master
 
 jobs:
   phpstan:
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -2,6 +2,7 @@ name: PHP Checks
 
 on:
   - push
+  - pull_request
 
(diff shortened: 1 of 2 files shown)


.github/workflows/push.yml
```diff
@@ -2,6 +2,7 @@ name: PHP Checks
 
 on:
   - push
+  - pull_request
```
staabm: This was the missing piece why the PR was not build..

staabm:
You can easily see that all github actions are already finished, while travis just started a single job at that time

.github/workflows/push.yml
Lctrs: Should be : 

```yaml
on:
  pull_request:
  push:
    branches:
      - master
```

if you don't want the CI to run twice on PR that comes from another branch in the repo.

staabm: Thx for the suggestion. Adjusted

TomasVotruba:
Thank you :+1:

PR #2592, opened 2020-01-06
Transition fatal-error scan to GithubAction

staabm:
https://github.com/rectorphp/rector/pull/2591 Should be merged before this PR, so we can see the status of the transition job in this PR prior merging


.github/workflows/push.yml
```diff
@@ -39,3 +39,15 @@ jobs:
...
+      name: Scan Fatal Errors
+      runs-on: ubuntu-latest
+      steps:
+          - uses: actions/checkout@v2
+          - name: Setup PHP
+            uses: shivammathur/setup-php@v1
+            with:
+                php-version: 7.3
+                coverage: none # disable xdebug, pcov
+          - run: composer install --no-progress && bin/rector scan-fatal-errors tests/Source/FatalErrors
```
TomasVotruba: && should never be used, it's not clear if the first fails && the second.

Could you move it to generic install section above somehow? The same way other CI's have it?

staabm: Done

staabm:
In case you want more jobs transitioned to GithubAction please give me a ping and I will see.

I would prefer moving those which have a similar setup like the ones we already moved.
Leaving some jobs on travis is still a good option because we have so much jobs in this project

TomasVotruba:
Thanks!

I was checking it yesterday and Github actions took around 3 mins in total. While Travis haven't started yet and takes about 5-7 extra minutes.

So you can continue moving to Github :+1:

staabm:
Travis also takes more time because of the defined stages. Those are build sequentially

TomasVotruba:
I wonder, why stay with Travis at all?

staabm:
Ok, I see what you are after now

PR #2600, opened 2020-01-07
[Symfony] fix process error of controller with Internationalized routing

### LINKED ISSUE #2585: [Symfony] Process error with Internationalized routing
| Rector version | v0.6.6          |
| PHP version    |  PHP 7.3.2                            |
| Full Command   | vendor/bin/rector process bin src --set symfony41 --dry-run  | 

## Current Behaviour 

```
 [ERROR] Could not process "/var/www/symfony/src/Controller/ContactController.php" file, due to:
         "Argument 1 passed to Rector\BetterPhpDocParser\PhpDocNode\Symfony\SymfonyRouteTagValueNode::__construct() must
         be of the type string, null given, called in
         /var/www/symfony/vendor/rector/rector/packages/BetterPhpDocParser/src/PhpDocNodeFactory/Symfony/SymfonyRoutePhp
         DocNodeFactory.php on line 40".
```

```php
class ContactController
{
    /**
     * @Route({
     *     "en": "/send-us-an-email",
     *     "nl": "/stuur-ons-een-email"
     * }, name="contact")
     */
    public function send()
    {
        // ...
    }
}
```

It looks like the problem is, that the constructor of the `SymfonyRouteTagValueNode` expects a string for the $path, but actually it can be an array.

ghostika:
Closes #2585

### DIFF
--- a/packages/BetterPhpDocParser/src/PhpDocNode/Symfony/SymfonyRouteTagValueNode.php
+++ b/packages/BetterPhpDocParser/src/PhpDocNode/Symfony/SymfonyRouteTagValueNode.php
@@ -26,7 +26,7 @@ final class SymfonyRouteTagValueNode extends AbstractTagValueNode
     public $name;
 
     /**
-     * @var string
+     * @var string|null
      */
     private $path;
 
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Thank you :+1:

PR #2603, opened 2020-01-07
Transition more jobs to GithubAction

staabm:


.github/workflows/push.yml
```diff
@@ -49,6 +49,35 @@ jobs:
...
+      steps:
+          - uses: actions/checkout@v2
+          - name: Setup PHP
+            uses: shivammathur/setup-php@v1
+            with:
+                php-version: 7.2
+                coverage: none # disable xdebug, pcov
+          - run: |
+              composer install --no-progress
+              composer check-docs
```
TomasVotruba: These few lines are repeated everywhere. Any way to exclude it somewhere above? Like Travis has?

.github/workflows/push.yml
```diff
@@ -49,6 +49,35 @@ jobs:
...
+              
+  simple-checks:
+      name: Simple checks
```
TomasVotruba: This stage is allowed to fail, that needs to be ported too

.github/workflows/push.yml
staabm: Not that I am aware of

.github/workflows/push.yml
staabm: Allow failure/requirements are configured in the branch protection settings of the repo

staabm: Per default every job is allowed to fail

TomasVotruba: I didn't find it there. Just require pass before merge

Could you share a link?

.github/workflows/push.yml
TomasVotruba: Well, that's kinda stupid to repeat it everywhere.

.github/workflows/push.yml
staabm: PNG-Bild

TomasVotruba: That's the title I saw, but it's kinda of inversed meaning.
I'll try that

.github/workflows/push.yml
staabm: Indeed

TomasVotruba: If that's kinda of Yaml, this might help:
https://github.com/rectorphp/rector/blob/master/.travis.yml#L122-L145

.github/workflows/push.yml
TomasVotruba: I tried to add it somehow

.github/workflows/push.yml
TomasVotruba: I tried it elsewhere and it's not support yet :/ damn

https://github.community/t5/GitHub-Actions/Support-for-YAML-anchors/m-p/31845/highlight/true#M919

PR #2605, opened 2020-01-08
more reliable way getting last stmt

Jaapze:
This is a more reliable way of getting the last array element (preserving the internal pointer), especially when the keys of the stmt gets scrambled by unsetting items.

### DIFF
--- a/packages/DeadCode/src/Rector/FunctionLike/RemoveDeadReturnRector.php
+++ b/packages/DeadCode/src/Rector/FunctionLike/RemoveDeadReturnRector.php
@@ -73,7 +73,7 @@ PHP
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


packages/DeadCode/src/Rector/FunctionLike/RemoveDeadReturnRector.php
```diff
@@ -73,7 +73,7 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
-        $lastStmt = $node->stmts[count($node->stmts) - 1];
+        $lastStmt = array_values(array_slice($node->stmts, -1))[0];
```
TomasVotruba: Could you decouple it to 2 lines with some variable names? It's very hard to read, too many things at once

PR #2618, opened 2020-01-09
More reliable way getting first stmt item

Jaapze:
The key values are not reliable, this problem is caused by unsetting the values in the stmts array of a node without resetting them. It's not only in this rector file and might need a bigger fix, but this one was bugging me whole day.


Jaapze:
I'm not really sure why the **Travis CI build** failed, can i re-trigger these checks?

TomasVotruba:
CI fail is unrelated, it's PHAR building.
I'll move it only to master today.

PR #2622, opened 2020-01-09
added link

C0pyR1ght:


README.md
```diff
@@ -92,7 +92,7 @@ Featured open-source projects have **prepared sets**. You can find them in [`/co
 vendor/bin/rector sets
 ```
 
-Let's say you pick the `symfony40` set and you want to upgrade your `/src` directory:
+Let's say you pick the [`symfony40`](https://github.com/rectorphp/rector/tree/master/config/set/symfony) set and you want to upgrade your `/src` directory:
```
staabm: A few lines below we use a relative path for such links.. maybe this should also be used here for consistency 

E.g. /config/set/symfony (Not tested)

PR #2627, opened 2020-01-10
[VarConstantCommentRector] Allow mixed[] in place of Rector's setting wrong infered types

gnutix:
I investigated a while but got stuck when I stumbled upon `PHPStan\Type\UnionType` on [`$newType->getItemType()`](https://github.com/rectorphp/rector/blob/master/packages/NodeTypeResolver/src/PhpDoc/NodeAnalyzer/DocBlockManipulator.php#L260). Any idea how to use that ?

### DIFF
--- a/packages/CodingStyle/tests/Rector/ClassConst/VarConstantCommentRector/Fixture/correct_mixed_to_specific_types.php.inc
+++ b/packages/CodingStyle/tests/Rector/ClassConst/VarConstantCommentRector/Fixture/correct_mixed_to_specific_types.php.inc
@@ -1,0 +1,35 @@
+<?php
+
+namespace Rector\CodingStyle\Tests\Rector\ClassConst\VarConstantCommentRector\Fixture;
+
+class CorrectMixedToSpecificTypes
+{
+    /**
+     * @var mixed[]
+     */
+    public const ARRAY_CONST = [
+        'key' => 'value',
+        'key2' => ['value2', 1234],
+        1 => null,
+    ];
+}
+
+?>
+-----
+<?php
+
+namespace Rector\CodingStyle\Tests\Rector\ClassConst\VarConstantCommentRector\Fixture;
+
+class CorrectMixedToSpecificTypes
+{
+    /**
+     * @var array<int|string, string|array<int|string>|null>
+     */
+    public const ARRAY_CONST = [
+        'key' => 'value',
+        'key2' => ['value2', 1234],
+        1 => null,
+    ];
+}
+
+?>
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Test looks correct. What is the bug?

gnutix:
This Rector shouldn't override an existing "mixed" phpdoc by trying to figure out all the possible values for it. That's what mixed stands for.

TomasVotruba:
`mixed` stands for unknown types, not for "more than one type". 
With constants it's constant value (by definition), so types are known, moreover in this case. So the solution is correct.

TomasVotruba:
The issues is, most people use mixed as "I don't care about these types, just want to silenece coding standards and phpstan".

That's what Rector handles for them in most correct way. It also helps PHPStan to work better with used constants :+1:

gnutix:
That's one way to look at it. I don't entirely agree, but even then the modified phpdoc proposed by Rector is wrong : `string|int|string[]|null[]`. The return value never will be a string nor an int.

TomasVotruba:
> The return value never will be a string nor an int.

That shoud be fixed then

TomasVotruba:
Could you add second half of the tests - what is expected?

gnutix:
Will do later then. Not sure I know how to describe such an array (and it's not even the most complex example a codebase could use...) though...

gnutix:
As I said [here](https://github.com/rectorphp/rector/issues/2632), I don't think there's a "currently supported everywhere" syntax for such an array except `mixed[]` (an array of many types). PHPStan 0.12+ syntax would look like this `array<int|string, string|array<string|int>|null>`, but I doubt we want Rector to use it ?

TomasVotruba:
> PHPStan 0.12+ syntax would understand something like this array<int|string, string|array<string|int>|null>

That should be the fix then

gnutix:
I'm out of my game to implement such a type infering mecanism. I can provide PRs to allow `mixed` as a workaround for now, but that's it.

TomasVotruba:
This one is correct, no need to implement anything:

image


Just decouple this to a new test with e.g. `make_mixed_more_precise.php.inc`

gnutix:
> the modified phpdoc proposed by Rector is wrong : string|int|string[]|null[]. The return value never for this function will never be a string nor an int. string[] or null[] "might be correct", but still does not depict the proper array structure.

:thinking:

TomasVotruba:
Then first add expected fixture, to make test fail correctly.

I can merge it and check what I can do locally then. I think is some union mix-up

gnutix:
Well this PR contains the proper fixture for solution 1 (allowing mixed[]).

Are you asking me to change it to go for solution 2 instead (better type infering) ?

TomasVotruba:
> Are you asking me to change it to go for solution 2 instead (better type infering) ?

:+1:

gnutix:
@TomasVotruba done

TomasVotruba:
I try it locally now at it's all correct, except one array level above.

Investigating..

PR #2630, opened 2020-01-10
[AddArrayReturnDocTypeRector] Allow mixed[] and iterable<mixed> in place of Rector's setting wrong infered types

gnutix:
Without this fix, it changes the return phpdoc into stuff like `@return int[][]|int[][][]|mixed[]`, which is a mess to read and is **plain wrong** (a correct one could be `@return array<int|int[]>` or `array<int|array<int>>` or `int[]|int[][]` (though I don't like that multiple `[]` syntax, hard to read IMO).


gnutix:
Not ready yet... I have other weird cases to fix.

packages/TypeDeclaration/tests/Rector/ClassMethod/AddArrayReturnDocTypeRector/Fixture/skip_mixed_array.php.inc
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Rector\TypeDeclaration\Tests\Rector\ClassMethod\AddArrayReturnDocTypeRector\Fixture;
+
+final class SkipMixedArray
+{
+    /**
+     * @return mixed[]
+     */
+    public function someMethod(): array
```
TomasVotruba: The more I see these issue, the more I think the mixed type should be configurable. Explicit conversion by default (*my* approach),
optionally preffered (*your* approach).

What do you think?

gnutix: I guess the question is : do people often use Rector to upgrade code which has many "mixed" type already written to get more specific types (guessed from the code itself) ? If yes, then it makes sense that Rector modifies existing mixed phpdocs. Otherwise, I would say that Rector shouldn't touch an existing mixed type (letting the developer be in control) and add specific types when the phpdoc is missing.

gnutix:
Do you need anything else to merge this PR ? IMHO, it's ready.

TomasVotruba:
This is hard to review.

It seems there are 2 mixed changes, skipping iterable and skipping mixed.
First one is ok, the next one would better to discuss and solve separately.

gnutix:
It touches exactly the same condition, so I did them both at once indeed. Would be a pain for me to separate them in two PRs. They are essentially the same things : allowing mixed for an array (`mixed[]`) or iterable (`iterable<mixed>`).

TomasVotruba:
> I made the fixtures more explicit. Should be easier to read the diff.

I just got into this. It it easier to read, thanks :+1:

PR #2631, opened 2020-01-10
[AddArrayReturnDocTypeRector] sets a less specific type in child method (mixed[]) than is defined in parent method (SomeObject[])

gnutix:
I'm curious to have your opinion / expertise on this one. I personally don't use `{@inheritdoc}` nor duplicate PHPDoc between parents and children, as PHPStorm and PHPStan both perfectly understand parent definitions. So here Rector is trying to add the PHPDoc to the child, and I don't think it should.

Not sure how to implement the fix though, if you've got a clue it might get me going (now that I'm getting the hang of it ;) ).


TomasVotruba:
Missing doc opens possible bug:

```diff
 public function someMethod(): array
 {
-    return ['test', 'test2'];
+    return ['test', 2];
 }
```

That's why fix is correct here

gnutix:
PHPStan would catch this; it's his job I think.

gnutix:
> don't rely too much on tooling

PHP won't ever break on a phpdoc ; so there's no other way but to rely on tools for this... Not sure I get your point. :thinking:

TomasVotruba:
Rectors' purpose is to ease every day development. Having look at other file to understand current one, is not easing users work.

gnutix:
Then it is overlapping with PhpStan here, right ? Is that something you want ? (I'm more and more feeling the need to have a clearer vision of Rector's purpose and boundaries as I get involved in fixes)

TomasVotruba:
The vision is best desribed here: https://getrector.org/

This feature is just one of many that can help with that. Many tool are overlapping, coding standard upgrade code, Rector completes types, that PHPStan and PHPStorm use for better static analysis.

In this specific case, the added annotation helps user to write code that is expected. Not manually, but deduced from the code from static analysis, a completed for the sure. That's the main goal of Rector.

gnutix:
Thanks! I realize now that I should have tried to get a more representative failed test. Here's what happens in my project : 

1. The parent interface sets a return type phpdoc of `@return Entity1[]` on a method `findByEntity2Id(Uuid $id): array;`.
2. The child implements the method without phpdoc, and Rector adds `@return mixed[]` in AddArrayReturnDocTypeRector.

My problem here isn't really the fact that the phpdoc gets written both in the Interface and the children (though it does, to some extends - I'd rather not have it twice), but what really bothers me is that the added type hint is less specific than the parent (`mixed[]` vs `Entity1[]`), which breaks auto-completion. I tried adding a failing test case with my code sample, but as often with code depending on third party libraries (Doctrine here), I couldn't get it to trigger.

I'll try one more time in `src/Whatever`...

gnutix:
I had to set it up in my "playground" repository : https://github.com/gnutix/rector-playground/pull/2/files

Output : 

```
Rector dev-master@a21b805
Config file: rector.yaml

[parsing] src/DoctrineRepository.php
[parsing] src/Entity1.php
[parsing] src/Entity1DoctrineRepository.php
[parsing] src/Entity1RepositoryInterface.php
[parsing] src/Entity2.php
[refactoring] src/DoctrineRepository.php
[refactoring] src/Entity1.php
[refactoring] src/Entity1DoctrineRepository.php
[refactoring] src/Entity1RepositoryInterface.php
[refactoring] src/Entity2.php
[printing] src/DoctrineRepository.php
[printing] src/Entity1.php
[printing] src/Entity1DoctrineRepository.php
[printing] src/Entity1RepositoryInterface.php
[printing] src/Entity2.php

1 file with changes
===================

1) src/Entity1DoctrineRepository.php

--- Original
+++ New
@@ -9,6 +9,9 @@
  */
 final class Entity1DoctrineRepository extends DoctrineRepository implements Entity1RepositoryInterface
 {
+    /**
+     * @return mixed[]
+     */
     public function findByEntity2(Entity2 $entity2): array
     {
         return $this->repository->findBy(['entity2Id' => $entity2->getId()]);

Applied rules:

 * Rector\TypeDeclaration\Rector\ClassMethod\AddArrayReturnDocTypeRector
 
 [OK] Rector is done! 1 file would have changed (dry-run).                           
```

TomasVotruba:
I'm getting lost in this. What is the issue now?

gnutix:
My last two comments explain the real issue I'm having. Ignore the PR's changes, it's an overly simplified version which doesn't really work.

TomasVotruba:
I didn't understand that comment. Diff of what you expect vs. what happen would help me more.

gnutix:
Said as simply as possible : 

### What Rector does

```
--- Original
+++ New
@@ -9,6 +9,9 @@
  */
 final class Entity1DoctrineRepository extends DoctrineRepository implements Entity1RepositoryInterface
 {
+    /**
+     * @return mixed[]
+     */
     public function findByEntity2(Entity2 $entity2): array
     {
         return $this->repository->findBy(['entity2Id' => $entity2->getId()]);
```

### What I would like 

```
--- Original
+++ New
@@ -9,6 +9,9 @@
  */
 final class Entity1DoctrineRepository extends DoctrineRepository implements Entity1RepositoryInterface
 {
     public function findByEntity2(Entity2 $entity2): array
     {
         return $this->repository->findBy(['entity2Id' => $entity2->getId()]);
```

### What would be acceptable

```
--- Original
+++ New
@@ -9,6 +9,9 @@
  */
 final class Entity1DoctrineRepository extends DoctrineRepository implements Entity1RepositoryInterface
 {
+    /**
+     * @return Entity1[]
+     */
     public function findByEntity2(Entity2 $entity2): array
     {
         return $this->repository->findBy(['entity2Id' => $entity2->getId()]);
```

TomasVotruba:
Btw, tip for syntax highlight. diff after the ``` makes it nice and clear for the eyes

gnutix:
Please, indulge me if I miss highlighting in one of the 294739 code sample I provide you ... Nobody's perfect. :/

gnutix:
Should I turn this into an issue instead of a PR, as the current code change makes no sense ? (a proper one being https://github.com/gnutix/rector-playground/pull/2/files )

gnutix:
Is there a Rector that has some "parent aware" logic for types ? (given a method, find if it has a parent and analyze it) If you point me out to an example, I could try to implement a fix.

TomasVotruba:
Sorry for delaying answer, I had to think hard about feature you needed. 

I don't think there is anything that does what you need, but you can get inspiration at VendorLockResolver - https://github.com/rectorphp/rector/blob/7c8a8618fe947d30baef655affa7ea9fb48655c1/packages/vendor-locker/src/VendorLockResolver.php

TomasVotruba:
I'll look on it

TomasVotruba:
The test case now shows correctly added `string[]` in #2866  

Possibly fixed by type refactoring

PR #2635, opened 2020-01-10
Test with Doctrine

TomasVotruba:
Helps with #2633


TomasVotruba:
@mssimi Hi, 

I tried to add your tip about running Rector on big project.
This one is for doctrine/orm. It took me hour to make dev dependencies right, but now it works in CI :)

PR #2637, opened 2020-01-11
[CountOnNullRector] fix Rector applying on properties with phpdocs array

### LINKED ISSUE #2474: [PHP] CountOnNull on properties with PHPDoc shouldn't apply

gnutix:
Fixes #2474.

### DIFF
--- a/packages/NodeTypeResolver/src/NodeTypeResolver.php
+++ b/packages/NodeTypeResolver/src/NodeTypeResolver.php
@@ -753,14 +753,7 @@ final class NodeTypeResolver
                 $propertyOwnerStaticType = $this->getStaticType($node->class);
             }
 
-            if (! $propertyOwnerStaticType instanceof ThisType && $propertyOwnerStaticType instanceof TypeWithClassName) {
-                if ($this->parsedNodesByType->findClass($propertyOwnerStaticType->getClassName()) === null) {
-                    // positive assumption about 3rd party code
-                    return true;
-                }
-            }
-
-            return false;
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Test case is needed in this PR so CI can verify it

gnutix:
I couldn't reproduce it within your Fixtures. But I could in my playground repo : https://github.com/gnutix/rector-playground/pull/3 (test is currently failing, but if you merge this in master it will pass).

I can't do better than that (tried for hours already), sorry.

TomasVotruba:
I see.

We should discover the point of origin, since it's probably cause of other related bugs.

gnutix:
I checked the history of this condition I removed before doing the change, and it's not used anywhere else and the feature it was added for works the same (read: the related test still passes) without this condition.

So I'm not sure there's much potential consequences here.

TomasVotruba:
> So I think there is not much risk of consequences to merge this (

I believe your testing, let's give it a try then.

Issue #2634, opened 2020-01-10
[AddDoesNotPerformAssertionToNonAssertingTestRector] Infinite loop


gnutix:
| Subject        | Details                                                      |
| :------------- | :----------------------------------------------------------- |
| Rector version | `0.6.x-dev@a21b805` |
| PHP version    | `7.3.13` |
| Full Command   | `php rector.phar process -c var/rector-hang.yaml --ansi --dry-run -vvv tests/Test.php` | 

## Current Behaviour 

Given : 

```
# var/rector-hang.yaml
services:
    Rector\PHPUnit\Rector\ClassMethod\AddDoesNotPerformAssertionToNonAssertingTestRector: ~
```

And adding `dump()` calls at the beginning of each methods of `AddDoesNotPerformAssertionToNonAssertingTestRector` : 

```
^ "public function __construct()"
Rector 0.6.x-dev@a21b805
Config file: var/rector-hang.yaml

[parsing] tests/Test.php
[refactoring] tests/Test.php
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
^ "public function getNodeTypes(): array"
    [applying] Rector\PHPUnit\Rector\ClassMethod\AddDoesNotPerformAssertionToNonAssertingTestRector
^ "public function refactor(Node $node): ?Node"
^ "private function shouldSkipClassMethod(ClassMethod $classMethod): bool"

^ "private function containsAssertCall(ClassMethod $classMethod): bool"
^ "private function hasDirectAssertCall(ClassMethod $classMethod): bool"
^ "private function hasNestedAssertCall(ClassMethod $classMethod): bool"
^ "private function findClassMethod(Node $node): ?ClassMethod"
^ "private function findClassMethodByParsingReflection(Node $node): ?ClassMethod"
[...]

PR #2640, opened 2020-01-11
[AddDoesNotPerformAssertionToNonAssertingTestRector] Add failing test for Prophecy assertions.

gnutix:
> [gnutix] Is this Rector developed with other testing framework than PHPUnit in mind ? (like Prophecy)
> [Tomas] Sure, whatever real case this rule should cover.
(from https://github.com/rectorphp/rector/issues/2634)

Here's a failing test case for this (currently it adds the annotation to the method yet it should not).

However, I'm not interested in implementing a fix myself, as it's ignored from my project anyway and I already wasted way too much time on this one : I'd rather focus on other issues (or take a break :smile: ). I could upgrade this Rector's documentation to state that it currently does not support Prophecy and any help in implementing it is welcome though.


TomasVotruba:
> However, I'm not interested in implementing a fix myself, as it's ignored from my project anyway and I already wasted way too much time on this one : I'd rather focus on other issues. I could upgrade this Rector's documentation to state that it currently does not support Prophecy and any help in implementing it is welcome though.

No troubles, I'll handle it. Thank you

PR #2646, opened 2020-01-11
Fix Union Array type StaticTypeMapper to string

TomasVotruba:
Closes #2627


TomasVotruba:
@gnutix This one was tricky to resolve. The problem is, that PHPStan and php-parser see this content in different ways:

```php
int[]|string[]
```

For one it is *array of union types*, something like `(int|string)[]`
for the other it's *union type of array*, something like `int[]|string[]`

Issue #2614, opened 2020-01-08
NotImplementedException due to: "Rector\NodeTypeResolver\StaticTypeMapper::mapPHPStanTypeToPhpParserNode for PHPStan\Type\IntersectionType"


UFTimmy:
| Subject        | Details                                                      |
| :------------- | :----------------------------------------------------------- |
| Rector version | Rector dev-master@870e6e7          |
| PHP version    | 7.2.12                              |
| Full Command   | ..\vendor\bin\rector.bat process <file> --dry-run --rule=Rector\TypeDeclaration\Rector\FunctionLike\ReturnTypeDeclarationRector  | 

## Current Behaviour 

```
    [applying] Rector\TypeDeclaration\Rector\FunctionLike\ReturnTypeDeclarationRector
PHP Fatal error:  Uncaught Rector\Exception\NotImplementedException: Rector\NodeTypeResolver\StaticTypeMapper::mapPHPStanTypeToPhpParserNode for PHPStan\Type\IntersectionType in vendor\rector\rector\packages\NodeTypeResolver\src\StaticTypeMapper.php:288
Stack trace:
#0 vendor\rector\rector\packages\TypeDeclaration\src\Rector\FunctionLike\ReturnTypeDeclarationRector.php(189): Rector\NodeTypeResolver\StaticTypeMapper->mapPHPStanTypeToPhpParserNode(Object(PHPStan\Type\IntersectionType))
#1 vendor\rector\rector\packages\TypeDeclaration\src\Rector\FunctionLike\ReturnTypeDeclarationRector.php(120): Rector\TypeDeclaration\Rector\FunctionLike\ReturnTypeDeclarationRector->isReturnTypeAlreadyAdded(Object(PhpParser\Node\Stmt\ClassMethod), Object(PHPStan\Type\IntersectionType))
#2 vendor\rector\rector\src\Rector\AbstractRector.php(116): Rector\TypeDeclaration\Rector in vendor\rector\rector\packages\NodeTypeResolver\src\StaticTypeMapper.php on line 288
Fatal error: Uncaught Rector\Exception\NotImplementedException: Rector\NodeTypeResolver\StaticTypeMapper::mapPHPStanTypeToPhpParserNode for PHPStan\Type\IntersectionType in vendor\rector\rector\packages\NodeTypeResolver\src\StaticTypeMapper.php:288
Stack trace:
[...]

PR #2649, opened 2020-01-12
Add PHPStanAttributeTypeSyncer

### LINKED ISSUE #2647: [DX] Add automatic generator for new PHPStan types with attribute aware support + check in CI
Something like:

```bash
bin/rector generate-attribute-phpstan-types
```

## Covers cases like

- https://github.com/rectorphp/rector/pull/2613/files
- https://github.com/rectorphp/rector/commit/b4e30172c8329a2d984c623036405e7938db67fe#diff-f1e612517030eb9a2ef4ba37ca8dfeaf

```
 [ERROR] Could not process                                                      
         "/home/runner/work/rector-prefixed/rector-prefixed/cakephp/src/Controll
         er/Component/FormProtectionComponent.php" file, due to:                
         "Implement "PHPStan\PhpDocParser\Ast\Type\ArrayShapeNode" to           
         "Rector\BetterPhpDocParser\Attributes\Ast\AttributeAwareNodeFactory::cr
         eateFromTypeNode"". 
```

TomasVotruba:
Closes #2647 

## Related Changes

- bump PHPStan to 0.12.5, break with number type resolution in https://github.com/rectorphp/rector/pull/2649/commits/d488d408f65d5ef289db6dfcda0bb8c36721d11e


TomasVotruba:
@gnutix This rather meta-utility, to generate AttributeAware nodes from PHPStan nodes. That way we'd know about new PHPStan nodes much faster then reported in the issues.

I'd like to know your feedback.

gnutix:
That would be awesome! I was wondering how many more types I would discover and add in PRs like these. :smile:

TomasVotruba:
Well, I hope that 0 :D should be automated by Github Actions Cron

packages/AttributeAwarePhpDoc/src/AttributeAwareNodeFactory/AttributeAwareGenericTagValueNodeFactory.php
```diff
@@ -0,0 +1,18 @@
...
+    }
+    public function isMatch(\PHPStan\PhpDocParser\Ast\Node $node): bool
+    {
+        return is_a($node, \PHPStan\PhpDocParser\Ast\PhpDoc\GenericTagValueNode::class, true);
+    }
+    public function create(\PHPStan\PhpDocParser\Ast\Node $node): \Rector\BetterPhpDocParser\Contract\PhpDocNode\AttributeAwareNodeInterface
+    {
+        return new \Rector\AttributeAwarePhpDoc\Ast\PhpDoc\AttributeAwareGenericTagValueNode();
+    }
+}
```
gnutix: This file doesn't seem to match coding standards. Is it on purpose ?

packages/AttributeAwarePhpDoc/src/AttributeAwareNodeFactoryCollector.php
```diff
@@ -0,0 +1,44 @@
...
+     * @return string[]
+     */
+    public function getSupportedNodeClasses(): array
+    {
+        $supportedNodeClasses = [];
+        foreach ($this->attributeAwareNodeFactories as $attributeAwareNodeFactory) {
+            $supportedNodeClasses[] = $attributeAwareNodeFactory->getOriginalNodeClass();
+        }
+
+        return $supportedNodeClasses;
```
gnutix: What about using `array_map` here ? Something like : 

```php
<?php
return array_map(
    static function (AttributeAwareNodeFactory $factory): string {
        return $factory->getOriginalNodeClass();
    },
    $this->attributeAwareNodeFactories
);
```

packages/AttributeAwarePhpDoc/src/AttributeAwareNodeFactory/AttributeAwareGenericTagValueNodeFactory.php
TomasVotruba: I didn't run coding standard in this commit.
It fixes it correctly, I tried it

utils/PHPStanAttributeTypeSyncer/src/ClassNaming/AttributeClassNaming.php
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public function createAttributeAwareClassName(string $nodeClass): string
+    {
+        return Paths::NAMESPACE_PHPDOC_NODE . '\\' . $this->createAttributeAwareShortClassName($nodeClass);
+    }
+
+    public function createAttributeAwareFactoryClassName(string $nodeClass): string
+    {
+        return Paths::NAMESPACE_NODE_FACTORY . '\\' . $this->createAttributeAwareFactoryShortClassName($nodeClass);
```
gnutix: Not using `sprintf` here? :)

packages/AttributeAwarePhpDoc/src/AttributeAwareNodeFactoryCollector.php
TomasVotruba: I find it very hard to read

utils/PHPStanAttributeTypeSyncer/src/Command/SyncTypesCommand.php
```diff
@@ -0,0 +1,104 @@
...
+            );
+
+            return Shell::CODE_SUCCESS;
+        }
+
+        $this->symfonyStyle->error('These classes are missing their attribute aware brother');
+
+        foreach ($missingNodeClasses as $missingNodeClass) {
+            // 1. generate node
+//            $this->attributeAwareNodeGenerator->generateFromPhpDocParserNodeClass($missingNodeClass);
```
gnutix: Is that not working yet?

packages/AttributeAwarePhpDoc/src/AttributeAwareNodeFactoryCollector.php
TomasVotruba: Also array walk, filter etc. have gifferent syntax, behavior, some reference, some return array. Often leads to bugs along with unclear code

utils/PHPStanAttributeTypeSyncer/src/ClassNaming/AttributeClassNaming.php
TomasVotruba: Apparently :D

utils/PHPStanAttributeTypeSyncer/src/Command/SyncTypesCommand.php
TomasVotruba: This one is
 I'm testing the 2nd one now

utils/PHPStanAttributeTypeSyncer/src/NodeFactory/AttributeAwareClassFactory.php
```diff
@@ -0,0 +1,65 @@
...
+{
+    /**
+     * @var string
+     */
+    private const ATTRIBUTE_AWARE_NODE_INTERFACE = 'Rector\BetterPhpDocParser\Contract\PhpDocNode\AttributeAwareNodeInterface';
+
+    /**
+     * @var string
+     */
+    private const ATTRIBUTE_TRAIT = 'Rector\BetterPhpDocParser\Attributes\Attribute\AttributeTrait';
```
gnutix: Why not using `::class` ?

TomasVotruba: Good point, I will

utils/PHPStanAttributeTypeSyncer/src/ValueObject/Paths.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    /**
+     * @var string
+     */
+    public const NAMESPACE_PHPDOC_NODE = 'Rector\AttributeAwarePhpDoc\Ast\PhpDoc';
+
+    /**
+     * @var string
+     */
+    public const NAMESPACE_NODE_FACTORY = 'Rector\AttributeAwarePhpDoc\AttributeAwareNodeFactory';
```
gnutix: I wish there was a PHP `::namespace` syntax (like `::class`) when I see this. ;)
Here's two alternatives : 

```php
<?php
$namespace = (new \ReflectionClass(SomeClassInThatNamespace::class))->getNamespaceName();
$namespace = (string) preg_replace('/(\\\(?!.*\\\).*)/', '', SomeClassInThatNamespace::class);
```

packages/AttributeAwarePhpDoc/src/AttributeAwareNodeFactoryCollector.php
gnutix: :+1: (we use a custom library of functional methods in our project to circumvent that.. I know Laravel offers that too)

gnutix:
Just went through the code quickly ; I didn't read the logic itself nor tested the feature, just added some general comments about code style.

Is there specific testing you'd like me to do as part of the review ?

utils/PHPStanAttributeTypeSyncer/src/ValueObject/Paths.php
TomasVotruba: Cool :+1: 
, I'll try to wrap it in some function.

TomasVotruba: Ha, I tried it, but in this case it would assume there is some class in the namespace. While there is 0 classes before generating and they could be overriden/removed later.

I'll stick with namespaces for now.

TomasVotruba:
Thank you for review.

#2614 is unrelated to this. This only creates Attribute aware mirror nodes for PhpDocParser nodes. Basically what PhpParser has for format preserving.

The #2614 is about new node, that needs to be converted by static mapper. 
There could be used similar approach too - wanna try to refactor it?

gnutix:
I will have very busy upcoming weeks, don't have much mental space to dig into these Rector/PhpStan/PhpParser layers of types (and frankly, its seriously missing documentation on these classes with complicated names - I can't understand the concepts behind it just looking at the code).

TomasVotruba:
> I will have very busy upcoming weeks, don't have much mental space to dig into these Rector/PhpStan/PhpParser layers of types

I think it takes time to understand AST and related mechanism. Don't rush on yourself. It took me 1-2 years to get from tokens into state, where the printed content looked as I needed.

> But I could yet do another PR to fix #2614... That would be fast and easy. 

:+1:

PR #2650, opened 2020-01-12
[CountOnNullRector] Should understand array/countable variable in trait method

gnutix:
This PR adds two fixtures relating to `trait` handling by `CountOnNullRector`. Currently : 

* `PropertyWithinTraitMethod` : passes :+1:
* `VariableWithinTraitMethod` : fails :-1: 

If I change the Fixture from a `trait` to a `class` they both pass, so this is related to how a variable within a trait is handled.

I spent some time digging and tried to use `\Rector\NodeTypeResolver\PerNodeTypeResolver\VariableTypeResolver::resolve()` that seems to be written to handle resolution of variables (though it's used nowhere in Rector's code, only in its own Test file :thinking: ) by calling : 

* `resolveTypesFromScope($node)`, which calls...
* `resolveNodeScope($node)`, which calls on `\Rector\NodeTypeResolver\PHPStan\Collector\TraitNodeScopeCollector`...
* `getScopeForTraitAndNode($traitName, $node)`)

But it returned `null`, which was resolved to `MixedType` in `resolveTypesFromScope` and bubbled up back to `\Rector\Php71\Rector\FuncCall\CountOnNullRector::refactor` which passes the `isNullableType` check and thus refactors the code.

@TomasVotruba Any idea how to deal with this one ?

### DIFF
--- a/packages/Php71/tests/Rector/FuncCall/CountOnNullRector/Fixture/property_within_trait_method.php.inc
+++ b/packages/Php71/tests/Rector/FuncCall/CountOnNullRector/Fixture/property_within_trait_method.php.inc
@@ -1,0 +1,16 @@
+<?php
+
+namespace Rector\Php71\Tests\Rector\FuncCall\CountOnNullRector\Fixture;
+
+trait PropertyWithinTraitMethod
+{
+    /**
+     * @var array
+     */
+    private $array = [];
+
+    public function run()
+    {
+        return count($this->array);
+    }
+}
--- a/packages/Php71/tests/Rector/FuncCall/CountOnNullRector/Fixture/variable_within_trait_method.php.inc
+++ b/packages/Php71/tests/Rector/FuncCall/CountOnNullRector/Fixture/variable_within_trait_method.php.inc
@@ -1,0 +1,20 @@
+<?php
+
+namespace Rector\Php71\Tests\Rector\FuncCall\CountOnNullRector\Fixture;
+
+use Rector\Php71\Tests\Rector\FuncCall\CountOnNullRector\Source\CountableClass;
+
+trait VariableWithinTraitMethod
+{
+    public function run()
+    {
+        $array = [];
+        $countable = new CountableClass();
+
+        return
+            count($array)
+            +
+            count($countable)
+        ;
+    }
+}


gnutix:
I tried the adding the following code at the beginning of `\Rector\NodeTypeResolver\NodeTypeResolver::getStaticType()` and run it against `packages/Php71/tests/Rector/FuncCall/CountOnNullRector/Fixture/variable_within_trait_method.php.inc`, but the output was `PHPStan\Type\MixedType` (where I expected it to be `ArrayType` or `ObjectType` : 

```php
<?php
if ($node instanceof Variable) {
    $paramStaticType = $this->variableTypeResolver->resolve($node);
    if ($paramStaticType !== null) {
        dump(get_class($paramStaticType));
    }
}
```

gnutix:
Could this be fixed by PHPStan 0.12.5 release containing https://github.com/phpstan/phpstan/issues/1601 ?

TomasVotruba:
> Could this be related to phpstan/phpstan#1601 (fixed in the new release by PHPStan 0.12.5) ?

Test would answer us

TomasVotruba:
`master` now has 0.12.5, try rebase

gnutix:
Last push force is the rebase on master. We'll see.

gnutix:
@TomasVotruba Not related to PHPStan 0.12.5. Any idea ?

TomasVotruba:
@gnutix What do you mean? Link might help

gnutix:
I mean that I've rebased this PR on master, which includes PHPStan 0.12.15, and it didn't fix the broken test. So the issue remains, and I have no lead / clue on why this is broken nor how to fix this. So I'm asking for your help / guidance. :)

TomasVotruba:
This might be more tricky. Traits sucks hard.

Are you ok with merging as it is?
I'll have to look at it locally with trial error.

gnutix:
Sure!

TomasVotruba:
Both passes after merge :+1: probably fixed by type refactoring

TomasVotruba:
Or not... false positive: paths have changed from `packages` to `rules`

TomasVotruba:
After bit of digging, traits are impossible to analyse without class by PHPStan. It provides many false positives on standalone analysis, so we better skip them.

If you find a stable way of using PHPStan for traits here, I'm open to the PR.

PR #2653, opened 2020-01-12
Enhancement: Normalize composer.json

localheinz:
This PR

* [x] normalizes `composer.json`

💁‍♂️ I ran

```
$ composer global require ergebnis/composer-normalize
```

followed by

```
$ composer normalize
```

For reference, see https://github.com/ergebnis/composer-normalize.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,14 +1,26 @@
 {
     "name": "rector/rector",
     "description": "Instant upgrade and refactoring of your PHP code",
-    "keywords": ["instant upgrades", "instant refactoring", "ast", "automated refactoring"],
+    "keywords": [
+        "instant upgrades",
+        "instant refactoring",
+        "ast",
+        "automated refactoring"
+    ],
     "homepage": "https://getrector.org",
     "license": "MIT",
     "authors": [
-        { "name": "Tomas Votruba", "email": "tomas.vot@gmail.com", "homepage": "https://tomasvotruba.com" },
-        { "name": "Jan Mikes", "email": "j.mikes@me.com", "homepage": "https://janmikes.cz" }
+        {
+            "name": "Tomas Votruba",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -1,14 +1,26 @@
 {
     "name": "rector/rector",
     "description": "Instant upgrade and refactoring of your PHP code",
-    "keywords": ["instant upgrades", "instant refactoring", "ast", "automated refactoring"],
+    "keywords": [
+        "instant upgrades",
+        "instant refactoring",
+        "ast",
+        "automated refactoring"
+    ],
```
TomasVotruba: These should be inlined, as added value for reader is minimal

composer.json
```diff
@@ -1,14 +1,26 @@
...
+        {
+            "name": "Tomas Votruba",
+            "email": "tomas.vot@gmail.com",
+            "homepage": "https://tomasvotruba.com"
+        },
+        {
+            "name": "Jan Mikes",
+            "email": "j.mikes@me.com",
+            "homepage": "https://janmikes.cz"
+        }
```
TomasVotruba: These should be inlined, as added value for reader is minimal

composer.json
```diff
@@ -1,14 +1,26 @@
...
     ],
-    "bin": ["bin/rector"],
```
TomasVotruba: By convention on the top, so everyone can see it's binary app

composer.json
```diff
@@ -199,46 +219,41 @@
             "packages/SOLID/tests/Rector/ClassMethod/UseInterfaceOverImplementationInConstructorRector/Source/Apple.php"
         ]
     },
+    "minimum-stability": "dev",
+    "prefer-stable": true,
```
TomasVotruba: Least valueable for reader, should be at the very bottom

PR #2655, opened 2020-01-13
Rename --rule argument into --only, add documentation.

### LINKED ISSUE #2651: --rule argument is not working
I've noticed that running `process --rule "[someRectorFQCN] [path]` throws an error (even though the `[-r|--rule RULE]` argument is described) : 

```
Rector 0.6.x-dev@a21b805

In RectorGuard.php line 41:
                                                    
  We need some rectors to run:                      
  * register them in rector.yaml under "services:"  
  * use "--set <set>"                               
  * or use "--config <file>.yaml"                   

process [-n|--dry-run] [-a|--autoload-file AUTOLOAD-FILE] [-e|--hide-autoload-errors] [--match-git-diff] [-r|--rule RULE] [-o|--output-format [OUTPUT-FORMAT]] [--no-progress-bar] [--] [<source>...]
```

Thrown by `\Rector\Guard\RectorGuard::ensureSomeRectorsAreRegistered()`.

I've tried to investigate, but I don't understand how `\Rector\PhpParser\NodeTraverser\RectorNodeTraverser` gets populated (in `\Rector\Console\Command\ProcessCommand::execute()` there's only a call to `$this->configuration->resolveFromInput($input);`, but the Configuration class is not aware of `RectorNodeTraverser` :thinking: and `\Rector\PhpParser\NodeTraverser\RectorNodeTraverser::getPhpRectorCount()` references a `$this->visitors` which is defined nowhere...). Is there some services / autowiring / compiler pass or whatnot going on here ?

gnutix:
Fixes #2651.


stloyd:
That name sound a bit confusing, what about `--use-only`?

TomasVotruba:
Good point! Maybe `--only-rule` would be better, as the value should be Rector rule

gnutix:
Is it a "rule" or a "rector" ? I think that should be clear and coherent everywhere. Filter or only, I don't mind.

TomasVotruba:
In that case --only might be the best solution
It's used on ECS in similar way

gnutix:
Done!

TomasVotruba:
Feature is ok :+1: 
CI should be  fixed thou

gnutix:
Done! I didn't fix the PhpStan inspections related to `master`

README.md
```diff
@@ -201,6 +201,20 @@ class SomeClass
 }
 ```
 
+### Filter Rectors
+
+If you have a configuration file for Rector including many sets and Rectors, you might want at times to run only a single
+Rector from them. The `--only` argument allows that, for example :
```
TomasVotruba: Here is extra space in sentence. It might create wierd spacing.
Better let Markdown/IDE handle that

README.md
```diff
@@ -201,6 +201,20 @@ class SomeClass
...
+Rector from them. The `--only` argument allows that, for example : 
+
+```bash
+vendor/bin/rector process --set solid --only "Rector\SOLID\Rector\Class_\FinalizeClassesWithoutChildrenRector" src/
+```
+
+Will only run `Rector\SOLID\Rector\Class_\FinalizeClassesWithoutChildrenRector`.
+
+Please note that the backslash in the Rector's fully-qualified class name needs to be properly escaped (by surrounding
+the string in double quotes).
```
TomasVotruba: Here is extra space in sentence. It might create wierd spacing.
Better let Markdown/IDE handle that

TomasVotruba:
Thank you :+1:

PR #2658, opened 2020-01-14
Introduce a CallReflectionResolver

Lctrs:


TomasVotruba:
Could elaborate on motivation and main features  for this PR?

Lctrs:
Main motivation was to reduce duplicated code and leverage PHPStan to create a class that is able to return the reflected function from any calls given, even on a variable that has been assigned a closure (eg. `$f = function(){}; $f();`)

TomasVotruba:
I see, thanks :+1:

Lctrs:
Worth noting that `Rector\PhpParser\Node\Manipulator\CallManipulator` class is now unused across the project. Didn't had the gut to remove it.

TomasVotruba:
Go for it

Issue #1107, opened 2019-02-22
[CodeQuality] If/else to early-return


TomasVotruba:
```diff
if (Strings::contains($nodeClass, '\\Scalar\\')) {
-            $category = 'Scalar nodes';
+            return 'Scalar nodes';
        } elseif (Strings::contains($nodeClass, '\\Expr\\')) {
-            $category = 'Expressions';
+            return 'Expressions';
        } elseif (Strings::contains($nodeClass, '\\Stmt\\')) {
-            $category = 'Statements';
+            return 'Statements';
-        } else {
-            $category = 'Various';
+            return 'Various';
-        }
-
-        return $category;
```

TomasVotruba:
@JanMikes I think this is also one you talked about :) Just FIY

JanMikes:
Yeah, this one is very close to it.

My example was like this?

```diff
if ($a == $b) {
    return $a;
- } else {
+ }
-     throw new \Exception();
- }
+ throw new \Exception();
```

TomasVotruba:
Another example I just created:

```php
if (count($classMethod->stmts) === 2) {
            if ($classMethod->stmts[0] instanceof If_) {
                /** @var If_ $if */
                $if = $classMethod->stmts[0];
                if ($if->cond instanceof Identical) {
                    /** @var Identical $identical */
                    $identical = $if->cond;
                    if ($this->isNull($identical->left) && $identical->right instanceof StaticPropertyFetch) {
                        if (count($if->stmts) === 1) {
                            if ($if->stmts[0] instanceof Expression) {
                                $stmt = $if->stmts[0]->expr;
                                // create self and assign to static property
                                if ($stmt instanceof Assign) {
                                    if ($this->areNodesEqual($identical->right, $stmt->var)) {
                                        if ($stmt->expr instanceof New_) {
                                            if ($this->getTypes($stmt->expr->class) === [$class]) {
                                                return true;
                                            }
                                        }
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }

        return false;
```

TomasVotruba:
Covered by https://github.com/rectorphp/rector/blob/master/packages/SOLID/src/Rector/If_/RemoveAlwaysElseRector.php

PR #2675, opened 2020-01-15
Added SplitIfsRector

jeroensmit:
Might be what @JanMikes meant in #1107


TomasVotruba:
Might be covered by https://github.com/rectorphp/rector/blob/master/packages/SOLID/src/Rector/If_/RemoveAlwaysElseRector.php already

Could you check?

We'll close #1107 then

TomasVotruba:
I see there are few more cases you've added.
Could you add them to `RemoveAlwaysElseRector` and `RemoveAlwaysElseRectorTest`?

Thanks

jeroensmit:
Yes, I will.

TomasVotruba:
Thanks! Feel free to open a new PR

PR #2687, opened 2020-01-17
Changed ChangeMethodVisibilityRector yaml config

C0pyR1ght:
The example config resulted in an invalid yaml file with the current rector version.
`The configuration key "FrameworkClass" is unsupported for definition`

### DIFF
--- a/docs/AllRectorsOverview.md
+++ b/docs/AllRectorsOverview.md
@@ -8722,7 +8722,8 @@ Change visibility of method from parent class.
 ```yaml
 services:
     Rector\Rector\Visibility\ChangeMethodVisibilityRector:
-        FrameworkClass:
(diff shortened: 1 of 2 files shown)


docs/AllRectorsOverview.md
```diff
@@ -8722,7 +8722,8 @@ Change visibility of method from parent class.
 ```yaml
 services:
     Rector\Rector\Visibility\ChangeMethodVisibilityRector:
-        FrameworkClass:
+        $methodToVisibilityByClass:
+            FrameworkClass:
             someMethod: protected
```
TomasVotruba: This file is generated from `ChangeMethodVisibilityRector`. This needs to be fixed in `getRectorDefinitiv
n()` method there

PR #2689, opened 2020-01-17
Document `import_short_classes` parameter.

Aerendir:
https://github.com/rectorphp/rector/pull/2077#issuecomment-564597626


TomasVotruba:
This file is generated. It's already in https://github.com/rectorphp/rector#import-use-statements

Aerendir:
😓Ok!

PR #2691, opened 2020-01-17
updated UnwrapFutureCompatibleIfRectorTest

C0pyR1ght:
added second test case to unittest


TomasVotruba:
Fix is needed

TomasVotruba:
See https://github.com/rectorphp/rector/blob/9d82c9e76a7b8f256a52cba9e9094d0e04131d1a/packages/Polyfill/src/FeatureSupport/FunctionSupportResolver.php#L14-L21

staabm:
Looks like the impl of the UnwrapFutureCompatibleIfRector needs tuning to work with if-cases without a else clause

TomasVotruba:
I'll look into it

PR #2711, opened 2020-01-19
Add SetcookieRector

zonuexe:
Convert to [`setcookie()`](https://www.php.net/setcookie) options array introduced in PHP 7.3.

This Rector just breaks backward compatibility with less than PHP 7.3 since PHP 7.3 actually supports old arguments. However, since the option array is required to set the `samesite` attribute, the changes are highlighted.

### DIFF
--- a/docs/AllRectorsOverview.md
+++ b/docs/AllRectorsOverview.md
@@ -1,4 +1,4 @@
-# All 442 Rectors Overview
+# All 443 Rectors Overview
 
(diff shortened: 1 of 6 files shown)


TomasVotruba:
CI needs fixing

zonuexe:
@TomasVotruba I rebase on master and fixed them.

TomasVotruba:
Looks good, thank you :+1:

Issue #2705, opened 2020-01-18
Cake 2.4 to Symfony 5.1


TomasVotruba:
## 1. Controller → HttpKernel

- [x] #2698 
- [ ] @Route annotations above actions
- [ ] request global object access

<br>

## 2. Disptacher → Event Dispacher

<br>

## 3. Template .ctp to Twig

- [x] #2709
- php templates

```diff
-<?php echo $this->Form->create('Right'); ?>
+{{ form_open(Right) }}
```

```diff
-<?php echo $this->Form->input('id'); ?>
+{{ form_row(Right.id) }}
```

```diff
-<?php echo $this->Form->end(__('Submit')); ?>
+{{ form_row(Right.Submit) }}
+{{ form_end(Right) }}
```

<br>

```php
<div class="rights form">
<?php echo $this->Form->create('Right'); ?>
    <fieldset>
        <legend><?php echo __('Edit Right'); ?></legend>
    <?php
        echo $this->Form->input('id');
        echo $this->Form->input('right_category_id');
        echo $this->Form->input('name');
        echo $this->Form->input('label');
        echo $this->Form->input('description');
        echo $this->Form->input('Role');
    ?>
    </fieldset>
<?php echo $this->Form->end(__('Submit')); ?>
</div>
<div class="actions">
    <h3><?php echo __('Actions'); ?></h3>
    <ul>

        <li><?php echo $this->Form->postLink(__('Delete'), ['action' => 'delete', $this->Form->value('Right.id')], null, __('Are you sure you want to delete # %s?', $this->Form->value('Right.id'))); ?></li>
        <li><?php echo $this->Html->link(__('List Rights'), ['action' => 'index']); ?></li>
        <li><?php echo $this->Html->link(__('List Right Categories'), ['controller' => 'right_categories', 'action' => 'index']); ?> </li>
        <li><?php echo $this->Html->link(__('New Right Category'), ['controller' => 'right_categories', 'action' => 'add']); ?> </li>
        <li><?php echo $this->Html->link(__('List Roles'), ['controller' => 'roles', 'action' => 'index']); ?> </li>
        <li><?php echo $this->Html->link(__('New Role'), ['controller' => 'roles', 'action' => 'add']); ?> </li>
    </ul>
</div>
```

<br>

## 4. Database Active Rector to Doctrine ORM

- [x] relations
- [x] repository method
- [ ] validatoin

## 5. Console commands to Symfony\Console

[...]

PR #2718, opened 2020-01-19
[CakePHPToSymfony] Add h function templates

TomasVotruba:
Ref #2705


afilina:
This rector doesn't actually work. `$this->doTestFileWithoutAutoload($file);` is a trick that makes the test pass, but there's no way to skip the `require_once` in a real run.

All templates are causing fatal errors. I tried to run `CakePHPTemplateHToTwigRector` on the exact same file as in the test:

```
<li>
    <?php echo $this->Html->link(__('List Rights'), ['action' => 'index']); ?>
</li>
```

and got this:
```
Fatal error: Uncaught Error: Call to a member function link() 
```

TomasVotruba:
> but there's no way to skip the require_once in a real run.

Apperantly it can work... What would you suggest to make this work?

afilina:
I have no idea how it works deep down. I don't have any design documents and there's too much code to randomly go digging. We need the ability for a file to be the target of a rector without being force to `require_once` it.

TomasVotruba:
Could add minimal failing test case to Github Actions? 
I'll try to dig it out

PR #2728, opened 2020-01-21
Merge pull request #2728 from C0pyR1ght/patch-4

C0pyR1ght:
added further description on default autoloading. Shows the difference between default & additional autoloading and which autloader is used by deafult


TomasVotruba:
Thanks

PR #2737, opened 2020-01-21
Docker build secured image for online demo

JanMikes:
After merging this PR, we can disable autobuild images on hub.docker.com, this way it way docker images will be deployed faster


JanMikes:
Solves https://github.com/rectorphp/getrector.org/issues/79

TomasVotruba:
Let's ship it :)

JanMikes:
Works great: https://github.com/rectorphp/rector/runs/402498876 :+1:

Issue #2611, opened 2020-01-08
The compiler is dumped when running the PHAR and systematically even outside of the PHAR


theofidry:
Reading https://www.tomasvotruba.com/blog/2019/12/02/how-to-box-symfony-app-to-phar-without-killing-yourself/ , especially [this line](https://www.tomasvotruba.com/blog/2019/12/02/how-to-box-symfony-app-to-phar-without-killing-yourself/#nobody-ever-used-symfony-in-phar-before), I'm certain there is something wrong here.

So checking locally, I with a few dump statements, I can confirm my suspicions. If you go to `vendor/symfony/http-kernel/Kernel.php`, you can do:

```php
protected function initializeContainer()
    {
        $class = $this->getContainerClass();
        $cacheDir = $this->warmupDir ?: $this->getCacheDir();
        $cache = new ConfigCache($cacheDir.'/'.$class.'.php', $this->debug);
        $cachePath = $cache->getPath();

        // Silence E_WARNING to ignore "include" failures - don't use "@" to prevent silencing fatal errors
        $errorLevel = error_reporting(\E_ALL ^ \E_WARNING);

        try {
            if (file_exists($cachePath) && \is_object($this->container = include $cachePath)
                && (!$this->debug || (self::$freshCache[$cachePath] ?? $cache->isFresh()))
            ) {
// Added debug statement here
                \file_put_contents('php://stdout', 'container is fresh'.\PHP_EOL);
                self::$freshCache[$cachePath] = true;
                $this->container->set('kernel', $this);
                error_reporting($errorLevel);

                return;
            }
        } catch (\Throwable $e) {
        }

// Added debug statement here
        \file_put_contents('php://stdout', 'init container'.\PHP_EOL);

        // ...
    }
```

And if I do:

```
$ bin/rector --version
boot
init container
Rector 0.6.x-dev@9ae9807
$ bin/rector --version
boot
init container
Rector 0.6.x-dev@9ae9807
```

Note that this happens _each time_. So even without talking about the PHAR here, there is already an issue: your container is being dumped _every single time_ instead of done once and re-used. Obviously not a performance killer, but that can definitely be improved!

[...]

PR #2748, opened 2020-01-23
Do not dump the compiler

theofidry:
This is just an idea, I don't think I can manage the finish the PR but I would say the gist of it is there.

Basically based on #2611 you have to me two solutions:

- Either embrace the huge performance cost that using the non-dumped Symfony container (which means which needs to be compiled every single time) but in which case you don't want to dump it (except in dev maybe) - it's another unnecessary overhead (the IO writing) + it will definitely cause issues in a PHAR
- Refactor `Rector` to switch from a Symfony container to another one, at the very least for the dynamic configuration

I think the second solution is the right one: the Symfony container is _not_ meant to be compiled at every request/CLI call: it has a huge service abstraction + compilation time which are enormous and only acceptable because in practice, you dump it once ready for prod and you only use that tiny optimized dumped container.

But I guess it can be quite a huge refactoring, so the first one might be a better solution at least on the short-term.

### DIFF
--- a/src/HttpKernel/RectorKernel.php
+++ b/src/HttpKernel/RectorKernel.php
@@ -4,17 +4,20 @@ declare(strict_types=1);
 
 namespace Rector\HttpKernel;
 
+use Phar;
 use Rector\Contract\Rector\RectorInterface;
 use Rector\DependencyInjection\Collector\RectorServiceArgumentCollector;
 use Rector\DependencyInjection\CompilerPass\MergeImportedRectorServiceArgumentsCompilerPass;
 use Rector\DependencyInjection\CompilerPass\RemoveExcludedRectorsCompilerPass;
 use Rector\DependencyInjection\Loader\TolerantRectorYamlFileLoader;
+use Symfony\Component\Config\ConfigCache;
 use Symfony\Component\Config\Loader\DelegatingLoader;
 use Symfony\Component\Config\Loader\GlobFileLoader;
 use Symfony\Component\Config\Loader\LoaderInterface;
 use Symfony\Component\Config\Loader\LoaderResolver;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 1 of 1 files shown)


TomasVotruba:
What are the benefits over current working solution?

theofidry:
Well your choice: IMO you're avoid one overhead (dumping the container which has a lot of IO) and hacking how PHAR works (to write in them or somewhere else).

TomasVotruba:
That not related to PHAR, it's matter of config invalidation. I use similar approach in other CLI Symfony Apps.

If you find better usage of Symfony container with cached unique param container dump, I'm open to it

theofidry:
>That not related to PHAR, it's matter of config invalidation. I use similar approach in other CLI Symfony Apps.

It is related to PHAR because PHAR are meant to be a read-only environment which is a strong infrastructure concern: you cannot have just any app working in any way in there you need to be strongly aware of this constraint for which your current usage of the Symfony Container is really ill suited in its current form.

>If you find better usage of Symfony container with cached unique param container dump, I'm open to it

There is 4 solutions IMO:

- Dump this part outside of the PHAR (when using the PHAR): point the cache directory to a system tmp directory or something. You still run under the issue of relying on the Symfony Container being able to compile itself within a PHAR which is not tested by Symfony hence there is no guarantee of it always working - as you experienced it first-hand
- Tweak the Symfony Kernel like started in this PR to not dump the compiler in the FS (which you don't benefit from anyway)
- Move the current app dynamic config away from the container:
  - Use a different container that don't requires to be dumped in the FS (and likely more performant that a full-pledge definition based DI Container like the Symfony one if not compiled)
  - Don't use a container for those dynamic parameters: put them as a complete dynamic config and treat it as such

On the last point you would at the very least be able to dump the SF container before shipping it into the PHAR and then not require to dump it ever again.

PR #2756, opened 2020-01-25
github-action: Annotate Github Pull Requests based on a Checkstyle X…

staabm:
…ML-report

That means you no longer search thru your GithubAction logfiles. No need to interpret messages which are formatted differently with every tool. Instead you can focus on your Pull Request, and you don't need to leave the Pull Request area.

See https://github.com/staabm/annotate-pull-request-from-checkstyle

### DIFF
--- a/.github/workflows/phpstan.yaml
+++ b/.github/workflows/phpstan.yaml
@@ -15,5 +15,7 @@ jobs:
                 with:
                     php-version: 7.3
                     coverage: none
-            -   run: composer install --no-progress
-            -   run: composer phpstan
+            -   run: |
+                    composer install --no-progress
+                    composer require staabm/annotate-pull-request-from-checkstyle:^1.0
+            -   run: composer phpstan --error-format=checkstyle | vendor/bin/cs2pr
--- a/.github/workflows/phpstan.yaml
+++ b/.github/workflows/phpstan.yaml
@@ -18,4 +18,5 @@ jobs:
             -   run: |
                     composer install --no-progress
                     composer require staabm/annotate-pull-request-from-checkstyle:^1.0
+            # turn the phpstan errors (formatted in checkstyle-format) into github pull request check annotations
             -   run: composer phpstan -- --error-format=checkstyle | vendor/bin/cs2pr
--- a/.github/workflows/phpstan.yaml
+++ b/.github/workflows/phpstan.yaml
@@ -16,7 +16,6 @@ jobs:
                     php-version: 7.3
                     coverage: none
                     tools: cs2pr
-            -   run: |
-                    composer install --no-progress
+            -   run: composer install --no-progress
             # turn the phpstan errors (formatted in checkstyle-format) into github pull request check annotations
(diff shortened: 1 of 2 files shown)


staabm:
Does any other tool within the GithubAction jobs support checkstyle?

phpstan.neon
```diff
@@ -16,7 +16,6 @@ parameters:
         - compiler/src
 
     excludes_analyse:
-        - "*/Expected/*"
```
staabm: This change is just for demoing purposes (and verification)

Will/need to be reverted before merge

.github/workflows/phpstan.yaml
```diff
@@ -15,5 +15,7 @@ jobs:
                 with:
                     php-version: 7.3
                     coverage: none
-            -   run: composer install --no-progress
-            -   run: composer phpstan
+            -   run: |
+                    composer install --no-progress
+                    composer require staabm/annotate-pull-request-from-checkstyle:^1.0
+            -   run: composer phpstan -- --error-format=checkstyle | vendor/bin/cs2pr
```
TomasVotruba: What is `|` needed for?

TomasVotruba: Some commend in code would help. Mix of phpstan and some style is confusing

staabm: its a regular unix command pipe. the output of phpstan is piped into the following command (it is passed in as stdin to the cs2pr)

TomasVotruba: Might be better to add this tool to require-dev and and have own composer.jsom script for it.

To prevent hidden install issue and keep consistency to other jobs

staabm: the tool itself is only usefull in github action context.
therefore I prefer installing it here.

TomasVotruba: I'm against introducing new format for no added value

staabm: not sure what you mean by 'format'. in case i would add it to the existing phpstan composer command the command could no longer be invoked from the command line.

I would have to create a separate new composer command, which can only be called within github action context (or at least it doesnt provide value if called on the workstation).

if you prefer a separate composer script, I can add it though.

staabm: The next release of shivammathur/setup-php@v1 will have cs2pr bundled, so we will no longer need to install it via composer require

staabm:
this PR is ready to merge now.

In case you like it, I can drop the "testing" change from https://github.com/rectorphp/rector/pull/2756/files#r370930092

In case there is something left todo - or you have open questions - feel free to ask

TomasVotruba:
Let's give it a try :+1:

TomasVotruba:
Ping me when it's ready

staabm:
ping @TomasVotruba 

ready to merge.

in case you like it - we can use this tool for every GithubAction which is able to provide a checkstyle-formatter report.

TomasVotruba:
Thank you!

PR #2757, opened 2020-01-25
Added phpunit problem matcher

staabm:
Similar to https://github.com/rectorphp/rector/pull/2756 this will enable reporting of phpunit test failures into the files-changed PR tab

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -18,9 +18,10 @@ jobs:
             -   uses: actions/checkout@v2
             -   uses: shivammathur/setup-php@v1
                 with:
-                    php-version: 7.2
+                    php-version: ${{ matrix.php }}
                     coverage: none # disable xdebug, pcov
             # report phpunit errors into files-changed PR tab
+            # https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md
             -   name: Setup Problem Matchers for PHPUnit
                 run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
                 
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -40,6 +40,8 @@ jobs:
                     php-version: 7.2
                     coverage: none
             # report phpunit errors into files-changed PR tab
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Now there are false positives. Error reported for skipped tests

staabm:
Reported the issue upstream in https://github.com/shivammathur/setup-php/issues/163

Not sure it is fixable thought

staabm:
After checking the phpunit sources, I wonder why we see the infos printed regarding skipped tests.
https://github.com/sebastianbergmann/phpunit/blob/b6fcd8290a37d301a4a87abaf2593d5f1ac5bd45/src/TextUI/DefaultResultPrinter.php#L168

These should only be printed in verbose-mode

staabm:
Guess I found it

phpunit.xml
```diff
@@ -3,7 +3,6 @@
     xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
     bootstrap="tests/bootstrap.php"
     colors="true"
-    verbose="true"
```
staabm: In case you want to stick with verbose mode: the upstream project was fixed and the next version of the gh-action should work then

staabm:
We got an upstream release and things are looking better now

staabm:
Any feedback?

.github/workflows/tests.yaml
```diff
@@ -20,6 +20,10 @@ jobs:
                 with:
                     php-version: 7.2
```
TomasVotruba: I just notice, here should be `${{ matrix.php }}`

.github/workflows/tests.yaml
```diff
@@ -20,6 +20,10 @@ jobs:
...
                     coverage: none # disable xdebug, pcov
+            # report phpunit errors into files-changed PR tab
```
TomasVotruba: Link to some knowledge base would help

staabm:
Here you are!

TomasVotruba:
Looks good!
Let's wait for CI and merge then

Thank you for your work :+1:

TomasVotruba:
I'll look how this can be added for Rector

PR #2775, opened 2020-01-29
[DX] Reducing technical debt of DocBlockManipulator with 705 lines

TomasVotruba:
## Main Improvements

* doc is not created ad hoc manually on every use, but with parsing decoration - the way parent nodes and names are resolved 
* constistent API with one point of origin
* PhpDocInfo becomes smart value object wich modifying method
* persistence (printing PhpDocInfo back to string) if still left on the printer, so no coupling to the service layer


packages/DoctrineGedmoToKnplabs/src/Rector/Class_/TranslationBehaviorRector.php
```diff
@@ -261,7 +261,6 @@ private function dumpEntityTranslation(Class_ $class, array $translatedPropertyT
 
         foreach ($translatedPropertyToPhpDocInfos as $translatedPropertyName => $translatedPhpDocInfo) {
             $property = $this->nodeFactory->createPrivateProperty($translatedPropertyName);
-            $this->docBlockManipulator->updateNodeWithPhpDocInfo($property, $translatedPhpDocInfo);
```
TomasVotruba: Probably use setAttribute instead

Issue #2749, opened 2020-01-23
Prefixed Rector PHAR not working


theofidry:
| Subject        | Details                                                               |
| :------------- | :-------------------------------------------------------------------- |
| Rector version | Rector 0.6.x-dev@262e8d8 |
| PHP version    | 7.2.18                                       |
| Full Command   | See below | 
| Demo link      | https://github.com/infection/infection | 
| rector.yaml    | See below |

```
$ php rector.phar process /Users/tfidry/Project/Humbug/infection/src/Config/Exception/InvalidConfigException.php --dry-run --set=dead-classes -vvv
```

<details>
<summary>rector.yaml</summary>

```yaml
parameters:
    paths:
        - 'src'
        - 'tests'

    exclude_rectors:
        - 'Rector\CodeQuality\Rector\If_\SimplifyIfReturnBoolRector'

    autoload_paths:
        - 'vendor/autoload.php'

    exclude_paths:
        - 'tests/e2e/**/*'

```
</details> 

## Current behaviour

```
Rector 0.6.x-dev@262e8d8
Config file: rector.yaml

[parsing] src/Config/Exception/InvalidConfigException.php
PHP Fatal error:  Uncaught Error: Class '_HumbugBox60f4f031e4cc\JetBrains\PHPStormStub\PhpStormStubsMap' not found in phar:///path/to/infection/rector.phar/vendor/ondrejmirtes/better-reflection/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php:67
Stack trace:
#0 phar:///path/to/infection/rector.phar/vendor/ondrejmirtes/better-reflection/src/SourceLocator/Type/PhpInternalSourceLocator.php(38): _HumbugBox60f4f031e4cc\Roave\BetterReflection\SourceLocator\SourceStubber\PhpStormStubsSourceStubber->generateClassStub('self')
#1 phar:///path/to/infection/rector.phar/vendor/ondrejmirtes/better-reflection/src/SourceLocator/Type/PhpInternalSourceLocator.php(31): _HumbugBox60f4f031e4cc\Roave\BetterReflection\SourceLocator\Type\PhpInternalSourceLocator->getClassSource(Object(_HumbugBox60f4f031e4cc\Roave\BetterReflection\Identifier\Identifier))
[...]

PR #2800, opened 2020-02-04
Update to PHPStan 0.12.9 and fix scoping deps

TomasVotruba:
Ref #2749

### DIFF
--- a/.github/workflows/test_with_doctrine.yaml
+++ b/.github/workflows/test_with_doctrine.yaml
@@ -27,6 +27,7 @@ jobs:
 
                     # remove phsptan config to prevent rector loading configs
                     rm phpstan.neon
+                    rm orm/phpstan.neon.dist
 
                     # do not intall doctrine/orm phpstan, it conflicts with Retor's one
                     composer install -d orm --no-dev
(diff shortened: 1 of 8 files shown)


TomasVotruba:
@enumag Btw, this was causing the FQN name changes. 

See Github Action fail before fix: https://github.com/rectorphp/rector/pull/2800/checks?check_run_id=428744569#step:5:12

image

.github/workflows/test_with_doctrine.yaml
```diff
@@ -31,4 +31,7 @@ jobs:
                     # do not intall doctrine/orm phpstan, it conflicts with Retor's one
                     composer install -d orm --no-dev
 
+                    # remove phpstan/phpstan to prevent autolaod conflicts
+                    composer remove phpstan/phpstan -d orm
```
enumag: Since phpstan is phar-ed and prefixed, can it still cause autoload conflicts?

TomasVotruba: Yes, like this:
https://github.com/rectorphp/rector/pull/2800/checks?check_run_id=430026084

enumag:
> @enumag Btw, this was causing the FQN name changes.

@TomasVotruba huh? not sure what are you trying to tell me... the link isn't showing anything...

TomasVotruba:
@enumag The history was removed now. See the screen

PR #2802, opened 2020-02-04
added --config parameter to README

C0pyR1ght:
`--config` parameter was nowhere metioned. Therefore I complemented it at a useful place.

Further I suggest a dedicated chapter for all possible parameters in the future because they are nowhere explained. Other params like the `-n` as synonym for `-dry-run` is also not mentioned and could lead to confusion

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ parameters:
 
 ### C. Custom Sets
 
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Thanks

PR #2809, opened 2020-02-07
rename package to use lowercased standard, prevent confusion with PSR4

TomasVotruba:
Basically CamelCase to dashes-naming


TomasVotruba:
Moved

PR #2836, opened 2020-02-10
WIP: #2753 Fix Auto Import Names

### LINKED ISSUE #2753: "Auto import names" breaks code with imported class as alias
| Subject        | Details                                                               |
| :------------- | :-------------------------------------------------------------------- |
| Rector version | v0.6.13                   |
| PHP version    | PHP 7.4.1                                       |
| Demo link      | https://getrector.org/demo/76fe6abf-1d46-4ef0-b3fd-721fbb3079f2#result  | 

## Current Behaviour 

Replaces aliased class name with original one in places where class name used but not in `use` section.

Details see in demo link.

## Expected Behaviour
Do not replace alias with original class name.

Great-Antique:
**WIP**

Ref  #2753.

Added 4 tests and 3 will fail. I need a help to understand expected result.

1) **class_imported_as_alias.php.inc**
`SecondOriginalClass` was imported but `SecondNamespace` was not removed from `use` section.

2) **class_imported_as_alias_with_namespace_with_false_condition.php.inc**
Passed however `NotExistingNamespace\ClassName` was not exist. Did `if (false) {` condition make it pass?

3) **class_imported_as_alias_without_namespace.php.inc**
Failed because `NotExistingNamespace\ClassName` didn't exist. Looks right.

4) **class_imported_as_alias_without_namespace_with_false_condition.php.inc**
This is the case from issue #2753. Alias replaced by original class name and alias was not removed from `use` section.
And one more question to this case: why it's not the same as in the second case? The only difference is presence of namespace.

### DIFF
--- a/packages/renaming/tests/Rector/Class_/RenameClassRector/FixtureAutoImportNames/class_imported_as_alias.php.inc
+++ b/packages/renaming/tests/Rector/Class_/RenameClassRector/FixtureAutoImportNames/class_imported_as_alias.php.inc
@@ -1,0 +1,23 @@
+<?php
+
+namespace Rector\Renaming\Tests\Rector\Class_\RenameClassRector\Fixture\AutoImportNamesParameter;
+
+use Rector\Renaming\Tests\Rector\Class_\RenameClassRector\Source\FirstNamespace\FirstOriginalClass as AliasedClass;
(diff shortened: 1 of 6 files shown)


TomasVotruba:
This could be delicate issue to fix, so we need to narrow scope to as minimal complexity possible.

First start with the easier snippet you think we can fix

TomasVotruba:
Merging to `master`, so I can work on it

PR #2839, opened 2020-02-11
[BUGFIX] Add cross platform compatible newline character

sabbelasichon:
We had issues with running tests on Windows. After short investigation we fixed this by providing the PHP_EOL constant.


TomasVotruba:
Thank you

PR #2840, opened 2020-02-11
Inject the coveralls token as a secret

ikvasnica:
1. Replace the Coveralls repo token in YAML with an environment variable.
2. Re-generate the token in Coveralls settings, since this one is already leaked online.
3. TODO: Set the new token as a `COVERALLS_REPO_TOKEN` variable in GitHub Actions Settings in the repository.

### DIFF
--- a/.github/workflows/code_coverage.yaml
+++ b/.github/workflows/code_coverage.yaml
@@ -25,4 +25,4 @@ jobs:
                     wget https://github.com/php-coveralls/php-coveralls/releases/download/v2.1.0/php-coveralls.phar
                     php php-coveralls.phar --verbose
                 env:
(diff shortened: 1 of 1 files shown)


TomasVotruba:
I miss reasoning.
What can be changed with the token?

ikvasnica:
From Coveralls documentation: 
> repo_token String
The secret repo token for your repository

I think anyone could send fake coveralls report on behalf of the repository or do some nasty things + it promotes a bad practice, someone might think it is a good idea to put a token into a configuration file and will put something more serious there, which is a big no-no.

TomasVotruba:
Better, thanks. Its always better to see why the change, rather than what has changed.

Often people try to work around some bug that should be fixed in the first place.

TomasVotruba:
Thank you Ivan!

TomasVotruba:
I tried to rebuild the token and it works for now. Let's see how it goes

PR #2784, opened 2020-01-31
feature: add rule to refactor exception methods

alfredbez:
see: https://github.com/sebastianbergmann/phpunit/issues/3775

PR #2843, opened 2020-02-13
add more checks to 'composer complete-check'

alfredbez:
see https://github.com/rectorphp/rector/pull/2784#issuecomment-580890319


TomasVotruba:
Thanks

PR #2848, opened 2020-02-13
add first OXID rector

alfredbez:
We created this initially in https://github.com/OXID-eSales/oxideshop_ce/pull/758, but I think this is useful for module developers as well. We plan to create a few other rectors in the future.

PR #2849, opened 2020-02-13
[Renaming] Add RenameFuncCallToStaticCallRector

TomasVotruba:
Ref #2848


alfredbez:
This would allow something like this if I understand this right:
```yaml
# config/set/oxid/oxid60.yaml
services:
    Rector\Renaming\Rector\FuncCall\RenameFuncCallToStaticCallRector:
        $functionsToStaticCalls:
            getStr:
                'OxidEsales\Eshop\Core\Str': 'getStr'
```

TomasVotruba:
Yes. We should have this months ago

PR #2860, opened 2020-02-14
Make `AnnotateThrowablesRector` continue on unhandled node types.

Aerendir:


Aerendir:
The branch is based on commit fcec6e30 and anyway fails 😓

TomasVotruba:
There is problem with code coverage only, skip it

Aerendir:
Ok, then it is ready to merge.

I simply removed two throws and returned a value anyway.

This way the rector doesn’t stop (it was for dev purposes only) and continues its work.

Later I will start implementing the other node types...

There is the problem of accessing the called function/method doc block, but I will describe the problem extensively in another issue/PR.

TomasVotruba:
Thanks

PR #2880, opened 2020-02-18
Fix PHP notice in ternary to spaceship rector

fsok:
Fixes generated PHP notice `Notice: Undefined property: PhpParser\Node\Expr\Isset_::$left in /home/fsok/src/rector/rules/php-70/src/Rector/Ternary/TernaryToSpaceshipRector.php on line 88`


TomasVotruba:
Thank you

PR #2882, opened 2020-02-18
WIP render zendview via compatibility layer

C0pyR1ght:
As mentioned before I am developing a package to replace Zend. I would like to get back on your offer to open a pull request and get some help.

Goal for this rector is to add `return $this->currentZendViewResult();` to all Zend action methods that do NOT call `$this->setNoRender()`

Example:
````diff
public function getArticleInfoAction()
{
}
public function exampleAction()
{
   $this->setNoRender();
}
---
public function getArticleInfoAction()
{
+   return $this->currentZendViewResult();
}
public function exampleAction()
{
   $this->setNoRender();
}
````

I stumpled upon the following problems:

1. As is created this PR with `bin/rector create` I was not able to execute the suggested unittests `vendor/bin/phpunit rules/zend2-rocket/tests/Rector/ClassMethod/RenderZendViewRector/`. I was wondering if this was just my fault before opening an issue.

````
PHP Warning: require_once(phar://vendor/phpstan/phpstan/phpstan.phar/vendor/jetbrains/phpstorm-stubs/PhpStormStubsMap.php): failed to open stream: phar error: "vendor/jetbrains/phpstorm-stubs/PhpStormStubsMap.php" is not a file in phar "vendor/phpstan/phpstan/phpstan.phar" in /mnt/c/Users/lfeiser/github/rector/tests/bootstrap.php on line 9
PHP Fatal error:  require_once(): Failed opening required 'phar://vendor/phpstan/phpstan/phpstan.phar/vendor/jetbrains/phpstorm-stubs/PhpStormStubsMap.php' (include_path='.:/usr/share/php') in /mnt/c/Users/lfeiser/github/rector/tests/bootstrap.php on line 9

````

2.  I managed so far to build the return statement but I was not able manipulate the the node to add the return statement to the classmethod. My problem is that I am listening for `ClassMethod` nodes to identify wether it is an actionmethod or not. But then I want to manipulate the subnodes.

### CODE REVIEW

File: rector.yaml [hunk 1]
```diff
@@ -23,3 +23,35 @@ parameters:
...
+            {
+                public function getArticleInfoAction()
+                {
+                    return $this->currentZendViewResult();
+                }
+            }
+
+        source: # e.g. link to RFC or headline in upgrade guide, 1 or more in the list
+            - ""
+        set: "Zend2Rocket" # e.g. symfony30, target config to append this rector to
```
TomasVotruba: This should not be parts of `rector.yaml` so it's not changed like `composer.lock`.

Better use local `create-rector.yaml`

rules/zend2-rocket/src/Rector/ClassMethod/RenderZendViewRector.php
```diff
@@ -0,0 +1,81 @@
...
+        $subnodes = $node->getSubNodeNames();
+        if ( in_array('setNoRender', $subnodes)) {
+            return $node;
+        }
+
+        // serve Zendview
+        return $this->refactorZendViewRender($node);
+    }
+
+    private function refactorZendViewRender(Node $node)
```
TomasVotruba:
```suggestion
    private function refactorZendViewRender(ClassMethod $node)
```

rules/zend2-rocket/src/Rector/ClassMethod/RenderZendViewRector.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        // build AST of 'return $this->currentZendViewResult();'
+        $methodcall = $this->createMethodCall('this', 'currentZendViewResult', []);
+        $return = new Return_($methodcall);
+        $node->setAttribute(AttributeKey::PARENT_NODE, $return);
```
TomasVotruba: Return replace class method, it's invalid PHP.

```diff
-public function someMethod()
-{
-}
+return ...;
```

Here should be something like:

```php
$classMethod->stmts = array_merge($classMethod->stmts, [$return]);
```

rules/zend2-rocket/tests/Rector/ClassMethod/RenderZendViewRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,60 @@
+<?php
+
```
TomasVotruba: Always use namespace to keep tests local-scoped

TomasVotruba:
As for 1., failing CI test would be great. We have this failure often reported, but without test case it's blind shooting. E.g. with standalone Github Action

rules/zend2-rocket/tests/Rector/ClassMethod/RenderZendViewRector/Fixture/fixture.php.inc
C0pyR1ght: see https://github.com/rectorphp/rector/pull/2882/commits/d752cbc576d6c2f3daf31184ebd9fd9f39d86f04

rector.yaml
C0pyR1ght: see https://github.com/rectorphp/rector/pull/2882/commits/d752cbc576d6c2f3daf31184ebd9fd9f39d86f04

rules/zend2-rocket/tests/Rector/ClassMethod/RenderZendViewRector/Fixture/fixture.php.inc
C0pyR1ght: The codebase which should be refactored is also global-scoped. Do you recommend to keep the test local-scoped anyways?

TomasVotruba: If that's not crutial for the rule (e.g. in case of Under_Score_Name to namespaced one), then keep tests always namespaced.

create-rector.yaml
```diff
@@ -0,0 +1,33 @@
+parameters:
```
TomasVotruba: This file is ignored by `.gitignore`, it should not be part of resposity. It's only for local development.

C0pyR1ght:
@TomasVotruba is Rector capable of displaying warnings / further instructions on manual steps to do next?

TomasVotruba:
@C0pyR1ght I'm not sure what you mean. Could you give me an example?

rules/zend2-rocket/src/Rector/ClassMethod/RenderZendViewRector.php
```diff
@@ -0,0 +1,81 @@
...
+        // identify if classmethod is an Actionmethod
+        if (! $this->controllerMethodAnalyzer->isAction($node)) {
+            return $node;
+        }
+
+        // check if classmethod calls setNoRender
+        $subnodes = $node->getSubNodeNames();
+        if ( in_array('setNoRender', $subnodes)) {
+            return $node;
+        }
```
C0pyR1ght: @TomasVotruba
this was my first approach of checking if $this->setNoRender(); gets called somewhere within the action-method, but this does not work. Do you know any simple approach like this instead of diving into Walking the AST with PHP-Parser

TomasVotruba: `getSubNodeNames()` is rather internal method that gives you list of properties in the node (in the object, not real properties), e.g. https://github.com/nikic/PHP-Parser/blob/a167aa206198570fc8957f7b3f2c6778b0a0e2ad/lib/PhpParser/Node/Expr/Cast.php#L10-L24

<br>

To find method call inside class method, you can do something like this:

```php
$hasSetNodeRenderMethodCall =$this->betterNodeFinder->findFirst((array) $classMethod->stmts, function (Node $node): bool {
    if (! $node instanceof MethodCall) {
        return false;
	} 

    return $this->isName($node->name, 'setNoRender');
});
```

C0pyR1ght:
> @C0pyR1ght I'm not sure what you mean. Could you give me an example?

There are some cases that cannot handled by rector because of context sensitivity and missing knowledge of the use case. Therefore we agreed on displaying further information for the dev. like: _rector detected an implicit return. look in method x / line x_
Then the dev should decide based on his knowledge and experience if this needs to be changed.

rules/zend2-rocket/src/Rector/ClassMethod/RenderZendViewRector.php
C0pyR1ght: works perfectly in general. but not if the method is empty.

```diff
-    public function getArticleDetailsAction()
+    public function getArticleDetailsAction
+    return $this->currentZendViewResult();()
     {
     }
```

if the method is empty rector wants to modify the name itself

TomasVotruba:
That's rather job for PHPStan, that handles this in much better way.

Rector's goal is to change code and work for user. If Rector would give suggestions and "maybes", it would do PHPStan's job, which is contraproductive for both tools and PHP community itself.

If you need any reporting, I recommend making PHPStan rule. I do it for my project and it works great.

rules/zend2-rocket/src/Rector/ClassMethod/RenderZendViewRector.php
C0pyR1ght: great thanks. didn't heard about the `betterNodeFinder` before.

TomasVotruba:
Please rebase on master to remove conflicts

PR #2883, opened 2020-02-18
[OXID] replace backwards-compatability classes in oxNew

alfredbez:
The OXID Framework has a wrapper for the `new` keyword, which allows loading a class like
```php
oxNew('oxarticle');
```

A few years ago we moved to namespaces
```php
oxNew(\OxidEsales\Eshop\Application\Model\Article::class);
```
... but many modules and projects out there still use the old syntax. This rule makes the transition to the new recommended way a breeze.

---

I plan to add a similar rector for replacing these classes also in docblocks (`@return` and `@param` statements) which should help to get static analysis like phpstan or psalm running. I will extract the class-map to a standalone class to make it usable in multiple rectors.


TomasVotruba:
Nice job :+1:

PR #2881, opened 2020-02-18
Callable type falling tests

snapshotpl:
Falling tests for #2841

ping @TomasVotruba

PR #2884, opened 2020-02-19
Callable type

TomasVotruba:
Covers #2881 

@snapshotpl This is the best I could do. The return `mixed` is removed, as its default value and it would polute all the callables by default.


snapshotpl:
Looks like missing mixed is a problem... https://github.com/phpstan/phpstan/issues/3008

snapshotpl:
@TomasVotruba did you see that ^^ ?

TomasVotruba:
tl;dr; I don't have time for this part anymore, but I can guide you in the possible fix

snapshotpl:
@TomasVotruba waiting for hints :)

TomasVotruba:
You need to detect format of original content in:
https://github.com/rectorphp/rector/blob/3dd1e7ca6929301fe5889d2bb721c3eba426f7d4/packages/attribute-aware-php-doc/src/Ast/Type/AttributeAwareCallableTypeNode.php

Then modify __toString() method to reflect it

Issue #2889, opened 2020-02-19
[WIP] AnnotateThrowables: support static methods calls


Aerendir:
General discussion: https://github.com/rectorphp/rector/issues/2861

... writing

Issue #2861, opened 2020-02-15
[RFC] Improvements suggestions for AnnotateThrowablesRector and problems related to their implementation


Aerendir:
# Summary
Currently the rector `AnnotateThrowablesRector` only supports a direct thrown `\Throwable` like:

```php
throw new \Exception
```

The statement `throw` is analyzed both in methods and in functions.

However, this is not sufficient to cover the full range of concrete possibilities that may lead to the throw of a `\Throwable`.

Following there are the possible kinds of `\Throwables` and the places in which is possible to find them.

# 1. Kinds of `\Throwables` (direct use of `throw`)

There are basically three types of possible thrown `\Throwable`s:

1. Direct class instantiation;
2. Use of a variable;
3. Use of a factory method.

## 1.1 - Direct class instantiation

```php
throw new \Exception();
```

### How to proceed

This is the current covered type of `\Throwable`.

## 1.2 - Use of a variable

*Example from a real world application that uses Guzzle and has to manage its exceptions*

```php
class VariableThrow
{
    /**
     * @param Throwable $e
     *
     * @throws Throwable
     * @throws \InvalidArgumentException
     *
     * @return string
     */
    public static function guessCode(Throwable $e): string
    {
        if ( ! $e instanceof Exception) {
            throw new InvalidArgumentException('Only an \Exception can be passed.');
        }

        // This is thrown if there is an error in connection and relates to problem of the app itself
        if ($e instanceof ConnectException) {
            switch ($e->getHandlerContext()['errno']) {
                case CURLE_OPERATION_TIMEOUTED:
                    return self::ERROR_CODE_OPERATION_TIMEOUT;
                case CURLE_COULDNT_RESOLVE_HOST:
                    return self::ERROR_CODE_COULDNT_RESOLVE_HOST;
                case CURLE_COULDNT_CONNECT:
                    return self::ERROR_CODE_COULDNT_CONNECT;
                case CURLE_SSL_CONNECT_ERROR:
                    return self::ERROR_CODE_SSL_CONNECT_ERROR;
                case CURLE_GOT_NOTHING:
                    return self::ERROR_CODE_GOT_NOTHING;
            }
        } elseif ($e instanceof TooManyRedirectsException) {
            return self::ERROR_CODE_TOO_MANY_REDIRECTS;
        } elseif ($e instanceof ClientException) {
            return self::ERROR_CODE_CLIENT_4XX;
        } elseif ($e instanceof ServerException) {
[...]

PR #2890, opened 2020-02-19
#2861: Support throw of static methods

Aerendir:
#2889 (general discussion: #2861).

# What this should support

```php
function throwThroughStaticMethod()
{
    throw ExcetionFactory::createException();
}
```

# Explanation

See the test case here: https://github.com/rectorphp/rector/blob/71ad810c8c777def532af6ff22be6d4d900b3bdd/rules/coding-style/tests/Rector/Throw_/AnnotateThrowablesRector/Fixture/factory_static_method_with_return_dockblock.php.inc

The problem is to read the Docblock of the invoked method.

One approach, as suggested in #2861  (https://github.com/rectorphp/rector/issues/2861#issuecomment-586970207) is to use the reflection.

Currently it is possible to obtain the Docblock using reflection, but the docblock is obtained as a simple string and it cannot be parsed.

Obtaining a simple string is useless, as it makes impossible to know which classes are actually returned.

```php
    /**
     * This is the DocComment of createExceptionEccolo().
     *
     * @param int $code
     *
     * @return TheException|TheExceptionTheSecond|TheExceptionTheThird|\RuntimeException
     */
```

Is `TheException` something like `Namespace1\TheException` or is it `Namespace2\TheExeception`?

Without being able to access a parsed version of the Docblock, it is impossible to know what is returned by the invoked static method.

@TomasVotruba , WDYT?

### DIFF
--- a/rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
+++ b/rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
@@ -4,9 +4,12 @@ declare(strict_types=1);
 
 namespace Rector\CodingStyle\Rector\Throw_;
 
+use Nette\Utils\Reflection;
+use Nette\Utils\Strings;
 use PhpParser\Node;
 use PhpParser\Node\Expr\New_;
 use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Name\FullyQualified;
 use PhpParser\Node\Stmt\Throw_;
 use PHPStan\PhpDocParser\Ast\PhpDoc\ThrowsTagValueNode;
 use PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode;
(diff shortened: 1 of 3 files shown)


rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -125,14 +125,32 @@ private function identifyThrownThrowables(Throw_ $throw): array
...
+    private function identifyThrownThrowablesInStaticCall(Throw_ $throw): array
     {
+        dumpf(null, $this, $throw);
+        // 1. Getting the DocBlock of the method using reflection
+        $thrownClass = $throw->expr->class;
+        $classFqn = implode('\\', $thrownClass->parts);
+        $methodNode = $thrownClass->getAttribute('nextNode');
+        $methodName = $methodNode->name;
+        $reflectedMethod = new \ReflectionMethod($classFqn, $methodName);
+        $methodDocblock = $reflectedMethod->getDocComment();
```
TomasVotruba: You can do something like this:

https://github.com/symplify/symplify/blob/eb4cfd2e17b564acf8ab3840c9207f470511c1ca/packages/autowire-array-parameter/src/DependencyInjection/CompilerPass/AutowireArrayParameterCompilerPass.php#L170-L186

Aerendir: I think this will work... I'll try...

Aerendir: 🎉It works like a charm! 🎉

I'm locally running the checks and then I'll update the PR.

Aerendir:
@TomasVotruba , ready to be merged! 💪

TomasVotruba:
Great job!

Issue #2879, opened 2020-02-18
rector-prefixed 0.7 throw '...PhpStormStubsMap' errors


shaal:
| Subject        | Details                                                               |
| :------------- | :-------------------------------------------------------------------- |
| Rector version | v0.7                   |
| PHP version    | PHP 7.3.11      |
| Full Command   | vendor/bin/rector process web/modules/contrib/config_filter --set dead-code --dry-run           | 
| Demo link      |  | 
| rector.yaml    | (see below) | 
```
parameters:
  autoload_paths:
    - 'web/core'
    - 'web/core/modules'
    - 'web/modules'
    - 'web/profiles'

  file_extensions:
    - module
    - theme
    - install
    - profile
    - inc
    - engine
```
## Current Behaviour 

Installing Drupal8 and adding rector-prefixed 0.6.13 works as expected
Steps to replicate:
- `composer create-project drupal/recommended-project drupal8`
- `cd drupal8`
- `composer require rector/rector-prefixed:0.6.13 --dev`
- (create the rector.yaml above)
- `vendor/bin/rector process web/core/modules/aggregator/src/Plugin/views/row/Rss.php --set deadcode --dry-run`

The output is correct and it display the diff of berfore/after the rector
```
1 file with changes
===================

1) web/core/modules/aggregator/src/Plugin/views/row/Rss.php

    ---------- begin diff ----------
--- Original
+++ New
@@ -66,14 +66,12 @@
         'attributes' => ['isPermaLink' => 'false'],
       ],
     ];
-
-    $build = [
+    return [
       '#theme' => $this->themeFunctions(),
       '#view' => $this->view,
       '#options' => $this->options,
       '#row' => $item,
     ];
-    return $build;
   }

 }
    ----------- end diff -----------

Applied rules:

 * Rector\CodeQuality\Rector\Return_\SimplifyUselessVariableRector
```

But upgrading to rector-prefix v0.7 is causing problems:
- `composer require rector/rector-prefixed:0.7 --dev`
- `vendor/bin/rector process web/core/modules/aggregator/src/Plugin/views/row/Rss.php --set deadcode --dry-run`

```                                                                                                             
 [ERROR] Could not process "web/core/modules/aggregator/src/Plugin/views/row/Rss.php" file, due to: "Class '_HumbugBox46c80243413c\JetBrains\PHPStormStub\PhpStormStubsMap' not found".                            
                                                                                                                
[...]

PR #2891, opened 2020-02-19
Adding tests for Rector with Drupal

shaal:
Follow up on https://github.com/rectorphp/rector/issues/2879


.github/workflows/test_with_drupal.yaml
```diff
@@ -0,0 +1,34 @@
...
+        steps:
+            -   uses: actions/checkout@v2
+            -   uses: shivammathur/setup-php@v1
+                with:
+                    php-version: 7.3
+                    coverage: none
+
+            -   name: Install Drupal project
+                run: |
+                    # Install latest Drupal version
```
TomasVotruba: `composer install` is needed, so Rector has it's own dependencies too

shaal: @TomasVotruba 
I thought `composer require rector/rector-prefixed:0.6.13 --dev` is what would bring rector-prefixed with all its dependencies...

Locally I run rector-prefixed from `vendor/bin/rector`, but through github actions I got an error.
In the meantime I switched it back to `bin/rector` as it was in the original script.

TomasVotruba: That installs remote version, which will be definitely different from local.

.github/workflows/test_with_drupal.yaml
```diff
@@ -0,0 +1,37 @@
...
+            -   uses: actions/checkout@v2
+            -   uses: shivammathur/setup-php@v1
+                with:
+                    php-version: 7.3
+                    coverage: none
+
+            -   name: Install Drupal project
+                run: |
+                    # bring rector dependencies
+                    composer install --no-progress --no-dev
```
TomasVotruba: The `--no-dev` needs to be removed, otherwise local `phpstan.neon` configs are missing
https://github.com/rectorphp/rector/pull/2891/checks?check_run_id=456122559#step:5:4

TomasVotruba:
The path seems wrong:

image

.github/workflows/test_with_drupal.yaml
```diff
@@ -0,0 +1,37 @@
...
+                    coverage: none
+
+            -   name: Install Drupal project
+                run: |
+                    # bring rector dependencies
+                    composer install --no-progress
+
+                    # Install latest Drupal version
+                    composer create-project drupal/recommended-project drupal
+                    cd drupal
```
TomasVotruba: The path issues might be caused by this line.

`cd ..` is only applied per `run:`, so next `run:` is again in root directory

TomasVotruba: Better have just one long `run:` with all the commands  than think about all the location issues.

shaal:
@TomasVotruba 
In this test, Rector 0.6.13 was running as it should with no issues.
but the test shows as 'fail'.

Can you help me understand what's wrong ?

```
 [WARNING] 1 nodes were removed                                                 

 [OK] Rector is done! 1 file would have changed (dry-run).                      

 3/3 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
##[error]Process completed with exit code 1.
```

TomasVotruba:
Closing as duplicated of https://github.com/rectorphp/rector/pull/2891 that is fixed now

PR #2928, opened 2020-02-25
[WIP] Support throwing functions

Aerendir:
The goal of this PR is to support functions and methods that use functions that may throw exceptions.

```
use function \Safe\sprintf;

class AClass {
    public function aMethod()
    {
        return sprintf('This method may throw an exception');
    }
}
```

# PROBLEM

Currently it is not possible to build the full namespace of the thrown class.

@TomasVotruba , any ideas about how to proceed to get the full namespace of the thrown classes?

Relevant code:

- https://github.com/rectorphp/rector/pull/2928/files#diff-7992343b214da5dce624ce2b1e55384aR262-R292

In the commit there is the dump in HTML format that may be useful.


TomasVotruba:
Why not use the same trick as with docblock and method reflection, like the last time?

Aerendir:
Because it doesn't work for functions: Nette doesn't support building the full namespace for functions, only for classes.

Aerendir:
@TomasVotruba ping

TomasVotruba:
Ah. In that case, you need to solve the behavior yourself with:

```php
$node->getAttribute(AttributeKey::USE_NODES)
```

Foreach it and match the propper name

TomasVotruba:
Here is some inspiration: https://github.com/rectorphp/rector/blob/fa5d62948abe5847b1b9deb516477b5def4316b5/rules/type-declaration/src/PHPStan/Type/ObjectTypeSpecifier.php#L58-L86

Just instead of `ObjectType`, you can use the short string name

Aerendir:
Ok, I will try this way...

composer.json
```diff
@@ -211,6 +211,7 @@
             "rules/cakephp-to-symfony/tests/Rector/Class_/CakePHPModelToDoctrineEntityRector/Source"
         ],
         "files": [
+            "packages/better-php-doc-parser/tests/PhpDocInfo/PhpDocInfoPrinter/AbstractPhpDocInfoPrinterTest.php",
```
TomasVotruba: Why this file? It should be auotloaded by PSR-4

Aerendir:
Too many errors: I will send a clean PR.

composer.json
Aerendir: Because it is a function and functions are not autoloaded by PHP.

Some references about this problem (that took me many days of researches before understanding what was that didn't work 🤬):

- "This is useful if your package includes PHP functions that cannot be autoloaded by PHP." (https://getcomposer.org/doc/04-schema.md#files)
- https://github.com/phpstan/phpstan/issues/95#issuecomment-274257189
- https://github.com/Roave/BetterReflection#limitations
- https://github.com/thecodingmachine/safe/blob/master/composer.json#L13-L99

PR #2952, opened 2020-02-28
ReturnTypeDeclarationRector: add broken test on array indexes (?)

gnutix:
I'm not sure exactly what's happening here ; might have to do with the fact that we access an array index using `[0]` ?

But I'm pretty sure it shouldn't replace `): \stdClass` by `): array` as it will break the code (PS: in my real code it's not \stdClass but some domain object).


TomasVotruba:
Nice test! :+1:

PR #2953, opened 2020-02-28
AssertRegExpRectorTest: add broken test on static method call

gnutix:
Could not reproduce in a demo set however ('guessing because there's no real PHPUnit in there) : https://getrector.org/demo/011085dd-1266-4a33-a8a9-44e355f9f709#result

Fails with error : 

```                                  
         "Pick more specific node than "PhpParser\Node\Expr\StaticCall"".                                                                        
```

I had the exact same error happen from these Rectors on my code base : 

* `Rector\PHPUnit\Rector\SpecificMethod\AssertTrueFalseToSpecificMethodRector`
* `Rector\PHPUnit\Rector\SpecificMethod\AssertFalseStrposToContainsRector`
* `Rector\PHPUnit\Rector\SpecificMethod\AssertPropertyExistsRector`

I guess the issue is the same, so I won't try to add tests for them all. We'll see later once a fix comes out if they still fail.

### DIFF
--- a/rules/phpunit/tests/Rector/SpecificMethod/AssertRegExpRector/Fixture/fixture_2.php.inc
+++ b/rules/phpunit/tests/Rector/SpecificMethod/AssertRegExpRector/Fixture/fixture_2.php.inc
@@ -1,0 +1,13 @@
+<?php
+
+namespace Rector\PHPUnit\Tests\Rector\SpecificMethod\AssertRegExpRector\Fixture;
+
+use PHPUnit\Framework\TestCase;
+
+final class Fixture2Test extends TestCase
+{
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Same as previous PR

gnutix:
There is still the issue on my code base with these two : 

- Rector\PHPUnit\Rector\SpecificMethod\AssertFalseStrposToContainsRector
- Rector\PHPUnit\Rector\SpecificMethod\AssertPropertyExistsRector

TomasVotruba:
Possibly similar solution might be applied. Look at stack trace, before going into `getName()`

PR #2954, opened 2020-02-28
RemoveDefaultArgumentValueRector: add broken test on static method call.

gnutix:
Might be related to https://github.com/rectorphp/rector/pull/2953 as it's the same error, but it might not (as the Rector is quite different).

Fails with error : 

```                                  
         "Pick more specific node than "PhpParser\Node\Expr\StaticCall"".                                                                        
```


TomasVotruba:
Same as previous PR

Issue #2955, opened 2020-02-28
rector-prefixed adds _HumbugBox prefix to FullyQualified


alfredbez:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | v0.7.2              |
| Installed as   | prefixed Rector PHAR                      |

In #2883 we added a rector to replace some deprecated classes with a namespaced one. It works if I install rector directly, but it adds the `HumbugBox`-prefix to the namespace when I use the prefixed-version of rector.

## Minimal PHP Code Causing Issue

Since the set is not available on https://getrector.org/demo/ yet, I added a small repo, to reproduce the issue, check this out: https://github.com/alfredbez/rector-issue
After running `composer install` you can run the set like:
```
vendor/bin/rector process src --set oxid60 --dry-run
```

## Actual Output

![image](https://user-images.githubusercontent.com/1001186/75565140-bc847380-5a4d-11ea-864c-0e6d2554cc1f.png)

or as text:
```
$ vendor/bin/rector process src --set oxid60 --dry-run
Rector v0.7.2

 3/3 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

1 file with changes
===================

1) src/Core/Foo.php

    ---------- begin diff ----------
--- Original
+++ New
@@ -6,6 +6,6 @@
 {
     public function makeSomethingWithCategories()
     {
-        $category = oxNew('oxcategory');
+        $category = oxNew(\_HumbugBoxaf3b4563b2b8\OxidEsales\Eshop\Application\Model\Category::class);
     }
 }
    ----------- end diff -----------

Applied rules:

 * Rector\Oxid\Rector\FuncCall\OxidReplaceBackwardsCompatabilityClassRector


                                                                                                                        
 [OK] Rector is done! 1 file would have changed (dry-run).                                                              
                                                                                                                        
```

## Expected Behaviour

I didn't expect to see the prefix `\_HumbugBoxaf3b4563b2b8` in front of `\OxidEsales\Eshop\Application\Model\Category`.

---

I'm not sure if this is related only to this Rector.

alfredbez:
Wow, that was super fast, thanks a lot!

PR #2956, opened 2020-02-28
AssertTrueFalseInternalTypeToSpecificMethodRector: add broken test method call

gnutix:
Might be related to #2954 and #2955. They all come from `Rector\NodeNameResolver\NodeNameResolver->getName()`. I won't open any more issues like this one for now as to not spam you if it all comes from the same source ; this should give you enough of a sample to work with. ;) 

Fails with error : 

```
         "Pick more specific node than "PhpParser\Node\Expr\MethodCall"".     
```


TomasVotruba:
If you run it with --debug option, it will show you where the getName() is called from.

Instead of getName($node), there should be something like getName($node->name)

TomasVotruba:
Miss-click

PR #2957, opened 2020-02-28
comparing results of rector 0.6.13 vs 0.72

shaal:


TomasVotruba:
Thanks for this demo! I was able to simulate it locally very quickly

PR #2986, opened 2020-03-02
Typed callable when mixed used

snapshotpl:
Covers #2881 and #2884

### DIFF
--- a/packages/better-php-doc-parser/tests/PhpDocInfo/PhpDocInfoPrinter/FixtureCallable/callable_mixed.txt
+++ b/packages/better-php-doc-parser/tests/PhpDocInfo/PhpDocInfoPrinter/FixtureCallable/callable_mixed.txt
@@ -1,1 +1,1 @@
-/** @var callable(array<mixed>) */
+/** @var callable(array<mixed>):mixed */
--- a/packages/attribute-aware-php-doc/src/Ast/Type/AttributeAwareCallableTypeNode.php
+++ b/packages/attribute-aware-php-doc/src/Ast/Type/AttributeAwareCallableTypeNode.php
@@ -34,7 +34,7 @@ final class AttributeAwareCallableTypeNode extends CallableTypeNode implements A
         }
 
         $parameterTypeString = $this->normalizeParameterType($parameterTypeString, $returnTypeAsString);
-        $returnTypeAsString = $this->normalizeReturnType($returnTypeAsString);
+        $returnTypeAsString = $this->normalizeReturnType($parameterTypeString, $returnTypeAsString);
 
         return sprintf('%s%s%s', $this->identifier->name, $parameterTypeString, $returnTypeAsString);
     }
@@ -51,9 +51,9 @@ final class AttributeAwareCallableTypeNode extends CallableTypeNode implements A
         return trim($parameterTypeString);
     }
 
-    private function normalizeReturnType(string $returnTypeAsString): string
+    private function normalizeReturnType(string $parameterTypeString, string $returnTypeAsString): string
     {
-        if ($returnTypeAsString === 'mixed') {
+        if ($returnTypeAsString === 'mixed' && $parameterTypeString === '') {
             return '';
         }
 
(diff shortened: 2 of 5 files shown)


packages/better-php-doc-parser/tests/PhpDocInfo/PhpDocInfoPrinter/FixtureCallable/callable_mixed.txt
```diff
@@ -1 +1 @@
-/** @var callable(array<mixed>) */
+/** @var callable(array<mixed>):mixed */
```
TomasVotruba: Please keep these 2 and add 2 new extra files, so it's clear there is no change in behavior

snapshotpl: But this behavior it's what we need. We need to disallow typed callable without return type via https://github.com/phpstan/phpstan/issues/3008

TomasVotruba: How would you explain people a rule that changes their phpdocs?

snapshotpl: IMO `mixed` means same `no return type`. So this change just improve `callable` definition.

Currently rector changes my phpdocs by remove mixed, so who would you explain me this change ;-)

TomasVotruba: That's just moving problem from to right.
The solution is to respect both cases.

snapshotpl: This change was introduced there https://github.com/rectorphp/rector/pull/2884 right?

snapshotpl: @ondrejmirtes maybe you can add something to discussion? Your voice is important because currently this trigger error in phpstan https://github.com/phpstan/phpstan/issues/3008

TomasVotruba: @snapshotpl Probably. I have focused on many issues since.

ondrejmirtes: `callable(X)` without return typehint in PHPDoc was never supported, because it's ambiguous for the AST parser. Better to use the `callable(X): mixed` syntax.

Is it `IdentifierTypeNode` with `(X)` description, or is it `CallableTypeNode`? There's no way to tell. I added some tests with the current behaviour based on the issue https://github.com/phpstan/phpdoc-parser/commit/e2b89c5119b378518c259ea4d5ff16c662281f84

snapshotpl: @TomasVotruba what are you thing now?

This change is introduced by my request, so probably nobody cares about "revert" this behavior.

If it's BC break you can start with 0.8, but IMO callable, if typed, should always contains return type.

TomasVotruba: Format must be preserved. It's the most discouraging to see no change yet 200 changed files

snapshotpl:
@TomasVotruba which rector should I disable to prevent this change?

TomasVotruba:
That's not possible, It's part of the printer.

snapshotpl:
So I'm blocked to 0.7 upgrade...

TomasVotruba:
Until it's fixed, yes

snapshotpl:
Any plan to move it forward? IMO This change is fine, however I see you have different opinion.

TomasVotruba:
Anything that preserves format is ok

TomasVotruba:
On second thought, let's give it a try and wait for community feedback.

Thank you for your work :+1:

snapshotpl:
:beers:

PR #2990, opened 2020-03-03
AssertTrueFalseToSpecificMethodRector: add broken test ('Pick more specific node than PhpParser\Node\Expr\StaticCall')

gnutix:
Same as https://github.com/rectorphp/rector/pull/2953.

### DIFF
--- a/rules/phpunit/src/Rector/SpecificMethod/AssertTrueFalseToSpecificMethodRector.php
+++ b/rules/phpunit/src/Rector/SpecificMethod/AssertTrueFalseToSpecificMethodRector.php
@@ -73,7 +73,9 @@ final class AssertTrueFalseToSpecificMethodRector extends AbstractPHPUnitRector
         }
 
         $firstArgumentValue = $node->args[0]->value;
-        if (! $this->isNames($firstArgumentValue, array_keys(self::OLD_TO_NEW_METHODS))) {
+        if ($firstArgumentValue instanceof StaticCall ||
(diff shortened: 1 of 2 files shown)


TomasVotruba:
If you look at stack trace, you see the problem is here: https://github.com/rectorphp/rector/blob/3e284b05111a67f1b7afbc0c648526a66098a04c/rules/phpunit/src/Rector/SpecificMethod/AssertTrueFalseToSpecificMethodRector.php#L76

The value must be `$node->name`, not `$node`. Try it

gnutix:
I did not understood your comment. But looking at the code of this Rector, it makes no sense to apply it on StaticCall methods, as it seems to be meant to replace PHP functions calls to PHPUnit specific assert methods.

rules/phpunit/src/Rector/SpecificMethod/AssertTrueFalseToSpecificMethodRector.php
```diff
@@ -73,7 +73,9 @@ public function refactor(Node $node): ?Node
         }
 
         $firstArgumentValue = $node->args[0]->value;
-        if (! $this->isNames($firstArgumentValue, array_keys(self::OLD_TO_NEW_METHODS))) {
+        if ($firstArgumentValue instanceof StaticCall ||
```
TomasVotruba: What about `! FuncCall` and early `return null;`?

PR #2993, opened 2020-03-03
AddArrayReturnDocTypeRector: add broken test on Assert::resource().

gnutix:
Rector now changes a return type `@return resource` to `@return resource|bool`, even though the code calls `Assert::resource()` to ensure it cannot be returning a boolean.

### DIFF
--- a/rules/type-declaration/tests/Rector/ClassMethod/AddArrayReturnDocTypeRector/Fixture/skip_assert.php.inc
+++ b/rules/type-declaration/tests/Rector/ClassMethod/AddArrayReturnDocTypeRector/Fixture/skip_assert.php.inc
@@ -1,0 +1,20 @@
+<?php
+
+namespace Rector\TypeDeclaration\Tests\Rector\ClassMethod\AddArrayReturnDocTypeRector\Fixture;
+
+use Webmozart\Assert\Assert;
+
+final class SkipAssert
+{
+    /**
+     * @return resource
+     */
+    public function stream()
+    {
+        $filePath = tempnam(sys_get_temp_dir(), 'test');
+        $stream = fopen($filePath, 'rb');
+        Assert::resource($stream, 'stream', sprintf('Could not stream content of local file: %s', $filePath));
+
+        return $stream;
+    }
+}


TomasVotruba:
This is false positive of PHPStan.

How would you handle it technically?

gnutix:
No idea... Maybe it should be reported back to PhpStan's repository then?

TomasVotruba:
I don't think such feature will be implemented any time soon there. It's pretty complex to follow workflow.

gnutix:
Hello @ondrejmirtes! Would you have any idea why this happens now (a "PHPStan false positive" apparently) ?

ondrejmirtes:
@gnutix Please open an issue in PHPStan - follow the Bug report template.

gnutix:
@ondrejmirtes Thing is, I don't really understand how PHPStan is used under the hood by Rector in this case and what makes it encounter a false positive. I guess @TomasVotruba would be more knowledgeable creating the issue and following it.

ondrejmirtes:
I've now checked the source code and I think you should:

1) Start analyzing your source code with PHPStan.
2) Install phpstan/phpstan-webmozart-assert extension (https://github.com/phpstan/phpstan-webmozart-assert) that makes PHPStan understand this code. I think that Rector somehow discovers your phpstan.neon (but I'm not sure if it works with phpstan/extension-installer, you might need to install the extension manually, see its README).

gnutix:
@ondrejmirtes I'm not sure reading you if you're talking about "me and my project", or "Tomas and Rector's code".

On my side, I am already using both PHPStan and webmozart extension. The issue here lies in when I'm running Rector, it tries to improve the return types of my methods (by using PHPStan's code internally to detect the types) and doesn't seem to understand the Assert call from webmozart (and probably as you stated, it's not using the webmozart extension internally).

ondrejmirtes:
@gnutix I'm talking about you. You should make sure that Rector discovers your PHPStan config. Try to uninstall extension-installer and include the extension manually. If it works, you should then open an issue so that Rector also works with phpstan/extension-installer.

gnutix:
I can't put that kind of time in that now. And as you might remember from the various issues I've opened on Rector/PHPStan, my project setup is still on PHP 7.2 with Docker containers and PHAR archives, and it's quite a mess to get it all working together. I'd rather wait for our upcoming infrastructure upgrade (PHP 7.4+) and try again then (if everything is installed via composer and in the project's root, it might just work then).

However, from Rector's point of view, I still think there's something wrong here ; this test case I've added should pass, independently of how my project is setup. Don't you think ? @TomasVotruba

ondrejmirtes:
It can't pass because Rector uses PHPStan's type inference for this, and PHPStan needs the webmozart/assert extension.

gnutix:
I see. And the way Assert works by triggering Exceptions on specific conditions cannot be interpreted by PHPStan nor Rector as would any other type of code - it's something that has to be "known". Let's drop it then.

PR #2998, opened 2020-03-03
AnnotateThrowables: support analysis of called functions and methods

Aerendir:
Was https://github.com/rectorphp/rector/pull/2928

Support
- [x] functions;
- [x] `$varName->method()`;
- [x] `$methodParam->method()`;
- [x] `$this->method()`;


Aerendir:
@TomasVotruba , Ready to be merged!

Before merging, please check this code:
https://github.com/rectorphp/rector/pull/2998/files#diff-7992343b214da5dce624ce2b1e55384aR367-R388

Aerendir:
Do not merge: some work still required...

I'm testing the Rector in an app and it causes exceptions...

TomasVotruba:
I'm waiting, no troubles

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -252,4 +338,55 @@ private function getThrowingStmtPhpDocInfo(Throw_ $throw): PhpDocInfo
...
+            }
+
+            $parts = $use->name->parts;
+            $uses[$parts[count($parts) - 1]] = implode('\\', $parts);
+        }
+
+        return $uses;
+    }
+
+    private function parseFunction(ReflectionFunction $reflectionFunction)
```
TomasVotruba: You should be albe to get parser service in constructor, no need to create it.

Also, never use `static $something`, its hidden fake property.

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -252,4 +338,55 @@ private function getThrowingStmtPhpDocInfo(Throw_ $throw): PhpDocInfo
...
+            return [];
+        }
+
+        $ast = $parser->parse($functionCode)[0];
+
+        if (! $ast instanceof Namespace_) {
+            return false;
+        }
+
+        return $ast;
```
TomasVotruba: Better return `null` instead of `false`, so return type can by strict, e.g. `?Type`

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: Ok, I will try...

In the mean time I'm refactoring the entire class as now it has a lot of repetitive code and doesn't pass the checks due to cognitive complexity.

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: I don't know anymore... I'm refactoring now... I will inspect this after the refactoring.

Aerendir:
Needs refactoring and checks to pass.

Aerendir:
@TomasVotruba , completely refactored and ready to be merged! 💪

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -33,19 +42,34 @@ final class AnnotateThrowablesRector extends AbstractRector
     /**
      * @var string
      */
-    private const RETURN_DOCBLOCK_TAG_REGEX = '#@return[ a-zA-Z0-9\|\\\t]+#';
+    private const RETURN_DOCBLOCK_TAG_REGEX = '#@return[ a-zA-Z0-9-_\|\\\t]+#';
+
+    /**
+     * @var string
```
TomasVotruba: Could you add http://regex101.com/ link? Just to make future refactoring easier

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -93,59 +117,77 @@ public function throwException(int $code)
         );
     }
 
-    /**
-     * @param Throw_ $node
-     */
     public function refactor(Node $node): ?Node
```
TomasVotruba: There should be same nodes as in `getNodeTypes()` method now

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -93,59 +117,77 @@ public function throwException(int $code)
...
+        return null;
     }
 
-    private function isThrowableAnnotated(Throw_ $throw): bool
+    private function hasThrowablesToAnnotate(Node $node): bool
     {
-        $phpDocInfo = $this->getThrowingStmtPhpDocInfo($throw);
-        $identifiedThrownThrowables = $this->identifyThrownThrowables($throw);
+        if ($node instanceof Throw_) {
+            $this->analyzeStmtThrow($node);
```
TomasVotruba: This method should return number of found throws, instead of using property that is filled by this method.

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -93,59 +117,77 @@ public function throwException(int $code)
...
-    private function identifyThrownThrowables(Throw_ $throw): array
+    private function analyzeStmtThrow(Throw_ $throw): void
     {
-        if ($throw->expr instanceof New_) {
-            return [$this->getName($throw->expr->class)];
+        $foundThrownThrowables = [];
+
+        // throw new \Throwable
+        if ($throw->expr instanceof New_ && $this->getName($throw->expr->class) !== null) {
+            $foundThrownThrowables[] = $this->getName($throw->expr->class);
```
TomasVotruba: Re-use a variable for duplicated `$this->getName($throw->expr->class)`.
It's convention to prevent multiple calls of the same code

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -93,59 +117,77 @@ public function throwException(int $code)
...
+            $foundThrownThrowables = $this->identifyThrownThrowablesInStaticCall($throw->expr);
         }
 
         if ($throw->expr instanceof MethodCall) {
-            return $this->identifyThrownThrowablesInMethodCall($throw->expr);
+            $foundThrownThrowables = $this->identifyThrownThrowablesInMethodCall($throw->expr);
         }
 
-        return [];
+        $alreadyAnnotatedThrowables = $this->extractAlreadyAnnotatedThrowables($throw);
```
TomasVotruba: :+1:

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -93,59 +117,77 @@ public function throwException(int $code)
...
+
+    private function analyzeStmtFuncCall(FuncCall $funcCall): void
+    {
+        $name = $funcCall->name;
+        if (! $name instanceof FullyQualified) {
+            return;
+        }
+        $foundThrownThrowables = $this->extractFunctionAnnotatedThrows($name);
+        $alreadyAnnotatedThrowables = $this->extractAlreadyAnnotatedThrowables($funcCall);
+        $this->compareThrowables($foundThrownThrowables, $alreadyAnnotatedThrowables);
```
TomasVotruba: I don't understand what `compareThrowables` really means. 
Could you think of better name?

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -173,83 +211,210 @@ private function identifyThrownThrowablesInStaticCall(StaticCall $staticCall): a
...
-        $classFqn = implode('\\', $thrownClass->parts);
-        $methodNode = $thrownClass->getAttribute('nextNode');
-        $methodName = $methodNode->name;
 
-        return $this->extractMethodReturnsFromDocblock($classFqn, $methodName);
+        return $this->extractMethodReturns($thrownClass, $thrownClass->getAttribute('nextNode'));
     }
 
-    private function extractMethodReturnsFromDocblock(string $classFqn, string $methodName): array
+    private function extractMethodReturns(FullyQualified $fullyQualified, $methodNode): array
```
TomasVotruba: `$methodNode` is missing type declaration

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -173,83 +211,210 @@ private function identifyThrownThrowablesInStaticCall(StaticCall $staticCall): a
...
     {
+        $classFqn = implode('\\', $fullyQualified->parts);
```
TomasVotruba: `$this->getName($fullyQualified)` should work

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -173,83 +211,210 @@ private function identifyThrownThrowablesInStaticCall(StaticCall $staticCall): a
...
             $returnClasses[] = Reflection::expandClassName($returnTag, $reflectedMethod->getDeclaringClass());
         }
 
-        $this->foundThrownClasses = $returnClasses;
-
         return $returnClasses;
     }
 
-    private function annotateThrowable(Throw_ $node): void
+    private function extractFunctionAnnotatedThrows(FullyQualified $fullyQualified): array
```
TomasVotruba: This method could be extracted to standalone service,
e.g. FunctionThrowsAnnotationResolver

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -173,83 +211,210 @@ private function identifyThrownThrowablesInStaticCall(StaticCall $staticCall): a
...
+        switch ($tagName) {
+            case '@throws':
+                $regEx = self::THROWS_DOCBLOCK_TAG_REGEX;
+                break;
+            case '@return':
+                $regEx = self::RETURN_DOCBLOCK_TAG_REGEX;
+                break;
+            default:
+                throw new ShouldNotHappenException();
+        }
```
TomasVotruba: This could be standalone method with return : string

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -173,83 +211,210 @@ private function identifyThrownThrowablesInStaticCall(StaticCall $staticCall): a
...
+        $uses = $this->getUses($functionNode);
+
+        $expandedClasses = [];
+        foreach ($classNames as $className) {
+            $shortClassName = $this->getShortName($className);
+            $expandedClasses[] = $uses[$shortClassName] ?? $className;
+        }
+
+        return $expandedClasses;
+    }
```
TomasVotruba: This has potential for standalone service, e.g. FunctionAnnotationExpander

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: Do you mean a link?

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: Yes, and it isn’t the only one 😅

Which namespace should have I to use?

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: Namespace?

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: Namespace?

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
TomasVotruba: Maybe `AnnotationReflection`

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
TomasVotruba: It's small method, so this file

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
TomasVotruba: `AnnotationReflection`

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: In which folder do I put the file?

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: In which folder do I put the file?

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: Done

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: Done

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: Done

Aerendir: It is not possible to reuse the variable without writing many lines of code.

`$this->getName($throw->expr->class)` works only if `$throw->expr instanceof New_`.

The code will become unreasonably long just to reuse a variable.

If `! $throw->expr instanceof New_`. an error occurs:

> Undefined property: PhpParser\Node\Expr\MethodCall::$class

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: I called it `diffThrowables()` and make it use `array_diff()` instead of the `foreach`: faster and more elegant.

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: Yes, I forgot it. Added.

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: Yes, it will.

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
TomasVotruba: With PSR-4, namespace == directory

so `src/Directory/...`

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
TomasVotruba: Not sure what you mean, for me it's clear:

```php
if ($throw->expr instanceof New_) {
    $name = $this->getName($throw->expr->class);
    // ...
}
```

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -421,6 +427,14 @@ private function reflectionFunctionGetNode(ReflectionFunction $reflectionFunctio
         return $ast;
     }
 
+    private function getClassFromMethodCall(MethodCall $methodCall):?FullyQualified
+    {
+        // $variable = new Class()
+        $class = $methodCall->getAttribute('previousExpression')->expr->expr->class;
+
+        return $class instanceof FullyQualified ? $class : null;
+    }
```
TomasVotruba: This looks very magical and unrealiable... What is the purpose?

```php
$staticType = $this->getStaticType($variableNode)
if ($staticType instanceof ObjectType) {
    // ...
}
```

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -43,13 +43,15 @@ final class AnnotateThrowablesRector extends AbstractRector
 {
     /**
      * @var string
+     * @link https://regex101.com/r/oEiq3y/3
      */
-    private const RETURN_DOCBLOCK_TAG_REGEX = '#@return[ a-zA-Z0-9-_\|\\\t]+#';
+    private const RETURN_DOCBLOCK_TAG_REGEX = '#@return[ a-zA-Z0-9_\|\\\t]+#';
```
TomasVotruba: Purfect :clap:

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
```diff
@@ -121,6 +121,11 @@ public function throwException(int $code)
         );
     }
 
+    /**
+     * @param Node|Throw_|MethodCall|FuncCall $node
+     *
+     * @return Node|null
```
TomasVotruba: Duplicated information, can be removed

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: Yes, but maybe it is something forgotten from the first version... Anyway, now I removed it...

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: Yes, it is not checked with style fixes... It is just to make you able to read the new code...

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
TomasVotruba: No troubles. Maybe I was too eager to review :D 

Just ping me next time, when you'll need my feedback.

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
TomasVotruba: Same answer https://github.com/rectorphp/rector/pull/2998#discussion_r388158583

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: I wrote this this way because in `refactor()` method it is more readable.

It sounds like `if hasThrowablesToAnnotate then annotateThrowables`.

The method name is a question and implies a true/false answer: "does this node have throwables to annotate? Yes or no?".

If I return the number of found throwables to annotate, then I need to change the name to something like `foundThrowablesToAnnotate` and then, in `refactor()`, change the `if` to something like `if (0 < $this->foundThrowablesToAnnotate()`.

but I think this way of writing code is less discursive and I think that when the code is discursive is easier to be read.

But if you think this is really important I will refactor the methods.

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
Aerendir: I don't like, but if you think it is important, I will refactor this way:

```php
        if ($throw->expr instanceof New_) {
            $className = $this->getName($throw->expr->class);
            if (null !== $className) {
                $foundThrownThrowables[] = $this->getName($throw->expr->class);
            }
        }
```

Aerendir:
I did a very deep refactoring.

I creating some classes and put them where I thought was appropriate and giving them a name that I think may be appropriate.

Please, review the code as I don't know if you prefer different namings/namespacings/etc...

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
TomasVotruba: * How would you test this method if decoupled to service?

It's better to use input/output way of coding that shows the result. Then work with result explicitly in this flow of code. Now I have to go inside the method, to figure out what "analyse" really does.

TomasVotruba: It might sound I'm making big deal of it. But other contributors might read your code and copy-paste the pattern to their rules, which would eventually lead to hard to maintain legacy code.

Aerendir: Reading again I’ve just understood you are speaking about analyze*() methods and not abou HasThrowablesToAnnotate() 😓

I will change them tomorrow...

rules/coding-style/src/Rector/Throw_/AnnotateThrowablesRector.php
TomasVotruba: That's how the rest of the code base is written. Do you have idea how to make it better, explicit and robust? I'm open to learning.

<br>

Better use early return:

```php
if (null === $className) {
    continue;
}
```

TomasVotruba: Also, I meant this way:

```diff
$className = $this->getName($throw->expr->class);
            if (null !== $className) {
-                $foundThrownThrowables[] = $this->getName($throw->expr->class);
+                $foundThrownThrowables[] = $className;
            }
```

TomasVotruba:
Looks good :+1:

TomasVotruba:
Thank you :clap:

PR #3001, opened 2020-03-04
comparing tests with drupal metatag

shaal:


TomasVotruba:
If it tests prefixed phar, it must be targeted at https://github.com/rectorphp/rector-prefixed

And use the rector version there

TomasVotruba:
I can't recall what is this about. 
I don't think we should external packages here. It should be part of yours CI

Issue #2994, opened 2020-03-03
DocBlocks are unnecessarily modified


localheinz:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | v0.7.3              |
| Installed as   | composer dependency                      |

After updating `rector/rector` from https://github.com/rectorphp/rector/compare/v0.6.13...v0.7.3, annotations in DocBlocks are unnecessarily modified

```diff
     }

     /**
-     * @return \Generator<array{0: null|string}>
+     * @return \Generator<array{0:null|string}>
      */
     public static function company(): \Generator
     {
```

or

```diff
 final class FooAction
 {
     /**
-     * @Route(
-     *     path="/foo/bar.{_format}",
-     *     name=BarAction::class,
-     *     methods="GET",
-     *     defaults={
-     *         "_format": "jsonld",
-     *         "_api_resource_class": Foo::class,
-     *         "_api_collection_operation_name": "foo",
-     *         "_api_receive": false,
-     *     }
-     * )
+     * @Route(     path="/foo/bar.{_format}",     name="App\Action\BarAction",     methods={"GET"},         defaults={"_format":"jsonld", "_api_resource_class":"App\Entity\Foo",         "_api_collection_operation_name":"foo", "_api_receive":false}         )
      */
     public function __invoke(Request $request): JsonResponse
     {
```

## Minimal PHP Code Causing Issue

Tried to reproduce on https://getrector.org/demo/c62d2af4-c14b-4761-9880-b8e2701efee6#result, but the result there is different from what we observe locally.

```diff
     }

     /**
-     * @return \Generator<array{0: null|string}>
+     * @return \Generator<array{0: (null | string)}>
      */
     public static function email(): \Generator
     {
```

Also see https://getrector.org/demo/009c98cd-0e4b-4a44-a811-622640f2a4ae#result.

❓ Perhaps it would make sense to indicate which version of `rector/rector` is available at https://getrector.org/demo?

## Expected Behaviour

Since no rules have been specified that indicate that DocBlocks should be touched, DocBlocks should not be touched.

Manuel-Linkener:
Same issue here. rector changes every method docblock in all classes, which make it really hard to review. A configuration to disable those changes (code style fixes in general) would be much appricated

TomasVotruba:
This bugs needs to be fixed instead.

[...]

Issue #2991, opened 2020-03-03
invalid phpdoc should not be changed


C0pyR1ght:
rector changes invalid phpdoc annotations instead of leaving them as they are.

This behavior seems to occur on serveral / all rectors.

**expected**: `@return multitype:number` should remain the same
````
	/**
    * count artfarbenids in aset of articles (used in basket and stock)
    *
    * @param unknown $array
    * @return multitype:number
    */
````

see: https://getrector.org/demo/ebb0f613-a6df-4231-9aa2-a43dc22aeea4

TomasVotruba:
Fixed

PR #3011, opened 2020-03-09
WIP: Testcase for unnecessary PHPdoc modification

C0pyR1ght:
This PR tries to provide Testcases for https://github.com/rectorphp/rector/issues/2994 and https://github.com/rectorphp/rector/issues/2991


TomasVotruba:
Better use explicit names... e.g. annotation_with_value_surrounded_by_space

C0pyR1ght:
I just picked a random one of your existing test cases and added the spaces. Rector will always change the phpdoc if it is not exactly in the required syntax. This applies to braces, spaces, pipe etc

TomasVotruba:
Do you think that's all covered from issues there?

C0pyR1ght:
I suppose that all these different unnecessary doc changes come from one single source of failure. This thesis is also supported by the fact that this bug seems appear on all cases that we pointed out in the issues between https://github.com/rectorphp/rector/compare/v0.6.13...v0.7.3

C0pyR1ght:
I think the next step would be to further investigate on the specific version this bug first occured

staabm:
I would suggest to `git bisect` to discover the offending commit

C0pyR1ght:
> I would suggest to `git bisect` to discover the offending commit

great idea. I will try that tomorrow

TomasVotruba:
The doc printing was added for each node in https://github.com/rectorphp/rector/pull/2775/files

That lead to exposition of such bugs, where format preserving is not finished yet. Now the goal is to cover these edge cases, so format preserving is close to reality as possible. Saying that, each php doc node has to check for possible spaces and try to deal with them.

C0pyR1ght:
In this case i have to correct my assumption that this testcase is representative for everything. If your solution this is to ignore spaces we have to come up with further cases for other characters like `()`

TomasVotruba:
The goal is to preserve format as much as possible, without going crazy :D

C0pyR1ght:
Okay, then I will just look into the cases mentioned in the issues

C0pyR1ght:
These three should be it. The breaking characters are spcaes, linebreaks and round brackets. Some other Characters I tried, had no negative impact.

TomasVotruba:
Thank you!

Would you take care of it with my guidance?

TomasVotruba:
Could you rebase on current master? I wonder what has changed now

C0pyR1ght:
had to open a new PR because i deleted the fork in the meantime
see https://github.com/rectorphp/rector/pull/3126

PR #3021, opened 2020-03-15
add phpunit 9 rector to convert non-strict assertContains

nightlinus:
As of https://github.com/sebastianbergmann/phpunit/issues/3426 assertContains() and assertNotContains() will perform strict comparisons starting with PHPUnit 9 where non-strict comparisons were performed in PHPUnit 8 and earlier; assertContainsEqual() and assertNotContainsEqual() should be used instead if needed non-strict comparison


TomasVotruba:
Nice job, thanks

Issue #2743, opened 2020-01-22
[Rector CI] Collect feedback from cakephp dead-code, fix edge cases, make useful


TomasVotruba:
See https://github.com/cakephp/cakephp/pull/14160

## Add `project_type` parameter - covered with https://github.com/rectorphp/rector/pull/2933 :heavy_check_mark: 

As for https://github.com/rectorphp/rector/pull/2926, maybe add parameter:

```yaml
parameters:
    project_type: proprietary # or "open_source"
```

First allows removing code from internal abstract classes or interface,
latter does not, because somebody else might be using it

TomasVotruba:
Closing as low priority now

PR #3031, opened 2020-03-18
fix asterisk indent

TomasVotruba:
Ref #2743 

Fixes 

image


TomasVotruba:
Without asterisk, it's probably bug in phpdoc-parser, that resolve the last constant `WHEN_UPDATE` indent the same way as previous one. Which is incorrect.

image

If you change the doc with `[]` → `['a']`, it works correctly:

image


<br>

This is the best hot fix I could think of ATM

PR #3036, opened 2020-03-19
Run cs after rector ci

JanMikes:


JanMikes:
@TomasVotruba it is now divided.

it creates 1st commit with rector changes, then run CS and if cs failed it will create another commit with different commit message (same prefix `fixup! [rector ci]`)

please let me know if you like this workflow and feel free to merge

TomasVotruba:
Great, that's exactly what I needed!
Thanks!

PR #3040, opened 2020-03-21
Proofread docs

greg0ire:


docs/HowItWorks.md
```diff
@@ -4,13 +4,15 @@
 
 ## 1. Finds all files and Load Configured Rectors
 
-- The application finds files in source you provide and registered Rectors - from `--set`, `--config` or local `rector.yaml`
+- The application finds files in the source code you provide and registered
+Rectors - from `--set`, `--config` or local `rector.yaml`
 - Then it iterates all found files and applies relevant Rectors to them.
```
TomasVotruba: Avoid line breaks as it mall-formats on various dimensions (pc, smart phone mobile...). Let browser handle it.

greg0ire: The line breaks are not taken into account while displaying the rendered html README, you can check that by clicking "Display the rich diff"

TomasVotruba: What is the reason then?

It's very hard to read and add extra work to control spaces everytime case content changes.

greg0ire: The reasons are making diffs easier to read, reducing the risk of conflicts, allow displaying the file side by side with another. Actually I wrote an entire blog post about that: https://dev.to/greg0ire/why-i-keep-pestering-everyone-about-long-lines
Also, it can spare you from having to scroll horizontally when reviewing (there is no wrapping in the editor on Github), and is just easier for humans to read when displayed in an editor and no wrapping occurs.

I would agree that for some lines that are a bit long but not that much, the benefits are not huge though. If you are still not convinced just tell me, and I will just drop that commit. For the other PR, it would be harder though, and the benefits a more visible (the lines are so long there that it is more worth it, also there is some code that really benefits from not having a horizontal scroll).

TomasVotruba: If you automate it, I don't mind. But I don't want to add extra maintenance cost.

greg0ire: I don't know of a way to automated that, be it for md files php files or any file, so I'll just drop the commit I guess.

TomasVotruba: FYI, In PHP there is: https://github.com/symplify/coding-standard/blob/master/src/Fixer/LineLength/LineLengthFixer.php

TomasVotruba:
Thank you :+1:

docs/HowItWorks.md
greg0ire: Good to know, but I don't know any fixer for markdown.

> It would me madness to ask developers to do it manually.

I don't know, I feel humans might often be better than robots at this, because they will split a line where it is meaningful to do so. Plus there are tools built in IDEs to make it a bit less manual I think

TomasVotruba: It's about the 0-entrance cost adoption. If you add CI tool, I don't mind and the change would be accepted.

TomasVotruba: Also, this PR is self-example of how human suck in this tasks (my English :D)

PR #3042, opened 2020-03-22
Create mock to create stub

greg0ire:


rector.yaml
```diff
@@ -23,3 +23,53 @@ parameters:
...
+                    $mock->expects($this->once())
+                        ->method('getMessage')
+                        ->willReturn('a message');
+                    self::assertSame('a message', $stub->getMessage());
+                    self::assertSame('a message', $mock->getMessage());
+                }
+            }
+        source:
+            - https://github.com/sebastianbergmann/phpunit/issues/3120
+        set: "phpunit90"
```
greg0ire: I will revert the commit that contains this file, it's just here for reference.

rules/phpunit/src/Rector/MethodCall/CreateMockToCreateStubRector.php
```diff
@@ -0,0 +1,102 @@
...
+    {
+        if (! $this->isName($node->name, 'createMock')) {
+            return null;
+        }
+        $parentNode = $node->getAttribute(AttributeKey::PARENT_NODE);
+        if (!$parentNode instanceof Assign) {
+            return null;
+        }
+
+        dump($this->methodCallManipulator->findMethodCallNamesOnVariable($parentNode->var));
```
greg0ire: I would expect to get all method calls made on `$stub` and `$mock` here.

rector.yaml
TomasVotruba: Tip: you can put this into `create-rector.yaml` file and it won't be commited.
https://github.com/rectorphp/rector/blob/master/.gitignore#L12

That's how I use it

rules/phpunit/src/Rector/MethodCall/CreateMockToCreateStubRector.php
```diff
@@ -0,0 +1,102 @@
...
+PHP
+                 ,
+                <<<'PHP'
+use PHPUnit\Framework\TestCase
+class MyTest extends TestCase {
+    public function testItBehavesAsExpected(): void
+    {
+        $stub = $this->createStub(\Exception::class);
+        $stub->method('getMessage')
+            ->willReturn('a message');
```
TomasVotruba: I thinks I see it

TomasVotruba:
Replaced by #3042 

I had to rework chain method caller resolution, as it was not reliable enough.
Thanks for the intro work and finding the edge case. Your work made it easier to spot the issue :

PR #3050, opened 2020-03-23
Fix assert choice tag value node with class constant reference

### LINKED ISSUE #3044: How to disable inlining of a constant in annotation
# Question
How to disable this?

```diff
    public const RATINGS_DISCRIMINATOR_MAP = [
        '5star' => FiveStar::class,
        '4star' => FourStar::class,
    ];
     /**
      * @Assert\NotBlank()
-     * @Assert\Choice(choices=App\Entity\Rating\Rating::RATINGS_DISCRIMINATOR_MAP)
+     * @Assert\Choice({"5star":"App\Entity\Rating\FiveStar", "4star":"App\Entity\Rating\FourStar"})
      */
```

TomasVotruba:
Closes #3044

### DIFF
--- a/packages/better-php-doc-parser/src/PhpDocNode/Symfony/Validator/Constraints/AssertChoiceTagValueNode.php
+++ b/packages/better-php-doc-parser/src/PhpDocNode/Symfony/Validator/Constraints/AssertChoiceTagValueNode.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Rector\BetterPhpDocParser\PhpDocNode\Symfony\Validator\Constraints;
 
+use Nette\Utils\Strings;
 use Rector\BetterPhpDocParser\Contract\PhpDocNode\ShortNameAwareTagInterface;
 use Rector\BetterPhpDocParser\Contract\PhpDocNode\TypeAwareTagValueNodeInterface;
 use Rector\Symfony\PhpDocParser\Ast\PhpDoc\AbstractConstraintTagValueNode;
@@ -26,6 +27,16 @@ final class AssertChoiceTagValueNode extends AbstractConstraintTagValueNode impl
     private $choices;
 
     /**
+     * @var string|null
+     */
+    private $originalChoiceContent;
+
+    /**
+     * @var bool
+     */
+    private $hasChoiceKeywordExplicit = false;
+
+    /**
      * @param mixed[]|string|null $callback
      */
     public function __construct($callback, ?bool $strict, string $annotationContent, ?array $choices)
@@ -34,6 +45,12 @@ final class AssertChoiceTagValueNode extends AbstractConstraintTagValueNode impl
         $this->strict = $strict;
         $this->resolveOriginalContentSpacingAndOrder($annotationContent);
         $this->choices = $choices;
+
+        // is default key
+        $this->hasChoiceKeywordExplicit = (bool) Strings::match($annotationContent, '#choices=#');
+
+        // hotfix for https://github.com/rectorphp/rector/issues/3044, the constant reference is actually parsed to values
+        $this->originalChoiceContent = $this->resolveOriginalChoiceContent($annotationContent);
     }
 
     public function __toString(): string
@@ -47,7 +64,7 @@ final class AssertChoiceTagValueNode extends AbstractConstraintTagValueNode impl
                 $contentItems['callback'] = sprintf('callback="%s"', $this->callback);
             }
         } elseif ($this->choices) {
-            $contentItems[] = $this->printArrayItem($this->choices);
+            $contentItems[] = $this->createChoices();
         }
 
         if ($this->strict !== null) {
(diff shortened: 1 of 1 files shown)


packages/better-php-doc-parser/src/PhpDocNode/Symfony/Validator/Constraints/AssertChoiceTagValueNode.php
```diff
@@ -71,4 +88,41 @@ public function getShortName(): string
...
+        if ($this->hasChoiceKeywordExplicit) {
+            $content = 'choices=';
+        }
+
+        return $content . $this->originalChoiceContent;
+    }
+
+    private function resolveOriginalChoiceContent(string $annotationContent): ?string
+    {
+        $matches = Strings::match($annotationContent, '#(\(choices=|\()(?<choice_content>.*?)(\)|,)#ms');
```
SergeC: I'm not sure `choices` must be hardcoded since it is related to my use-case and it may not work for others.

TomasVotruba: How would you implement it better?

SergeC: https://getrector.org/demo/54a3163e-13d1-4a01-8101-f32340bcd095 I changed `choices ` to `groups` and issue is back as expected. I would try to remove `choices` from regexp and detect `=` sign. If `=CONST`.

TomasVotruba: Could you send PR?

SergeC: I'm not familiar with how Rector works under the hood. I can provide use cases and do manual testing. My point is not to hardcode `choices` because variable name could be any other or not set.

TomasVotruba: This part is not in Rector, but in doctrine/annotations.
The best would be to disable constant value resolution in there.
This is just a bad band-aid on doctrine design choice.

If we can somehow disable this line: https://github.com/doctrine/annotations/blob/904dca4eb10715b92569fbcd79e201d5c349b6bc/lib/Doctrine/Common/Annotations/DocParser.php#L981
and `return $identifier;` instead, it would solve all these cases.

SergeC: I suggest reverting this Pull Request, reopen an issue and think about solving this issue together. May be Rector can parse a file, find annotations with constants and exclude them with `setIgnoredAnnotationNames` https://github.com/doctrine/annotations/blob/904dca4eb10715b92569fbcd79e201d5c349b6bc/lib/Doctrine/Common/Annotations/DocParser.php#L257 ?

SergeC: Here is a test https://github.com/doctrine/annotations/blob/dc296632db0500c7bd5144161cf96708f61059ff/tests/Doctrine/Tests/Common/Annotations/DocParserTest.php#L1055

TomasVotruba: It seems unrelated.
Could you update your annotation code with that?

SergeC: Sorry, I didn't get your point. What should I do? Please provide more details.

TomasVotruba: Working solution for your use case with "choice" annotation.

Any gist would do.

SergeC: The `setIgnoredAnnotationNames` method doesn't work. Even Doctrine's test doesn't work properly. There is a patch that looks relevant but it is not released yet. https://github.com/doctrine/annotations/pull/244

SergeC: Opened an issue to find it out https://github.com/doctrine/annotations/issues/312

TomasVotruba: How did you verify this would help in our case?

Any reproducible code would help

TomasVotruba:
Surpassed by #3134

PR #3054, opened 2020-03-24
skip if not used with the `array []` operator fixes #3053

### LINKED ISSUE #3053: ListSwapArrayOrderRector should only apply in conjunction with the array [] operator
# Bug Report

| Rector version | v0.7.6              |

The `ListSwapArrayOrderRector` should [only apply in conjunction with the `array []` operator](https://www.php.net/manual/en/migration70.incompatible.php#migration70.incompatible.variable-handling.list).

    list($a['one'], $a['two']) = [1, 2];
https://getrector.org/demo/d7bdc7b5-d3bb-41fc-9270-389e49cd792a#result

## Expected Behaviour

It should skip if an arary with keys is supplied

derflocki:
Closes #3053

### DIFF
--- a/rules/php-70/src/Rector/List_/ListSwapArrayOrderRector.php
+++ b/rules/php-70/src/Rector/List_/ListSwapArrayOrderRector.php
@@ -56,7 +56,7 @@ final class ListSwapArrayOrderRector extends AbstractRector
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Why this should be skipped?
Could you provide 3v4l.org example?

TomasVotruba:
I see. Better add the "fixes #..." description.

PR #3058, opened 2020-03-24
[PHP 7.4] Add default null type on properties

### LINKED ISSUE #2894: TypedPropertyRector: add null assign?
Hi there,

Given this:
https://getrector.org/demo/8a5be694-957f-4981-9189-5c7d835698da

What do you think about making this the result instead?
```php
final class DemoFile
{
    private ?int $foo = null;
}
```

it adds the null assignment, which would be compatible with the old code (implicit null).

without the assignment, there is the possibility for
`Fatal error: Uncaught Error: Typed property DemoFile::$foo must not be accessed before initialization...
`

TomasVotruba:
Closes #2894


rodnaph:
This is a big help, thanks @TomasVotruba and @bendavies !

PR #3066, opened 2020-03-25
[DeadCode] Add RemoveUnusedNonEmptyArrayBeforeForeachRector

TomasVotruba:
Closes #3060

PR #3070, opened 2020-03-26
[DeadCode] Add empty() + count($values) > 0 checks to RemoveUnusedNonEmptyArrayBeforeForeachRector

TomasVotruba:
Follow up to #3066 

image


TomasVotruba:
@afilina Any more samples we should check here?

afilina:
@TomasVotruba I'll append examples should I stumble upon something new.

PR #3072, opened 2020-03-26
[PHP 8.0] Add StrContainsRector

TomasVotruba:
Ref #3025 

```diff
 class SomeClass
 {
     public function run()
     {
-        return strpos('abc', 'a') !== false;
+        return str_contains('abc', 'a');
     }
 }
```

### DIFF
--- a/rules/php80/tests/Rector/NotIdentical/StrContainsRector/Fixture/preg_match.php.inc
+++ b/rules/php80/tests/Rector/NotIdentical/StrContainsRector/Fixture/preg_match.php.inc
@@ -1,0 +1,27 @@
+<?php
+
+namespace Rector\Php80\Tests\Rector\NotIdentical\StrContainsRector\Fixture;
+
+class PregMatchFunction
+{
+    public function run($needle, $haystack)
+    {
+        return preg_match('/' . $needle . '/', $haystack) !== 0;
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\Php80\Tests\Rector\NotIdentical\StrContainsRector\Fixture;
+
+class PregMatchFunction
+{
+    public function run($needle, $haystack)
+    {
+        return str_contains($haystack, $needle);
+    }
+}
+
+?>
--- a/docs/AllRectorsOverview.md
+++ b/docs/AllRectorsOverview.md
@@ -1,4 +1,4 @@
-# All 470 Rectors Overview
+# All 471 Rectors Overview
 
 - [Projects](#projects)
 - [General](#general)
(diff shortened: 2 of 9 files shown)


rules/php80/tests/Rector/NotIdentical/StrContainsRector/Fixture/preg_match.php.inc
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Rector\Php80\Tests\Rector\NotIdentical\StrContainsRector\Fixture;
+
+class PregMatchFunction
+{
+    public function run($needle, $haystack)
+    {
+        return preg_match('/' . $needle . '/', $haystack) !== 0;
```
staabm: https://wiki.php.net/rfc/str_contains Does not mention that str_contains supports all the preg_match wildcards etc.. are you sure this works in this generic fashion?

TomasVotruba: I never used it. It was mentioned in the RFC on internals, where I got inspired.

How would you narrow scope so it works?

staabm: Either you know the $needle at analysis time and can check whether it contains preg-meta chars.. 

When $needle contents is unknown at analysis time, the rector should not refactor the code site

TomasVotruba: Great. I can imagine unknown value.

But could you provide a code sample for first case, where we know the code?
I can think of:

```php
needle = 'abc'; // refactor
needle = '\d'; // skip
preg_match('/' . $needle . '/', $haystack) !== 0;

preg_match('/abc/', $haystack) !== 0; // refactor
preg_match('/\d/', $haystack) !== 0; // skip
```

staabm: e.g. when needle is taken from a unknown source like

```
$needle = $_GET['searchTerm'];

if (preg_match($needle, $haystack)) { // skip
}
```

TomasVotruba: I meant the first case

> Either you know the $needle at analysis time and can check whether it contains preg-meta chars..

staabm: maybe a pretty simple heuristic could be: 

if ($needle === preg_quore($needle, $delim)) {
  // nothing was escaped, it seems to be save to use with `str_contains`
}

TomasVotruba: We cannot run the code, not get the values if not static.

TomasVotruba: Thinking of this, we might skip the `preg_match()` at all. It might be too ambiguous to solve correctly.

TomasVotruba: See https://github.com/rectorphp/rector/pull/3076

PR #3074, opened 2020-03-27
Fix 'auto_import_names' tests fixtures path

Great-Antique:


TomasVotruba:
Thank you!

I'll look on it tonight

PR #3081, opened 2020-03-28
[TypeDeclaration] Add class method param type resolving by property

### LINKED ISSUE #2887: infer types of getters/setters from properties
would be great if rector could infer the param-type of a setter and the return-type of a getter

before
```php

final class DemoFile
{
	/**
	 * @var int
	 */
	private $id;
	
	public function setId($id) {
		$this->id = $id;
	}
	public function getId() {
		return $this->id;
	}
}
```

after
```php

final class DemoFile
{
	/**
	 * @var int
	 */
	private $id;
	
	public function setId(int $id) {
		$this->id = $id;
	}
	public function getId():int {
		return $this->id;
	}
}
```

TomasVotruba:
Closes #2887

### DIFF
--- a/rules/type-declaration/tests/Rector/FunctionLike/ParamTypeDeclarationRector/Fixture/infer_from_property_type.php.inc
+++ b/rules/type-declaration/tests/Rector/FunctionLike/ParamTypeDeclarationRector/Fixture/infer_from_property_type.php.inc
@@ -1,0 +1,37 @@
+<?php
+
+namespace Rector\TypeDeclaration\Tests\Rector\ClassMethod\ParamTypeDeclarationRector\Fixture;
+
+class InferFromPropertyType
+{
+    /**
+     * @var int
+     */
+    private $id;
+
+    public function setId($id)
+    {
+        $this->id = $id;
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\TypeDeclaration\Tests\Rector\ClassMethod\ParamTypeDeclarationRector\Fixture;
+
+class InferFromPropertyType
+{
+    /**
+     * @var int
+     */
+    private $id;
+
+    public function setId(int $id)
+    {
+        $this->id = $id;
+    }
+}
+
+?>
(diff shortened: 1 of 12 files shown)


TomasVotruba:
cc @staabm

rules/type-declaration/tests/Rector/FunctionLike/ParamTypeDeclarationRector/Fixture/infer_from_property_type.php.inc
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Rector\TypeDeclaration\Tests\Rector\ClassMethod\ParamTypeDeclarationRector\Fixture;
+
+class InferFromPropertyType
```
staabm:
```suggestion
class InferFromPropertyPhpdocType
```

rules/type-declaration/tests/Rector/FunctionLike/ParamTypeDeclarationRector/Fixture/infer_from_property_type.php.inc
```diff
@@ -0,0 +1,37 @@
...
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\TypeDeclaration\Tests\Rector\ClassMethod\ParamTypeDeclarationRector\Fixture;
+
+class InferFromPropertyType
```
staabm:
```suggestion
class InferFromPropertyPhpdocType
```

staabm:
Maybe its a different topic, but I would also add the following example (Same idea but with typed property)

Before

```
class InferFromPropertyType
 {
     private int $id;

     public function setId($id)
     {
         $this->id = $id;
     }
 }
```

After

```
class InferFromPropertyType
 {
     private int $id;

     public function setId(int $id)
     {
         $this->id = $id;
     }
 }
```

TomasVotruba:
> Thx for working on it. The test case shows what I expected. Cannot judge the implementation... not much AST experience on my end

Thanks, any review and feedback helps me :+1: . I'm blinded by my own ego.

TomasVotruba:
> Maybe its a different topic, but I would also add the following example (Same idea but with typed property)

Good idea, I'll add new type inferer.

PR #3087, opened 2020-03-28
Fix ArrayKeysAndInArrayToArrayKeyExistsRectorTest

paslandau:
- fixed typo in RectorDefinition code sample
- fixed argument order of array_key_exists
- renamed test folder from ArrayKeysAndInArrayToIssetRector
  to ArrayKeysAndInArrayToArrayKeyExistsRector


TomasVotruba:
Thanks :+1:

paslandau:
@TomasVotruba 

FYI: I went trough some other rectors and spotted 
`\Rector\CodeQuality\Rector\FuncCall\InArrayAndArrayKeysToArrayKeyExistsRector`
which seems to be quite similar to the new 
`\Rector\CodeQuality\Rector\FuncCall\ArrayKeysAndInArrayToArrayKeyExistsRector`
==> does it even make sense to keep both?

TomasVotruba:
If there is no difference, it's possible I made them twice :D

PR #3089, opened 2020-03-28
Add RunnableTestCase to run fixed code in a test

paslandau:
- added RunnableTestCase::assertOriginalAndFixedFileYieldSameResult($file)
  that takes a fixture file and expects fixture classes that implement the
  new RunnableInterface which exposes a run() method
- the fixture class is dynamically renamed to avoid naming conflicts
  and loaded via eval()
- fixtures that don't implement the RunnableInterface are ignored
  otherwise the run() method is called on the original class as well
  as on the fixed one and the output is expected to be equal
  (via assertSame) by default
- the default assertion behavior can be changed via $assertion callable

### DIFF
--- a/rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/ArrayKeysAndInArrayToArrayKeyExistsRectorTest.php
+++ b/rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/ArrayKeysAndInArrayToArrayKeyExistsRectorTest.php
@@ -6,9 +6,9 @@ namespace Rector\CodeQuality\Tests\Rector\FuncCall\ArrayKeysAndInArrayToArrayKey
 
 use Iterator;
 use Rector\CodeQuality\Rector\FuncCall\ArrayKeysAndInArrayToArrayKeyExistsRector;
-use Rector\Core\Testing\PHPUnit\AbstractRectorTestCase;
+use Rector\Core\Testing\PHPUnit\AbstractRunnableRectorTestCase;
 
-final class ArrayKeysAndInArrayToArrayKeyExistsRectorTest extends AbstractRectorTestCase
+final class ArrayKeysAndInArrayToArrayKeyExistsRectorTest extends AbstractRunnableRectorTestCase
 {
     /**
      * @dataProvider provideData()
@@ -16,6 +16,7 @@ final class ArrayKeysAndInArrayToArrayKeyExistsRectorTest extends AbstractRector
     public function test(string $file): void
     {
         $this->doTestFile($file);
+        $this->assertOriginalAndFixedFileResultEquals($file);
     }
 
     public function provideData(): Iterator
--- a/rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/Fixture/fixture.php.inc
+++ b/rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/Fixture/fixture.php.inc
@@ -1,11 +1,16 @@
 <?php
 
-namespace Rector\CodeQuality\Tests\Rector\FuncCall\ArrayKeysAndInArrayToIssetRector\Fixture;
+namespace Rector\CodeQuality\Tests\Rector\FuncCall\ArrayKeysAndInArrayToArrayKeyExistsRector\Fixture;
 
-class SomeClass
+use Rector\Core\Testing\PHPUnit\RunnableInterface;
+
+class SomeClass implements RunnableInterface
 {
-    public function run($packageName, $values)
+    public function run()
     {
+        $packageName = "foo";
+        $values      = ["foo" => "bar"];
+        
         $keys = array_keys($values);
         return in_array($packageName, $keys, true);
     }
@@ -15,12 +20,16 @@ class SomeClass
 -----
 <?php
 
-namespace Rector\CodeQuality\Tests\Rector\FuncCall\ArrayKeysAndInArrayToIssetRector\Fixture;
+namespace Rector\CodeQuality\Tests\Rector\FuncCall\ArrayKeysAndInArrayToArrayKeyExistsRector\Fixture;
 
-class SomeClass
+use Rector\Core\Testing\PHPUnit\RunnableInterface;
+
+class SomeClass implements RunnableInterface
 {
-    public function run($packageName, $values)
+    public function run()
     {
+        $packageName = "foo";
+        $values      = ["foo" => "bar"];
         return array_key_exists($packageName, $values);
     }
 }
--- a/src/Testing/PHPUnit/AbstractRunnableRectorTestCase.php
+++ b/src/Testing/PHPUnit/AbstractRunnableRectorTestCase.php
@@ -1,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Core\Testing\PHPUnit;
+
+use Nette\Utils\Strings;
+use Rector\Core\Testing\ValueObject\SplitLine;
+use ReflectionClass;
+use Symplify\SmartFileSystem\SmartFileInfo;
+
+abstract class AbstractRunnableRectorTestCase extends AbstractRectorTestCase
+{
+    protected function assertOriginalAndFixedFileResultEquals(string $file): void
+    {
+        /**
+         * Todo: Duplicate from
+         *
+         * @see FixtureSplitter::splitContentToOriginalFileAndExpectedFile
+         * ==> refactor in a method
+         */
+        $smartFileInfo = new SmartFileInfo($file);
+        if (Strings::match($smartFileInfo->getContents(), SplitLine::SPLIT_LINE)) {
+            [$originalContent, $expectedContent] = Strings::split($smartFileInfo->getContents(), SplitLine::SPLIT_LINE);
+        } else {
+            $originalContent = $smartFileInfo->getContents();
+            $expectedContent = $originalContent;
+        }
+
+        $originalInstance = $this->loadClass($originalContent);
+        if ($originalInstance !== null) {
+            $expectedInstance = $this->loadClass($expectedContent);
+            if ($expectedInstance !== null) {
+                $actual = $originalInstance->run();
+                $expected = $expectedInstance->run();
+
+                $this->assertSame($actual, $expected);
+            }
+        }
+    }
+
+    protected function getTemporaryClassName(): string
+    {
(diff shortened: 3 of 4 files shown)


paslandau:
@TomasVotruba 
Is there anything else apart from `composer run complete-check` that I can run locally to verify the code? Was kinda suprised to see the GH actions fail because everything looked fine locally :)

TomasVotruba:
The composer command doesn't include all the actions.
Could you add missing scripts there?

rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/ArrayKeysAndInArrayToArrayKeyExistsRectorTest.php
```diff
@@ -6,16 +6,16 @@
...
-final class ArrayKeysAndInArrayToArrayKeyExistsRectorTest extends AbstractRectorTestCase
+final class ArrayKeysAndInArrayToArrayKeyExistsRectorTest extends RunnableTestCase
 {
     /**
      * @dataProvider provideData()
      */
     public function test(string $file): void
     {
         $this->doTestFile($file);
+        $this->assertOriginalAndFixedFileYieldSameResult($file);
```
TomasVotruba: yield is PHP keyword that reminds something related to collections

What about assert...ResultEquals?

rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/Fixture/fixture.php.inc
```diff
@@ -1,28 +1,48 @@
...
         return in_array($packageName, $keys, true);
     }
+
+    public function run()
+    {
+        $packageName = "foo";
+        $values      = ["foo" => "bar"];
+
+        return $this->test($packageName, $values);
+    }
```
TomasVotruba: Having 2+ methods in test fixture adds huge complexity to maintenance.

Could there be another way having just one method?

rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/RunnableTestCase.php
```diff
@@ -0,0 +1,75 @@
...
+
+namespace Rector\CodeQuality\Tests\Rector\FuncCall\ArrayKeysAndInArrayToArrayKeyExistsRector;
+
+use Nette\Utils\Strings;
+use Rector\Core\Testing\PHPUnit\AbstractRectorTestCase;
+use Rector\Core\Testing\ValueObject\SplitLine;
+use ReflectionClass;
+use Symplify\SmartFileSystem\SmartFileInfo;
+
+class RunnableTestCase extends AbstractRectorTestCase
```
TomasVotruba: Should be abstract

rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/RunnableTestCase.php
```diff
@@ -0,0 +1,75 @@
...
+{
+    protected function assertOriginalAndFixedFileYieldSameResult(string $file, ?callable $assertion = null): void
+    {
+        $defaultAssertion = function (RunnableInterface $original, RunnableInterface $fixed): void {
```
TomasVotruba: Callable type is weak type. Either use contract of no functions at all. Here should be just result assertSame()

paslandau:
> Could you add missing scripts there?
Had a quick look at the `.github/workflows` but I don't quite understand the setup and I don't feel comfortable changing the composer script without fully understanding the consequences. 

Is this something that you can provide quickly on master?

Happy to fix any open issues in my PR as long as I get the corresponding feedback locally as the feedback loop via github actions only takes too long for that :/

rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/Fixture/fixture.php.inc
paslandau: Sure, just I can just use `run()` directly.

rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/RunnableTestCase.php
paslandau: kk, Works for me

rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/ArrayKeysAndInArrayToArrayKeyExistsRectorTest.php
paslandau: makes sense

TomasVotruba:
It's usually the last `run` in each workflow.
E.g.
https://github.com/rectorphp/rector/blob/23b08fc2b15ad16231e3bce8f58ab97d067e03c2/.github/workflows/check_fixtures.yaml#L19

I might get into it in 1-2 hours and update the `composer.json` with full check

TomasVotruba:
It seems most of issues are gone.

To apply Rector, just run:

```
composer rector-ci-fix
```

It's seems there bugs in few old Rector rules though, I'll check it.

paslandau:
How long does a `composer complete-check` now take for you? Running all rector sets is pretty slow for me and actually just failed with 

```
[...]
Set "safe07" is OK
Set "shopware55" is OK
Set "shopware56" is OK
Set "silverstripe" is OK

  [Symfony\Component\Process\Exception\ProcessTimedOutException]
  The process "php ci/run_all_sets.php" exceeded the timeout of 300 seconds.

```

TomasVotruba:
I never run it locally, so I don't know.

How could be the performance improved there?

rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/AbstractRunnableTestCase.php
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Rector\CodeQuality\Tests\Rector\FuncCall\ArrayKeysAndInArrayToArrayKeyExistsRector;
+
+use Nette\Utils\Strings;
+use Rector\Core\Testing\PHPUnit\AbstractRectorTestCase;
+use Rector\Core\Testing\ValueObject\SplitLine;
+use ReflectionClass;
+use Symplify\SmartFileSystem\SmartFileInfo;
+
+abstract class AbstractRunnableTestCase extends AbstractRectorTestCase
```
TomasVotruba:
```suggestion
abstract class AbstractRunnableRectorTestCase extends AbstractRectorTestCase
```

rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/AbstractRunnableTestCase.php
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    protected function loadClass(string $classContent): ?RunnableInterface
+    {
+        $className = $this->getTemporaryClassName();
+        $loadable = (string) Strings::replace($classContent, '#\\s*<\\?php#', '');
+        $loadable = (string) Strings::replace($loadable, '#\\s*namespace.*;#', '');
+        $loadable = (string) Strings::replace($loadable, '#class\\s+(\\S*)\\s+#', sprintf('class %s ', $className));
+        eval($loadable);
+        if (is_a($className, RunnableInterface::class, true)) {
```
TomasVotruba: This should be right under `$className = $this->getTemporaryClassName();` with early reaturn,
as it can skipp all the `Strings::replace()`

rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/ArrayKeysAndInArrayToArrayKeyExistsRectorTest.php
```diff
@@ -6,16 +6,16 @@
...
-final class ArrayKeysAndInArrayToArrayKeyExistsRectorTest extends AbstractRectorTestCase
+final class ArrayKeysAndInArrayToArrayKeyExistsRectorTest extends AbstractRunnableTestCase
 {
     /**
      * @dataProvider provideData()
      */
     public function test(string $file): void
     {
         $this->doTestFile($file);
+        $this->assertOriginalAndFixedFileResultEquals($file);
```
TomasVotruba: :+1: Nice DX!

rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/AbstractRunnableTestCase.php
paslandau: Don't think that would work. 
`is_a` will only be able to recognize the class after `eval` has made it available to php

TomasVotruba: Ah, I didn't notice. Is there a way to handle this withou `eval()`?
That opens big black door.

paslandau: Well, to be fair you already have an unchecked `require_once` in `\Rector\Core\Testing\PHPUnit\FixtureSplitter::splitContentToOriginalFileAndExpectedFile` (which is basically the same) :)

Only way to work around this is to split the "orignal" file and the "fixed" files into separate files and also use different class names and proper autoloading. However, this clashes with the current way fixtures are written and checked (e.g. it won't be possible to do a full "equality" check of the contents any longer because the class names will be different - which I think is very helpful).

If you keep the "goal" in mind (i.e. "reuse existing test files and simply add an additional layer of validation), I'm afraid there is no way around using something that dymically injects code.

paslandau:
@TomasVotruba 
anything still open/unclear here?

TomasVotruba:
Looks good me to me :+1:

PR #3094, opened 2020-03-29
Add parallel execution to ci/run_all_sets

paslandau:
- add a task runner to run the sets in parallel
- number of cores is determined automatically via https://gist.github.com/divinity76/01ef9ca99c111565a72d3a8a6e42f7fb
- added github action at `ci/sets_check_parallel.yaml` to make sure it works in CI

==> roughly 3 times faster on my local machine with 8 cores

Usage:

`php ci/run_all_sets_parallel.php`

Output:
```
www-data@f273549c9848:/codebase/rector$ time php ci/run_all_sets_parallel.php
Running 135 sets with 8 parallel processes

✔ Set "cakephp-24-to-symfony-51" is OK ( 1 / 135)
✔ Set "cakephp34" is OK ( 2 / 135)
✔ Set "action-injection-to-constructor-injection" is OK ( 3 / 135)
✔ Set "array-str-functions-to-static-call" is OK ( 4 / 135)
✔ Set "cakephp30" is OK ( 5 / 135)
✔ Set "cakephp35" is OK ( 6 / 135)
✔ Set "cakephp36" is OK ( 7 / 135)
✔ Set "cakephp37" is OK ( 8 / 135)
✔ Set "cakephp38" is OK ( 9 / 135)
✔ Set "cakephp40" is OK (10 / 135)
✔ Set "constructor-injectin-to-action-injection" is OK (11 / 135)
✔ Set "celebrity" is OK (12 / 135)
✔ Set "code-quality" is OK (13 / 135)
✔ Set "codeigniter-40" is OK (14 / 135)
✔ Set "contributte-to-symfony" is OK (15 / 135)
✔ Set "coding-style" is OK (16 / 135)
✔ Set "dead-classes" is OK (17 / 135)
✔ Set "doctrine" is OK (18 / 135)
✔ Set "dead-code" is OK (19 / 135)
✔ Set "doctrine-behaviors-20" is OK (20 / 135)
✔ Set "doctrine-code-quality" is OK (21 / 135)
✔ Set "doctrine-common-20" is OK (22 / 135)
✔ Set "doctrine-dbal-210" is OK (23 / 135)
✔ Set "doctrine-dbal-30" is OK (24 / 135)
✔ Set "doctrine-id-to-uuid-step-1" is OK (25 / 135)
✔ Set "doctrine-gedmo-to-knplabs" is OK (26 / 135)
✔ Set "doctrine-id-to-uuid-step-2" is OK (27 / 135)
[...]

PR #3097, opened 2020-03-29
move ci validation scripts to objectivy ProjectValidator package

TomasVotruba:
Follow up to #3094


TomasVotruba:
cc @paslandau

TomasVotruba:
@paslandau There you go!

I took more time than I expected. I got lost in `&` references a bit.

Could you check it?

utils/project-validator/src/Process/ParallelTaskRunner.php
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * @var SymfonyStyle
+     */
+    private $symfonyStyle;
+
+    /**
+     * @var Process[]
+     */
+    private $runningProcesses = [];
```
paslandau: On the one hand I feel that this is better than using references - on the other it feels kinda weird to have this as a "stateful property" in the class. Same for
```
    private $isSuccessful = true;
    private $finishedProcessCount = 0;
    $remainingTasks = [];
```
I feel that those things are tightly coupled to the `run()` method and should not be available outside of it. I'd propose to create one more class that specifies the actual "Run" as in "Execution of a set of tasks", e.g. `TaskRun`. The ParallelTaskRunner would instantiate that class with `$maxProcesses`  and `$sleepInSeconds` and the above mentioned "stateful properties" would only live in that class.

But that's icing on the cake. I'll add a POC that shows the concept and then you can decide to go with it or keep it like this.

TomasVotruba: References are hard to see, refactor and it cryples static analysis and Rector.

You're right. It would break in case of re-use. There should be a `reset()` method call in the `run()` start.
I'll add it.

utils/project-validator/src/Command/ValidateSetsCommand.php
```diff
@@ -0,0 +1,100 @@
...
+        $this->setName(CommandNaming::classToName(self::class));
+        $this->setDescription('[CI] Validate each sets by running it on simple file');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $setTasks = $this->createSetTasks();
+        $cpuCoreCount = $this->cpuCoreCountResolver->resolve();
+
+        $this->parallelTaskRunner->run($setTasks, $cpuCoreCount, self::SLEEP_IN_SECONDS);
```
paslandau: Shouldn't this return `ShellCode::ERROR;` if ->run returns false? E.g.

```
$noErrors = $this->parallelTaskRunner->run($setTasks, $cpuCoreCount, self::SLEEP_IN_SECONDS);

        if (!$noErrors) {
            return ShellCode::ERROR;
        }

TomasVotruba: Indeed. I'll fix that

utils/project-validator/src/Process/ParallelTaskRunner.php
paslandau: FYI: https://github.com/rectorphp/rector/pull/3102

TomasVotruba: I see all the PRs here :)

PR #3102, opened 2020-03-29
Add ParallelTaskRun class to handle a full run of rector sets

paslandau:


composer.json
```diff
@@ -263,7 +264,8 @@
         ],
         "rector-ci-fix": "bin/rector process --config rector-ci.yaml --ansi --no-progress-bar",
         "rector-ci": "bin/rector process --config rector-ci.yaml --ansi --no-progress-bar --dry-run",
-        "rector": "bin/rector process --config rector-ci.yaml --ansi"
+        "rector": "bin/rector process --config rector-ci.yaml --ansi",
+        "rector-dry": "bin/rector process --config rector-ci.yaml --ansi --dry-run"
```
TomasVotruba: Same as `rector-ci` above

utils/project-validator/config/config.yaml
```diff
@@ -8,3 +8,4 @@ services:
         resource: '../src'
         exclude:
             - '../src/ValueObject/*'
+            - '../src/Process/ParallelTaskRun.php'
```
TomasVotruba: If this is value object, should be in `ValueObject` namespace. Thus the excluded case above

utils/project-validator/src/Command/ValidateSetsCommand.php
```diff
@@ -76,9 +76,12 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
         $cpuCoreCount = $this->cpuCoreCountResolver->resolve();
 
-        $this->parallelTaskRunner->run($setTasks, $cpuCoreCount, self::SLEEP_IN_SECONDS);
+        $noErrors = $this->parallelTaskRunner->run($setTasks, $cpuCoreCount, self::SLEEP_IN_SECONDS);
 
-        return ShellCode::SUCCESS;
+        if ($noErrors) {
+            return ShellCode::SUCCESS;
+        }
+        return ShellCode::ERROR;
```
TomasVotruba: :+1: 
```suggestion
        return $noErrors ? ShellCode::ERROR : ShellCode::SUCCESS;
```

utils/project-validator/src/Process/ParallelTaskRun.php
```diff
@@ -0,0 +1,206 @@
...
+        $this->symfonyStyle = $symfonyStyle;
+        $this->total = $total;
+        $this->maxProcesses = $maxProcesses;
+        $this->sleepInSeconds = $sleepInSeconds;
+    }
+
+    /**
+     * @param Iterator<Process> $processes
+     */
+    public function __invoke(Iterator $processes): bool
```
TomasVotruba: This is mix of object + services. Should be definitely separated to 2 classes

TomasVotruba:
I don't see much benefit in this. For me it's finished and working. Just FYI

paslandau:
> I don't see much benefit in this. For me it's finished and working. Just FYI

Sure, no problem. Guess it's more of a preference thing. I'll close the PR :)

TomasVotruba:
Thanks :+1: 

It's also about maintainability. The more magic we have in the code, the less helpful are tools like PHPStan and Rector. And I'm very lazy with maintaining code so the less work the better :D

PR #3106, opened 2020-03-30
Require symfony 5.0.6 or 4.4.6

### LINKED ISSUE #3067: Permission denied ... Kernel.php on line 486
# Bug Report

| Rector version | 0.7.7              |
| Installed as   | both                   |

When running rector I started getting the following errors:

When run regular (rector/rector)
```
vendor\bin\rector process src --set=php74 --dry-run
PHP Notice:  include(): read of 4096 bytes failed with errno=13 Permission denied in \rector-test-2\vendor\symfony\http-kernel\Kernel.php on line 486
Notice: include(): read of 4096 bytes failed with errno=13 Permission denied in \rector-test-2\vendor\symfony\http-kernel\Kernel.php on line 486
```

When run prefixed:

```
vendor\bin\rector process src --set=php74 --dry-run
PHP Notice:  include(): read of 4096 bytes failed with errno=13 Permission denied in phar://.../rector-test/vendor/rector/rector-prefixed/rector/vendor/symfony/http-kernel/Kernel.php on line 487
Notice: include(): read of 4096 bytes failed with errno=13 Permission denied in phar://.../rector-test/vendor/rector/rector-prefixed/rector/vendor/symfony/http-kernel/Kernel.php on line 487
```

I made a new folder and then included rector via composer and nothing else:
composer require rector/rector-prefixed OR composer require rector/rector

Then I made a super simple PHP file in src called HelloWorld.php that's blank.

We started getting this error in our real setup when upgrading rector, but I hope the above is a simple enough way to replicate it.

So right now it's happening in our real, more complex project, and in these two super simple test cases. It's happening on multiple computers.

I've tried setting windows permissions to give full control of the files to everyone, but it hasn't helped.

These are all Windows computers. I do not get the error when doing same thing from Ubuntu.

If there are changes to be made, Rector will still make them. But it always put those two lines of error at the top, which breaks JSON parsing of the output.

UFTimmy:
Makes sure the symfony/http-kernel package is a high enough version to fix #3067


TomasVotruba:
Thanks

Issue #3110, opened 2020-03-31
[Cache] Add result cache based on nodes and files, similar to PHPStan


TomasVotruba:

TomasVotruba:
Closed by #3118

PR #3118, opened 2020-04-01
Add cache for un-changed files

TomasVotruba:
Depends on #3197 

## Enable in Config

```yaml
# rector.yaml
parameters:
    is_cache_enabled: true # false by default
```

<br>

- [x] add feature flag, disabled by default for testing

## Without cache (~ 18 seconds on 200 files)

wihtout_cache

## With cache (~ 1 second on 200 files)

with_cache

Ref #3110 

## Notes

How to approach rules, that needs more scope for analysis, like:

- `PrivatizeLocalOnlyMethodRector`
- `RemoveUnusedClassConstantRector`

etc. Extend dependency files resolver? Probably best would be:


ruudboon:
Nice work!

bendavies:
phpstan and psalm both to dependency tracking, which sounds like you want to do.

TomasVotruba:
@bendavies This PR uses part of PHPStan dependency resolver. It works quite well. 

Rector just sometimes needs full scope of the application, e.g. for unused public class method. For that I came up with interface that needs full cache clear. I wonder if that can be done in some better way. Ideas?

Aerendir:
Wow, great work Tomas!

I will try to understand the implementation, but in those days it is really complex for me due to Covid limitations: I'm forced at home with my wife and my three years old son: it is really really complex to work...

TomasVotruba:
@Aerendir Oh, I can't imagine how you're feeling while coding in this environment. I wish we all survive this in both physical and mental health. I sometime go for a run in the park or coding there. It helps me to lay of some steam from lockdown. Also hangouting or calling with friends. If you want one, we can do :)

Anyway, if you find something you don't understand, just go for it and write a comment. It's a first draft and I believe most of it can be written much better way. That's why appreciate feedback so much.

TomasVotruba:
Let's :ship: it'

PR #3137, opened 2020-04-03
Fixup has same commit message

JanMikes:


.github/workflows/rector_ci.yaml
```diff
@@ -68,7 +69,7 @@ jobs:
 
             -   name: Commit CS changes
                 if: steps.cs-git-check.outputs.modified == 'true'
-                run: git commit -am "fixup! [Rector CI] Fixed CS issues"
+                run: git commit -am "fixup! fixup! ${COMMIT_MESSAGE}"
```
TomasVotruba, line 72: This might not be needed, as the previous commit massage already starts with "fixup! ..."

JanMikes: it is needed, because `$COMMIT_MESSAGE` contains original message only.

And because we have split rector fix run and CS fixes run into 2 separate commits and $COMMIT_MESSAGE does not get updated meanwhile (it stays still same) it would end up with commits like this:

- `Original`
- `fixup! Original` (rector fix)
- `fixup! Original` (cs fix)

JanMikes: with current setup the output will be 

- `Original`
- `fixup! Original` (rector fix)
- `fixup! fixup! Original` (cs fix)

is this behavior desired?

PR #3146, opened 2020-04-07
Make sure doctrine alias rector works in symfony controllers

acrobat:
While using this rector on a project, I noticed it didn't work on symfony controllers. Because the `$this->getDoctrine()->getManager()` returns an instance of `Doctrine\Persistence\ObjectManager` (or the deprecated interface `Doctrine\Common\Persistence\ObjectManager`). So I've debugged and adjusted the check so it would also work in symfony controllers.

The only strange thing is that the original example (test fixture) does work in the test suite with the old code, but it didn't work in the project I was debugging 🤷‍♂


TomasVotruba:
Thank you, great job with the test :+1:

PR #3155, opened 2020-04-10
Add extra allowed interface to EntityAliasToClassConstantReferenceRector

acrobat:
This is an extension to the changes in #3146. I've discovered some other usages which were not covered by this rector. So these extra interfaces will allow to fix these cases also.

```
$this->getDoctrine()->getRepository();
```

This is also possible as the manager registry will select to correct manager based on the provided entity. 

https://github.com/doctrine/persistence/blob/fdaaa50cb23760c9da2d6dcab9f10caa34d92b61/lib/Doctrine/Persistence/AbstractManagerRegistry.php#L226-L233

### DIFF
--- a/rules/doctrine/src/Rector/MethodCall/EntityAliasToClassConstantReferenceRector.php
+++ b/rules/doctrine/src/Rector/MethodCall/EntityAliasToClassConstantReferenceRector.php
@@ -4,8 +4,10 @@ declare(strict_types=1);
 
 namespace Rector\Doctrine\Rector\MethodCall;
 
+use Doctrine\Common\Persistence\ManagerRegistry as DeprecatedManagerRegistry;
 use Doctrine\Common\Persistence\ObjectManager as DeprecatedObjectManager;
 use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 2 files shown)


rules/doctrine/src/Rector/MethodCall/EntityAliasToClassConstantReferenceRector.php
```diff
@@ -65,7 +67,13 @@ public function refactor(Node $node): ?Node
...
         if (! $this->isObjectTypes(
             $node->var,
-            [EntityManagerInterface::class, ObjectManager::class, DeprecatedObjectManager::class]
+            [
+                EntityManagerInterface::class,
+                ObjectManager::class,
+                DeprecatedObjectManager::class,
+                ManagerRegistry::class,
+                DeprecatedManagerRegistry::class,
+            ]
```
TomasVotruba, line 76: Should be extracted to local constant

acrobat: Fixed 👍

TomasVotruba:
Thank you very much

Issue #3223, opened 2020-04-23
Rector crash with [ Fatal error: Uncaught Error: Call to undefined method Symfony\Component\Routing\Annotation\Route::getLocalizedPaths() ]


atompulse:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.8.x-dev@e02410b             |
| Installed as   | docker rector/rector:latest                      |

Project code base "symfony/symfony": "3.4.12"
PHP version 7.1
Running with -vvv
Having more than 3k classes in /src

Whenever rector is being run and analyzes symfony controllers an error is being thrown:
```
[parsing] ../project/src/XXX/WebsiteBundle/Controller/YYY/ZZZ/DashboardController.php
Fatal error: Uncaught Error: Call to undefined method Symfony\Component\Routing\Annotation\Route::getLocalizedPaths() in /rector/packages/better-php-doc-parser/src/PhpDocNodeFactory/Symfony/SymfonyRoutePhpDocNodeFactory.php:41
Stack trace:
#0 /rector/packages/better-php-doc-parser/src/PhpDocParser/BetterPhpDocParser.php(195): Rector\BetterPhpDocParser\PhpDocNodeFactory\Symfony\SymfonyRoutePhpDocNodeFactory->createFromNodeAndTokens(Object(PhpParser\Node\Stmt\ClassMethod), Object(PHPStan\PhpDocParser\Parser\TokenIterator))
#1 /rector/packages/better-php-doc-parser/src/PhpDocParser/BetterPhpDocParser.php(169): Rector\BetterPhpDocParser\PhpDocParser\BetterPhpDocParser->parseTagValue(Object(PHPStan\PhpDocParser\Parser\TokenIterator), '@Route')
#2 /rector/vendor/phpstan/phpdoc-parser/src/Parser/PhpDocParser.php(54): Rector\BetterPhpDocParser\PhpDocParser\BetterPhpDocParser->parseTag(Object(PHPStan\PhpDocParser\Parser\TokenIterator))
#3 [internal function]: PHPStan\PhpDocParser\Parser\PhpDocParser->parseChild(Object(PHPStan\PhpD in /rector/packages/better-php-doc-parser/src/PhpDocNodeFactory/Symfony/SymfonyRoutePhpDocNodeFactory.php on line 41
```
The curious fact is that if I move this controller from /project/src/XXX/WebsiteBundle/Controller/YYY/ZZZ/
to 
[...]

PR #3224, opened 2020-04-23
https://github.com/rectorphp/rector/issues/3223

atompulse:
Should not call getLocalizedPaths if method doesnt exists.


TomasVotruba:
Thank you :+1:

PR #3246, opened 2020-04-24
Fix empty name in @Route

### LINKED ISSUE #3234: Rector removes empty name from @Route annotation
# Bug Report

| Rector version | 0.7.17              |
```

In Symfony you can annotate a name and path in a Controller class, which is actually a prefix of controller actions in that class:

/**
 * @Route("front", name="front")
 */
final class FrontController extends Controller
{
    /**
     * @Route("/", name="")
     */
    public function homepage() : Response
    {
         //...
    }
}
attribute
>  bin/console debug:router
`  front                                   GET|POST            ANY      ANY    /front/`

Rector removes the empty name attribute, which means the route is removed.

```

https://getrector.org/demo/9dee6b7b-b443-4432-915b-5c22a80a4106#result

## Expected Behaviour

An empty attribute is required in this case so it should be untouched.

TomasVotruba:
- Closes #3240
- Closes #3234﻿

### DIFF
--- a/packages/better-php-doc-parser/src/PhpDocNode/Symfony/SymfonyRouteTagValueNode.php
+++ b/packages/better-php-doc-parser/src/PhpDocNode/Symfony/SymfonyRouteTagValueNode.php
@@ -118,9 +118,8 @@ final class SymfonyRouteTagValueNode extends AbstractTagValueNode implements Sho
             'path' => $this->printValueWithOptionalQuotes('path', $this->path, $this->localizedPaths),
         ];
 
-        if ($this->name) {
-            $contentItems['name'] = $this->printValueWithOptionalQuotes('name', $this->name);
-        }
+        // required
+        $contentItems['name'] = $this->printValueWithOptionalQuotes('name', $this->name);
 
         if ($this->methods !== []) {
             $contentItems['methods'] = $this->printArrayItem($this->methods, 'methods');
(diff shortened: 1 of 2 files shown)


packages/better-php-doc-parser/src/PhpDocNode/Symfony/SymfonyRouteTagValueNode.php
```diff
@@ -118,9 +118,8 @@ public function __toString(): string
             'path' => $this->printValueWithOptionalQuotes('path', $this->path, $this->localizedPaths),
         ];
 
-        if ($this->name) {
```
stephanvierkant, line 121: Maybe `if ($this->name !== null) {`?

TomasVotruba: I'm not sure if annotation parser from Doctrine makes any difference about these 2 cases.

Why do you need empty attribute?

stephanvierkant: That's the point of this issue, isn't it? Name may be empty, but in this case an empty name is removed.

TomasVotruba: I mean why do you need empty attribute in your code.

```php
$this->redirect('')
```

It might not be technically possible to differentiate, so if that has no effect, it should be removed in your code.

stephanvierkant: No, a method `@Route` can have an empty name (`""`, not `null`) if the class `@Route` has a name.

TomasVotruba: How can you use that?

stephanvierkant: I'm not sure I understand you correctly. 

When you remove the name attribute from a method, the method is removed from routing. It should be empty, not removed. See my example: https://github.com/rectorphp/rector/pull/3240/files

That file contains valid code and should'nt be changed.

stephanvierkant: See #3248

TomasVotruba:
Replaced by #3248

PR #3248, opened 2020-04-24
Fix @Route name can be empty

### LINKED ISSUE #3234: Rector removes empty name from @Route annotation

stephanvierkant:
* Closes #3240
* Closes #3234
* Closes #3246


TomasVotruba:
Thank you :+1:

Issue #3222, opened 2020-04-23
Additional (code standard?) rules are running by default


shaal:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | v0.7.16 (invoke `vendor/bin/rector --version`)              |
| Installed as   | prefixed Rector PHAR                      |

## Minimal PHP Code Causing Issue

In this example a docblock is being removed, because there's an empty docblock after it:
```
-/**
- * This docblock is being deleted when there's an empty docblock after it.
- */
-
-/**
- * 
- */
-
```
https://getrector.org/demo/9dbe8d1d-b062-4b16-97d2-605769834cbc#result

In this example empty comment lines are being removed, and 2 lines are being removed and added back the same way:
```

 /**
  * The empty line after this docblock is being deleted.
- */
-
+ */
 /**
  * Another dockblock.
- * 
+ *  
  *
  * first_function()
- * 
+ *  
  *
- * second_function()
- * third_function()
+ * second_function()
+ * third_function()
  * fourth_function()
- */
-
+ */
```
https://getrector.org/demo/0737a1e7-0b83-4e50-9e64-72f2e33476c8#result

## Expected Behaviour

I'm expecting rector to run only the rules that are set in rector.yml, and if it's not explicitly called it shouldn't run.

shaal:
@TomasVotruba Can you please help us?
This became a major/urgent bug for us.

We are running [Drupal Rector](https://github.com/palantirnet/drupal-rector) against 10,000 projects (Drupal modules), to help them upgrade their code to the upcoming Drupal 9.

The Rector rules we implemented can already help 3,000 of these projects, as you can see in this list (rightmost column):
https://dev.acquia.com/drupal9/deprecation_status?name=&result=&error=&type=&group=&plan=&patch=available&op=Submit

But because Rector is running these additional code style rules, we cannot send the patches to all these projects that could benefit from it.

Here's an example with a mix of expected Rector rules and unexpected code style rules -
(lines 91, 99-260, 270, 278-395)
https://dispatcher.drupalci.org/job/phpstan/1462/artifact/phpstan-results/google_analytics.2.4.rector.patch/*view*/

(same file uploaded to github)
[google_analytics-drupal-module--rector-generated-patch.txt](https://github.com/rectorphp/rector/files/4524774/google_analytics-drupal-module--rector-generated-patch.txt)

TomasVotruba:
[...]

PR #3250, opened 2020-04-24
Add examples tests for issues we have seen.

damontgomery:
Right now the newline test fails, but the others pass. We have seen issues with the docblocks and `{`, so not sure why those aren't showing up.

See #3222 for discussion.

https://getrector.org/demo/5796cb68-d654-4ffb-9773-c70aeabe9bad#result shows the docblock issue that happens online, but not on the test. 🤷


tests/Standalone/ClassWithDifferentCodeStyleTest.php
```diff
@@ -0,0 +1,34 @@
...
+ * Class ClassWithDifferentCodeStyleTest
+ *
+ * We expect that Rector will not make changes to the code style for things it doesn't touch.
+ * In this example, the `{` should not be moved.
+ */
+final class ClassWithDifferentCodeStyleTest extends TestCase
+{
+    public function test(): void
+    {
+        $rectorStandaloneRunner = RectorStandaloneRunnerStaticFactory::create();
```
TomasVotruba, line 20: This static factory is designed for extreme/legacy/weird cases where rector cannot be run via command line.

Better use this approach:
https://github.com/rectorphp/rector/tree/master/tests/Issues

TomasVotruba:
It seems there are multiple issues at one PR. Better have separated PR per issues, as one can be solve in few minutes, while other in days.

TomasVotruba:
Closing as many mixed concepts. See #3311 for inspiration how to write such integraiton test.

PR #3255, opened 2020-04-25
Test with Drupal-Rector project to prevent BC breaks early on

shaal:
This is similar to a previous PR - https://github.com/rectorphp/rector/pull/3187, but for some reason the Drupal Rector test was removed later on.

Today I found out that rector-prefixed v0.7.19 is breaking Drupal Rector.
This PR has the same test from the previous PR + 2 temporary tests that show v0.7.18 is working well, but v0.7.19 breaks.


shaal:
This is the Drupal Rector test like in the previous PR (currently fails because of rector-prefix v.0.7.19:
https://github.com/rectorphp/rector/pull/3255/checks?check_run_id=617111805#step:8:6

These are the temporary tests:
1. Specifically requiring rector-prefix:0.7.18, test is successful:
https://github.com/rectorphp/rector/pull/3255/checks?check_run_id=617111887#step:6:4

1. Specifically requiring rector-prefix:0.7.19, test fail:
https://github.com/rectorphp/rector/pull/3255/checks?check_run_id=617111851#step:6:4

TomasVotruba:
The prefixed Rector is already tested here: https://github.com/rectorphp/rector-prefixed

I removed the test case, because it's responsibility of the external tool to test it's dependencies. 
It's like if Symfony would test all its bundles in Symfony CI.

PR #3281, opened 2020-04-30
moving to generic tag node

TomasVotruba:


packages/better-php-doc-parser/src/PhpDocNode/AbstractTagValueNode.php
```diff
@@ -293,4 +303,20 @@ private function createQuotedKeyPattern(?string $silentKey, string $key, string
...
+     * @param object|mixed[] $annotationOrItems
+     */
+    private function resolveItems($annotationOrItems): array
+    {
+        if (is_array($annotationOrItems)) {
+            return $annotationOrItems;
+        }
+
+        if (is_object($annotationOrItems)) {
+            return get_object_vars($annotationOrItems);
```
stephanvierkant, line 317: @TomasVotruba For some reason, this breaks Rector in my application. A `Symfony\Component\Routing\Annotation\Route` object returns `[]` in my case.

stephanvierkant: I'll try to find out why it fails in my application while the tests are OK, but it's hard to keep up with your development speed at this moment.

TomasVotruba: Thanks for such a fast testing, wow!

That's caused by Route annotation having private properties (made fake `public` in tests here for now) and `get_object_vars()` works only on public ones.

Do you have idea how to get into private properties of an object? Effectively.

TomasVotruba: This PR tests this bug, see: https://github.com/rectorphp/rector/pull/3285

Issue #3020, opened 2020-03-14
Getting Rector\Core\Exception\ShouldNotHappenException on php5.6 code


AntonioCS:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | v0.7.6             |
| Installed as   | prefixed Rector PHAR                      |

I am on PHP 7.4.3 (cli) (built: Feb 26 2020 12:17:01)
I am running this command: `vendor/bin/rector process . --set php74 --dry-run --verbose`
After rector finishes parsing all the files I get the following:
```
[refactoring] admin/CategorySelection/CategorySelection.php

In NodeNameResolver.php line 81:
                                                                 
  [Rector\Core\Exception\ShouldNotHappenException]               
  Pick more specific node than "PhpParser\Node\Expr\MethodCall"  
   
Exception trace:
  at phar:///projectName/common/vendor/rector/rector-prefixed/rector/packages/node-name-resolver/src/NodeNameResolver.php:81
 Rector\NodeNameResolver\NodeNameResolver->getName() at phar:///projectName/common/vendor/rector/rector-prefixed/rector/src/PhpParser/Node/Manipulator/ClassMethodPropertyFetchManipulator.php:47
 Rector\Core\PhpParser\Node\Manipulator\ClassMethodPropertyFetchManipulator->Rector\Core\PhpParser\Node\Manipulator\{closure}() at phar:///projectName/common/vendor/rector/rector-prefixed/rector/src/PhpParser/NodeTraverser/CallableNodeTraverser.php:42
 class@anonymous\phar:///projectName/common/vendor/rector/rector-prefixed/rector/src/PhpParser/NodeTraverser/CallableNodeTraverser.php:26$23b->enterNode() at phar:///projectName/common/vendor/rector/rector-prefixed/rector/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:113
 PhpParser\NodeTraverser->traverseNode() at phar:///projectName/common/vendor/rector/rector-prefixed/rector/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:196
 PhpParser\NodeTraverser->traverseArray() at phar:///projectName/common/vendor/rector/rector-prefixed/rector/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:85
[...]

PR #3287, opened 2020-05-01
[DX] drop confusing --only option to promote config

TomasVotruba:
## Why?

- it's confusing, people are unable to use it https://github.com/rectorphp/rector/issues/3020#issuecomment-621765729
- the `--only="Namespace\Class"` must be quoted right - slashes? single? double? + OS differences
- unable to configure, e.g. rename class method
- no CLI tool has this *feature*, so API is not intuitive

<br>

## Prefer Config

- configurable
- [YAML autocomplete](https://www.tomasvotruba.com/blog/2018/08/23/9-features-of-symfony-plugin-you-should-not-miss-in-gifs/#4-instant-service-autocomplete-in-yaml-yaml)
- easy to reproduce in [demo](getrector.org/demo)

```yaml
# rector.yaml
services:
    Namespace\Class: null
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,18 +213,6 @@ class SomeClass
 }
 ```
 
-### Filter Rectors
-
-If you have a configuration file for Rector including many sets and Rectors, you might want at times to run only a single Rector from them. The `--only` argument allows that, for example :
-
-```bash
-vendor/bin/rector process --set solid --only "Rector\SOLID\Rector\Class_\FinalizeClassesWithoutChildrenRector" src/
-```
-
-Will only run `Rector\SOLID\Rector\Class_\FinalizeClassesWithoutChildrenRector`.
-
-Please note that the backslash in the Rector's fully-qualified class name needs to be properly escaped (by surrounding the string in double quotes).
-
 ### Provide PHP Version
 
 By default Rector uses the language features matching your system version of PHP. You can configure it for a different PHP version:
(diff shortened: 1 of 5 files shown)


afilina:
In most of my projects, it makes no sense to maintain dozens of separate config files. I want to be able to programmatically apply a single rector or any combination. The CLI option gave me this ability.

Now, I need to write a script to chop up the full config and stitch it together on the fly in order to only keep the rectors that I want to apply. The DX on that just plummeted for me. `-r` was fine also, but now I'm out of _sane_ options.

TomasVotruba:
There is a better way to achieve the same result. 

I have similar way of using Rector, but instead of dozens of config files I have 1 `rector.yaml`.
As I use single rule, the refactoring is done and I simply use a new rule.

If the rules is applied in huge, I put it into `rector-ci.yaml` config, that checks the code periodically in CI.

This way code is:

- refactored step by step
- checked for any step-backs on per-commit bases without any manual work

afilina:
I don't get how you can apply a single rule if both `--only` and `--rule` have been dropped. I also used to have a single config, but that relied on the `--rule` option.

TomasVotruba:
Like this:

```yaml
# rector.yaml
services:
	OnlyRule: null
```

```yaml
# rector-ci.yaml
services:
   AllTheRules: null
   IHaveUsed: null
   Before: null
```

PR #3297, opened 2020-05-03
[Utils] PHPStan rule improvements

TomasVotruba:
Every Rector should have `@see` annotation with test case reference, so it's easy to jump in and create a fixture. 

Also will be useful in the future for: https://github.com/rectorphp/getrector.org/issues/155

<br>

:no_entry_sign: 

```php
final class NetteAssertToPHPUnitAssertRector extends AbstractRector
```

<br>

:+1: 

```php
/**
 * @see \Rector\NetteTesterToPHPUnit\Tests\Rector\Class_\NetteTesterClassToPHPUnitClassRector\NetteTesterPHPUnitRectorTest
 */
final class NetteAssertToPHPUnitAssertRector extends AbstractRector
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,11 +65,9 @@ It supports all versions of PHP from 5.2 and many open-source projects:
 
 ## How to Apply Coding Standards?
 
-The AST libraries that Rector uses aren't well-suited for coding standards, so it's better to let coding standard tools do that.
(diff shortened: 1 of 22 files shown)


TomasVotruba:
Thanks for inspiration in https://github.com/sabbelasichon/typo3-rector/pull/1387 to @sabbelasichon

Issue #3291, opened 2020-05-02
Does RemoveUnusedPrivatePropertyRector remove to much? 


tomasnorre:
# Question

In TYPO3 we have the `PublicPropertyDeprecationTrait` and `PublicMethodDeprecationTrait` which used two private Properties see example below. 

The rector rule `Rector\DeadCode\Rector\Property\RemoveUnusedPrivatePropertyRector` want to remove them as they are not directly used in the code, but by the trait. 

Is this something that I need to handle on my site, or is this something in Rector that doesn't fit TYPO3 (yet)? 

```
2) Classes/Service/ProcessService.php

    ---------- begin diff ----------
--- Original
+++ New
@@ -53,26 +53,6 @@
     use PublicPropertyDeprecationTrait;

     /**
-     * @var string[]
-     */
-    private $deprecatedPublicProperties = [
-        'queueRepository' => 'Using queueRepository is deprecated since 9.0.1 and will be removed in v10.x',
-        'crawlerController' => 'Using crawlerController is deprecated since 9.0.1 and will be removed in v10.x',
-        'countInARun' => 'Using countInARun is deprecated since 9.0.1 and will be removed in v10.x',
-        'processLimit' => 'Using processLimit is deprecated since 9.0.1 and will be removed in v10.x',
-        'verbose' => 'Using verbose is deprecated since 9.0.1 and will be removed in v10.x',
-    ];
-
-    /**
-     * @var string[]
-     */
-    private $deprecatedPublicMethods = [
-        'multiProcess' => 'Using ProcessService::multiProcess() is deprecated since 9.0.1 and will be removed in v10.x',
-        'reportItemStatus' => 'Using ProcessService::reportItemStatus() is deprecated since 9.0.1 and will be removed in v10.x',
-        'startRequiredProcesses' => 'Using ProcessService::startRequiredProcesses() is deprecated since 9.0.1 and will be removed in v10.x',
-    ];
-
-    /**
      * @var int
      */
     private $timeToLive;
    ----------- end diff -----------

Applied rules:

 * Rector\DeadCode\Rector\Property\RemoveUnusedPrivatePropertyRector
```

TomasVotruba:
I don't think it's possible to resolve traits using some private properties (too much nested magic).

Either use `@noRector` for these cases or better refactor to better design. Regardless tooling, reading such code would be WTF for me.

tomasnorre:
Thanks for your input, will see what I do.

[...]

PR #3299, opened 2020-05-03
Tests for @noRector 

tomasnorre:
This PR doesn't contain any fix yet, only tests to have it testable with the whole pipeline. 

As I don't have the pipeline setup on my site, I decided to do a PR even though no fix is included yet.

Tests for issue #3291


tests/Issues/Issue3291/config3291.yaml
```diff
@@ -0,0 +1,3 @@
+parameters:
+    sets:
+        - "dead-code"
```
TomasVotruba, line 3: Better use just the single rule if you it's 1 rule. It's easier to debug and maintain later.

tests/Exclusion/Check/Fixture/docblock-on-private-property.php.inc
```diff
@@ -0,0 +1,33 @@
...
+final class SomeController
+{
+    /**
+     * @noRector \Rector\DeadCode\Rector\Property\RemoveUnusedPrivatePropertyRector
+     */
+    private $anotherUnusedProperty = [];
+
+}
+
+?>
```
TomasVotruba, line 33: What's benefit of this class over the `fixture3291_keep_private_property.php.inc`? I think it can be removed.

It's better to have 1 test covering 1 issue.

tomasnorre: The is no benefit as I see, just understood that you wanted it like that. I can remove it.

tests/Issues/Issue3291/config3291.yaml
tomasnorre: is rule and sets the same in this context? 

Should I replace "deadcode" with "Rector\DeadCode\Rector\Property\RemoveUnusedPrivatePropertyRector" ? 

Not sure I get you.

tomasnorre:
Unittests are not working, will check.

tests/Issues/Issue3291/config3291.yaml
TomasVotruba: Sets is a set of rules. Here we test just 1 rule - `RemoveUnusedPrivatePropertyRector`.

So it can be registered as a service.

```yaml
services:
    FullNamespace...\RemoveUnusedPrivatePropertyRector: null
```

TomasVotruba: Check this: https://github.com/rectorphp/rector#b-standalone-rules

TomasVotruba:
Looks like unit tests are passing.

Could you fix PHPStan as well?

tomasnorre:
I don't get why the tests are ok, when the demo and my first unit-test are running red.

TomasVotruba:
That's the thing we need to figure out. At least we know Rector works. 
There will be some difference in your setup or the test is wrong.

I'd try:

1) using dead-code set in this tests's config again
2) try same test fixture code locally on your machine 

```bash
vendor/bin/rector process  tests/Issues/Issue3291/Fixture/fixture3291_keep_private_property.php.inc --set dead-code --dry-run
```

3) compare versions
4) in case your version is phar, add the same [test case there](https://github.com/rectorphp/rector-prefixed), there might be an issue in annotation process in PHAR

tomasnorre:
My setup is composer based, so should not be a big deal running with latest. 
I'll test this tomorrow somewhen and let you know. 

Thanks for helping me.

tomasnorre:
My PR is based on master, justed rebased this morning. I'll try later with the tests locally.

tomasnorre:
I'm having problems with the autoloading locally for my test. 

```
➜ bin/rector process --set dead-code --dry-run tests/Issues/Issue3291/Fixture/fixture3291.php.inc 
Rector 0.8.x-dev@d89b92f
Config file: rector.yaml

 3/3 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ! [NOTE] Ruleset contains 5 rules that need "--clear-cache" option to analyse full project                            

                                                                                                                       
 [ERROR] Could not process "tests/Issues/Issue3291/Fixture/fixture3291.php.inc" file, due to:                          
         "Analyze error: "Class Rector\Core\Tests\Issues\Issue3291\Fixture\Issue3291 was not found while trying to     
         analyse it - autoloading is probably not configured properly.". Include your files in "parameters >           
         autoload_paths".                                                                                              
         See https://github.com/rectorphp/rector#extra-autoloading".                                                   
                                                                           
```

I have tried adding following
```
#File config3291.yaml
parameters:
    autoload_paths:
        - 'Fixture/'

#File rector-ci.yaml 
autoload_paths:
        - 'tests/Issues/Issue3291/Fixture'
```

But both doesn't work for me.

TomasVotruba:
Include the file absolute path in config, or use -a option in cli with the same file

TomasVotruba:
I'm just trying this test case locally...

It seems one rule was missed in the config.

image

TomasVotruba:
Now it fails nicely, thanks! :+1:

tomasnorre:
Don't you squash commits on merge?

TomasVotruba:
Not really. I let contributors handle that on their side.

tomasnorre:
> Not really. I let contributors handle that on their side.

Then I'll make sure it squashed next time I contribute. Like that more my self.

TomasVotruba:
I do the same before sending to review. That would be great, thanks!

tomasnorre:
In my case I had multiple commits and had the PR more or less directly to get the pipeline benefits, so I simply didn't think about it as I thought it wasn't ready for merge.

TomasVotruba:
I see, no troubles. Each person has different coding habbits, we just need to sync.

tomasnorre:
I really don't get this. It's still failing in my project, both ci and locally. 
https://travis-ci.org/github/AOEpeople/crawler/jobs/683417146#L575
https://github.com/AOEpeople/crawler/pull/562/files#diff-bf56d9779b9916e69b145ec4d789c571R57

PR #3324, opened 2020-05-06
Depre Zend 1 to Symfony 4 set

### LINKED ISSUE #3323: Deprecate Zend to Symfony as unstarted project

TomasVotruba:
The migration project didn't start, so set is very basic and useles. 

Resurect in case of new project.

Closes #3323


TomasVotruba:
See https://github.com/rectorphp/rector/pull/1980

PR #3343, opened 2020-05-11
Alter command name

ddziaduch:
This change will suggest to use command given in the output of the vendor/bin/rector


TomasVotruba:
:+1:

PR #3344, opened 2020-05-11
Fix link and it's name

ddziaduch:
It was linking to non existing page.

I've also changed the link name to match my another pull request :)


TomasVotruba:
Thank you

PR #3345, opened 2020-05-11
Fix invalid path

ddziaduch:
Generator was not working

### DIFF
--- a/packages/rector-generator/src/Node/NodeClassProvider.php
+++ b/packages/rector-generator/src/Node/NodeClassProvider.php
@@ -26,6 +26,8 @@ final class NodeClassProvider
         $pathUsedAsStandaloneProject = __DIR__ . '/../../../../vendor/nikic/php-parser/lib/PhpParser/Node';
         $pathUsedAsDependency = __DIR__ . '/../../../../../../nikic/php-parser/lib/PhpParser/Node';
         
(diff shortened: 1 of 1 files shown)


packages/rector-generator/src/Node/NodeClassProvider.php
```diff
@@ -14,7 +14,7 @@ final class NodeClassProvider
     public function getNodeClasses(): array
     {
         $robotLoader = new RobotLoader();
-        $robotLoader->addDirectory(__DIR__ . '/../../../../vendor/nikic/php-parser/lib/PhpParser/Node');
+        $robotLoader->addDirectory(__DIR__ . '/../../../../../../nikic/php-parser/lib/PhpParser/Node');
```
TomasVotruba, line 17: It is working in the main repository. 
It's probably because it's installed as dependency. The best would be to support both locations.

ddziaduch: Got it @TomasVotruba 

Should I add condition here with `file_exists()` call?

TomasVotruba: Whatever works

TomasVotruba:
Btw, what do you need this for outside the main repository?

ddziaduch:
@TomasVotruba I wanted to create new Rector for my project :)

TomasVotruba:
Thanks :+1:

PR #3357, opened 2020-05-12
[PHPUnit] Fix CreateMockToCreateStubRector outside PHPUnit test case

### LINKED ISSUE #3157: createMockToCreateStub rector fixes too many methods
# Bug Report

| Rector version | 0.7.11             |
| Installed as   | rector/rector-prefixed                      |

I tried using the rector on https://github.com/sonata-project/SonataAdminBundle, and here is the result: https://github.com/sonata-project/SonataAdminBundle/pull/6043#discussion_r407072534

Could not reproduce it on the demo, because it would imply defining several interfaces
```php
<?php

declare(strict_types=1);

namespace Sonata\AdminBundle\Tests\Admin;

use PHPUnit\Framework\TestCase;
use Sonata\AdminBundle\Model\ModelManagerInterface;
use Sonata\DatagridBundle\Datagrid\DatagridInterface;

class AdminTest extends TestCase
{
    public function testGetFilterFieldDescription(): void
    {
        $modelManager = $this->createStub(ModelManagerInterface::class);
        $modelManager
            ->method('getNewFieldDescriptionInstance')
            ->willReturnCallback(static function () {
            });

        $datagrid = $this->createMock(DatagridInterface::class);
        $datagrid->expects($this->once())
            ->method('getPager');
    }
}

```
## Expected Behaviour

The change to `createStub` should not be done

TomasVotruba:
Closes #3157

### DIFF
--- a/rules/phpunit/tests/Rector/MethodCall/CreateMockToCreateStubRector/Fixture/skip_sonata_expects.php.inc
+++ b/rules/phpunit/tests/Rector/MethodCall/CreateMockToCreateStubRector/Fixture/skip_sonata_expects.php.inc
@@ -1,0 +1,17 @@
+<?php
+
+namespace Rector\phpunit\Tests\Rector\MethodCall\CreateMockToCreateStubRector\Fixture;
+
+use PHPUnit\Framework\TestCase;
+
+class SkipSonataExpects extends TestCase
+{
+    public function testItBehavesAsExpected(): void
+    {
+        $datagrid = $this->createMock(DatagridInterface::class);
+
+        $datagrid->expects($this->once())
+            ->method('getPager')
+            ->willReturn($pager);
+    }
+}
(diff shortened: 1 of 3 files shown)


rules/phpunit/tests/Rector/MethodCall/CreateMockToCreateStubRector/Fixture/skip_sonata_expects.php.inc
```diff
@@ -0,0 +1,17 @@
...
+{
+    public function testItBehavesAsExpected(): void
+    {
+        $datagrid = $this->createMock(DatagridInterface::class);
+
+        $datagrid->expects($this->once())
+            ->method('getPager')
+            ->willReturn($pager);
+    }
+}
```
TomasVotruba, line 17: @greg0ire I've just added test case for:
https://github.com/rectorphp/rector/issues/3157

And it is skipped.

If the problem still remains, we need either failing test case (like this fixture, just failing) or failing demo. We're unable to fix that without any replicable proof.

PR #3359, opened 2020-05-12
Rule "Rector\ClassConst\VarConstantCommentRector" keeps throwing "could not process" errors.

MetalArend:
If we don't redeclare the `$phpDocInfo` variable, it will fail at the next line, as calling a method on null will fail.


TomasVotruba:
Thanks

Issue #3336, opened 2020-05-09
Prevent rewrite of partial namespace imports


lstrojny:
Given this input

```php
<?php
use Symfony\Component\Validator\Constraints as Assert;

$constraint = new Assert\Blank();
```

Rector would apply this change
```diff
--- Original
+++ New
@@ -1,4 +1,5 @@
 <?php
+use Symfony\Component\Validator\Constraints\Blank;
 use Symfony\Component\Validator\Constraints as Assert;

-$constraint = new Assert\Blank();
+$constraint = new Blank();
```

How to prevent rector to refactor towards fully qualified imports?

TomasVotruba:
We need failing test case for that here: https://github.com/rectorphp/rector/tree/c915743c3e3a4eb21dc33204a4c485b0afba36c9/rules/coding-style/tests/Rector/Namespace_/ImportFullyQualifiedNamesRector/Fixture

lstrojny:
What can I do to help? Open a PR with the test case?

TomasVotruba:
Yes :+1:

lstrojny:
@TomasVotruba here we go: #3101

PR #3360, opened 2020-05-12
#3336 test case for partial use

lstrojny:
Adds a test-case for use statements that are referencing a partial namespace only.

Fails as expected:

```diff
1) Rector\CodingStyle\Tests\Rector\Namespace_\ImportFullyQualifiedNamesRector\ImportFullyQualifiedNamesRectorTest::test with data set #5 ('/Users/lstrojny/Projects/gith...hp.inc')
rules/coding-style/tests/Rector/Namespace_/ImportFullyQualifiedNamesRector/Fixture/partial_use.php.inc
Failed asserting that string matches format description.
--- Expected
+++ Actual
@@ @@
 <?php
 namespace Rector\CodingStyle\Tests\Rector\Namespace_\ImportFullyQualifiedNamesRector\Fixture;

+use PhpParser\Node\Stmt\Expression;
 use PhpParser\Node\Stmt;

-$class = Stmt\Expression::class;
+$class = Expression::class;

 ?>
```

### DIFF
--- a/rules/coding-style/tests/Rector/Namespace_/ImportFullyQualifiedNamesRector/Fixture/partial_use.php.inc
+++ b/rules/coding-style/tests/Rector/Namespace_/ImportFullyQualifiedNamesRector/Fixture/partial_use.php.inc
@@ -1,0 +1,17 @@
+<?php
+namespace Rector\CodingStyle\Tests\Rector\Namespace_\ImportFullyQualifiedNamesRector\Fixture;
+
+use PhpParser\Node\Stmt;
+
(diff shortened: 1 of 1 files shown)


TomasVotruba:
This is something different - a partial namespace. It should be imported, as there is no clear way to define rule, how many of name parts should be imported. It's not deterministic. Behavior in tests is correct.

In the issues, you mentioned alias: 

```php
use Something as SomethingElse;
```

We need test case for that

TomasVotruba:
Actually solved by #3680

PR #3374, opened 2020-05-15
Reduce size of docker images

hollodotme:
## Proposed changes

* Change base image to `php:7.4-cli-alpine`
* Add image cleanup that removes not needed software after installing dependencies

## Building images

```bash
docker build --no-cache --squash --pull \
            --target=rector \
            -t "hollodotme/rector" \
            -f ./Dockerfile .

# secured version
docker build --no-cache --squash --pull \
            --target=rector-secured \
            -t "hollodotme/rector-secured" \
            -f ./Dockerfile .
```

### Effect

```bash
$ docker images | grep rector
# With new Dockerfile
hollodotme/rector-secured ... 547MB
hollodotme/rector ... 184MB

# With current Dockerfile
rector/rector-secured ... 818MB
rector/rector ... 818MB
```

So the `rector` image is only 22,5% of the size of the current image (184MB vs. 818MB).

### Downsides

The only negative effect (as far as I can see) is that `iconv()` does not allow `//TRANSLIT` and/or `//IGNORE` in the conversion charset on alpine linux and the image emits the following PHP notice on startup:
```
$ docker run --rm -v $(pwd):/project hollodotme/rector process /project/src \
                --set php74 \
                --autoload-file /project/vendor/autoload.php \
                --dry-run

Notice: iconv(): Wrong charset, conversion from `UTF-8' to `ASCII//TRANSLIT//IGNORE' is not allowed in /rector/vendor/nette/utils/src/Utils/Strings.php on line 175
```

In case this is not an actual problem and the notice can be fixed, you could use this PR to reduce the size of the rector images.

### Further notes

This is just a suggestion, I am not a docker image professional. :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,8 @@
-FROM php:7.4-cli as rector
+FROM php:7.4-cli-alpine as rector
 WORKDIR /rector
 
-# Install php extensions
-RUN apt-get update && apt-get install -y \
-        git \
-        unzip \
-        g++ \
-        libzip-dev \
-    && pecl -q install \
-        zip \
-    && docker-php-ext-configure \
-        opcache --enable-opcache \
-    && docker-php-ext-enable \
-        zip \
-        opcache
+RUN apk update && apk upgrade && apk add --no-cache $PHPIZE_DEPS git libzip-dev \
+    && docker-php-ext-install zip opcache
 
 # Installing composer and prestissimo globally
 COPY --from=composer:latest /usr/bin/composer /usr/bin/composer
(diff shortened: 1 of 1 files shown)


TomasVotruba:
cc @JanMikes

JanMikes:
Hi @hollodotme overall MR looks good, i will do some testing today.

I heard about some problems with alpine, but i think they were related to fpm+nginx, so far personally i have not encounter any of these.

> Notice: iconv(): Wrong charset, conversion from `UTF-8' to `ASCII//TRANSLIT//IGNORE' is not allowed in /rector/vendor/nette/utils/src/Utils/Strings.php on line 175

Yeah, this is unfortunate and has to be fixed, im going to dive into it

TomasVotruba:
Alpine has a very bad reputation in using PHP with the ascii/string issues. This happens both in Nette and Symfony, see https://github.com/symfony/symfony/issues/37213#issuecomment-643154434
So switching to it is not possible.

> Add image cleanup that removes not needed software after installing dependencies

Could you send that in standalone PR?

PR #3386, opened 2020-05-19
Fix RenameAnnotationRector for null phpdoc

### LINKED ISSUE #3381: RenameAnnotationRector using hasByName() on null
# Bug Report

| Rector version | v0.7.26              |

rector fails with message: *[ERROR] Could not process "tests/SomeTest.php" file by "Rector\Renaming\Rector\Annotation\RenameAnnotationRector", due to:                                                                                                                     "Call to a member function hasByName() on null".*

```php
<?php
// file tests/SomeTest.php
namespace tests;

class SomeTest extends \PHPUnit\Framework\TestCase
{
    /*
     * This is a comment and not a docblock
     */
    public function testSomething()
    {
        $this->assertTrue(false, 'Expected failure!');
    }
}
```

running: `rector process --set phpunit70 --dry-run tests/SomeTest.php`

The error is not thrown if instead of a comment enclosed by `/* .. */` is replaced by `/** ... */`.

The error is also present if `// ...` is used

## Expected Behaviour

Do not fail on `RenameAnnotationRector` rule.

## More information

Related to `ExceptionAnnotationRector`, a similar issue is present.
On commit #3199 issue #3179 is solved by checking if the instance returned by `$node->getAttribute(AttributeKey::PHP_DOC_INFO);` is not null, I think that the same check should be placed for `RenameAnnotationRector`.

IMO this is an easy PR. If you would like me to solve the issue by adding the null check after `$node->getAttribute()` I can make it, but I'm not familiar with rector development. Please advise.

PD: Thanks for rector and all this work!

eclipxe13:
Fixes #3381

Including the fixture produces error:

```
$ phpunit rules/renaming/tests/Rector/Annotation/RenameAnnotationRector/RenameAnnotationRectorTest.php 
PHPUnit 9.1.4 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.3.15-3
Configuration: /home/eclipxe/work/github/rector/phpunit.xml

.E                                                                  2 / 2 (100%)

Time: 00:02.103, Memory: 82.36 MB

There was 1 error:

1) Rector\Renaming\Tests\Rector\Annotation\RenameAnnotationRector\RenameAnnotationRectorTest::test with data set #1 ('/home/eclipxe/work/github/rec...hp.inc')
Error: Call to a member function hasByName() on null

/home/eclipxe/work/github/rector/rules/renaming/src/Rector/Annotation/RenameAnnotationRector.php:106
...
```

The null check and return make the test pass again.

When creating the fixture to test I notice that it only fails when the class extends `\PHPUnit\Framework\TestCase` and classname ends with `Test`.

I didn't find a way to reproduce the error without the previous condition.

PD: running `composer complete-check` changes `compiler/src/HttpKernel/RectorCompilerKernel.php`. I didn't include it.

Please, let me know if you would like some changes.

### DIFF
--- a/rules/renaming/src/Rector/Annotation/RenameAnnotationRector.php
+++ b/rules/renaming/src/Rector/Annotation/RenameAnnotationRector.php
@@ -90,8 +90,11 @@ PHP
         /** @var Class_ $class */
         $class = $node->getAttribute(AttributeKey::CLASS_NODE);
 
-        /** @var PhpDocInfo $phpDocInfo */
+        /** @var PhpDocInfo|null $phpDocInfo */
         $phpDocInfo = $node->getAttribute(AttributeKey::PHP_DOC_INFO);
+        if ($phpDocInfo === null) {
+            return null;
+        }
 
         foreach ($this->classToAnnotationMap as $type => $annotationMap) {
             /** @var string $type */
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thank you :+1:

PR #3394, opened 2020-05-20
Keep an empty docblock as is

shaal:
To maintain the assumption, that rector would only change what it was instructed to, when there's an empty docblock, we want to preserve it.

Removing empty docblocks should be a specific Rector rule that people can opt-in to run, instead of automatically removed every time Rector runs.


shaal:
I think the PR is failing because now there is nothing that calls the function `isPhpDocNodeEmpty()`.

TomasVotruba:
One of them is coding standard and other's are failing tests.

shaal:
Yikes, 43 tests are failing https://github.com/rectorphp/rector/pull/3394/checks?check_run_id=695669034

shaal:
@TomasVotruba how would you like to resolve this?
Should this PR update all the failing tests to keep the empty docblocks as is?

TomasVotruba:
When the docblock is:

```php
/**
 * @var int
 */
```

And then `@var` is removed by some rule (e.g. PHP 7.4 typed properties), the output canno be:

```
/**
 */
```

but:

```
```

<br>

So the docblock must be cleaned from start/end **if changed**. If there is no change, it can be printed in original way.

TomasVotruba:
This is surpassed by https://github.com/rectorphp/rector/pull/5251

Thank you for your work :+1:

shaal:
Thank you @TomasVotruba

Issue #3352, opened 2020-05-12
The string '<?php  ?>' is being mistakenly removed


shaal:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | v0.7.25 (prefixed)              |
| Installed as   |  prefixed Rector PHAR                      |

## Minimal PHP Code Causing Issue

https://getrector.org/demo/f1bc0f86-1bfa-4527-80de-3ab3839347f0#result

## Expected Behaviour

Rector should leave that string `'<?php  ?>'` as is and not remove it.

TomasVotruba:
This is part of php-parser empty print cleanup.

It's dead code, that doesn't do any behavior and should not be in your code.

What troubles it causes you?

shaal:
The assumption is that Rector will not change anything in the code, unless it was specifically instructed to do so through rector.yml

`<?php ?>' is used in various Drupal projects, when a string is sent to Drupal multilingual system (the t() function), like this:
```
$description .= ' ' . $this->t('If the PHP option is chosen, enter PHP code between %php. Note that executing incorrect PHP code can break your Drupal site.', ['%php' => '<?php ?>']);
```

List of projects using it:
http://grep.xnddx.ru/search?text=%3C%3Fphp+%3F%3E&filename=

TomasVotruba:
I see. I'm not sure how to solve this in the printer though.

shaal:
PR #3395 resolved this issue.

PR #3395, opened 2020-05-20
Keep empty php code as is

shaal:
(Resolving the same type of issue as in https://github.com/rectorphp/rector/pull/3394)
To maintain the assumption, that rector would only change what it was instructed to, when there's an empty php code `<?php ?>`, we want to preserve it.
You can see an example of when it's necessary here: https://github.com/rectorphp/rector/issues/3352#issuecomment-631691576

Removing empty php code should be a specific Rector rule that people can opt-in to run, instead of automatically being removed every time Rector runs.

### DIFF
--- a/src/PhpParser/Printer/BetterStandardPrinter.php
+++ b/src/PhpParser/Printer/BetterStandardPrinter.php
@@ -82,13 +82,8 @@ final class BetterStandardPrinter extends Standard
         // php attributes
         $content = $this->contentPhpAttributePlaceholderReplacer->decorateContent($content);
 
-        // remove dead <?php structures
-        $clearContent = Strings::replace($content, '#\<\?php(\s)\?\>\n?#');
-
-        $clearContent = Strings::replace($clearContent, '#\<\?php(\s+)\?\>#');
-
         // keep EOL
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Just for reference: it was added for CakePHP to Symfony rules, that are no longer present: https://github.com/rectorphp/rector/commit/81531b26696179fa8e7c673dd328d6541739c86c#diff-560bcf8e3b306989b3cb6718d0403d0c
 
Thanks :+1:

shaal:
@TomasVotruba It seems that this PR is not merged yet to `master`, due to: 
`kodiakhq: status 🛑 cannot merge (missing automerge_label: 'automerge')`
 https://github.com/rectorphp/rector/runs/694553608

TomasVotruba:
Why do you think so? 

I see it right there:

image

Also commit is in the `master`
https://github.com/rectorphp/rector/commit/b01eb496d92f3574260a3a7af729e5efb51776a2

shaal:
Sorry, must be something in my internet connection :(

It kept showing me `master` without this PR merged to it, but now I see the latest `master` with this PR merged. All good now :)

shaal:
Closes #3352

PR #3407, opened 2020-05-24
Fix SimplifyArraySearchRector w.r.t. "strictness"

### LINKED ISSUE #3404: SimplifyArraySearchRector is incorrect w.r.t. "strictness"
# Bug Report

| Rector version | v0.7.26              |

Rector does the following change:

```diff
-    array_search($needle, $haystack) !== false
+    in_array($needle, $haystack, true)
```

that adds `, true` which causes a different result e.g. for `$needle = 42; $haystack = ['42'];`:

```diff
-    true
+    false
```

(https://3v4l.org/ZM5aI)

https://getrector.org/demo/e226c0b9-4efd-4253-aac5-d87bf457e568

## Expected Behaviour

Rector should do the following changes (note: each variable might also be a literal value):

```diff
-    array_search($needle, $haystack) !== false
+    in_array($needle, $haystack)
```

```diff
-    array_search($needle, $haystack, $strict) !== false
+    in_array($needle, $haystack, $strict)
```

and their negative counterparts:

```diff
-    array_search($needle, $haystack) === false
+    !in_array($needle, $haystack)
```

```diff
-    array_search($needle, $haystack, $strict) === false
+    !in_array($needle, $haystack, $strict)
```

i.e. preserve the optional third argument (https://3v4l.org/E44t1).

**Moreover**, it should _not_ do this refactor for comparisons other than `!==` and `===` (e.g. `!=` or `==`) nor for values other than `false` (e.g. `true` or `0`). Akin to StrposToStringsContainsRector.

References:
- https://www.php.net/manual/en/function.array-search.php
- https://www.php.net/manual/en/function.in-array.php

guilliamxavier:
Fixes #3404


TomasVotruba:
Thank you :+1:

PR #3415, opened 2020-05-24
[CodingStyle] UnderscoreToCamelCaseVariableAndPropertyNameRector

### LINKED ISSUE #3414: [CodingStyle] under_score to camelCase variable names
```diff
-$value_name
+$valueName
```

<br>

See failed cs attemps:

- https://github.com/squizlabs/PHP_CodeSniffer/issues/1228
- https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2362

TomasVotruba:
Closes #3414

### DIFF
--- a/config/set/coding-style/coding-style.yaml
+++ b/config/set/coding-style/coding-style.yaml
@@ -35,3 +35,4 @@ services:
             pi: M_PI
     Rector\CodingStyle\Rector\Function_\CamelCaseFunctionNamingToUnderscoreRector: null
     Rector\CodingStyle\Rector\Use_\SplitGroupedUseImportsRector: null
+    Rector\CodingStyle\Rector\Variable\UnderscoreToPascalCaseVariableAndPropertyNameRector: null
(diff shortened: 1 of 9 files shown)


popovserhii:
@TomasVotruba As I understand the `UnderscoreToCamelCaseVariableAndPropertyNameRector` property is not available any more in [the latest release](https://github.com/rectorphp/rector/releases/tag/0.10.3)? What should we use instead?

TomasVotruba:
Yes, that's correct. These rules are more about subjective naming conventions, that should be handled by  coding standards.
See references coding standard rules https://github.com/rectorphp/rector/issues/3414

popovserhii:
This rule was only the reason why I've started to use Rector and was waiting for 4 months while two related issues will be fixed ( #4620, #5392), and now I need to discover another solution.

TomasVotruba:
That was my mistake, that original upgrade/refactoring shifted way too close to coding standards. I'm sorry about that.

The phpdoc fix is still in progress, see https://github.com/rectorphp/rector/issues/4334

josep11:
Hello. It's not on main, right?

samsonasik:
@josep11 yes, it removed, if you need it, you can check on `CodeIgniter4` custom rule for it with customize on based on CodeIgniter4 coding standard, see https://github.com/codeigniter4/CodeIgniter4/blob/develop/utils/Rector/UnderscoreToCamelCaseVariableNameRector.php

PR #3426, opened 2020-05-26
introducing fixture for choises with ceros

bitgandtter:
validate that does not change the values

originally the issue was https://getrector.org/demo/bebc3973-a15e-4843-b478-a510003b45d7#result


bitgandtter:
@TomasVotruba any guidance on where should I look to make the fix? Sorry, I don know the whole architecture of the project.

TomasVotruba:
Great question!

The tag is represented by [`AssertChoiceTagValueNode`](https://github.com/rectorphp/rector/blob/master/packages/better-php-doc-parser/src/PhpDocNode/Symfony/Validator/Constraints/AssertChoiceTagValueNode.php).

It has stored values in `$items` propeperty, that is printed back to string code by [`printedItems()`](https://github.com/rectorphp/rector/blob/92ddbfc6e360a98a9a07de5946d6b74513e1fdcd/packages/better-php-doc-parser/src/PhpDocNode/AbstractTagValueNode.php#L88-L95) method.

As `"choices"` is an array of values, it is [printed by `printArrayItem()` method](https://github.com/rectorphp/rector/blob/92ddbfc6e360a98a9a07de5946d6b74513e1fdcd/packages/better-php-doc-parser/src/PhpDocNode/AbstractTagValueNode.php#L97-L133). I assume it will be something wrong there with regular expressions.

TomasVotruba:
Thank you for the test case. It helped me to find the bug more quickly

I'll fix this locally.

PR #3453, opened 2020-05-31
Future-proof ScopeFactory

ondrejmirtes:
As MutatingScope constructor can change often and is not covered by backwards compatibility promise, this is a much better way to construct it. As you can notice, it's just a bunch of code deleted as you can simply call `$this->phpStanScopeFactory->create()` for the same effect.

Thanks.


TomasVotruba:
I see, thank you Ondra.

Issue #3127, opened 2020-04-02
[PHP 8.0] [Meta] 


TomasVotruba:
# New Accepted Features 

### `get_resource_id()`

- PR: https://github.com/php/php-src/pull/5427

### `mixed` type

- RFC: https://wiki.php.net/rfc/mixed_type_v2

### constructor promotion

- RFC: https://wiki.php.net/rfc/constructor_promotion

<br>

## Generic sources

- https://www.reddit.com/r/PHP/comments/g5w52s/php_8_will_add_union_types_the_jit/
- https://wiki.php.net/rfc/deprecations_php_8_0

<br>

### 1. `str_starts_with()` / `str_ends_with()`  :heavy_check_mark: 

- RFC: https://wiki.php.net/rfc/add_str_starts_with_and_ends_with_functions
- Issue: https://github.com/rectorphp/rector/issues/3221

### 2. `str_contains()`  :heavy_check_mark: 

- RFC: https://wiki.php.net/rfc/str_contains
- Issue: https://github.com/rectorphp/rector/issues/3025

### 3. `Stringable` interface for classes with `__toString()`  :heavy_check_mark: 

- RFC: https://wiki.php.net/rfc/stringable

### 4. Attributes v2 :heavy_check_mark: 

- RFC: https://wiki.php.net/rfc/attributes_v2

### 5. `static` type  :heavy_check_mark: 

- RFC: https://wiki.php.net/rfc/static_return_type
- in e.g. Rector\TypeDeclaration\Rector\FunctionLike\ReturnTypeDeclarationRector

### 6. `::class` on objects  :heavy_check_mark: 

- RFC: https://wiki.php.net/rfc/class_name_literal_on_object

### 7. `get_debug_type` :heavy_check_mark: 

- RFC: https://wiki.php.net/rfc/get_debug_type

### 8. `token_get_all` to `PhpToken`  :heavy_check_mark: 

- RFC: https://wiki.php.net/rfc/token_as_object

### 9. `catches` without $variable   :heavy_check_mark: 

- RFC: https://wiki.php.net/rfc/non-capturing_catches

### 10. `match` :heavy_check_mark: 

- RFC: https://wiki.php.net/rfc/match_expression_v2
- Rector PR: https://github.com/rectorphp/rector/pull/3569

<br>

### New features, without no path to upgrade

- https://wiki.php.net/rfc/throw_expression

staabm:
[...]

PR #3458, opened 2020-06-02
[PHP 8.0] Add TokenGetAllToObjectRector

TomasVotruba:
Ref #3127 

/cc @staabm This is going to be a bit wobbly, as Rector doesn't have API for tracking variable flow, but it will do the job.

### DIFF
--- a/rules/php80/src/Rector/FuncCall/TokenGetAllToObjectRector.php
+++ b/rules/php80/src/Rector/FuncCall/TokenGetAllToObjectRector.php
@@ -1,0 +1,172 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Php80\Rector\FuncCall;
+
+use PhpParser\Node;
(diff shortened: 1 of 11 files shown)


rules/php80/src/Rector/FuncCall/TokenGetAllToObjectRector.php
```diff
@@ -0,0 +1,192 @@
...
+
+            if ($node->else === null) {
+                return null;
+            }
+
+            // 1. is array
+            $this->ifArrayTokenManipulator->refactorIsArrayOnToken($node->stmts, $singleToken);
+
+            // 2. is not array
+            $this->ifArrayTokenManipulator->refactorIsNotArrayOnToken($node->else->stmts, $singleToken);
```
TomasVotruba, line 146: Try phpstan type of `$token` variable in here

PR #3460, opened 2020-06-03
Update docker commands to remove container on exit

codereviewvideos:
Just a small tweak to ensure there are no left over containers when running the respective Docker commands. 

Thanks for making / sharing Rector.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ See [the contribution guide](/CONTRIBUTING.md).
 You can run Rector on your project using Docker:
 
 ```bash
-docker run -v $(pwd):/project rector/rector:latest process /project/src --set symfony40 --dry-run
+docker run --rm -v $(pwd):/project rector/rector:latest process /project/src --set symfony40 --dry-run
 
 # Note that a volume is mounted from `pwd` (the current directory) into `/project` which can be accessed later.
 ```
@@ -293,7 +293,7 @@ docker run -v $(pwd):/project rector/rector:latest process /project/src --set sy
 Using `rector.yaml`:
 
 ```bash
-docker run -v $(pwd):/project rector/rector:latest process /project/app \
+docker run --rm -v $(pwd):/project rector/rector:latest process /project/app \
 --config /project/rector.yaml \
 --autoload-file /project/vendor/autoload.php \
 --dry-run


TomasVotruba:
cc @janmikes

TomasVotruba:
Could you explain my why is this good?

I'm just practical Docker user, no expert.

a6software:
Sure - the way I see it is it's good housekeeping to remove exited containers after they run. 

I want to keep around the Docker image itself, but as an example, if I wanted to run rector to convert phpspec to phpunit for a project, I'd probably only do this once. After that, I'd never need that particular container again.

If I do need to run the command again, I can re-run the `docker run ...` command, re-using the local Docker rector image, but get a new container. And again, I'd probably want to remove that particular container instance when that command has finished, too.  

If you do a `docker ps -a` you may find a bunch of left over containers from previous times you have run a `docker run...` command without the `--rm` flag. You would need to clear these up yourself ('bulk' command pretty easy to find on google - e.g. https://www.digitalocean.com/community/tutorials/how-to-remove-docker-images-containers-and-volumes).

Personal preference I suppose. If I'm going to re-run the container many times I would not remove it each time. But for one off commands, personally I always remove them. 

Feel free to disregard this PR, I won't be offended ! :D

TomasVotruba:
I see, thanks for explaining :+1: 

Are there any performance changes? I support with removing the container, it will take longer to run it repeatadly.

codereviewvideos:
There are (almost) no performance issues, to the very best of my knowledge. 

To clarify, removing the container does not remove the `rector/rector` Docker image from your system. 

When you use `docker run` it creates a container from the image.

All this change does is remove the container after the command running inside the container finishes executing.

The image the container was created from remains on your system, so other containers can still be created from the image very quickly.

The only performance issue I am aware of is it would be marginally more performant to resume an exited container than create a new one, but on any modern PC that difference should be essentially unnoticeable.  

To remove the `rector/rector` image, you would need something like `docker image rm rector/rector:latest`, and that would mean the next time you created a container from the image then yes, you would need to download the image again which could take a while.

TomasVotruba:
I see :) I feel really smart thanks to your explanation. It makes sense, is clear, short and only relevant information I need. Very rare combination :+1: 

Thanks for taking time and effort to write all this.

JanMikes:
Hi, it makes sense to me as I used this locally anyways (but I am a power user and knows that --rm exists 😄).

I did not want to mention it in docs to not confuse and keep the command as simple as possible, but still, it is good practice and i see no real reason to not mention `--rm`, i checked and phpstan has it in docs too, for example.

Issue #2271, opened 2019-11-06
running composer complete-check on windows


orklah:
| Subject        | Details                                                      |
| :------------- | :----------------------------------------------------------- |
| Rector version | current master         |
| PHP version    | PHP 7.3                       |
| Full Command   | composer complete-check --debug  | 

## Current Behaviour 

C:\Users\username\PhpstormProjects\rector>composer complete-check
> vendor/bin/ecs check bin packages src tests utils --ansi
'vendor' n’est pas reconnu en tant que commande interne
ou externe, un programme exécutable ou un fichier de commandes.
Script vendor/bin/ecs check bin packages src tests utils --ansi handling the check-cs event returned with error code 1

The error message is in french. It can be roughly translated to
'vendor' was not recognized as an internal or external command nor an executable program or command file

While trying to fiddle in this, I managed to make it work for ECS by replacing 
"vendor/bin/ecs" by "vendor\bin\ecs.bat" but I'm not sure how this can be fixed within composer...

## Expected Behaviour

Being able to launch the command or having a fallback command to run all the tests and code style to contribute

TomasVotruba:
This one is simple. Just update the `composer.json` paths like this:

```diff
"vendor/bin/ecs
"ecs
```

For all commands there. If it helps, please send PR with this change

orklah:
You're right, it works fine. I'll try to do the PR

Is it normal that I get more than a thousand suggested changes from ecs?

Here are the checkers applied:
 * PhpCsFixer\Fixer\Whitespace\LineEndingFixer
 * Symplify\CodingStandard\Fixer\Commenting\BlockPropertyCommentFixer

TomasVotruba:
> Is it normal that I get more than a thousand suggested changes from ecs?

Probably line endings from Windows. On Windows it enforces windows endings.

I quick workaround is to exclude these 2 files locally in `ecs.yaml`

orklah:
I'll try to see tomorrow if I can make it OS agnostic. Ideally, everything should be able to be run on windows

TomasVotruba:
Agreed. That would be amazing :+1:

TomasVotruba:
@orklah I just found out Travis supports Windows for a couple of months.
I'm trying to add Windows testsuite here: https://github.com/rectorphp/rector/pull/2278

orklah:
Nice! Just checked, the path seems to work! I'll take a look to the line feed issue.

orklah:
[...]

PR #3465, opened 2020-06-03
Don't remove null if the property is type hinted to nullable

### LINKED ISSUE #3464: RemoveNullPropertyInitializationRector should not remove null on typed properties
# Bug Report

| Rector version | Rector 0.8.x-dev@5c82ca6              |
| Installed as   | composer dependency                 |

Rector should not remove `= null` from typed properties, as assigning null to typed properties prevents the property not initialized error in PHP 7.4.

Without the `= null`, you cannot access the variables before giving them a value due to the error talked about here: https://stitcher.io/blog/typed-properties-in-php-74#uninitialized

You could give them a value in the constructor, sure, but that seems to go against other rules Rector has, like giving every property typed to an array an initial value of [].

I am opening this issue because of your message to me here: https://github.com/rectorphp/rector/pull/1880#issuecomment-621330867

https://getrector.org/demo/5693607b-b520-4610-948c-2459a1f2b7f4#result

The corrected code will error if you call `getSomeProperty` whereas it would not error in the uncorrected version.

## Expected Behaviour

Rector should skip it.

UFTimmy:
Fixes #3464

This is my first time delving into a Rector, so I'm not sure of the internals. I assume that the previous node should always be defined, but I'd be happy to add a guard if not. The example I found in another Rector didn't check first, so I didn't check here.


TomasVotruba:
Looks good :+1: 

Could you also add test fixture, that shows the problem?

TomasVotruba:
Any updates?

UFTimmy:
I hope to have a look at it early next week. I haven't done much with tests previously, but I'll give it a shot.

TomasVotruba:
Great :+1: Let me know if you need any help

UFTimmy:
I tried briefly, but ran into this issue here: https://github.com/rectorphp/rector/issues/2271

So at this time, all I can contribute is the fix, and not the tests, unfortunately.

TomasVotruba:
Before/after fixture should be enough: https://github.com/rectorphp/rector/tree/master/rules/dead-code/tests/Rector/Property/RemoveNullPropertyInitializationRector/Fixture

Just add single file with:

```
<?php

before php code

?>
------
<?php

after php code

?>
```

If there is **no changed desired**, just:

```
<?php

before php code
```

UFTimmy:
How do I run the fixtures? I thought they would run as part of the complete-check process.

TomasVotruba:
I always test just the single rule:

```
vendor/bin/phpunit rules/dead-code/tests/Rector/Property/RemoveNullPropertyInitializationRector/RemoveNullPropertyInitializationRectorTest.php
```

UFTimmy:
I appreciate your patience with this, I know this is something you could have done yourself many times faster than trying to help me.

When I run just that one rule I get the following:

```
<path>\rector-patch-1>vendor\bin\phpunit.bat rules/dead-code/tests/Rector/Property/RemoveNullPropertyInitializationRector/RemoveNullPropertyInitializationRectorTest.php
PHPUnit 9.2.2 by Sebastian Bergmann and contributors.

PHP Fatal error:  Cannot declare class Rector\DeadCode\Tests\Rector\Property\RemoveNullPropertyInitializationRector\Fixture\SomeClass, because the name is already in use in <path>\rector_temp_tests\original_16286651993a54b_fixture.php on line 30
-----

Fatal error: Cannot declare class Rector\DeadCode\Tests\Rector\Property\RemoveNullPropertyInitializationRector\Fixture\SomeClass, because the name is already in use in <path>\rector_temp_tests\original_16286651993a54b_fixture.php on line 30
```

TomasVotruba:
> I appreciate your patience with this, I know this is something you could have done yourself many times faster than trying to help me.

I'm trying, thanks for noticing. I know it takes time to get first step into new project that you never saw before. Legacy or open-source, that's same.

> When I run just that one rule I get the following:

That class name should be unique per fixture file. It's good practise to name class as the topic it tests. E.g. `SkipNullableIfTypedProperty`

rules/dead-code/tests/Rector/Property/RemoveNullPropertyInitializationRector/TypedPropertiesRemoveNullPropertyInitializationRectorTest.php
```diff
@@ -0,0 +1,36 @@
...
+use Iterator;
+use Rector\Core\Testing\PHPUnit\AbstractRectorTestCase;
+use Rector\Core\ValueObject\PhpVersionFeature;
+use Rector\DeadCode\Rector\Property\RemoveNullPropertyInitializationRector;
+
+final class TypedPropertiesRemoveNullPropertyInitializationRectorTest extends AbstractRectorTestCase
+{
+    /**
+     * @dataProvider provideData()
+     */
```
TomasVotruba, line 16: To run this test only on PHP 7.4, just add:

```suggestion
     * @requires PHP >= 7.4
     */
```

See [PHPUnit documentation of this feature](https://phpunit.readthedocs.io/en/9.2/incomplete-and-skipped-tests.html#skipping-tests-using-requires).

TomasVotruba: Great job with separated test case!

TomasVotruba:
Thank you :+1: 

Merged locally

PR #3488, opened 2020-06-07
PHPStan compatibility fixes

ondrejmirtes:
PHPStan 0.12.26 makes user's lives easier by a mile. See [the release notes](https://github.com/phpstan/phpstan/releases/tag/0.12.26) and [the accompanying article](https://phpstan.org/blog/zero-config-analysis-with-static-reflection).


TomasVotruba:
Thanks for update Ondra :+1: 

Does this mean Rector can use static reflection to get all static types too? 
If so, what should be changed in current PHP-reflection to make it work?

ondrejmirtes:
I’ll open a new issue to discuss that, give me some time to process everything 😊

PR #3494, opened 2020-06-08
Fix typo in rector_rules_overview.md

berezuev:
Change classname in ReturnThisRemoveRector example.
Looks like it should be the same with config in example above.


TomasVotruba:
This file is generated from `getRectorDefinition()`, so this needs to be updated in the responsible Rector class too

berezuev:
Here it is!

TomasVotruba:
Thank you :+1:

PR #3503, opened 2020-06-09
I found a bug with the ChangeNestedForeachIfsToEarlyContinueRector

derrickschoen:
When there is a comment inside the if statement before the brace, it misses a brace creating the early continue statement.

https://getrector.org/demo/9757fa4a-ed15-4130-8bdc-ae150305360e


TomasVotruba:
Thanks :+1:

PR #3512, opened 2020-06-11
Don't type hint to traits

### LINKED ISSUE #3431: Rector tries to type hint to traits
# Bug Report

| Rector version | 0.8.x-dev@7286f34              |
| Installed as   | Get Rector demo                |

When you have a parameter that is type hinted via docblock to a trait then Rector tries to bring that type hint to a trait into the function signature. Traits cannot be type hinted in PHP.

I understand the correct way is to create an interface and type hint to that, but Rector should still be aware that the parameter is a trait and not try to type hint to it.

Additionally, it appears this type hint is being added from a Rector in the type-declaration set, but as you can see from the get rector demo, the rector making the change isn't reporting itself. It's the same when I run it locally.

https://getrector.org/demo/ea869f83-9113-44e5-8268-b22228847a27#result

```php
<?php

declare(strict_types=1);

trait SomeTrait
{
    public function someMethod() : int
    {
        return 5;
    }
}

final class DemoFile
{
    /**
     * @param SomeTrait $someTrait
     */
    public function run($someTrait) : int
    {
        return $someTrait->someMethod();
    }
}
```

## Expected Behaviour

Do nothing.

UFTimmy:
Fixes #3431

I'm not sure about my logic here of checking if it's a trait at all. I tried to use the inferred type's getClassReflection method, but that was null, and I felt like it should not be.

### DIFF
--- a/rules/type-declaration/src/Rector/FunctionLike/ParamTypeDeclarationRector.php
+++ b/rules/type-declaration/src/Rector/FunctionLike/ParamTypeDeclarationRector.php
@@ -10,10 +10,12 @@ use PhpParser\Node\Param;
 use PhpParser\Node\Stmt\ClassMethod;
 use PhpParser\Node\Stmt\Function_;
 use PHPStan\Type\MixedType;
+use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thank you :+1:

PR #3535, opened 2020-06-16
[Experimental] Add safe_types option

TomasVotruba:
Do you want to use only explicit PHP type declaration? Enable `safe_types`:

```yaml
# rector.yaml
parameters:
    safe_types: true
```

Then, **docblocks are skipped** and only pure PHP is source of type truth:

```diff
 <?php

 class ValueObject
 {
     public $value;

-    public $count;
+    public int $count;

     /**
      * @param string $value
      */
     public function __construct($value, string $count)
     {
         $this->value = $value;
         $this->count = $count
     }
 }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,67 @@ parameters:
     php_version_features: '7.2' # your version is 7.3
 ```
 
+### Safe Types
+
+In defualts setting:
+
+```yaml
+# rector.yaml
+parameters:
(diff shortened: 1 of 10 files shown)


orklah:
I think you meant to have int $count in signature :)

TomasVotruba:
Yes ;) I'll fix that, thanks

PR #3552, opened 2020-06-18
[PHP] Handle arrow functions in AddDefaultValueForUndefinedVariableRector 

fsok:
* do not try to set variables in arrow functions, fixes error `Undefined property: PhpParser\Node\Expr\ArrowFunction::$stmts at AddDefaultValueForUndefinedVariableRector.php:118`
* consider that arrow functions inherit the parent scope


TomasVotruba:
Thank you :+1:

PR #3559, opened 2020-06-19
"$this->"" needs to become "static::" instead of "self::"

derrickschoen:
If a class could be inherited, "$this->"" needs to become "static::" instead of "self::"

### DIFF
--- a/rules/php70/tests/Rector/MethodCall/ThisCallOnStaticMethodToStaticCallRector/Fixture/fixture.php.inc
+++ b/rules/php70/tests/Rector/MethodCall/ThisCallOnStaticMethodToStaticCallRector/Fixture/fixture.php.inc
@@ -24,7 +24,7 @@ class SomeClass
 {
     public static function run()
     {
-        self::eat('something');
+        static::eat('something');
     }
 
     public static function eat(string $thing)
--- a/rules/php70/tests/Rector/MethodCall/ThisCallOnStaticMethodToStaticCallRector/Fixture/fixture.php.inc
+++ b/rules/php70/tests/Rector/MethodCall/ThisCallOnStaticMethodToStaticCallRector/Fixture/fixture.php.inc
@@ -30,7 +30,7 @@ class SomeRunnable implements RunnableInterface
     {
         $expectedString = "child";
         $c = new SomeChild();
-        return $c->speak() === $expectedString;
+        return $c->speak();
     }
 }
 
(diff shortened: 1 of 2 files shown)


TomasVotruba:
I need more information about this.

derrickschoen:
> I need more information about this.

```
<?php

class SomeClass
{
    public function run()
    {
        $this->eat();
    }
    public static function eat()
    {
        echo "parent";
    }
}

class SomeChild extends SomeClass
{

    public static function eat()
    {
        echo "child";
    }
}

$c = new SomeChild();
$c->run();
```

This echos "child". 

If I replace  ```$this->eat();``` with ```self::eat();```: it echos "parent" which is different than the original output.


However, if I replace  ```$this->eat();``` with ```static::eat();```: it matches the output of the original code: "child"

TomasVotruba:
I see. We should have this covered with a test case.

Could you update the test with Runnable interface? 
https://github.com/rectorphp/rector/blob/f2eb777f1da530db0c5f9201cf6d6248301505bb/rules/code-quality/tests/Rector/FuncCall/ArrayKeysAndInArrayToArrayKeyExistsRector/Fixture/fixture.php.inc#L7

The `run()` method should return value. It will be compared with value after the change.

rules/php70/tests/Rector/MethodCall/ThisCallOnStaticMethodToStaticCallRector/Fixture/fixture.php.inc
```diff
@@ -20,15 +40,35 @@ class SomeClass
...
+}
 
-    public static function eat(string $thing)
+class SomeRunnable implements RunnableInterface
+{
+    public function run()
     {
+        $expectedString = "child";
+        $c = new SomeChild();
+        return $c->speak() === $expectedString;
```
TomasVotruba, line 71: This should return the value:

```suggestion
        return $c->speak();
```

rules/php70/tests/Rector/MethodCall/ThisCallOnStaticMethodToStaticCallRector/Fixture/fixture.php.inc
```diff
@@ -2,15 +2,35 @@
...
```
TomasVotruba, line 33: So should this:

```suggestion
        return $c->speak();
```

derrickschoen:
I'm not really sure what I'm doing here. 
Mostly just tried to copy what I saw in Rector\CodeQuality\Tests\Rector\FuncCall\ArrayKeysAndInArrayToArrayKeyExistsRector\Fixture

Wondering why I am getting the error:
```
Cannot declare class Rector\Php70\Tests\Rector\MethodCall\ThisCallOnStaticMethodToStaticCallRector\Fixture\ClassName_t3377givb1400qffpo7a, because the name is already in use in /tmp/rector_temp_tests/runnable_

```

TomasVotruba:
That will be something related to autoloading. I'll handle it tomorrow, don't worry

Thanks for great work :+1:

TomasVotruba:
It took a bit more improvements on runnable testing side, but all works now :)

PR #3569, opened 2020-06-22
[PHP 8.0] Add ChangeSwitchToMatchRector

TomasVotruba:
Covers https://github.com/nikic/PHP-Parser/pull/672, https://wiki.php.net/rfc/match_expression_v2

```diff
 class SomeClass
 {
     public function run()
     {
-        $statement = switch ($this->lexer->lookahead['type']) {
-            case Lexer::T_SELECT:
-                $statement = $this->SelectStatement();
-                break;
-            default:
-                $this->syntaxError('SELECT, UPDATE or DELETE');
-                break;
-        }
+        $statement = match ($this->lexer->lookahead['type']) {
+            Lexer::T_SELECT => $this->SelectStatement(),
+            default => $this->syntaxError('SELECT, UPDATE or DELETE'),
+        };
     }
 }
```

Ref #3127

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,9 +18,9 @@
         "jean85/pretty-package-versions": "^1.2",
         "nette/robot-loader": "^3.2",
         "nette/utils": "^3.1",
-        "nikic/php-parser": "4.6",
+        "nikic/php-parser": "^4.7",
         "ondram/ci-detector": "^3.4",
-        "phpstan/phpdoc-parser": "^0.4.7",
+        "phpstan/phpdoc-parser": "^0.4.8",
         "phpstan/phpstan-phpunit": "^0.12.10",
         "psr/simple-cache": "^1.0",
         "sebastian/diff": "^3.0|^4.0",
@@ -37,7 +37,7 @@
         "symplify/set-config-resolver": "^8.1.15",
         "symplify/smart-file-system": "^8.1.15",
         "tracy/tracy": "^2.7",
-        "phpstan/phpstan": "^0.12.33"
+        "phpstan/phpstan": "^0.12.34"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.16",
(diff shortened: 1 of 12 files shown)


TomasVotruba:
I've just added separated jobs for:

- php-parser 4.6 - works 
- php-parser 4.7 - breaks

Everything else in the code is the same

ondrejmirtes:
I probably know what this is about:

* Here's the phpstan/phpstan autoloader: https://github.com/phpstan/phpstan/blob/master/bootstrap.php
* It's used when you use `phpstan/phpstan` but use PHPStan not by running `vendor/phpstan/phpstan/phpstan` executable but by asking for some internal classes, like NodeScopeResolver or RuleTestCase.
* At some point this line will be executed:

```
$composerAutoloader = require 'phar://' . __DIR__ . '/phpstan.phar/vendor/autoload.php';
```

* Which means that potentially the PHP-Parser library from inside the PHAR will be loaded some time later.
* But it has to be this way as PHPStan internals also rely on PHP-Parser.
* My conclusion is that if a project uses PHPStan internals but also requires PHP-Parser on its own, the PHP-Parser has to be the same version as it's bundled in the PHPStan's PHAR.

I really don't know the way out of this, it's a hard problem.

TomasVotruba:
Great! I came to exact same conclussion.

TomasVotruba:
The only solution that worked for me in the past is to use the same php-parser as PHPStan does.
Last bug was with property promotion, so we had to wait till new version of phpstan uses 4.6. It does now, but there is already 4.7 with more breaks.

ondrejmirtes:
I'll be releasing PHPStan 0.12.34 with PHP-Parser 4.7 later today 🚀

TomasVotruba:
That's way over my expectations. Epic, thank you :clap: :heart:

TomasVotruba:
condrejmirtes I can confirm 0.12.34 works purfectly! Thanks :clap:  :clap: :

TomasVotruba:
@ondrejmirtes I've just updated Rector and can confirm 0.12.34 works purfectly! Thanks :clap:  :clap:

TomasVotruba:
@ondrejmirtes I've just updated Rector and can confirm 0.12.34 works purfectly! Thanks :clap:  :clap:

ondrejmirtes:
Cool :)

PR #3576, opened 2020-06-23
fix return type in abstract class

TomasVotruba:
See https://3v4l.org/CVO6m


rules/type-declaration/src/TypeInferer/ReturnTypeInferer/ReturnedNodesReturnTypeInferer.php
```diff
@@ -30,15 +31,18 @@ public function inferFunctionLike(FunctionLike $functionLike): Type
     {
         /** @var Class_|Trait_|Interface_|null $classLike */
         $classLike = $functionLike->getAttribute(AttributeKey::CLASS_NODE);
+        if ($classLike === null) {
```
TomasVotruba, line 34: Skip anonymous classes

PR #3579, opened 2020-06-23
Boolean Operands cause ChangeNestedForeachIfsToEarlyContinueRector to…

derrickschoen:
… produce code that gives a different output from the original code.

### DIFF
--- a/rules/solid/tests/Rector/Foreach_/ChangeNestedForeachIfsToEarlyContinueRector/Fixture/multi_exprs_with_bool_oprnds.php.inc
+++ b/rules/solid/tests/Rector/Foreach_/ChangeNestedForeachIfsToEarlyContinueRector/Fixture/multi_exprs_with_bool_oprnds.php.inc
@@ -38,7 +38,7 @@ class MultiExprsWithBoolOprnds implements RunnableInterface
 {
     public function run()
     {
-        $partPackageList = [ new \StdClass(), null, 'string', 0, [] ];
+        $partPackageList = [ new \StdClass(), 'string' ];
         $payload = [];
 
         foreach ($partPackageList as $partPackage) {
--- a/rules/solid/tests/Rector/Foreach_/ChangeNestedForeachIfsToEarlyContinueRector/Fixture/multi_exprs_with_bool_oprnds.php.inc
+++ b/rules/solid/tests/Rector/Foreach_/ChangeNestedForeachIfsToEarlyContinueRector/Fixture/multi_exprs_with_bool_oprnds.php.inc
@@ -8,7 +8,7 @@ class MultiExprsWithBoolOprnds implements RunnableInterface
 {
     public function run()
     {
-        $partPackageList = [ new \StdClass(), null, 'string', 0, [] ];
+        $partPackageList = [ new \StdClass(), 'string' ];
         $payload = [];
 
         foreach( $partPackageList as $partPackage )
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Do you have an idea how to fix it?

derrickschoen:
> Do you have an idea how to fix `ChangeNestedForeachIfsToEarlyContinueRector`?

This codebase is way over my head.

My best guess would be to simplify what the rector tries to do.
Make it so that it only tries to wrap the if logic in parenthesis, then negate it, then early continue.

Ex
```
<?php
$loaded = true;
$usePackages = true;
$notBad = true;
$partPackageList = [ new \StdClass(), null, 'string', 0, [] ];
$payload = [];

foreach( $partPackageList as $partPackage )
{
    if( ! is_null( $partPackage ) && ! is_object($partPackage) )
    {
        if ($loaded || ($usePackages && $notBad))
        {
            $payload[] = $partPackage;
        }
    }
}
```

## Expected Behaviour
```
<?php
$loaded = true;
$usePackages = true;
$notBad = true;
$partPackageList = [ new \StdClass(), null, 'string', 0, [] ];
$payload = [];

foreach( $partPackageList as $partPackage )
{
	if (! (
           ! is_null( $partPackage ) && ! is_object($partPackage))
        ) {
		continue;
	}
	if (! (
            $loaded || ($usePackages && $notBad))
        ) {
		continue;
	}
	$payload[] = $partPackage;
}
```

TomasVotruba:
Related demo link: https://getrector.org/demo/88a502a5-5c37-4c2c-af8b-63c137cdfb4d

rules/solid/tests/Rector/Foreach_/ChangeNestedForeachIfsToEarlyContinueRector/Fixture/multi_exprs_with_bool_oprnds.php.inc
```diff
@@ -0,0 +1,65 @@
...
+            if(! is_null( $partPackage ) && ! is_object($partPackage) )
+            {
+                if ($loaded || ($usePackages && $notBad))
+                {
+                    $payload[] = $partPackage;
+                }
+            }
+        }
+
+        return json_encode($payload);
```
TomasVotruba, line 28: The snippet is quite large to cover with test only the problematic spot. 

2 variables and ~5 lines of code in method would be ideal.

<br>

What lines can be remove to still show the bug?

derrickschoen: Is it better now?

TomasVotruba: Better. Is that the least code needed?

E.g.?

```diff
-$partPackageList = [ new \StdClass(), null, 'string', 0, [] ];
+$partPackageList = [];
```

derrickschoen: I've stripped it down as much as I can while reproducing the bug.

TomasVotruba: Great, thanks!

TomasVotruba:
Merged locally. It was missing `continue;` statements :laughing: 

Thanks for the test case!

It might need more coverage with `==`, as they're impossible to inverse, e.g. this is not correct inversion: 

```diff
-1 != '1'
+1 == '1'
```

But provided test now passes.

derrickschoen:
> Merged locally. It was missing `continue;` statements 😆
> 
> Thanks for the test case!
> 
> It might need more coverage with `==`, as they're impossible to inverse, e.g. this is not correct inversion:
> 
> ```diff
> -1 != '1'
> +1 == '1'
> ```
> 
> But provided test now passes.

I messed up the bug replication when I removed the variables.
I made a new commit to my branch, not sure if I open a new PR or what.

There is a bug where if there is a nested if with an "OR" and all conditions evaluate to true, the rector changes the output.

TomasVotruba:
> I made a new commit to my branch, not sure if I open a new PR or what.

The branch is merged now. Open new PR

PR #3582, opened 2020-06-23
Update create_own_rule.md

Philosoft:
fix MyFirstRector example, because it doesn't work as is now.

Or use `$node->name->toString()`. I'm not sure which one is more appropriate

### DIFF
--- a/docs/create_own_rule.md
+++ b/docs/create_own_rule.md
@@ -53,7 +53,7 @@ final class MyFirstRector extends AbstractRector
             return null;
         }
 
-        $methodCallName = $this->getName($node);
+        $methodCallName = (string)$node->name;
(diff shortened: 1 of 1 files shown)


docs/create_own_rule.md
```diff
@@ -53,7 +53,7 @@ final class MyFirstRector extends AbstractRector
             return null;
         }
 
-        $methodCallName = $this->getName($node);
+        $methodCallName = (string)$node->name;
```
TomasVotruba, line 56: Ah, you're right. It should be called on `name` property:

```suggestion
        $methodCallName = $this->getName($node->name);
```

TomasVotruba:
Thank you

PR #3594, opened 2020-06-25
Symfony FormTypeInstanceToClassConstRector: Include AbstractController in allowed object types

### LINKED ISSUE #3570: Symfony: Cannot use FormTypeInstanceToClassConstRector within instances of AbstractController
# Bug Report

| Rector version | v0.7.37             |

I noticed that the `Rector\Symfony\Rector\MethodCall\FormTypeInstanceToClassConstRector` rector isn't working for controllers which extend `Symfony\Bundle\FrameworkBundle\Controller\AbstractController`. 

Having a look at the code, I think this is because the rector only runs on controllers which extend `Symfony\Bundle\FrameworkBundle\Controller\Controller`

Is is possible to get the rector to run on instances of `Symfony\Bundle\FrameworkBundle\Controller\AbstractController`?

```php
<?php

namespace App\Controller;

use App\Type\RectorTestType;

class RectorTestController extends \Symfony\Bundle\FrameworkBundle\Controller\AbstractController
{
    public function rectorTestAction()
    {
        $this->createForm(new RectorTestType());
    }
}
```

## Expected Behaviour

```php
<?php

namespace App\Controller;

use App\Type\RectorTestType;

class RectorTestController extends \Symfony\Bundle\FrameworkBundle\Controller\AbstractController
{
    public function rectorTestAction()
    {
        $this->createForm(RectorTestType::class);
    }
}
```

andyexeter:
Fixes #3570 by adding `Symfony\Bundle\FrameworkBundle\Controller\AbstractController` as a valid object type.

### DIFF
--- a/rules/symfony/src/Rector/MethodCall/FormTypeInstanceToClassConstRector.php
+++ b/rules/symfony/src/Rector/MethodCall/FormTypeInstanceToClassConstRector.php
@@ -86,11 +86,20 @@ PHP
      */
     public function refactor(Node $node): ?Node
     {
-        if ($this->isObjectTypes($node->var, ['Symfony\Bundle\FrameworkBundle\Controller\Controller', 'Symfony\Bundle\FrameworkBundle\Controller\AbstractController']) && $this->isName($node->name, 'createForm')) {
+        if ($this->isObjectTypes(
+            $node->var,
+            [
+                'Symfony\Bundle\FrameworkBundle\Controller\Controller',
+                'Symfony\Bundle\FrameworkBundle\Controller\AbstractController',
+            ]
+        ) && $this->isName($node->name, 'createForm')) {
             return $this->processNewInstance($node, 0, 2);
         }
 
-        if (! $this->isObjectTypes($node->var, ['Symfony\Component\Form\FormBuilderInterface', 'Symfony\Component\Form\FormInterface'])) {
+        if (! $this->isObjectTypes(
+            $node->var,
+            ['Symfony\Component\Form\FormBuilderInterface', 'Symfony\Component\Form\FormInterface']
+        )) {
             return null;
         }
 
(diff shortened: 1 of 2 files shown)


rules/symfony/src/Rector/MethodCall/FormTypeInstanceToClassConstRector.php
```diff
@@ -90,7 +91,7 @@ public function getNodeTypes(): array
      */
     public function refactor(Node $node): ?Node
     {
-        if ($this->isObjectType($node->var, Controller::class) && $this->isName($node->name, 'createForm')) {
+        if ($this->isObjectTypes($node->var, [Controller::class, AbstractController::class]) && $this->isName($node->name, 'createForm')) {
```
TomasVotruba, line 94: It might be fix for this case, but this can be acually broken when `Symfony\` gets prefixed in `rector.phar`.

Could you update the args so its strings?

```suggestion
        if ($this->isObjectTypes($node->var, ['Symfony\Bundle\FrameworkBundle\Controller\AbstractController', 'Symfony\Bundle\FrameworkBundle\Controller\Controller']) && $this->isName($node->name, 'createForm')) {
```

andyexeter: @TomasVotruba I will update.

There are other Symfony imports in this class:

Symfony\Component\Form\FormBuilderInterface
Symfony\Component\Form\FormInterface
Symfony\Component\OptionsResolver\OptionsResolver

Do these also need to be updated?

TomasVotruba: Yes. That would be great!

TomasVotruba:
Thank you

PR #3600, opened 2020-06-28
Mockery to Prophecy

jaapio:
From: https://github.com/mockery/mockery
To: https://github.com/phpspec/prophecy

I'm not sure if these kinds of contributions are welcome. But I was looking for it for a while. It is far from finished but I would like to have some feedback on the approach. Maybe you have some hints where to look for examples how to write rules like this. 
Things are getting complicated with mocking frameworks :-D


TomasVotruba:
Hey, this is very welcomed :+1: 

You can get some inspiration from already existing sets that also work with unit tests/mocking tools:
https://github.com/rectorphp/rector/tree/master/config/set/framework-migration

- phpspec-to-phpunit.yaml
- mockista-to-mockery.yaml

TomasVotruba:
If you need any help, just ping me at specific lines in the pull-request. I'll help

rules/mockery-to-prophecy/src/Rector/MethodCall/MockeryPredictionsToProphizeRector.php
```diff
@@ -0,0 +1,220 @@
...
+            if ($node instanceof MethodCall) {
+                /** @var $node MethodCall */
+                if ($this->isName($node->name, 'shouldReceive')) {
+                    if (!isset($node->args[0]) || !$node->args[0]->value instanceof String_) {
+                        throw new ShouldNotHappenException();
+                    }
+
+                    $methodCall = new MethodCall($node->var, $node->args[0]->value->value);
+
+                    $nextCall = $node->getAttribute('nextNode');
```
jaapio, line 169: @TomasVotruba methods there are chained. But I could not really find how to rewrite the chain properly. This is just a first test case. using `$mock->shouldReceive('create')->never();` to `$mock->create()->shouldNotBeCalled();` 

I found that `methodCall of `shouldReceive` has a nextNode of the type `Identifier` Like I wrote here. But the current result is that `create` is appended with `shouldNotBeCalled` like I want it to be. But `never` persists. So that's not removed from the tree somehow?

TomasVotruba: Use this service:
https://github.com/rectorphp/rector/blob/71247e15877c043fbd2a251e16eb299576b23b2c/rules/magic-disclosure/src/NodeAnalyzer/ChainMethodCallNodeAnalyzer.php

TomasVotruba: Here is how it's used in one rule: https://github.com/rectorphp/rector/blob/master/rules/magic-disclosure/src/Rector/MethodCall/DefluentMethodCallRector.php#L108-L112

TomasVotruba:
It would help to split to smaller PRs. That way the simple ones can be already merged, and we could narrow focus on more complicated parts.

jaapio:
We could split this, what do you think of this:
- create a PR to remove the mockery close
- create a PR to replace mock creation (without expectations) 
- PR to replace simple expectations 

Mainly just the test cases I added up til now, and see those as steps to get merged? People, who execute the rule will break their tests, but will at least know where things should be changed? Is that acceptable for you?

TomasVotruba:
Exactly my thoughts. Baby steps and iterate fast.
Go for it :+1:;

jaapio:
Thanks to your suggestion I got a first draft version ready now. I'm going to extract the commits here and get this thing prepared for a review and merge. The First 3 bullets I suggested are now implemented.

PR #3628, opened 2020-07-01
Fix PHPStan Reflection break from 0.12.26

TomasVotruba:


src/Stubs/StubLoader.php
```diff
@@ -35,6 +35,7 @@ public function loadStubs(): void
         $robotLoader->addDirectory($stubDirectory);
         $robotLoader->setTempDirectory(sys_get_temp_dir() . '/_rector_stubs');
         $robotLoader->register();
+        $robotLoader->refresh();
```
TomasVotruba, line 38: This is the saving line :partying_face:

PR #3648, opened 2020-07-03
update rule to support multiple occurrences of the class in the string

### LINKED ISSUE #3389: [CodeQuality] Add rule to replace in-string classes
Similar to: https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#splitstringclassconstanttoclassconstfetchrector

```diff
>write(
-                '<?php if (App\SomeClass::createCache($netteCacheStorage, %var,  $this->global->cacheStack, %node.array?)) { ?>',
+                '<?php if (' . SomeClass::class . '::createCache($netteCacheStorage, %var,  $this->global->cacheStack, %node.array?)) { ?>',
                \md5($node->args)
            );
```

Just for classes in `String_` node

dobryy:
Closes #3389

### DIFF
--- a/rules/coding-style/src/Rector/String_/UseClassKeywordForClassNameResolutionRector.php
+++ b/rules/coding-style/src/Rector/String_/UseClassKeywordForClassNameResolutionRector.php
@@ -50,33 +50,19 @@ PHP
      */
     public function refactor(Node $node): ?Node
     {
-        // @see https://regex101.com/r/Vv41Qr/1/
-        $matches = Strings::matchAll($node->value, '#([\\\\a-zA-Z0-9_\\x80-\\xff]*)::#', PREG_PATTERN_ORDER);
+        $classNames = $this->getExistentClasses($node);
 
-        $classNames = array_filter($matches[1], function (string $className) {
-            return class_exists($className);
-        });
-
-        if (empty($classNames)) {
+        if ($classNames === []) {
             return $node;
         }
 
-        $classNames = array_map(function (string $className) {
-            return preg_quote($className);
-        }, $classNames);
+        $parts = $this->getParts($node, $classNames);
 
-        // @see https://regex101.com/r/8nGS0F/1
-        $splits = Strings::split($node->value, '#(' . implode('|', $classNames) . ')#');
-
-        foreach ($splits as $split) {
-            if (empty($split)) {
-                continue;
-            }
-
-            if (class_exists($split)) {
-                $returnNode = new ClassConstFetch(new FullyQualified(ltrim($split, '\\')), 'class');
+        foreach ($parts as $part) {
+            if (class_exists($part)) {
+                $returnNode = new ClassConstFetch(new FullyQualified(ltrim($part, '\\')), 'class');
             } else {
-                $returnNode = new String_($split);
+                $returnNode = new String_($part);
             }
 
             $concat = ! isset($concat) ? $returnNode : new Concat($concat, $returnNode);
@@ -87,5 +73,29 @@ PHP
         }
 
         return $concat;
+    }
+
+    public function getExistentClasses(String_ $string): array
+    {
+        // @see https://regex101.com/r/Vv41Qr/1/
+        $matches = Strings::matchAll($string->value, '#([\\\\a-zA-Z0-9_\\x80-\\xff]*)::#', PREG_PATTERN_ORDER);
+
+        return array_filter($matches[1], function (string $className) {
+            return class_exists($className);
+        });
+    }
+
+    public function getParts(String_ $string, array $classNames): array
+    {
+        $classNames = array_map(function (string $className) {
+            return preg_quote($className);
+        }, $classNames);
+
+        // @see https://regex101.com/r/8nGS0F/1
+        $parts = Strings::split($string->value, '#(' . implode('|', $classNames) . ')#');
+
+        return array_filter($parts, function (string $className) {
+            return $className !== '';
+        });
     }
 }
--- a/rules/coding-style/src/Rector/String_/UseClassKeywordForClassNameResolutionRector.php
+++ b/rules/coding-style/src/Rector/String_/UseClassKeywordForClassNameResolutionRector.php
@@ -50,23 +50,40 @@ PHP
      */
     public function refactor(Node $node): ?Node
     {
-        if (substr_count($node->value, '::') !== 1) {
-            return null;
-        }
-        // a possible static call reference
         // @see https://regex101.com/r/Vv41Qr/1/
-        [$before, $possibleClass, $after] = Strings::split($node->value, '#([\\\\a-zA-Z0-9_\\x80-\\xff]*)::#');
+        $matches = Strings::matchAll($node->value, '#([\\\\a-zA-Z0-9_\\x80-\\xff]*)::#', PREG_PATTERN_ORDER);
 
-        if (! $possibleClass || ! class_exists($possibleClass)) {
-            return null;
+        $classNames = array_filter($matches[1], function (string $className) {
+            return class_exists($className);
+        });
+
+        if (empty($classNames)) {
+            return $node;
         }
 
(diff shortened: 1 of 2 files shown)


rules/coding-style/src/Rector/String_/UseClassKeywordForClassNameResolutionRector.php
```diff
@@ -50,23 +50,40 @@ public function getNodeTypes(): array
...
-        [$before, $possibleClass, $after] = Strings::split($node->value, '#([\\\\a-zA-Z0-9_\\x80-\\xff]*)::#');
+        $matches = Strings::matchAll($node->value, '#([\\\\a-zA-Z0-9_\\x80-\\xff]*)::#', PREG_PATTERN_ORDER);
+
+        $classNames = array_filter($matches[1], function (string $className) {
+            return class_exists($className);
+        });
 
-        if (! $possibleClass || ! class_exists($possibleClass)) {
-            return null;
+        if (empty($classNames)) {
```
TomasVotruba, line 60: `empty` is very vague, as it includes many values null, false, 0...

Better use strict comparison
```suggestion
        if ($classNames === []) {
```

rules/coding-style/src/Rector/String_/UseClassKeywordForClassNameResolutionRector.php
```diff
@@ -50,23 +50,40 @@ public function getNodeTypes(): array
...
-            return null;
-        }
-        // a possible static call reference
         // @see https://regex101.com/r/Vv41Qr/1/
-        [$before, $possibleClass, $after] = Strings::split($node->value, '#([\\\\a-zA-Z0-9_\\x80-\\xff]*)::#');
+        $matches = Strings::matchAll($node->value, '#([\\\\a-zA-Z0-9_\\x80-\\xff]*)::#', PREG_PATTERN_ORDER);
+
+        $classNames = array_filter($matches[1], function (string $className) {
+            return class_exists($className);
+        });
```
TomasVotruba, line 58: The refactor method seems pretty big now.

Could you decouple 2 methods - one for getting class names, the other for getting splits?

rules/coding-style/src/Rector/String_/UseClassKeywordForClassNameResolutionRector.php
```diff
@@ -50,23 +50,40 @@ public function getNodeTypes(): array
...
+            return preg_quote($className);
+        }, $classNames);
+
+        // @see https://regex101.com/r/8nGS0F/1
+        $splits = Strings::split($node->value, '#(' . implode('|', $classNames) . ')#');
+
+        foreach ($splits as $split) {
+            if (empty($split)) {
+                continue;
+            }
```
TomasVotruba, line 74: Again, empty is dangerous here. Could you think of way how to avoid it completely here?

rules/coding-style/src/Rector/String_/UseClassKeywordForClassNameResolutionRector.php
```diff
@@ -50,23 +50,40 @@ public function getNodeTypes(): array
...
+
+            if (class_exists($split)) {
+                $returnNode = new ClassConstFetch(new FullyQualified(ltrim($split, '\\')), 'class');
+            } else {
+                $returnNode = new String_($split);
+            }
+
+            $concat = ! isset($concat) ? $returnNode : new Concat($concat, $returnNode);
```
TomasVotruba, line 82: This looks very crazy :smile:  What is going on here? 
It's some node stacking in a loop?

Better collect concat nodes in an array, e.g. to $contacts

```php
$concats[] = $concat;
```

And resolve later in in go, e.g.

```php
if ($concats === []) {
    return ...;
}

return ...;
```


It's much easier to debug and test in case of troubles.

dobryy: This solution helped me to keep myself away from becoming crazy :D Believe me or not but is a million percents better than a couple of the solution I had before this one :D

Maybe it is not clear but this solution is able to process the random amount of class name occurrences in the string (not only one or two) that is why concatenation in the loop is required.

When it comes to the condition it helps to avoid the first empty Node. If I won't use it I'll have to create the first Node with an empty value outside of the loop for first concatenation which will then produce empty node which will in final result look like 

```
'' . \App\ClassName()::class . 'staticCall()'
```

While it is not a problem to create the array of nodes with the current solution... I can't imagine what I should implement under `...` in your code to cover the cases in the test fixtures.

TomasVotruba: I see, I think what you mean. Nodes can break human brain sometimes, I've been there :D. 
Don't worry about it :+1: , I'll check if there are is some space after we merge this.

rules/coding-style/src/Rector/String_/UseClassKeywordForClassNameResolutionRector.php
dobryy: The only way to avoid it completely here is to filter it out with `array_filter` in one of the methods you suggested. Unfortunately String::split() produces empty match group in case the class name is present at the very beginning (and I believe the very end) of the string.

Let me know if it makes sense to you to filter empty match groups out in the newly created method. To me it makes sense.

rules/coding-style/src/Rector/String_/UseClassKeywordForClassNameResolutionRector.php
dobryy: ok,ok, I was kinda surprised that phpstan or rector didn't shout on me :D

TomasVotruba: You're right, they should. I'll add a new rule :wink: 


I already added no isset/empty on objects: https://github.com/symplify/symplify/pull/1985

rules/coding-style/src/Rector/String_/UseClassKeywordForClassNameResolutionRector.php
TomasVotruba: > Let me know if it makes sense to you to filter empty match groups out in the newly created method. To me it makes sense.

Yes :+1:

rules/coding-style/src/Rector/String_/UseClassKeywordForClassNameResolutionRector.php
dobryy: Not sure how much you care about perfomance but to make this loop a bit nicer I can also 
```diff
+            $returnNode = new String_($part);
+            if (class_exists($part)) {
+                $returnNode = new ClassConstFetch(new FullyQualified(ltrim($part, '\\')), 'class');
+            }
-            if (class_exists($part)) {
-                $returnNode = new ClassConstFetch(new FullyQualified(ltrim($part, '\\')), 'class');
-            } else {
-                $returnNode = new String_($part);
-            }
```

TomasVotruba: Unless a memory/performance leak, it's not an issue as this is CLI app, not a response/request one.

Overriding variable like this is confusing. Personally I'd extract it to own method too. E.g. `createReturnNode()` with early returns.

But current state is also ok

dobryy: > I see, I think what you mean. Nodes can break human brain sometimes, I've been there :D.
> Don't worry about it  , I'll check if there are is some space after we merge this.

Not sure what do you mean by that. Can I leave it as it looks like at the moment (after today's changes)?

dobryy:
I've applied the changes according to the comments but for some reason, checks are failing... locally it looks fine...

TomasVotruba:
That's Symplify update I made today. I'll fix it.

Thanks for you work!

PR #3649, opened 2020-07-04
decouple createConcat() method

TomasVotruba:


src/PhpParser/Node/NodeFactory.php
```diff
@@ -352,6 +352,25 @@ public function createGetterClassMethodFromNameAndType(string $propertyName, ?No
...
+        }
+
+        /** @var Expr $previousConcat */
+        $previousConcat = array_shift($exprs);
+
+        foreach ($exprs as $expr) {
+            $previousConcat = new Concat($previousConcat, $expr);
+        }
+
+        return $previousConcat;
```
TomasVotruba, line 371: cc @dobryy This is the foreach magic I meant :+1:

PR #3650, opened 2020-07-04
Fix typo in rector_rules_overview.md

Gymnasiast:


TomasVotruba:
Thanks for contribution!

This must be fixed in the rule itself too, as it's used to generate the docs.

Gymnasiast:
OK, updated that one as well. Hope I did it correctly.

TomasVotruba:
Great, thanks!

Issue #3436, opened 2020-05-28
Split UnderscoreToPascalCaseVariableAndPropertyNameRector to Variable only rector and Property only rector


tigitz:
# Feature Request

I inherited a laravel legacy code base and would like to apply pascal case rector only on variables as there's a lot of eloquent model usage like `$user->first_name` that I obviously don't want to touch for now.

That would mean splitting the current `UnderscoreToPascalCaseVariableAndPropertyNameRector` into:

- UnderscoreToPascalCaseVariableRector
- UnderscoreToPascalCasePropertyNameRector

Does it seem reasonable to you @TomasVotruba  ?

Remaining question would be "Does the `UnderscoreToPascalCaseVariableAndPropertyNameRector` be kept for backward-compatibility ?"

TomasVotruba:
Hi, if it's related only to magic call on certain type of object, this rule should skip it.

Other property fetches are not related to Eloquent, right?

tigitz:
Hey Tomas !

So I haven't dove deep enough in the big diff to check if there are "legit" use case of non-magic property fetches but if that's the case, indeed, that would be great to ignore the magical ones rather than simply discard every properties fetch.

So when you say 

> this rule should skip it.

You mean the current behaviour is supposed to skip them already and it's a bug or it's a feature that must be implemented ?

TomasVotruba:
>  it's a feature that must be implemented ?

This one :+1:

tigitz:
Alright, thanks to this legacy project I have a ~10 items list of edge cases where rector falls short. I'll try to work on them including this one.

Do you have any tips on how I'll be able to guess if the property is a magical one or not ?  

Property declaration or lack of ? Seems fragile as it don't necessarily mean it's a magic one.
If a __set() method exists ? Property can be left untouched by the method.

TomasVotruba:
> Do you have any tips on how I'll be able to guess if the property is a magical one or not ?

There is a rule, that completes properties, that are called without definition (magically created).

See: https://github.com/rectorphp/rector/blob/master/rules/code-quality/src/Rector/Class_/CompleteDynamicPropertiesRector.php#L54


> Property declaration or lack of ? Seems fragile as it don't necessarily mean it's a magic one.
> If a __set() method exists ? Property can be left untouched by the method.

Could you provide demo link for both of these cases? https://getrector.org/demo

ishan-biztech:
[...]

PR #3653, opened 2020-07-05
[Naming] Split UnderscoreToPascalCaseVariableAndPropertyNameRector

dobryy:
Closes https://github.com/rectorphp/rector/issues/3436


rules/coding-style/src/Rector/Variable/UnderscoreToPascalCasePropertyNameRector.php
```diff
@@ -0,0 +1,111 @@
...
+
+        return $node;
+    }
+
+    /**
+     * @param PropertyProperty|PropertyFetch|StaticPropertyFetch $node
+     */
+    private function createPascalName(string $nodeName, Node $node): string
+    {
+        $camelCaseName = StaticRectorStrings::underscoreToPascalCase($nodeName);
```
dobryy, line 99: @TomasVotruba, don't you think that `StaticRectorStrings::underscoreToPascalCase` shouldn't uppercase first letter in case there is _ before it? Or it might be a good idea to create StaticRectorVariable::underscoreToPascalCase which will care of it...

rules/coding-style/src/Rector/Variable/UnderscoreToPascalCasePropertyNameRector.php
```diff
@@ -0,0 +1,111 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\CodingStyle\Rector\Variable;
```
dobryy, line 5: Is it a good idea to move this rule to new `Rector\CodingStyle\Rector\Property` namespace?

TomasVotruba: Yes

rules/coding-style/src/Rector/Variable/UnderscoreToPascalCasePropertyNameRector.php
TomasVotruba: By coding standar, property and variable names should always start with lower letter.

TomasVotruba: It's probably incorrectly user here. Could you add test fixture for it and fix it?

dobryy: Not sure what do you mean. 
Here is how it works at the moment
```diff
-    public $some_property;
+    public $someProperty;

-    public static $some_static_property;
+    public static $someStaticProperty;

-    public $_first_and_multiple_underscore_property;
+    public $_firstAndMultipleUnderscoreProperty;

-    public $_underscore;
+    public $_underscore;
```
Is it correct behavior?
This behavior is covered by test fixtures.

rules/coding-style/src/Rector/Variable/UnderscoreToPascalCaseVariableNameRector.php
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * @return string[]
+     */
+    public function getNodeTypes(): array
+    {
+        return [Variable::class];
+    }
+
+    /**
+     * @param Variable $node
```
dobryy, line 58: @TomasVotruba  I changed @param 
```diff
     * @param Variable|PropertyProperty|PropertyFetch|StaticPropertyFetch $node
     * @param Variable $node
```
And now I see phpstan error
```
 rules/coding-style/src/Rector/Variable/UnderscoreToPascalCaseVariableNameRector.php:78
 - '#Property PhpParser\\Node\\Expr\\Variable\:\:\$name \(PhpParser\\Node\\Expr\|string\) does not accept PhpParser\\Node\\Identifier#'
```
Can you advise what is the correct way to fix it?

TomasVotruba: I think the message says to remove `new Identifier` and use bare string

rules/coding-style/src/Rector/Variable/UnderscoreToPascalCasePropertyNameRector.php
TomasVotruba: I mean like this:

```diff
-    public $_first_and_multiple_underscore_property;
+    public $firstAndMultipleUnderscoreProperty;

-    public $_underscore;
+    public $underscore;
```

dobryy: Just check the PR :)

rules/coding-style/src/Rector/Property/UnderscoreToPascalCasePropertyNameRector.php
```diff
@@ -71,32 +71,39 @@ public function refactor(Node $node): ?Node
...
         $node->name = new Identifier($camelCaseName);
 
         return $node;
     }
 
     /**
-     * @param Variable|PropertyProperty|PropertyFetch|StaticPropertyFetch $node
+     * @param PropertyProperty|PropertyFetch|StaticPropertyFetch $node
      */
     private function createPascalName(string $nodeName, Node $node): string
```
TomasVotruba, line 97: This method should be in `StaticRectorStrings` service with others

rules/coding-style/src/Rector/Property/UnderscoreToPascalCasePropertyNameRector.php
```diff
@@ -71,32 +71,39 @@ public function refactor(Node $node): ?Node
...
         }
 
-        if (! Strings::contains($nodeName, '_')) {
+        // @see https://regex101.com/r/dXc89K/1
+        // matches if underscore is present but excluding first underscore
+        if (! Strings::match($nodeName, '#(?<!^)\_#')) {
             return null;
         }
 
         $camelCaseName = $this->createPascalName($nodeName, $node);
```
TomasVotruba, line 87: Probably incorrect name on the left or on the right :D

```diff
-$camelCaseName = $this->createPascalName($nodeName, $node);
+$pascalCaseName = $this->createPascalName($nodeName, $node);
```

or 

```diff
-$camelCaseName = $this->createPascalName($nodeName, $node);
+$camelCaseName = $this->createCamelCaseName($nodeName, $node);
```

rules/coding-style/tests/Rector/Property/UnderscoreToPascalCasePropertyNameRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,51 @@
...
+
+final class Property
+{
+    public $someProperty;
+
+    public static $someStaticProperty;
+
+    public $_firstAndMultipleUnderscoreProperty;
+
+    public $_underscore;
```
TomasVotruba, line 39: Should be `$underscore`

rules/coding-style/src/Rector/Variable/UnderscoreToPascalCasePropertyNameRector.php
TomasVotruba: Just checked, it's missing there.

dobryy: In original rule properties and variable names with _ in the beginning, are skipped. That is why I left the original behavior and made these rule skip properties which are accessed using magic calls. So for now as far as I understand the rule should skip only "magic accesses'?

rules/coding-style/src/Rector/Property/UnderscoreToPascalCasePropertyNameRector.php
dobryy: :D it was in original rule and I didn't pay attention too much to this name.

TomasVotruba: No troubles. I just noticed it ;)

rules/coding-style/src/Rector/Property/UnderscoreToPascalCasePropertyNameRector.php
dobryy: 1. I have just notice that it is possible to pass only the node as a param in this method and to resolve node name using `$this->getName($node)`. Is there a way to resolve the name of the node within `StaticRectorStrings`?

rules/coding-style/src/Rector/Property/UnderscoreToPascalCasePropertyNameRector.php
TomasVotruba: There should be a PHPStan/Rector rule for that, right?

rules/coding-style/src/Rector/Variable/UnderscoreToPascalCasePropertyNameRector.php
dobryy: Also... by removing underscore in the beginning there might be a potential problem. In case, in the code (for some crazy reason) there are two variables/properties one `$_some_value` and another `$some_value`  the rule will rename to the same name `$someValue`

rules/coding-style/src/Rector/Property/UnderscoreToPascalCasePropertyNameRector.php
dobryy: It's gonna be too strict then! :D

rules/coding-style/src/Rector/Property/UnderscoreToPascalCasePropertyNameRector.php
dobryy: 2. I moved it but, in my opinion, there should be better place or name for it. There is already a method `underscoreToPascal`, also it is not clear from which naming convention it is going to change the name.
The current `StaticRectorStrings` seems like some helper class for strings. The new method is more related to the Node itself... at least I have such a feeling. Maybe it will be nice to create StaticRectorPropertyVariable and place this new method there with self explanotary name like `createPascalCaseNameFromUnderscore`...

PR #3681, opened 2020-07-10
[Legacy] Adding RemoveIncludeRector

phpfui:
The example uses comments because I noticed it was deleting comments attached as attributes to the Include_ node.  If there is a comment before the include, there will be a blank line where the include was, otherwise the line is removed.

### DIFF
--- a/rules/legacy/tests/Rector/Include_/RemoveIncludeRector/Fixture/fixture.php.inc
+++ b/rules/legacy/tests/Rector/Include_/RemoveIncludeRector/Fixture/fixture.php.inc
@@ -2,8 +2,9 @@
 
 namespace Rector\Legacy\Tests\Rector\Include_\RemoveIncludeRector\Fixture;
 
+$files = glob(__DIR__ . '/expect*.php');
 // comment before include
-include 'somefile.php';
+include array_pop($files);
 // comment after include
 
 ?>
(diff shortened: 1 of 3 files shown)


phpfui:
So interesting problem here.  I need to include a file, which the tests are executing.  But the tests run in the temp directory.  I not sure I can construct a path to a known file path.  I originally did not see this because PHPUnit was crashing with a memory issue on my Windows machine.  I can run the test individually, which works, but can't be sure which file gets included when the whole suite runs, as I take a random one from the current temp directory, and the one selected could throw an error.

Any ideas?

TomasVotruba:
As the test code is executed, it's the best practise go wrap it in class method. That way the test code is scoped. You can see it others tests

rules/legacy/tests/Rector/Include_/RemoveIncludeRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Rector\Legacy\Tests\Rector\Include_\RemoveIncludeRector\Fixture;
+
+file_put_contents('somefile.php', '<?php // some comment');
+// comment before include
+include 'somefile.php';
+// comment after include
+unlink('somefile.php');
```
TomasVotruba, line 9: There should be no code run in fixtures.
Wrap it in the class method

TomasVotruba:
Looks good! :+1: 

Could you rebase it to 1 commit?

I'll merge it then

phpfui:
New PR on squashed commit issued.  Closing this.

Issue #3660, opened 2020-07-06
[Downgrade] PHP 7.4 to 5.6 set


TomasVotruba:
Set for downgrading any code down the PHP version.

Useful for more comfortable development in the latest PHP, while allowing composer install on older PHP versions.

E.g. @phpstan https://github.com/phpstan/phpstan-src/blob/master/bin/transform-source.php from untyped properties

The partial goal might be to run Rector on itself to get to version PHP 5.6  :1st_place_medal: 

## Gradual Steps

- [ ] add Rector to CI here on PHP 5.6 and allow failure

## PHP 7.1 → 7.0

- [ ] move void type to docs

## PHP 7.0 → 5.6

- [ ] move param/return scalar types to docs
- [ ] remove strict_types

TomasVotruba:
Re-opening, as the set is not complete at all.

The goal might be to run Rector on itself, so version running PHP 5.6 is made  :1st_place_medal: 

## Gradual Steps

- add Rector to CI here on PHP 5.6

## PHP 7.1 → 7.0

- move void type to docs


## PHP 7.0 → 5.6

- move param/return scalar types to docs
- remove strict_types

staabm:
It would be great to use such a process to downgrade the requirements for a lib/tool to a older version while the upstream project can require latest php versions.

The process could/should also adjust the composer.json php version requirement, I guess.

staabm:
e.g. some tools of ocramius are not usable for use, because he defines very recent version requirements which do not fit into a world where we need ubuntu lts servers with not so recent libs like php etc

TomasVotruba:
Exactly. I work mostly with newest PHP, so I'm not sure what are the most common blockers to downgrade code.

Could you think of exact steps per each PHP version and extend the list I started?

ondrejmirtes:
I think the main thing you need is to make sure all users are using the PHAR version. This is why I made the switch so that phpstan/phpstan is distributed as PHAR and it's used by everyone, it allows me to have a build step to change the source before releasing.

Otherwise I'd install the dependencies as if they're for PHP 7.4. You then need to run the "downgrade Rector" on vendor too. Also, Symfony polyfills are useful for the missing functions from newer PHP versions.

BTW I think that PHP 5.6 is too ambitious. I'd opt-in for 7.4 -> 7.1 and then maybe 7.0.

staabm:
I might have misinterpreted the initial question:

[...]

PR #3686, opened 2020-07-10
[Downgrade] PHP 7.4 to 7.1 (Property  Type)

dobryy:
Closes https://github.com/rectorphp/rector/issues/3660


rules/downgrade/src/Rector/Property/TypedPropertyRector.php
```diff
@@ -0,0 +1,104 @@
...
+     * @param Identifier|Name|NullableType|UnionType $type
+     */
+    private function printType($type): string
+    {
+        if ($type instanceof NullableType) {
+            return $this->printType($type->type) . '|null';
+        }
+
+        if ($type instanceof UnionType) {
+            throw new Exception('UnionType not yet supported');
```
dobryy, line 86: @TomasVotruba is it a good idea to throw an exception? Or is it much better to return nulls? In case of nulls, it is much easier to test it using the existing testing mechanism...

TomasVotruba: Why not add support for it? It's already in php-parser.

Should be as simple as foreach it and connect with |

rules/downgrade/src/Rector/Property/TypedPropertyRector.php
```diff
@@ -0,0 +1,104 @@
...
+            }
+
+            return $name;
+        }
+
+        if ($type instanceof Identifier) {
+            return $type->name;
+        }
+
+        throw new Exception('Unsupported type class');
```
dobryy, line 102: There is another exception here.

This code is actually a cope-paste from https://github.com/phpstan/phpstan-src/blob/master/bin/transform-source.php

rules/downgrade/src/Rector/Property/TypedPropertyRector.php
```diff
@@ -0,0 +1,104 @@
...
+            return null;
+        }
+
+        $docComment = $node->getDocComment();
+        if ($docComment !== null) {
+            $node->type = null;
+            return $node;
+        }
+
+        $node->setDocComment(new Doc(sprintf("/**\r\n* @var %s\r\n*/", $propertyTypeDefinition)));
```
TomasVotruba, line 74: Actually, Rector uses own object for doc operations.

You can fidn example here: https://github.com/rectorphp/rector/blob/05d518bce33e60cdd7ebec31e01e1b701b30f4d1/rules/solid/src/Rector/ClassMethod/ChangeReadOnlyVariableWithDefaultValueToConstantRector.php#L215-L221

dobryy: Yeah, yeah, I saw... playing with it at the moment.

dobryy:
@TomasVotruba, ready.

TomasVotruba:
Could you update the nodes docs? It was removed in this PR

rules/downgrade/tests/Rector/Property/TypedPropertyRector/Fixture/classname_nullable_type.php.inc
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Rector\Downgrade\Tests\Rector\Property\TypedPropertyRector\Fixture;
+
+class ClassNameNullableTypeClass {
+    private ?\Rector\Downgrade\Tests\Rector\Property\TypedPropertyRector\Fixture\AnotherClassNameNullableTypeClass $property;
+}
+
+class AnotherClassNameNullableTypeClass {
+}
```
TomasVotruba, line 10: These extra classes, that should be autolaoded but do not change at all, should be located in `/Source` directory, on the same level as `/Fixtures`

dobryy: Changed

dobryy:
> Could you update the nodes docs? It was removed in this PR
> 
> ```shell
> composer docs
> ```

Hmmmm, weird. Recovered.

TomasVotruba:
Thank you :+1:

PR #3688, opened 2020-07-11
[Legacy] Add AddTopIncludeRector

phpfui:
This will add an include type statement to any plain PHP file, but not class definitions.  It does add a new include statement each time it is run, but I think that is OK, as you may want to add more than one (but not sure why). Can be combined with the RemoveIncludeRector to nuke the old includes.

### DIFF
--- a/rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
+++ b/rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
@@ -11,13 +11,12 @@ use PhpParser\Node\Stmt\Nop;
 use PhpParser\Parser;
 use PhpParser\ParserFactory;
 use PhpParser\PrettyPrinter\Standard;
+use Rector\Core\Exception\Rector\InvalidRectorConfigurationException;
 use Rector\Core\RectorDefinition\CodeSample;
 use Rector\Core\RectorDefinition\RectorDefinition;
-use Rector\Core\Exception\Rector\InvalidRectorConfigurationException;
 use Rector\FileSystemRector\Rector\AbstractFileSystemRector;
 use ReflectionClass;
 use Symplify\SmartFileSystem\SmartFileInfo;
-
 /**
  * @see https://github.com/rectorphp/rector/issues/3679
  *
@@ -63,7 +62,7 @@ final class AddTopIncludeRector extends AbstractFileSystemRector
     public function __construct(array $settings = [])
     {
         $this->settings = $settings;
-        $this->parser = (new ParserFactory)->create(ParserFactory::PREFER_PHP7);
+        $this->parser = (new ParserFactory())->create(ParserFactory::PREFER_PHP7);
         $this->prettyPrinter = new Standard();
     }
 
--- a/rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
+++ b/rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
@@ -7,6 +7,7 @@ namespace Rector\Legacy\Rector\Include_;
 use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\Include_;
 use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\Expression;
 use PhpParser\Node\Stmt\Nop;
 use PhpParser\Parser;
 use PhpParser\ParserFactory;
@@ -46,11 +47,6 @@ final class AddTopIncludeRector extends AbstractFileSystemRector
     private $patterns = [];
 
     /**
-     * @var array
-     */
-    private $settings = [];
-
-    /**
      * @var Parser
      */
     private $parser;
@@ -60,25 +56,42 @@ final class AddTopIncludeRector extends AbstractFileSystemRector
      */
     private $prettyPrinter;
 
-    public function __construct(array $settings = [])
+    public function __construct(string $type = 'TYPE_INCLUDE', string $file = '"autoload.php"', array $match = [])
     {
-        $this->settings = $settings;
-        $this->parser = (new ParserFactory())->create(ParserFactory::PREFER_PHP7);
-        $this->prettyPrinter = new Standard();
+		$this->parser = (new ParserFactory())->create(ParserFactory::PREFER_PHP7);
+		$this->prettyPrinter = new Standard();
+		$reflection = new ReflectionClass(Include_::class);
+		$constants = $reflection->getConstants();
+
+		if (! isset($constants[$type])) {
+			throw new InvalidRectorConfigurationException('Invalid type: must be one of ' . implode(
+				', ',
+				array_keys($constants)
+			));
+			$this->parser = (new ParserFactory())->create(ParserFactory::PREFER_PHP7);
+			$this->prettyPrinter = new Standard();
+		}
+
+		$this->type = $constants[$type];
+		$this->file = $file;
+		$this->fileExpression = $this->getExpressionFromString($this->file);
+		// normalize the file we are including
+		$this->file = $this->getStringFromExpression($this->fileExpression);
+  		$this->patterns = $match;
     }
 
     public function getDefinition(): RectorDefinition
     {
         return new RectorDefinition('Adds an include file at the top of matching files but not class definitions', [
             new CodeSample(
-                <<<PHP
+                <<<'PHP'
 <?php
 
 if (isset($_POST["csrf"])) {
     processPost($_POST);
 }
 PHP,
-                <<<PHP
+                <<<'PHP'
 <?php
 
 include "autoloader.php";
@@ -86,15 +99,13 @@ include "autoloader.php";
 if (isset($_POST["csrf"])) {
     processPost($_POST);
 }
-PHP,
+PHP
             ),
         ]);
     }
 
     public function refactor(SmartFileInfo $smartFileInfo): void
     {
-        $this->initialize();
-
         if (! $this->matchFile($smartFileInfo->getRelativeFilePath()) || $this->fileExpression === null) {
             return;
         }
@@ -117,7 +128,7 @@ PHP,
 
         // add the include to the statements and print it
         array_unshift($nodes, new Nop());
-        array_unshift($nodes, new Include_($this->fileExpression, $this->type));
+        array_unshift($nodes, new Expression(new Include_($this->fileExpression, $this->type)));
 
         $this->printNodesToFilePath($nodes, $smartFileInfo->getRelativeFilePath());
     }
@@ -138,34 +149,6 @@ PHP,
         }
 
         return false;
-    }
-
-    private function initialize(): void
-    {
-        if ($this->type) {
-            return;
-        }
-        $reflection = new ReflectionClass(Include_::class);
-        $constants = $reflection->getConstants();
-
-        if (! isset($constants[$this->settings['type'] ?? ''])) {
-            throw new InvalidRectorConfigurationException('Invalid type: must be one of ' . implode(
-                ', ',
-                array_keys($constants)
-            ));
-        }
-
-        if ('' === ($this->settings['file'] ?? '')) {
-            throw new InvalidRectorConfigurationException('Invalid parameter: file must be provided');
-        }
-        $this->type = $constants[$this->settings['type']];
-        $this->file = $this->settings['file'];
-        $this->fileExpression = $this->getExpressionFromString($this->file);
-        // normalize the file we are including
-        $this->file = $this->getStringFromExpression($this->fileExpression);
-        if (isset($this->settings['match']) && is_array($this->settings['match'])) {
-            $this->patterns = $this->settings['match'];
-        }
     }
 
     private function getExpressionFromString(string $source): ?Expr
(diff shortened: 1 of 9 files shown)


rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getNodeTypes(): array
+    {
+        $this->initialize();
+
+        return [Node::class];
```
TomasVotruba, line 115: If you want to use somethign like `Node`, that's a sing this should be a `AbstractFileSystemRector`

phpfui: So finally had some time to look at this.  If I were to make it a AbstractFileSystemRector, then I would have to use BetterNodeFinder.  I can find a class node with that, but I also want to find the first executable statement and insert the include before that.  I currently just filter out nodes that are non-executable, then insert when I find a node that is not in that list. Not sure how I would do that using BetterNodeFinder.  There are some filters available, but not sure it is any better than what I am doing now.

rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
```diff
@@ -0,0 +1,200 @@
...
+if (isset($_POST["csrf"])) {
+    processPost($_POST);
+}',
+                '<?php
+
+include "autoloader.php";
+
+if (isset($_POST["csrf"])) {
+    processPost($_POST);
+}',
```
TomasVotruba, line 96: In multine-samples, there should be now/heredoc format used: 
https://github.com/rectorphp/rector/blob/965cef98e218163cb43049c1e5ef0b32fa635eef/rules/renaming/src/Rector/Class_/RenameClassRector.php#L55-L79

This prevents spacing issues and missing newlines

rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
TomasVotruba: There is no need to use `BetterNodeFinder`, if you want to update top-level statements.

`foreach` over of nodes should do.

Could you update test fixtures with ~2 more samples that will cover edge-cases you discovered? 

Also 1 fixture with `skip_xxx.php.inc`, that checks the include is already there and to prevent double adding it.


<br>

I'll prepare short code sample for `AbstractFileSystemRector`, so it's easier to understand

rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
```diff
@@ -0,0 +1,200 @@
...
+            $this->completed = false;
+        }
+
+        if (! $this->matchFile($path)) {
+            $this->completed = true;
+
+            return null;
+        }
+
+        $name = get_class($node);
```
TomasVotruba, line 134:
```suggestion
        $nodeClass = get_class($node);
```

rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
```diff
@@ -0,0 +1,200 @@
...
+        $this->addNodeBeforeNode($includeNode, $node);
+        $this->addNodeBeforeNode(new Nop(), $node);
+
+        return $node;
+    }
+
+    /**
+     * Match file against matches, no patterns provided, then it matches
+     */
+    private function matchFile(string $path): bool
```
TomasVotruba, line 161: :+1:

rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
TomasVotruba: Here is a good example:

https://github.com/rectorphp/rector/blob/965cef98e218163cb43049c1e5ef0b32fa635eef/rules/legacy/src/Rector/FileSystem/FunctionToStaticMethodRector.php#L70-L73



```php
    public function refactor(SmartFileInfo $smartFileInfo): void
    {
        $nodes = $this->parseFileInfoToNodes($smartFileInfo);
        $fileStmts = $this->getFileOrNamespaceStmts($nodes);

        // 0. has already include? → skip
        foreach ($fileStmts as $fileStmt) {
           if ($this->isTopFileInclude(fileStmt)) {
               return;
           }
		}
      
        // 1. add include to right location
        foreach ($fileStmts as $fileStmt) {
           if (! $this->isLocationToAddInclude(fileStmt)) {
               continue;
           }
             
           // add it
           $fileStmts = ...;
		}

        // 2. print stmts back
        // ... mimic the process from the rule
   }
```

To add a stmt betweent other statements, you can use `array_splice`, but that was too magic for me and very hard to read/debug. So I use simple foreach:

```php
    $newFileStmts = [];
    foreach ($fileStmts as $fileStmt) {
        if ($this->isLocationToAddInclude(fileStmt)) {
           // add it
           $newFileStmts[] = new Expression(new Include_(...));               continue;
        }

        $newFileStmts[] = $fileStmt;
    }


// $this->print...( $newFileStmts)
```

rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
phpfui: So you are saying if I move to AbstractFileSystemRector, I can look for class and include (of the same file name), and if neither is found, I can insert before the first node and that should work?

I'll will check to see if that would handle InlineHTML before the first executable statement.  That is the only exception that might be an issue.

rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
phpfui: OK, thanks for the example!  They really help to understand the code.

rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
TomasVotruba: > I can look for class and include (of the same file name), and if neither is found, I can insert before the first node and that should work?

Partially. It should also work with other includes of different files, e.g.

```diff
+include __DIR__ . '/autoloader.php' # be sure the paths is absolute, to prevent magic with relative paths 

 include 'some_other_file.php;

 if (...) {
    include 'yet_another.php;
 }
```

rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
TomasVotruba: If that's too much to absorb now and you'll feel stressed, just add the test fixture and I'll merge it and do the change myself for this 1st case.

phpfui: Let me take another whack at it.  I think I understand what needs to be done.  The biggest issue for me is the PHPStan checks don't seem to run well under Windows 10. They also seem to reconfigure the method single line into multiple lines, which fail on github.  I prefer the single line format myself, so I was confused when it multi lined them.

I was also wondering about the string expression that could be in a include statement.  How would I specify something like __DIR__ . '/../file.php' in the config and get it converted into nodes?

phpfui: Hmm, I could probably just send the string through PhpParser, then add that expression to the include node.

TomasVotruba: @phpfui 

> Hmm, I could probably just send the string through PhpParser, then add that expression to the include node.

Exactly! :+1:

TomasVotruba: > The biggest issue for me is the PHPStan checks don't seem to run well under Windows 10.

I'm sorry to hear that. I actually don't know many CLI tools that work on Windows. It's often quite risky. I used Windows for PHP, but I do use Ubuntu for last 10 years and it's much easier as it just works. You might find more troubles with Rector on Windows in the future, just so you're aware of it.

phpfui:
I figured out how to call PHPStan from Windows.  Next commit will have that fixed.

Question is how do I save the nodes in this mode?  I called ~~~$this->print($nodes)~~ but that does not seem to work.

TomasVotruba:
> Question is how do I save the nodes in this mode? I called ~~~$this->print($nodes)~~ but that does not seem to work.

If you check the other rules that extends `AbstratcFileSystemRector`, there should be a method for that

rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
```diff
@@ -181,20 +154,48 @@ private function initialize(): void
...
+        // normalize the file we are including
+        $this->file = $this->getStringFromExpression($this->fileExpression);
+        if (isset($this->settings['match']) && is_array($this->settings['match'])) {
+            $this->patterns = $this->settings['match'];
+        }
+    }
+
+    private function getExpressionFromString(string $source): ?Expr
+    {
+        $ast = $this->parser->parse("<?php {$source};\n");
```
TomasVotruba, line 179: Better use `$nodes` convention, so we have the same naming in all files

TomasVotruba: Also, the "in string {$source}" should be prevented too.

```suggestion
        $nodes = $this->parser->parse('<?php ' . $source . ';');
```

 It makes PHPStan + Rector refactoring more complicated.


PHPStan missed that?

rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
```diff
@@ -181,20 +154,48 @@ private function initialize(): void
...
+        if (! isset($constants[$this->settings['type'] ?? ''])) {
+            throw new InvalidRectorsettingsException('Invalid type: must be one of ' . implode(
                 ', ',
                 array_keys($constants)
             ));
         }
 
-        if ('' === ($this->configuration['file'] ?? '')) {
-            throw new InvalidRectorConfigurationException('Invalid parameter: file must be provided');
+        if ('' === ($this->settings['file'] ?? '')) {
```
TomasVotruba, line 164: 3 operations at one line. This is bit confusing to read... what is the goal of this ?

My guess: 

```suggestion
        if (! isset($this->settings['file'])) {
```

phpfui: I originally had empty($this->settings['file']) but PHPStan complained "Use strict comparison instead of empty".  empty is actually correct in this situation. If the index is not set, it is an error and empty.  If the index is set, but empty ('', 0, etc), it is also an error.  It needs to be a string of some sort.  isset would need an additional test.

So I went with the null coalesce operator, but you have to wrap it in () to make sure the operator precedence is always correct, then type compare it to string to make sure it is not empty.  I could reverse the lhs and rhs, but it is my style to put the scalar constant on the lhs so if you mistype a comparison as as assign, you get an error. I have noticed your style guide does not do this.  I would recommend it.

Other suggestions?

TomasVotruba: I wonder how to make it simple to prevent such a rocket engeneering over one variable.

What about makint it a parameter? The file name is required anyway, right?

rules/legacy/tests/Rector/Include_/AddTopIncludeRector/AddTopIncludeRectorTest.php
```diff
@@ -0,0 +1,38 @@
...
+
+    protected function getRectorsWithConfiguration(): array
+    {
+        return [
+            AddTopIncludeRector::class => [
+                '$settings' => [
+                    'type' => 'TYPE_INCLUDE',
+                    'file' => '__DIR__ . "/../autoloader.php"',
+                ],
+            ],
```
TomasVotruba, line 35: Ah, I see it now. Array makes sense only in case of multiple instances, like rename these 30 classes to these 30 classes.

With this 1-level setup, you can use them directly:

```suggestion
            AddTopIncludeRector::class => [
                '$type' => 'TYPE_INCLUDE',
                '$file' => '__DIR__ . "/../autoloader.php"',
            ],
```

rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
TomasVotruba: See: https://github.com/rectorphp/rector/pull/3688#discussion_r454419900

rules/legacy/tests/Rector/Include_/AddTopIncludeRector/AddTopIncludeRectorTest.php
phpfui: OK, that is what I was trying to figure out, but could not seem to get it to work.  Will change.

TomasVotruba:
I guess you've tried that and it fails on Windows... but if not, you can fix all coding standard issues before commiting like this:

```bash
composer fix-cs
```

phpfui:
fix-cs does not find all issues, and makes changes that if checked in will
not pass the Linux version, like multiline method declarations.  Very
strange.

Seems to run under Windows though, so not sure why it is finding and fixing
different things.

TomasVotruba:
Don't worry about it :+1: . I think it's ready now.  

Could you squash it to 1 commit?

I'll merge it then and review locally

phpfui:
Yes, I finally think it is good. Squash coming up.

I did notice that PHPUnit seems to think the tests are risky with no
assertions made.  Not sure why that is, you may want to look into it and
fix.  I just don't know enough about the unit tests to fix.  I think I got
it right, but obviously not.

phpfui:
Closing.  Opening squashed PR.

Issue #3656, opened 2020-07-05
[Naming] Name variable after get/create method


TomasVotruba:
```diff
-$value = $this->getSomeValue();
+$someValue = $this->getSomeValue();
```

Based on https://github.com/rectorphp/rector/pull/3653#discussion_r449898288

dobryy:
Reopening to do not forget about `create`, and maybe `resolve`, rule.

TomasVotruba:
Closing as not a priority ATM

PR #3706, opened 2020-07-13
[Naming] Name variable after get method

dobryy:
Partially closes https://github.com/rectorphp/rector/issues/3656

Renames only after get method. It seems to me that for create method we should create a separate rule.

### DIFF
--- a/config/set/naming/naming.yaml
+++ b/config/set/naming/naming.yaml
@@ -1,0 +1,2 @@
+services:
+    Rector\Naming\Rector\Assign\RenameVariableToMatchGetMethodNameRector: null
--- a/rules/naming/src/Naming/ExpectedNameResolver.php
+++ b/rules/naming/src/Naming/ExpectedNameResolver.php
@@ -5,13 +5,19 @@ declare(strict_types=1);
 namespace Rector\Naming\Naming;
 
 use Nette\Utils\Strings;
+use PhpParser\Node;
+use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\Assign;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\New_;
+use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Name;
 use PhpParser\Node\Param;
 use PhpParser\Node\Stmt\Property;
 use Rector\BetterPhpDocParser\PhpDocInfo\PhpDocInfo;
+use Rector\Core\PhpParser\Node\BetterNodeFinder;
 use Rector\NodeNameResolver\NodeNameResolver;
 use Rector\NodeTypeResolver\Node\AttributeKey;
 use Rector\PHPStan\Type\FullyQualifiedObjectType;
@@ -34,14 +40,21 @@ final class ExpectedNameResolver
      */
     private $staticTypeMapper;
 
+    /**
+     * @var BetterNodeFinder
+     */
+    private $betterNodeFinder;
+
     public function __construct(
         NodeNameResolver $nodeNameResolver,
         PropertyNaming $propertyNaming,
-        StaticTypeMapper $staticTypeMapper
+        StaticTypeMapper $staticTypeMapper,
+        BetterNodeFinder $betterNodeFinder
     ) {
         $this->nodeNameResolver = $nodeNameResolver;
         $this->propertyNaming = $propertyNaming;
         $this->staticTypeMapper = $staticTypeMapper;
+        $this->betterNodeFinder = $betterNodeFinder;
     }
 
     public function resolveForPropertyIfNotYet(Property $property): ?string
@@ -146,6 +159,32 @@ final class ExpectedNameResolver
         $fullyQualifiedObjectType = new FullyQualifiedObjectType($className);
 
         return $this->propertyNaming->getExpectedNameFromType($fullyQualifiedObjectType);
+    }
+
+    public function resolveForGetCallExpr(Expr $expr): ?string
+    {
+        /** @var MethodCall|StaticCall|FuncCall|null $callNode */
+        $callNode = $this->betterNodeFinder->findFirst($expr, function (Node $node) {
+            return $node instanceof MethodCall || $node instanceof StaticCall || $node instanceof FuncCall;
+        });
+
+        if ($callNode === null) {
+            return null;
+        }
+
+        $name = $this->nodeNameResolver->getName($callNode->name);
+        if ($name === null) {
+            return null;
+        }
+
+        // @see https://regex101.com/r/hnU5pm/2/
+        $matches = Strings::match($name, '#^get(([A-Z]).+)#');
+
+        if ($matches === null) {
+            return null;
+        }
+
+        return lcfirst($matches[1]);
     }
 
     /**
--- a/rules/naming/src/Rector/Assign/RenameVariableToMatchGetMethodNameRector.php
+++ b/rules/naming/src/Rector/Assign/RenameVariableToMatchGetMethodNameRector.php
@@ -1,0 +1,239 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Naming\Rector\Assign;
+
+use PhpParser\Comment\Doc;
+use PhpParser\Node;
+use PhpParser\Node\Expr\ArrowFunction;
+use PhpParser\Node\Expr\Assign;
+use PhpParser\Node\Expr\Closure;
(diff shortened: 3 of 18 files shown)


config/set/naming/naming.yaml
```diff
@@ -1,3 +1,3 @@
 services:
```
dobryy, line 1: I noticed that there is one rule missing in this configuration file.
Rector\Naming\Rector\ClassMethod\RenameVariableToMatchNewTypeRector.

Do you want me to add it withing PR or let's create a new issue for it?

dobryy: I guess I know why it is missing there :) because this rules are applied in CI for Rector itself...
@TomasVotruba, let me know what is the correct solution here, please. To remove it from set, or to setup CI somehow to skip this rule?

TomasVotruba: > Rector\Naming\Rector\ClassMethod\RenameVariableToMatchNewTypeRector.

I guess it still needs some work. Should be added and ideally, those cases fixed :+1:

rules/naming/src/Rector/MethodCall/RenameVariableToMatchGetMethodNameRector.php
```diff
@@ -0,0 +1,97 @@
...
+
+        // @see https://regex101.com/r/hnU5pm/1/
+        $matches = Strings::match($name, '#get(([A-Z]).+)#');
+
+        if ($matches === null) {
+            return null;
+        }
+
+        return lcfirst($matches[1]);
+    }
```
TomasVotruba, line 96: Could you decouple this logic to 
https://github.com/rectorphp/rector/blob/master/rules/naming/src/Naming/ExpectedNameResolver.php ?

It's a service that should handle all these variable renames based on type/name. It can also re-use it's methods without duplication.

config/set/naming/naming.yaml
dobryy: I wasn't that clear... At the moment, the CI fails because of the new rule that has been added to the set. I'm not sure this issue should be fixed.

TomasVotruba: I get it. The code should always pass for Rector ruleset. If it's ignored, it kills the point of Rector rules helping us with code-reviews.

This should be enough to fix code locally:

```bash
bin/rector p --config rector-ci.yaml
```

rules/naming/src/Rector/MethodCall/RenameVariableToMatchGetMethodNameRector.php
dobryy: Ready

rules/naming/src/Naming/ExpectedNameResolver.php
```diff
@@ -148,6 +150,26 @@ public function resolveForAssignNew(Assign $assign): ?string
         return $this->propertyNaming->getExpectedNameFromType($fullyQualifiedObjectType);
     }
 
+    /**
+     * @param MethodCall $node
+     */
+    public function resolveForMethodCall(Node $node): ?string
```
TomasVotruba, line 156:
```suggestion
    public function resolveForMethodCall(MethodCall $methodCall): ?string
```

If it's really a `MethodCall`, it can be inlined... maybe next Rector rule? :D

config/set/naming/naming.yaml
dobryy: Thanks. Found an issue with the rule :) I'll change the PR status to Ready For Review once this rule is ready for your review.

rules/naming/src/Naming/ExpectedNameResolver.php
dobryy: My bad here, it also could be StaticCall

rules/naming/src/Rector/Assign/RenameVariableToMatchGetMethodNameRector.php
```diff
@@ -0,0 +1,118 @@
...
+        if ($callNode === null) {
+            return null;
+        }
+
+        $newName = $this->expectedNameResolver->resolveForMethodCall($callNode);
+        if ($newName === null || $this->isName($node, $newName)) {
+            return null;
+        }
+
+        /** @var Node\FunctionLike $classLikeNode */
```
TomasVotruba, line 91:
```suggestion
        /** @var Node\FunctionLike|null $classLikeNode */
```

rules/naming/src/Rector/Assign/RenameVariableToMatchGetMethodNameRector.php
```diff
@@ -0,0 +1,118 @@
...
+        $functionLikeNode = $this->betterNodeFinder->findFirstParentInstanceOf($node, Node\FunctionLike::class);
+        if ($functionLikeNode !== null) {
+            $nodes = $functionLikeNode->getStmts();
+        } else {
+            $functionLikeNode = $node->getAttribute(AttributeKey::PARENT_NODE);
+            $nodes = [];
+            /** @var Node\Stmt\Expression $expression */
+            while ($expression = $functionLikeNode->getAttribute(AttributeKey::NEXT_NODE)) {
+                $nodes[] = $expression;
+            }
```
TomasVotruba, line 101: What exactly is doing this? Maybe extra a method with clear name

rules/naming/src/Rector/Assign/RenameVariableToMatchGetMethodNameRector.php
```diff
@@ -0,0 +1,118 @@
...
+
+        /** @var Variable $variableNode */
+        $variableNode = $node->var;
+        $originalName = $variableNode->name;
+
+        $this->traverseNodesWithCallable($nodes, function ($node) use ($originalName, $newName): void {
+            if ($this->isVariableName($node, $originalName)) {
+                $node->name = $newName;
+            }
+        });
```
TomasVotruba, line 112: This is probably variable rename? Could be standalone method too, just to show in the name

rules/naming/tests/Rector/Assign/RenameVariableToMatchGetMethodNameRector/Fixture/arrow_function.php.inc
```diff
@@ -0,0 +1,19 @@
+<?php
+
+function testArrowFunction() {
```
TomasVotruba, line 3: Better always use current namespace in the fixture files.

It makes classs/functions unique and it prevents somebody else using the same `function test()` in another fixture file.

rules/naming/tests/Rector/Assign/RenameVariableToMatchGetMethodNameRector/Fixture/clousure.php.inc
```diff
@@ -0,0 +1,19 @@
+<?php
+
+$datetime = new DateTime();
+array_map(function ($element) use ($datetime) {
+    $a = $datetime->getTimestamp() + $element;
+    return $a;
+}, [1, 2, 3])
+
+?>
```
TomasVotruba, line 9: Having this file outside any function/class method means it will be included in test case. It's included, because reflection/function_exists/PHPStan needs it to reflect types correctly.

I usually wrap it like this:

```php
<?php

class ClosureClass
{
    public function run()
	{
		$datetime = new DateTime();
		array_map(function ($element) use ($datetime) {
		    $a = $datetime->getTimestamp() + $element;
		    return $a;
		}, [1, 2, 3])
	}
}
```

It also allows to add runnable interface in the future, to compare results.

```diff
<?php

-class ClosureClass
+class ClosureClass implements RunnableInterface
{
    public function run()
	{
		$datetime = new DateTime();
-		array_map(function ($element) use ($datetime) {
+		return array_map(function ($element) use ($datetime) {
		    $a = $datetime->getTimestamp() + $element;
		    return $a;
		}, [1, 2, 3])
	}
}
```

rules/naming/src/Rector/Assign/RenameVariableToMatchGetMethodNameRector.php
dobryy: This is for plain php script. In plain php script there is no parent node (or at least I didn't find how to get it) that is why here I'm getting all Expr this way after the node where there is an assignment.

I'll think about how to make it more understandable after two glasses of wine ;)

TomasVotruba: > This is for plain php script.

Btw, no need to cover these cases. They're pretty edge ones and most of rules don't handle spaghetti code. 

There are rules that help to migrate spaghetti code to classes and PSR-4, then the best practise is to apply rules like these. So you can remove this case + test fixture.

PR #3713, opened 2020-07-14
[Legacy] Correct test for AddTopIncludeRector

TomasVotruba:
Follow up to #3712


rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
```diff
@@ -27,74 +24,34 @@
...
-if (isset($_POST["csrf"])) {
+if (isset($_POST['csrf'])) {
     processPost($_POST);
 }
 PHP,
                 <<<'PHP'
-<?php
-
-include "autoloader.php";
+require_once __DIR__ . '/../autoloader.php';
```
TomasVotruba, line 52: Looking at include/require, the `require_once` makes the only sense.

- `include` allows missing the file, that would be false positive
- and `require` (without `_once`) can include the file twice, so it would break it

rules/legacy/src/Rector/Include_/AddTopIncludeRector.php
```diff
@@ -27,74 +24,34 @@
...
-     */
-    private $parser;
-
-    /**
-     * @var Standard
-     */
-    private $standard;
-
-    public function __construct(string $type = 'TYPE_INCLUDE', string $file = '"autoload.php"', array $match = [])
+    public function __construct(string $autoloadFilePath = '/autoload.php', array $match = [])
```
TomasVotruba, line 36: The only relevant string is relative path to the file, `__DIR__` is added by default, so no need to add it in the string, parse it with parse, extract the expression and it again.

File string is enough here

rules/legacy/tests/Rector/Include_/AddTopIncludeRector/AddTopIncludeRectorTest.php
```diff
@@ -4,33 +4,47 @@
...
-     * @dataProvider provideData()
-     */
-    public function test(SmartFileInfo $fileInfo): void
+    public function test(): void
+    {
+        $fileInfo = new SmartFileInfo(__DIR__ . '/Fixture/fixture.php.inc');
+        $temporaryFileInfo = $this->doTestFileInfo($fileInfo);
+
+        $this->assertStringEqualsFile(__DIR__ . '/Expected/expected_autoload.php', $temporaryFileInfo->getContents());
+    }
```
TomasVotruba, line 19: File system is a bit more complicated than normal tests, I've update it here so the file is tested.

rules/legacy/tests/Rector/Include_/AddTopIncludeRector/Expected/has_include.php.inc
```diff
@@ -1,13 +0,0 @@
...
-
-class SomeClass
-{
-
-    private function someMethod() : void
-    {
-        include 'somefile.php';
-    }
-
-}
```
TomasVotruba, line 13: If the file doesn't change at all, we only test the file is the same as input file. So this fixture can be removed

TomasVotruba: Actually, this fixture was testing false postive for `Class_`. The method that looked for include was never called.

I've chaged it only to function, so the `require_once` is checked

TomasVotruba:
@phpfui I left few notes that you might find useful for next rule.

Is there anything you'd change?

phpfui:
It all looks good, but I would not hard code __DIR__, as your include file is not always relative to the current directory of the source code.  My idea was to pass a valid string expression.  I generally do something relative to the document root. My use case is adding a common include to all plain php files that spit out html, but they are in various directories at different levels, so I can't use an absolute path.  Luckily all my files are served, so I always have document root available.  The class files have all includes stripped.  I my real code, I only use one include statement obviously where it is also handy for quick one off scripts.

Good idea on REQUIRE_ONCE, as that is really the only option that makes any sense for our purposes.

TomasVotruba:
> your include file is not always relative to the current directory of the source code.

I see, that's a good point.

So basically the behavior should be like this:

```diff
# someFile.php
+require_once __DIR__ . '/autoload.php';
```

```diff
# src/Nested/someFile.php
+require_once __DIR__ . '/../../autoload.php';
```

Is that correct or would you change something?

phpfui:
No, I would just let people insert what ever they feel like.  The job of the Rector is to add an include whatever;  We really don't care what they want to include, up to them.  Hard coding anything makes it less flexible.

When I first started this, I was not thinking about inserting an expression.  I was just going to do a global search and replace in my editor, but since we can insert a string expression, we should.

TomasVotruba:
The example is related to path nesting, not to inserted string.
Writing nodes in string configuratoin could lead broken code.

phpfui:
Yes, you could include something that is not valid PHP.  The first test would be to insure what is passed in can be parsed.  It should throw an exception if not.

Next, that valid code may not be valid in an include for sure.  But that is not our job, that is for the developer using Rector of figure out they are responsible for what they do.  I don't think the developer needs hand holding here (but OK for specifying type of include, we can do the right thing for them there).  I often run Rectors to see what they do, then throw out the changes.

But I do agree in general, that Rector should not output invalid code.  In this case, we are doing an insert and I am fine letting the developer hang themselves if they really want to.

TomasVotruba:
This is rather question of Rector design. Rector configuration should be strict in what it allows. If you can insert nodes in there, it would soon be YAML full of PHP code in strings. That's my main concern.

phpfui:
Yes, we are just inserting an include statement only.  I was originally just going to insert text after the include, but that is not how Rector (or PhpParser) work. You need to insert valid nodes, which is a good thing.  So the idea was to parse the PHP that came in through the include and add that as an Expression node.  So I think makes it type safe enough.

Ultimately PHP needs to be able to parse whatever Rector produces, and if the user misconfigures Rector, it is on them, not us.  PHP does not just parse anything, you have to give it valid PHP.  That is the job of the developer, not the tool.  You can misuse anything, We can put (and have put) guardrails on things, but at some point you have to trust the developer to not do stupid stuff.  God knows we are only here because developers do stupid stuff! But you can't fix stupid.  You can only guide them, but ultimately they are responsible for themselves.

In my first job, we had a saying, "given them enough rope", meaning they can hang themselves, but also you need a lot of rope to tie up a ship. You have to give the user the flexibility to do weird stuff.  I think Include_ with an Expression is enough rope but also limited to do the right thing for the stupid among us.

TomasVotruba:
> That is the job of the developer, not the tool

I disaggree about this. From DX of the point of view, they don't care about PHP code in configuration. Thats mixing concerns. 

The result in tests is the same, now there is just less space to do it wrong.

phpfui:
OK, easy enough to go a global search and replace once the Rector is done.

Thanks for adding this Rector!

TomasVotruba:
Thank you :+1:

phpfui:
Great working with you.

I am doing the Rector to add namespaces to new statements.  When I get it
working, I'll reconnect and we can figure out if you think it would be good
to add.  I think it is new, but not 100% sure yet.  My source is so bad, I
can't use the PSR4 Rectors in the current state of the source.

Like doing a puzzle!

TomasVotruba:
I look forward to it. I've been meeting such code bases more and more often.

In my experience, any super complex and shitty code base can be refactored, if we take just enough small piece of the puzzle. I often fight with myself, not to rush into first obvious pain, but rather step back, breathe and just take that little brick a put it on the other brick.

This often snowballs for a month, then we can do massive refactorings, that are as big as the code base itself. It really a zen job 😄

phpfui:
100% agreed!

Looks like I have something working.  Very simple.  But needs a config file.

Here is the issue.

No namespacing in ANY code.

Code is all:

$someClass = new SomeClass();
$someOtherClass = new SomeOtherClass();

I want to change it to:

$someClass = new \App\Class\SomeClass();
$someOtherClass = new \App\OtherNameSpace\SomeOtherClass();

Or even:

$someClass = new \App\Class\SomeClass();
$someOtherClass = new \App\Other\BetterName();

So I wrote a rector to find all class declarations.  Dumped the class name,
any existing namespace, and the file path into a CSV file.

I then converted to Excel, sorted, etc.  Added new_class_name and
new_namespace column, converted back to CSV

Then the Rector is simple to go find all New_ statements, get the class
name with namespace, look up the new class name and namespace, then rewrite
the node with the new information.

Works great!

Let me know if you are interested in it.

TomasVotruba:
Could you use formatting here? It's hard to read code as text.

Like this: ``` php (without space)
Code
Again: ```

phpfui:
I see the problem, I replied via email, and that does not support markdown.  But looks properly spaced.  I'll try to reply on the web.

PR #3717, opened 2020-07-15
[CodingStyle] Create new TernaryConditionVariableAssignmentRector

u01jmg3:
After discussion at https://github.com/symplify/symplify/issues/2006, here is a new Rector to convert the following:

```php
$value ? $a = 1 : $a = 0; // Before

$a = $value ? 1 : 0; // After
```

### DIFF
--- a/rules/coding-style/src/Rector/Ternary/TernaryConditionVariableAssignmentRector.php
+++ b/rules/coding-style/src/Rector/Ternary/TernaryConditionVariableAssignmentRector.php
@@ -1,0 +1,85 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\CodingStyle\Rector\Ternary;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\Assign;
+use PhpParser\Node\Expr\Ternary;
+use PhpParser\Node\Expr\Variable;
+use Rector\Core\Rector\AbstractRector;
+use Rector\Core\RectorDefinition\CodeSample;
+use Rector\Core\RectorDefinition\RectorDefinition;
+use Rector\NodeTypeResolver\Node\AttributeKey;
+
+/**
+ * @see \Rector\CodingStyle\Tests\Rector\Ternary\TernaryConditionVariableAssignmentRector\TernaryConditionVariableAssignmentRectorTest
+ */
+final class TernaryConditionVariableAssignmentRector extends AbstractRector
+{
+    public function getDefinition(): RectorDefinition
+    {
+        return new RectorDefinition('Assign outcome of ternary condition to variable, where applicable', [
+            new CodeSample(
+                <<<'PHP'
+function ternary($value)
+{
+    $value ? $a = 1 : $a = 0;
+}
+PHP
+                ,
+                <<<'PHP'
+function ternary($value)
+{
+    $a = $value ? 1 : 0;
+}
+PHP
+            ),
+        ]);
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getNodeTypes(): array
+    {
+        return [Ternary::class];
+    }
+
+    /**
+     * @param Ternary $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        $nodeIf = $node->if;
+        $nodeElse = $node->else;
+
+        if (! $nodeIf instanceof Assign || ! $nodeElse instanceof Assign) {
+            return null;
+        }
+
+        $nodeIfVar = $nodeIf->var;
+        $nodeElseVar = $nodeElse->var;
+
+        if (! $nodeIfVar instanceof Variable || ! $nodeElseVar instanceof Variable) {
+            return null;
+        }
+
+        if ($nodeIfVar->name !== $nodeElseVar->name) {
+            return null;
+        }
+
+        $previousNode = $node->getAttribute(AttributeKey::PREVIOUS_NODE);
+        if ($previousNode !== null) {
+            return null;
+        }
+
+        $node->if = $nodeIf->expr;
+        $node->else = $nodeElse->expr;
+
+        $variable = new Variable($nodeIfVar->name);
+
+        return new Assign($variable, $node);
+    }
+}
--- a/rules/coding-style/tests/Rector/Ternary/TernaryConditionVariableAssignmentRector/Fixture/fixture.php.inc
+++ b/rules/coding-style/tests/Rector/Ternary/TernaryConditionVariableAssignmentRector/Fixture/fixture.php.inc
@@ -1,0 +1,21 @@
+<?php
+
+namespace Rector\CodingStyle\Rector\Ternary\TernaryConditionVariableAssignmentRector\Fixture;
+
+function ternary($value)
+{
+    $value ? $a = 1 : $a = 0;
+}
+
+?>
+-----
+<?php
+
+namespace Rector\CodingStyle\Rector\Ternary\TernaryConditionVariableAssignmentRector\Fixture;
+
+function ternary($value)
+{
+    $a = $value ? 1 : 0;
+}
+
+?>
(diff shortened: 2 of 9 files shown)


TomasVotruba:
Thanks for the contribution, that was fast!

Could you look at failing CI? 

It's good practise to check PHPStan and ECS locally, like this:

```bash
composer phpstan
composer fix-cs
```

u01jmg3:
That should be it now

---

I did notice when trying to run `composer phpstan` and `composer fix-cs` on Windows that those commands don't work. Reason being the path needs to be quoted (e.g. `"vendor/bin/phpstan" analyse --ansi --error-format symplify` and `"vendor/bin/ecs" check --fix --ansi`).

I also had to add `--memory-option 2G` for PHPStan to complete. Furthermore, did you know there are 2 other errors that come up when you run PHPStan (not related to this PR)?

TomasVotruba:
> Furthermore, did you know there are 2 other errors that come up when you run PHPStan (not related to this PR)?

The CI passes, so I don't know about those errors. Probably Windows related.

Feel free to fix it, if that helps.

----


Could you squash it into 1 commit?

rules/coding-style/src/Rector/Ternary/TernaryConditionVariableAssignmentRector.php
```diff
@@ -0,0 +1,93 @@
...
+        $nodeIfExpr = $nodeIf->expr;
+        $nodeElseExpr = $nodeElse->expr;
+
+        if (! $nodeIfExpr instanceof LNumber || ! $nodeElseExpr instanceof LNumber) {
+            return null;
+        }
+
+        $node->if = new LNumber($nodeIfExpr->value);
+        $node->else = new LNumber($nodeElseExpr->value);
+
```
TomasVotruba, line 90: Does it work only with nubmers? What about strings, floats, basically any other value?

rules/coding-style/tests/Rector/Ternary/TernaryConditionVariableAssignmentRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,21 @@
...
+<?php
+
+namespace Rector\CodingStyle\Rector\Ternary\TernaryConditionVariableAssignmentRector\Fixture;
+
+function ternary()
+{
+    $a = true ? 1 : 0;
+}
+
+?>
```
TomasVotruba, line 21: Could you also always provide a fixture that is skipped? So have `return null;` varians covered too

TomasVotruba:
Ah, no need to squash now. I've found few spots to improve

u01jmg3:
Hopefully good to go now 🤞

As for squashing commits, you should be able to do so when you merge this PR: https://github.blog/2016-04-01-squash-your-commits/

rules/coding-style/tests/Rector/Ternary/TernaryConditionVariableAssignmentRector/Fixture/fixture.php.inc
u01jmg3: Fixed

rules/coding-style/src/Rector/Ternary/TernaryConditionVariableAssignmentRector.php
u01jmg3: Fixed

rules/coding-style/src/Rector/Ternary/TernaryConditionVariableAssignmentRector.php
```diff
@@ -0,0 +1,105 @@
...
+    public function createScalarWithValue(Expr $nodeExpr)
+    {
+        if (! $nodeExpr instanceof DNumber && ! $nodeExpr instanceof EncapsedStringPart && ! $nodeExpr instanceof LNumber && ! $nodeExpr instanceof String_) {
+            return null;
+        }
+
+        $nodeExprInstance = get_class($nodeExpr);
+
+        return new $nodeExprInstance($nodeExpr->value);
```
TomasVotruba, line 103: I wonder, why is this limit needed? What prevent using any Expr?

rules/coding-style/src/Rector/Ternary/TernaryConditionVariableAssignmentRector.php
```diff
@@ -0,0 +1,105 @@
...
+use PhpParser\Node\Scalar\String_;
+use Rector\Core\Rector\AbstractRector;
+use Rector\Core\RectorDefinition\CodeSample;
+use Rector\Core\RectorDefinition\RectorDefinition;
+use Rector\NodeTypeResolver\Node\AttributeKey;
+
+/**
+ * @see \Rector\CodingStyle\Tests\Rector\Ternary\TernaryConditionVariableAssignmentRector\TernaryConditionVariableAssignmentRectorTest
+ */
+final class TernaryConditionVariableAssignmentRector extends AbstractRector
```
TomasVotruba, line 24: The rule should be added in the set, `coding-style.yaml`

TomasVotruba:
> As for squashing commits

I'm aware of this feature, but I don't use it because creates 2 different histories - on in master (1 commit) one in here (zillion commits usually).

If you look at old PR like 2 years back and there is 20 tiny commity without any value, it's really painful to try to find out what happened and why.A

-----

As for PHPStan errors, I saw such errors before. They seem related to PHPStan phar autoloading mixed with our project. These 2 errors are not that relevant, so you can skip them. Yet you can try deleting PHPStan cache, not sure if that helps.

rules/coding-style/src/Rector/Ternary/TernaryConditionVariableAssignmentRector.php
u01jmg3: Good spot! 👀

u01jmg3:
No problem, shall squash once you are happy.

---

> _As for PHPStan errors, I saw such errors before. They seem related to PHPStan phar autoloading mixed with our project. These 2 errors are not that relevant, so you can skip them._

Added this locally to `phpstan.neon`, not sure if you want it in the core.

```yml
    ignoreErrors:
        -
            message: '#Class with base \"[a-zA-Z0-9\\_]+\" name is already used in#'
            paths:
                - src/Console/Application.php
                - src/PhpParser/Parser/Parser.php
```

TomasVotruba:
> Added this locally to phpstan.neon, not sure if you want it in the core.

Sure :+1:

u01jmg3:
🤞 🤞 🤞

Good to squash?

Lctrs:
Personally, I find the later quite confusing. Would it be possible to at least wrap the ternary inside parenthesis (eg. `$a = (true ? 1 : 0);` ?

TomasVotruba:
@Lctrs That has the same meaning as without brackets. The parsers doesn't print any clutter code. You'd  have to make own coding standard rule to handle that.

rules/coding-style/src/Rector/Ternary/TernaryConditionVariableAssignmentRector.php
```diff
@@ -0,0 +1,85 @@
...
+{
+    public function getDefinition(): RectorDefinition
+    {
+        return new RectorDefinition('Assign outcome of ternary condition to variable, where applicable', [
+            new CodeSample(
+                <<<'PHP'
+function ternary()
+{
+    true ? $a = 1 : $a = 0;
```
TomasVotruba, line 28: The `true` value is a bit confusing here, as it could be any expr (falseable).

Adding a function arg `$value` and using it would make it more clear.

TomasVotruba:
For me it's ready to merge :+1:

u01jmg3:
Super - got the next Rector lined up to PR once this is merged.

TomasVotruba:
Could you squah it to 1 commit?

u01jmg3:
Done

TomasVotruba:
Splendid! Thank you for new rule :+1:

PR #3732, opened 2020-07-17
[CodingStyle] Create new WrapVariableVariableNameInCurlyBracesRector

u01jmg3:
After discussion at https://github.com/symplify/symplify/issues/2006, here is a new Rector to convert the following:

```php
function run($foo)
{
    // Before: valid in PHP 5 only
    global $$foo->bar;

    // After: valid in PHP 5 and 7
    global ${$foo->bar};
}
```

The Rector also wraps other [variable variables](https://www.php.net/manual/en/language.variables.variable.php) in curly braces.

```php
$$value = true; // Before

${$value} = true; // After
```

### DIFF
--- a/config/set/coding-style/coding-style.php
+++ b/config/set/coding-style/coding-style.php
@@ -31,6 +31,7 @@ use Rector\CodingStyle\Rector\Ternary\TernaryConditionVariableAssignmentRector;
 use Rector\CodingStyle\Rector\Use_\RemoveUnusedAliasRector;
 use Rector\CodingStyle\Rector\Use_\SplitGroupedUseImportsRector;
 use Rector\CodingStyle\Rector\Variable\UnderscoreToPascalCaseVariableNameRector;
+use Rector\CodingStyle\Rector\Variable\WrapVariableVariableNameInCurlyBracesRector;
 use Rector\Php55\Rector\String_\StringClassNameToClassConstantRector;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 
@@ -98,6 +99,8 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     $services->set(UnderscoreToPascalCasePropertyNameRector::class);
 
     $services->set(UnderscoreToPascalCaseVariableNameRector::class);
+
+    $services->set(WrapVariableVariableNameInCurlyBracesRector::class);
 
     $services->set(RemoveDoubleUnderscoreInMethodNameRector::class);
 };
--- a/rules/coding-style/src/Rector/Variable/WrapVariableVariableNameInCurlyBracesRector.php
+++ b/rules/coding-style/src/Rector/Variable/WrapVariableVariableNameInCurlyBracesRector.php
@@ -1,0 +1,74 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\CodingStyle\Rector\Variable;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\PropertyFetch;
+use PhpParser\Node\Expr\Variable;
+use Rector\Core\Rector\AbstractRector;
+use Rector\Core\RectorDefinition\CodeSample;
+use Rector\Core\RectorDefinition\RectorDefinition;
+
+/**
+ * @see \Rector\CodingStyle\Tests\Rector\Variable\WrapVariableVariableNameInCurlyBracesRector\WrapVariableVariableNameInCurlyBracesRectorTest
+ * @see https://www.php.net/manual/en/language.variables.variable.php
+ */
+final class WrapVariableVariableNameInCurlyBracesRector extends AbstractRector
+{
+    public function getDefinition(): RectorDefinition
+    {
+        return new RectorDefinition('Ensure variable variables are wrapped in curly braces', [
+            new CodeSample(
+                <<<'PHP'
+function run($foo)
+{
+    global $$foo->bar;
+}
+PHP
+                ,
+                <<<'PHP'
+function run($foo)
+{
+    global ${$foo->bar};
+}
+PHP
+            ),
+        ]);
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getNodeTypes(): array
+    {
+        return [Variable::class];
+    }
+
+    /**
+     * @param Variable $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        $nodeName = $node->name;
+
+        if (! $nodeName instanceof PropertyFetch && ! $nodeName instanceof Variable) {
+            return null;
+        }
+
+        if ($node->getEndTokenPos() !== $nodeName->getEndTokenPos()) {
+            return null;
+        }
+
+        if ($nodeName instanceof PropertyFetch) {
+            return new Variable(new PropertyFetch($nodeName->var, $nodeName->name));
+        }
+
+        if ($nodeName instanceof Variable) {
+            return new Variable(new Variable($nodeName->name));
+        }
+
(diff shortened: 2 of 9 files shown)


u01jmg3:
2 failing tests appear to the same issue related to Git name/email

rules/coding-style/src/Rector/Variable/VariableVariablesRector.php
```diff
@@ -0,0 +1,87 @@
...
+                return null;
+            }
+
+            if (! is_string($node->name)) {
+                return null;
+            }
+
+            $node->name = new Variable($node->name);
+
+            return $node;
```
TomasVotruba, line 80: Could you extract this to own method? To separate this big long method complexity

rules/coding-style/tests/Rector/Variable/VariableVariablesRector/VariableVariablesRectorTest.php
```diff
@@ -0,0 +1,31 @@
...
+     * @dataProvider provideData()
+     */
+    public function test(SmartFileInfo $fileInfo): void
+    {
+        $this->doTestFileInfo($fileInfo);
+    }
+
+    public function provideData(): Iterator
+    {
+        return $this->yieldFilesFromDirectory(__DIR__ . '/Fixture');
```
TomasVotruba, line 24: Tests look good.

Could you also add 2 skipped fixture files? E.g. already migrated or ignored cases

rules/coding-style/src/Rector/Variable/VariableVariablesRector.php
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * @param Variable $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $node instanceof Variable) {
+            return null;
+        }
+
```
TomasVotruba, line 58:
```suggestion

```

This can be removed, as `getNodeTypes()` provides only `Variable` nodes

rules/coding-style/src/Rector/Variable/VariableVariablesRector.php
```diff
@@ -0,0 +1,87 @@
...
+{
+    public function getDefinition(): RectorDefinition
+    {
+        return new RectorDefinition('Ensure variable variables are wrapped in curly braces', [
+            new CodeSample(
+                <<<'PHP'
+function f()
+{
+    global $$foo->bar;
+}
```
TomasVotruba, line 29: Better use longer funciton name than `f`, could be reserved in the future, like `fn()` is now.

We use `run()` as convention here

u01jmg3: Changed - sorry, it came from the PHP docs!

rules/coding-style/tests/Rector/Variable/VariableVariablesRector/VariableVariablesRectorTest.php
u01jmg3: More than the 2 that exist already?

TomasVotruba: Yes, maybe some completely normal variable or variable with 3 $$$ (if that's possible)

rules/coding-style/src/Rector/Variable/VariableVariablesRector.php
u01jmg3: Turns out the code I had was more complicated than was needed! Simplified it to the point a separate method is not required

config/set/coding-style/coding-style.php
```diff
@@ -99,5 +100,7 @@
 
     $services->set(UnderscoreToPascalCaseVariableNameRector::class);
 
+    $services->set(VariableVariablesRector::class);
```
TomasVotruba, line 103: Btw, this name is not very clear IMO. Could you think of better name that described what is being changed from to?

u01jmg3: `WrapVariableVariables`? `WrapVariableVariablesInBraces`?

Unfortunately "variable variables" is exactly what this Rector applies to.

Open to other suggestions...

u01jmg3:
Are you fixing the PHPUnit tests?

config/set/coding-style/coding-style.php
u01jmg3: `UnwrappedVariableVariableNameToCurlyBracesVariableVariableNameRector`?

TomasVotruba: The first word should be a verb, since Rector is about changing from to.

But the rest is good :+1:

u01jmg3: In that case it should be using the word "wrap": `WrapVariableVariableNameInCurlyBracesRector`

TomasVotruba:
On the `master` branch test are passing. Look at the details:

```
1) Rector\CodingStyle\Tests\Rector\Variable\WrapVariableVariableNameInCurlyBracesRector\WrapVariableVariableNameInCurlyBracesRectorTest::test with data set #3 (Symplify\SmartFileSystem\SmartFileInfo Object (...))
ParseError: syntax error, unexpected '->' (T_OBJECT_OPERATOR), expecting ',' or ';'
```

rules/coding-style/src/Rector/Variable/WrapVariableVariableNameInCurlyBracesRector.php
```diff
@@ -0,0 +1,62 @@
...
+     * @param Variable $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if ($node->name instanceof PropertyFetch) {
+            return new Variable(new PropertyFetch($node->name->var, $node->name->name));
+        } elseif ($node->name instanceof Variable) {
+            return new Variable(new Variable($node->name->name));
+        }
+
```
TomasVotruba, line 59: It seems like huge amount of previous code is missing.
What happened?

This will re-create nodes from already correct code. That should not happen.

It's conventions to `return null;` on no-change, so it can report what rule has changed what node.

u01jmg3: > _It seems like huge amount of previous code is missing.
What happened?_

This is because I tested with both of my examples in the same file and found that only the property fetch was corrected. I had to run Rector again (second pass) before example 2 was corrected. It appears `getStartTokenPos()` gives a different result for `$$value = true;` depending on whether example 1 is `global $$foo->bar;` or `global ${$foo->bar};`

```php
global $$foo->bar; // Example 1

$$value = true; // Example 2
```

u01jmg3:
PHPUnit was failing because `global $$foo->bar;` is only valid in PHP 5. I've therefore removed the test.

TomasVotruba:
Great job, thank you :+1:

u01jmg3:
Not a problem. I see the total number of Rectors should be 527 but with all the rebases is still at 526.

TomasVotruba:
It usually requires to update the docs:

```bash
composer docs
```

u01jmg3:
> `composer docs`

Doesn't seem to work on Windows even properly quoting the path.

---

Also, it dawned on me that perhaps this Rector should have gone under the set `php70` rather than `coding-style` 🤷‍♂️

TomasVotruba:
> Doesn't seem to work on Windows even properly quoting the path.

Well, Windows. Are you able to fix it?


> Also, it dawned on me that perhaps this Rector should have gone under the set php70 rather than coding-style man_shrugging

Good one :+1:  Could you send PR?

PR #3738, opened 2020-07-19
[Naming] Skip non-object type returns, classes with children and typical naming patterns

TomasVotruba:
Follow up to #3737, #3706


rules/naming/tests/Rector/Assign/RenameVariableToMatchGetMethodNameRector/Fixture/clousure.php.inc
```diff
@@ -20,13 +22,15 @@ class ClosureClass
...
+use DateTime;
+
 class ClosureClass
 {
     public function run()
     {
         $datetime = new DateTime();
         array_map(function ($element) use ($datetime) {
-            $timestamp = $datetime->getTimestamp() + $element;
+            $timestamp = $datetime->getTimestamp();
```
TomasVotruba, line 33: This case shoud be skipped, as:

```php
apple + banana != 'apple'
```

It creates something new, so we can't guess name of the new thing

rules/naming/src/Rector/Assign/RenameVariableToMatchGetMethodNameRector.php
```diff
@@ -175,4 +190,25 @@ private function renameInDocComment(Node $node, string $originalName, string $ne
...
+    /**
+     * Keep cases like:
+     *
+     * $someNameSuffix = $this->getSomeName();
+     * $prefixSomeName = $this->getSomeName();
+     * $someName = $this->getSomeName();
+     *
+     * @param FuncCall|StaticCall|MethodCall $expr
+     */
+    private function shouldSkipForNamingConvention(Expr $expr, string $currentName, string $expectedName): bool
```
TomasVotruba, line 203: decouple

PR #3739, opened 2020-07-19
[Naming] Apply RenameVariableToMatchGetMethodNameRector on code + fix docblock rename

TomasVotruba:
Follow up to #3738, #3737, #3706

### DIFF
--- a/rules/naming/src/PhpDoc/VarTagValueNodeRenamer.php
+++ b/rules/naming/src/PhpDoc/VarTagValueNodeRenamer.php
@@ -38,9 +38,9 @@ final class VarTagValueNodeRenamer
     {
         $phpDocInfo = $node->getAttribute(AttributeKey::PHP_DOC_INFO);
 
-        $parent = $node->getAttribute(AttributeKey::PARENT_NODE);
-        if ($parent instanceof Expression) {
-            $expressionPhpDocInfo = $parent->getAttribute(AttributeKey::PHP_DOC_INFO);
+        $expression = $node->getAttribute(AttributeKey::CURRENT_STATEMENT);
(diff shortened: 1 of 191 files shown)


rules/naming/src/PhpDoc/VarTagValueNodeRenamer.php
```diff
@@ -0,0 +1,48 @@
...
+     * Expression doc block has higher priority
+     */
+    private function resolvePhpDocInfo(Node $node): ?PhpDocInfo
+    {
+        $phpDocInfo = $node->getAttribute(AttributeKey::PHP_DOC_INFO);
+
+        $parent = $node->getAttribute(AttributeKey::PARENT_NODE);
+        if ($parent instanceof Expression) {
+            $expressionPhpDocInfo = $parent->getAttribute(AttributeKey::PHP_DOC_INFO);
+        }
```
TomasVotruba, line 44: @dobryy So this was the issue :face_with_head_bandage: 

The PhpDocInfo is used twice, so it's accessible on the `Assign` node (so you don't have to always traverse up to `Expression`).

<br>

`Assign` (no `;` in the end)

```php
/** @var Node $name */
$name = $this->getName();
```

That was printer correctly. But overriden in the print by parent expression.

<br>

`Expression`

```php
/** @var Node $a */
$name = $this->getName();
```

Only the `Expression` is printed.

TomasVotruba: After I decopuled all classes I could (~3 hours), it took me around 20 minutes to find this one. It's easier for me to work only with current minimal scope of what is going on.

PR #3757, opened 2020-07-22
Quote paths so commands work on Windows machines

u01jmg3:
💻

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -265,9 +265,9 @@
             "bin/rector sync-types",
             "bin/rector check-static-type-mappers"
         ],
-        "check-cs": "vendor/bin/ecs check --ansi",
-        "fix-cs": "vendor/bin/ecs check --fix --ansi",
-        "phpstan": "vendor/bin/phpstan analyse --ansi --error-format symplify",
+        "check-cs": "\"vendor/bin/ecs\" check --ansi",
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Hi, thanks for your time.

Windows really suck at this.
We used this approach in the past, but quoting quotes usually goes to mess in very bad way. Like 4 hours of looking for the issue in the Docker, loking for single config way.

Another way is to use just `ecs` over `vendor/bin/ecs` path, since composer autocompletes the path.
But that lead to copy-paste WTFs while using script on CI, Docker etc, because in `ecs` itself without composer doesn' work.

I know this would a quick fix that helps you. But it proven to be very costly in maintenance for everyone else. Saying that, we'll keep as it is.

Still, there is a way to resolve this :) 
I recommend using own shortcuts in bash command line.

Eg. I have `cs` that:


```bash
# .bashrc
alias cs="vendor/bin/ecs ..."
alias fs="vendor/bin/ecs ... --fix"
```

You can use it any project, regardless their custom `composer` shorcuts.

u01jmg3:
Not a problem

I already use doskeys which is very similar to the aliases you suggest.

My only recommendation would be checking if a contributor has a Windows machine before mentioning `composer phpstan`, etc. otherwise they will run into problems!

Issue #3726, opened 2020-07-16
Incorrect unused assignment dedection


GrahamCampbell:
# Bug Report

<img width="643" alt="image" src="https://user-images.githubusercontent.com/2829600/87694338-f8627000-c785-11ea-9e38-4d1f369a4f6f.png">

Applied rules:

 * `Rector\DeadCode\Rector\Assign\RemoveUnusedAssignVariableRector`

TomasVotruba:
We need reproducible demo: https://getrector.org/demo

GrahamCampbell:
Here you go: https://getrector.org/demo/7ef107d7-452b-46d4-bec7-2189f758e3fc.

TomasVotruba:
Thanks!

Btw, why is input mix of tabs and spaces? It makes diff hard to read

dobryy:
@TomasVotruba, the issue occurs because there are multiple assignments for the same variable name. At the moment, the rule removes assignments and leaves just the last one. Here is the [test fixture](https://github.com/rectorphp/rector/blob/master/rules/dead-code/tests/Rector/Assign/RemoveUnusedAssignVariableRector/Fixture/assign_multi_same_line.php.inc) that describes the logic and explains the current issue.

The solution might be: Add additional check that the variable is not used in the statement (in FunctionLike params, `use` of closure and a regular statement like `$a = $a + 1;`). Is there any rule that has something similar?

TomasVotruba:
> the issue occurs because there are multiple assignments for the same variable name

I see. Here is just used in the expression right before assigned. While in the linked fixture its only on the left side.

This should be skipped too:

```php
$values = [1, 2];
$values = $this->whatever($values);
```

> Is there any rule that has something similar?

This is rule of its own kind. I don't think so.

PR #3759, opened 2020-07-22
[DeadCode] Fix variable usage detection

dobryy:
Closes https://github.com/rectorphp/rector/issues/3726

### DIFF
--- a/rules/dead-code/src/NodeFinder/NextVariableUsageNodeFinder.php
+++ b/rules/dead-code/src/NodeFinder/NextVariableUsageNodeFinder.php
@@ -6,9 +6,12 @@ namespace Rector\DeadCode\NodeFinder;
 
 use PhpParser\Node;
 use PhpParser\Node\Expr\Assign;
+use PhpParser\Node\Expr\Variable;
 use PhpParser\NodeTraverser;
+use Rector\Core\PhpParser\Node\BetterNodeFinder;
 use Rector\Core\PhpParser\NodeTraverser\CallableNodeTraverser;
 use Rector\Core\PhpParser\Printer\BetterStandardPrinter;
+use Rector\NodeNameResolver\NodeNameResolver;
 use Rector\NodeNestingScope\ParentScopeFinder;
 use Rector\NodeTypeResolver\Node\AttributeKey;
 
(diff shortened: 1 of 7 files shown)


rules/dead-code/src/NodeFinder/NextVariableUsageNodeFinder.php
```diff
@@ -80,4 +81,25 @@ public function find(Assign $assign): ?Node
 
         return $nextUsageOfVariable;
     }
+
+    private function findInParentExpression(Assign $assign, Variable $variable): bool
```
TomasVotruba, line 85: `find*` usually return object/array, e.g. `findProduct` returns `Product`

Here should be `is*/has*` prefix... 

I hear you, this should be already in PHPStan rules :thinking: 

<br>

A better position for this method might be `BetterNodeFinder`

rules/dead-code/tests/Rector/Assign/RemoveUnusedAssignVariableRector/Fixture/skip_assign_expression_class_use_variable.php.inc
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\DeadCode\Tests\Rector\Assign\RemoveUnusedAssignVariableRector\Fixture;
+
+final class SkipAssignExpressionClassUseVariable
+{
+    public function run(string $foo = null, string $bar = null)
```
TomasVotruba, line 9:
```suggestion
    public function run()
```

Test fixture case should be as minimal as possible to make it easier in 2 years to debug them.

From the code I assume this is never used to test as tested variable name is `datetime`

rules/dead-code/src/NodeFinder/NextVariableUsageNodeFinder.php
dobryy: How often do you hear voices? :D

rules/dead-code/tests/Rector/Assign/RemoveUnusedAssignVariableRector/Fixture/skip_assign_expression_class_use_variable.php.inc
dobryy: You're right, missed during some testing.

PR #3760, opened 2020-07-22
[Nette] Add form dim access to standalone Form Control

TomasVotruba:


config/config.php
```diff
@@ -28,8 +28,6 @@
 
     $parameters->set(Option::AUTOLOAD_PATHS, []);
 
-    $parameters->set('rector_recipe', []);
```
TomasVotruba, line 31: Already covered in split config

PR #3768, opened 2020-07-23
Test to skip foreach rewrite on IteratorAggregate

bohacpetr:
Simplifying foreach should be skipped on IteratorAggregate

### DIFF
--- a/rules/code-quality/tests/Rector/Foreach_/ForeachItemsAssignToEmptyArrayToAssignRector/Fixture/skip_iterator_aggregate.php.inc
+++ b/rules/code-quality/tests/Rector/Foreach_/ForeachItemsAssignToEmptyArrayToAssignRector/Fixture/skip_iterator_aggregate.php.inc
@@ -1,0 +1,28 @@
+<?php
+declare(strict_types=1);
+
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Hi, thanks for test fixture!

Could you provide *why* this shoud be skipped? I don't work with iterable object types, so I have no idea

bohacpetr:
1) It's `TypeError`. `SkipIterator::fetchAll()` is expected to return `array`
2) It's not equivalent. This will not work:
```
$values = (new SkipIterator())->fetchAll();
$values[0]; // $values is object without ArrayAccess
```

I also opened new issue https://github.com/rectorphp/rector/issues/3769

bohacpetr:
Oups, I'm taking back, on actual version this works :) I swear, that some older versions had problems with that. Sorry to bother

TomasVotruba:
We can actually test results of tests before and after :)

If you add the runnable interface and return, they will be compared. That way we'd know for sure.

See

https://github.com/rectorphp/rector/blob/6182320da435247ac8f94848ba25be7e5e381a55/rules/php70/tests/Rector/MethodCall/ThisCallOnStaticMethodToStaticCallRector/Fixture/fixture.php.inc

Issue #3716, opened 2020-07-15
[PHP7.4] Invalid type with ::class in ORM


shyim:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | latest              |
| Installed as   | Docker                      |

## Minimal PHP Code Causing Issue

```php
class Download
{
    /**
     * @ORM\Id()
     * @ORM\GeneratedValue()
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @ORM\ManyToOne(targetEntity=Package::class, inversedBy="downloads")
     * @ORM\JoinColumn(nullable=false)
     */
    private $package;
}
```
getting to
```php
class Download
{
    /**
     * @ORM\Id()
     * @ORM\GeneratedValue()
     * @ORM\Column(type="integer")
     */
    private int $id;

    /**
     * @ORM\ManyToOne(targetEntity=Package::class, inversedBy="downloads")
     * @ORM\JoinColumn(nullable=false)
     */
    private \Package::class $package;
}
```

## Expected Behaviour

I guess we should just replace `::class`

TomasVotruba:
- Could you provide demo link?
- What is Rector version?
- OS?

shyim:
https://getrector.org/demo/24a6eb87-7be6-4aad-af9e-4a3434dead6e#result

Version: 0.8.x-dev@ae3adef
OS: NixOs (Its wayne its docker)

TomasVotruba:
Interesting. Could you send PR with failing test fixture to the rule?

I can fix it right away then.

PR #3770, opened 2020-07-23
[PHP7.4] Invalid type with class keyword in Doctrine ORM property

dobryy:
Related: https://github.com/rectorphp/rector/issues/3716

### DIFF
--- a/packages/better-php-doc-parser/src/PhpDocNodeFactory/AbstractPhpDocNodeFactory.php
+++ b/packages/better-php-doc-parser/src/PhpDocNodeFactory/AbstractPhpDocNodeFactory.php
@@ -49,6 +49,7 @@ abstract class AbstractPhpDocNodeFactory
 
     protected function resolveFqnTargetEntity(string $targetEntity, Node $node): string
     {
+        $targetEntity = $this->getCleanedUpTargetEntity($targetEntity);
         if (class_exists($targetEntity)) {
             return $targetEntity;
         }
@@ -75,5 +76,10 @@ abstract class AbstractPhpDocNodeFactory
         $closingSpace = $match['closing_space'] ?? '';
 
         return [$openingSpace, $closingSpace];
+    }
+
+    private function getCleanedUpTargetEntity(string $targetEntity): string
+    {
+        return Strings::replace($targetEntity, '#::class#', '');
     }
 }
--- a/rules/php74/tests/Rector/Property/TypedPropertyRector/FixtureDoctrine/doctrine_one_to_one_class_keyword.php.inc
+++ b/rules/php74/tests/Rector/Property/TypedPropertyRector/FixtureDoctrine/doctrine_one_to_one_class_keyword.php.inc
@@ -1,0 +1,31 @@
+<?php
+
+namespace Rector\Php74\Tests\Rector\Property\TypedPropertyRector\Fixture;
+
+use Doctrine\ORM\Mapping as ORM;
+
+class DoctrineOneToManyClassKeywordClass
+{
+    /**
+     * @ORM\OneToOne(targetEntity=Rector\Php74\Tests\Rector\Property\TypedPropertyRector\Source\SomeParent::class)
+     * @ORM\JoinColumn(nullable=false)
+     */
+    private $package;
+}
+?>
+-----
+<?php
+
+namespace Rector\Php74\Tests\Rector\Property\TypedPropertyRector\Fixture;
+
+use Doctrine\ORM\Mapping as ORM;
+
+class DoctrineOneToManyClassKeywordClass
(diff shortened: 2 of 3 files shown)


rules/type-declaration/src/TypeInferer/PropertyTypeInferer/DoctrineRelationPropertyTypeInferer.php
```diff
@@ -88,6 +89,7 @@ private function processToOneRelation(
 
         $fullyQualifiedTargetEntity = $toOneTagNode->getFullyQualifiedTargetEntity();
```
TomasVotruba, line 90: I think method should return class name without `::class`, as it's in the name.
Why not use it?

rules/php74/tests/Rector/Property/TypedPropertyRector/FixtureDoctrine/doctrine_one_to_one_class_keyword.php.inc
```diff
@@ -0,0 +1,31 @@
...
+
+use Doctrine\ORM\Mapping as ORM;
+
+class DoctrineOneToManyClassKeywordClass
+{
+    /**
+     * @ORM\OneToOne(targetEntity=Rector\Php74\Tests\Rector\Property\TypedPropertyRector\Source\SomeParent::class)
+     * @ORM\JoinColumn(nullable=false)
+     */
+    private \Rector\Php74\Tests\Rector\Property\TypedPropertyRector\Source\SomeParent $package;
```
TomasVotruba, line 29:
```suggestion
    /**
     * @ORM\OneToOne(targetEntity=SomeParent::class)
     * @ORM\JoinColumn(nullable=false)
     */
    private \Rector\Php74\Tests\Rector\Property\TypedPropertyRector\Source\SomeParent $package;
```

There should be 1 case for imported class too

TomasVotruba:
Solution looks elegant :+1:

rules/type-declaration/src/TypeInferer/PropertyTypeInferer/DoctrineRelationPropertyTypeInferer.php
dobryy: https://github.com/rectorphp/rector/pull/3770/commits/24d7b66a87c6adf5b9b8c23ecc03998eaa87dc59#diff-5641698a38562059eecf39ef68b4b9efL62-R62

What about this place to cleanup the FQCN?

TomasVotruba: Great!

packages/better-php-doc-parser/src/PhpDocNodeFactory/AbstractPhpDocNodeFactory.php
```diff
@@ -49,6 +49,7 @@ protected function resolveContentFromTokenIterator(TokenIterator $tokenIterator)
 
     protected function resolveFqnTargetEntity(string $targetEntity, Node $node): string
     {
+        $targetEntity = $this->getCleanedUpTargetEntity($targetEntity);
```
dobryy, line 52: I moved it to this place, it might be better to clean it up in the very beginning

rules/php74/tests/Rector/Property/TypedPropertyRector/FixtureDoctrine/doctrine_one_to_one_class_keyword.php.inc
dobryy: Are you sure? :) I checked some tests and all of them use FQCN... When I use the imported class name it always adds `\` before the class name to make it kinda FQCN and do not resolve it correctly :/ and the bug is actually here
https://github.com/rectorphp/rector/pull/3770/files#diff-5641698a38562059eecf39ef68b4b9efR57

`$node->getAttribute(AttributeKey::NAMESPACE_NAME)` returns the namespace of the current test fixture class and not the namespace of the imported class, thus `class_exists` will always return `false`

```diff
-Rector\Php74\Tests\Rector\Property\TypedPropertyRector\Fixture
+Rector\Php74\Tests\Rector\Property\TypedPropertyRector\Source
```

TomasVotruba: Oh, it's broken in the same as in input. I haven't noticed that.

If it works, then it's ok :+1:

dobryy:
@TomasVotruba, ready for merging

TomasVotruba:
Thank you :+1:

PR #3802, opened 2020-07-27
misc

TomasVotruba:


phpstan.neon
```diff
@@ -202,9 +191,9 @@ parameters:
 
         # mixed
         - '#in iterable type Iterator#'
-        - '#with no typehint specified#'
+#        - '#with no typehint specified#'
         - '#return type has no value type specified in iterable type array#'
-        - '#has no return typehint#'
+#        - '#has no return typehint#'
```
TomasVotruba, line 196: @dobryy Do you recall as you've often reported that PHPStan should report the missing type? I've noticed it's ignored and conflicting mixed type removal from PHP CS Fixer. We had like 115 missing cases where type-hint could be used :)

Now it's fixed and next time PHPStan will till us :+1:

Issue #3807, opened 2020-07-27
Order of elements is changed at each new run of Rector


Aerendir:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | v0.7.54              |
| Installed as   | composer dependency                       |

Each time I run Rector, it applies:

 * `Rector\Order\Rector\Class_\OrderClassConstantsByIntegerValueRector`
 * `Rector\Order\Rector\Class_\OrderPropertyByComplexityRector`

## Minimal PHP Code Causing Issue

Demo is impossible to use due to alert about the use of yaml configuration.

Anyway the source code is available here: https://github.com/Aerendir/component-console-styles/blob/dev/src/Console/Style/SerendipityHQStyle.php

***NOTE: Reactivate the set `ORDER` in `rector.php`.***

If you like, you can clone locally, then run

```
composer install && vendor/bin/rector process src/Console/Style/SerendipityHQStyle.php
```

Then call again `vendor/bin/rector process src/Console/Style/SerendipityHQStyle.php` and again and again: each time the linked class is changed.

You can be sure you will get the same environment and dependencies I have as the versions are fixed.

Those are some examples of some consecutive run:

**RUN 1**
```diff
    ---------- begin diff ----------
--- Original
+++ New
@@ -41,6 +41,8 @@
  */
 final class SerendipityHQStyle extends OutputStyle
 {
+    /** @var SymfonyQuestionHelper $questionHelper */
+    private $questionHelper;
     /** @var int */
     private const MAX_LINE_LENGTH = 120;

@@ -49,9 +51,6 @@

     /** @var InputInterface $input */
     private $input;
-
-    /** @var SymfonyQuestionHelper $questionHelper */
-    private $questionHelper;

     /** @var ProgressBar $progressBar */
     private $progressBar;
    ----------- end diff -----------
```

**RUN 2**
```diff
    ---------- begin diff ----------
--- Original
+++ New
@@ -41,6 +41,8 @@
  */
 final class SerendipityHQStyle extends OutputStyle
 {
+    /** @var InputInterface $input */
+    private $input;
     /** @var SymfonyQuestionHelper $questionHelper */
     private $questionHelper;
     /** @var int */
@@ -48,9 +50,6 @@

     /** @var string */
     private const CAUTION = 'CAUTION';
-
-    /** @var InputInterface $input */
-    private $input;

     /** @var ProgressBar $progressBar */
     private $progressBar;
[...]

PR #3826, opened 2020-07-29
[Order] Order properties and constants with respect of complexity and position

dobryy:
Closes https://github.com/rectorphp/rector/issues/3807

### DIFF
--- a/rules/order/src/Rector/Class_/OrderPropertyByComplexityRector.php
+++ b/rules/order/src/Rector/Class_/OrderPropertyByComplexityRector.php
@@ -27,6 +27,14 @@ final class OrderPropertyByComplexityRector extends AbstractRector
      * @var PropertyRanker
      */
     private $propertyRanker;
+    /**
+     * @var string
+     */
+    private const RANK = 'rank';
+    /**
+     * @var string
+     */
+    private const POSITION = 'position';
 
     public function __construct(PropertyRanker $propertyRanker, StmtOrder $stmtOrder)
     {
@@ -109,8 +117,8 @@ PHP
                 $propertyName = $this->getName($property);
 
                 $propertyPositionByName[$position] = $propertyName;
-                $propertyNameToRank[$propertyName]['rank'] = $this->propertyRanker->rank($property);
-                $propertyNameToRank[$propertyName]['position'] = $position;
+                $propertyNameToRank[$propertyName][self::RANK] = $this->propertyRanker->rank($property);
+                $propertyNameToRank[$propertyName][self::POSITION] = $position;
             }
 
             $sortedPropertyByRank = $this->getSortedPropertiesByRankAndPosition($propertyNameToRank);
@@ -159,9 +167,9 @@ PHP
         uasort(
             $propertyNameToRank,
             function (array $firstArray, array $secondArray): int {
-                return [$firstArray['rank'], $firstArray['position']] <=> [
-                    $secondArray['rank'],
-                    $secondArray['position'],
+                return [$firstArray[self::RANK], $firstArray[self::POSITION]] <=> [
+                    $secondArray[self::RANK],
+                    $secondArray[self::POSITION],
                 ];
             }
         );
--- a/rules/order/src/Rector/Class_/OrderPropertyByComplexityRector.php
+++ b/rules/order/src/Rector/Class_/OrderPropertyByComplexityRector.php
@@ -19,6 +19,16 @@ use Rector\Order\StmtOrder;
 final class OrderPropertyByComplexityRector extends AbstractRector
 {
     /**
+     * @var string
+     */
+    private const RANK = 'rank';
+
+    /**
+     * @var string
+     */
+    private const POSITION = 'position';
+
+    /**
      * @var StmtOrder
      */
     private $stmtOrder;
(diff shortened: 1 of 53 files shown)


dobryy:
@TomasVotruba This is not the final version but I'd like to share the progress. Some decoupling is required.

To make it work correctly I had to sort by visibility, complexity, and name. There is still one problem with Traits. I created a test fixture for it and will look into it tomorrow.

TomasVotruba:
There are 3 fixes by Rector CI now. There should be just 2:

- 1 for Rector fixes
- and 1 for coding standard fixing the changes

image

That probably means Rector is needed to run twice to work, which is not good. I should be just once.
Could you fix that before review?

dobryy:
> There are 3 fixes by Rector CI now. There should be just 2:
> 
> * 1 for Rector fixes
> * and 1 for coding standard fixing the changes
> 
> image
> 
> That probably means Rector is needed to run twice to work, which is not good. I should be just once.
> Could you fix that before review?

Everything is fine, it runs only two times. It looks like three times because I squashed two previous fixups into one and then it created another two after the rebase. I'll reset the branch to only my changes to make sure it runs fine.

dobryy:
> That probably means Rector is needed to run twice to work, which is not good. I should be just once.

Is this relevant to all rules? It seems like for ordering rules (e.g. OrderPrivateMethodsByUseRector) it is very possible that Rector will need to run couple of times.

TomasVotruba:
> Everything is fine, it runs only two times. It looks like three times because I squashed two previous fixups into one and then it created another two after the rebase. I'll reset the branch to only my changes to make sure it runs fine.

Ah, I see. Thanks for resetting!

TomasVotruba:
> Is this relevant to all rules? It seems like for ordering rules (e.g. OrderPrivateMethodsByUseRector) it is very possible that Rector will need to run couple of times.

It's bad practise and should be avoided if possible. All Rector rules should be deterministics. If it's not the case, it's mostly likely lack of effort :D

dobryy:
> > Is this relevant to all rules? It seems like for ordering rules (e.g. OrderPrivateMethodsByUseRector) it is very possible that Rector will need to run couple of times.
> 
> It's bad practise and should be avoided if possible. All Rector rules should be deterministic:
> 
> * A → B  heavy_check_mark
> * ~A → B → C~ no_entry
> 
> If it's not the case, it's mostly likely lack of effort :D

Lack of effort? It's not about us! :D

rules/order/src/Rector/Class_/OrderPropertyByComplexityRector.php
```diff
@@ -100,50 +102,53 @@ public function refactor(Node $node): ?Node
...
 
-            foreach ($propertyByPosition as $position => $property) {
+            foreach ($resolvePropertyByPosition as $position => $property) {
                 /** @var string $propertyName */
                 $propertyName = $this->getName($property);
 
                 $propertyPositionByName[$position] = $propertyName;
-                $propertyNameToRank[$propertyName] = $this->propertyRanker->rank($property);
+                $propertyNameToRank[$propertyName]['rank'] = $this->propertyRanker->rank($property);
+                $propertyNameToRank[$propertyName]['position'] = $position;
```
TomasVotruba, line 115: I see. This is the spot that fixes it all? :)

rules/order/src/Rector/Class_/OrderPropertyByComplexityRector.php
```diff
@@ -100,50 +102,53 @@ public function refactor(Node $node): ?Node
...
     /**
-     * @return Property[][]
+     * @return Property[][]|ClassConst[][]
      */
     private function resolvePropertyByVisibilityByPosition(Class_ $class): array
     {
         $propertyByVisibilityByPosition = [];
         foreach ($class->stmts as $position => $classStmt) {
-            if (! $classStmt instanceof Property) {
+            if (! $classStmt instanceof Property && ! $classStmt instanceof ClassConst) {
```
TomasVotruba, line 151: I see `OrderPropertyByComplexityRector` but here it orderes `ClassConst` as well? That's confusing.

It should be 2 separated rules or ClassConst skipped here.

rules/order/src/Rector/Class_/OrderPropertyByComplexityRector.php
```diff
@@ -153,4 +158,13 @@ private function resolvePropertyByVisibilityByPosition(Class_ $class): array
...
     }
+
+    private function getSortedPropertiesByRankAndPosition(array $propertyNameToRank): array
+    {
+        $rank = array_column($propertyNameToRank, 'rank');
+        $position = array_column($propertyNameToRank, 'position');
+        array_multisort($rank, SORT_ASC, $position, SORT_ASC, $propertyNameToRank);
+
+        return array_keys($propertyNameToRank);
+    }
```
TomasVotruba, line 169: This is very hard read to me. What should it do?

Wouldn't battleship do it?

See the last code snippet here: https://blog.martinhujer.cz/clever-way-to-sort-php-arrays-by-multiple-values/

rules/order/src/StmtOrder.php
```diff
@@ -44,7 +44,7 @@ public function reorderClassStmtsByOldToNewKeys(Class_ $class, array $oldToNewKe
             // reorder here
             $newKey = $oldToNewKeys[$key];
 
-            $reorderedStmts[$newKey] = $stmt;
+            $reorderedStmts[$key] = $class->stmts[$newKey];
```
TomasVotruba, line 47: This looks like a bug fix. Why this change?

TomasVotruba:
I've added few comments

Issue #3764, opened 2020-07-22
RemoveUnusedAliasRector incorrectly removing a used alias


u01jmg3:
# Bug Report

| Subject        | Details             |
| :------------- | :-------------------|
| Rector version | `v0.7.48`           |
| Installed as   | composer dependency |

[`Rector\CodingStyle\Rector\Use_\RemoveUnusedAliasRector`](https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#removeunusedaliasrector) is incorrectly removing a *used* alias

```diff
<?php

namespace App\Helpers;

-use Stripe\PaymentIntent as StripePaymentIntent;
+use Stripe\PaymentIntent;

class StripeHelper
{
    public static function createPaymentIntent()
    {
        return StripePaymentIntent::create([
            'receipt_email' => auth()->user()->email,
        ]);
    }
}
```

## Minimal PHP Code Causing Issue

- https://getrector.org/demo/675867b4-7d2a-40cc-9abf-50cb78daa96c
- (Also, there looks to be an issue with the result output on getrector.org warning about switching from the YAML format)

## Expected Behaviour

Rector should *not* be removing the alias as it is in use

TomasVotruba:
Hi, thanks for report :+1:

PR #3845, opened 2020-07-31
Add failing fixture for RemoveUnusedAliasRector

u01jmg3:
Ref #3764

### DIFF
--- a/rules/coding-style/src/Rector/Use_/RemoveUnusedAliasRector.php
+++ b/rules/coding-style/src/Rector/Use_/RemoveUnusedAliasRector.php
@@ -184,6 +184,10 @@ PHP
             return true;
         }
 
+        if (isset($this->useNamesAliasToName[$lowercasedLastName], $this->useNamesAliasToName[$lowercasedLastName][0]) && $this->useNamesAliasToName[$lowercasedLastName][0] === $aliasName) {
+            return true;
+        }
+
         // part of some @Doc annotation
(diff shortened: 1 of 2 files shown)


u01jmg3:
@TomasVotruba: on attempting to fix #3764 I've also upset 4 other previously passing tests. 

Therefore, what is the best way of determining that a static method call is present and thus an alias should be kept?

u01jmg3:
@TomasVotruba: on attempting to fix #3764 I've also upset 4 other previously passing tests. 

Therefore, what is the best way of determining that a static method call is present and thus an alias should be kept?

TomasVotruba:
I'm sorry I missed this PR.

> Therefore, what is the best way of determining that a static method call is present and thus an alias should be kept?

Why not change the name of the alias class? The alias is not needed anymore in this case.

You can either use `NodeFinder` or `CallableNodeTraverser` to find particular nodes.

TomasVotruba:
I'll look into it

TomasVotruba:
Yes, see https://github.com/rectorphp/rector/pull/4068

u01jmg3:
Thanks - I am finding that the rector is still not quite working for constants.

e.g.

```diff
-use Stripe\PaymentIntent as StripePaymentIntent;
+use Stripe\PaymentIntent;

if (StripePaymentIntent::STATUS_SUCCEEDED) {
    return;
}
```

TomasVotruba:
I see. I linked you in the #4068 PR to ask for a fix

PR #3850, opened 2020-07-31
method call remover rector

### LINKED ISSUE #3843: Chain of methods replacer rector
# Question

Hey, I was just looking into all rectors documented and I cannot find a way how to replace chain of methods, is there any way or there is no rector that can do that so far?

Example:

Before
```php
final class Car
{
    private CarType $carType;
    
    public function getCarType() :CarType
    {
        return $this->carType;
    }
}

final class CarType
{
    private string $type;

    public function getType(): string
    {
        return $this->type;
    }
}

final class SomeClass
{
    public function something()
    {
        $car = new Car();
        $car->getCarType()->getType();
    }
}
```

After:

```php
final class Car
{
    private string $type;
    
    public function getType() :string
    {
        return $this->type;
    }
}

final class SomeClass
{
    public function something()
    {
        $car = new Car();
        $car->getType();
    }
}
```

Does not even have to remove CarType class and replace attribute in Car class, as that is simple, I am just looking for way how to refactor all `$car->getCarType()->getType()` to `$car->getType()`.

mssimi:
Closes #3843

As stated in issue I have no idea how to retrieve variable name.

### DIFF
--- a/rules/generic/src/Rector/MethodCall/MethodCallRemoverRector.php
+++ b/rules/generic/src/Rector/MethodCall/MethodCallRemoverRector.php
@@ -8,6 +8,7 @@ use PhpParser\Node;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\Variable;
 use Rector\Core\Contract\Rector\ConfigurableRectorInterface;
+use Rector\Core\PhpParser\Node\Manipulator\MethodCallManipulator;
 use Rector\Core\Rector\AbstractRector;
 use Rector\Core\RectorDefinition\ConfiguredCodeSample;
 use Rector\Core\RectorDefinition\RectorDefinition;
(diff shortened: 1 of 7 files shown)


mssimi:
Next steps will be remove **CarType**, add property `string $type `to **Car**, remove Car's method `getCarType():CarType` and add `getType: string` so you understand the big picture. But next steps are quite simple so I don't even need rector for that.

rules/generic/src/Rector/MethodCall/MethodCallRemoverRector.php
```diff
@@ -0,0 +1,85 @@
...
+            if (! $this->isObjectType($node->var, $className)) {
+                continue;
+            }
+
+            if (! $this->isName($node->name, $methodName)) {
+                continue;
+            }
+
+            // todo resolve variable name
+            return new Variable('this');
```
TomasVotruba, line 75: Maybe this would help: https://github.com/rectorphp/rector/blob/8c5fdca38737e45e55f7fca95d743fece567d29e/src/PhpParser/Node/Manipulator/MethodCallManipulator.php#L191

Make the method `public` if needed.

There 2 more services, that are handy for fluent method calls:

- https://github.com/rectorphp/rector/blob/master/rules/magic-disclosure/src/NodeManipulator/ChainMethodCallRootExtractor.php
- https://github.com/rectorphp/rector/blob/master/rules/magic-disclosure/src/NodeAnalyzer/ChainMethodCallNodeAnalyzer.php

mssimi: Great thx, it is what I was looking for!

mssimi:
Yes, thx :)

TomasVotruba:
Thank you, the code is joy to read :)

PR #3881, opened 2020-08-03
lazy command bug

mssimi:
LazyCommandRector does not work for some cases, wrote test that shows it.

### DIFF
--- a/rules/symfony/tests/Rector/Class_/MakeCommandLazyRector/Fixture/set_name_fluent.php.inc
+++ b/rules/symfony/tests/Rector/Class_/MakeCommandLazyRector/Fixture/set_name_fluent.php.inc
@@ -6,8 +6,9 @@ class SetNameFluent extends \Symfony\Component\Console\Command\Command
 {
     public function configure()
     {
-        $this->setName('some:command')
-            ->setDescription('Description');
+        $this
+            ->setDescription('Description')
+            ->setName('some:command');
     }
 }
 


TomasVotruba:
This actually removes one test case.
Could you send separate fixture so we have them covered both, in case of some weird regression?

mssimi:
Done.

rules/symfony/tests/Rector/Class_/MakeCommandLazyRector/Fixture/set_name_fluent.php.inc
```diff
@@ -6,7 +6,8 @@ class SetNameFluent extends \Symfony\Component\Console\Command\Command
 {
     public function configure()
     {
-        $this->setName('some:command')
+        $this
+            ->setName('some:command')
```
TomasVotruba, line 10:
```suggestion
        $this->setName('some:command')
```

TomasVotruba:
Thanks!

Any idea what went wrong?

What error do you see with `--debug`?

How is https://getrector.org/demo handling it?

mssimi:
Well it is implemented not to work for chain calls [here](https://github.com/rectorphp/rector/blob/master/packages/post-rector/src/Collector/NodesToRemoveCollector.php#L105).

TomasVotruba:
CI doesn't always show as good message as debug options.

In that case, the fix is very similar so your in-fluent method removal. Could you fix it?

mssimi:
Sadly I found a bug there as well, I need something that removes method in chain, not something that removes everything on left side of method chain.

TomasVotruba:
I see. There is trick for fluent method call removal.

Basically override the methodCall->var that has the method call that needs to be removed, by parent car/method call

mssimi:
No idea how to do that with collector, sorry. Could take hours for me to find a way.

TomasVotruba:
Sure :+1: 

If you're in a rush, I recommend using `defluent` set for these cases. Removes fluent + make this rule work

mssimi:
I am not in hurry, I was also thinking about de-fluent rector, but I have not found a way to exclude/include single rector just for one dir. I have only found a way to exclude dirs for all rectors and how to exclude file by annotation inside.

PR #3882, opened 2020-08-03
move services by suffix bug demonstration

mssimi:
`MoveServicesBySuffixToDirectoryRector` does not work, actually it seems any `AbstractFileMovingFileSystemRector` does not. Tests shows false negative, I cannot reproduce it by test, but I can reproduce it by this "dirty" example. I have put two dumps to demonstate that when rector starts, configuration is set, but is not afterwards. Configuration variable is lost somewhere during rector run.. Here is screen:

rector-move-services-bug

### DIFF
--- a/rules/autodiscovery/src/Rector/FileSystem/MoveServicesBySuffixToDirectoryRector.php
+++ b/rules/autodiscovery/src/Rector/FileSystem/MoveServicesBySuffixToDirectoryRector.php
@@ -77,6 +77,7 @@ PHP
 
     public function configure(array $configuration): void
     {
+		dump($configuration);
         $this->groupNamesBySuffix = $configuration[self::GROUP_NAMES_BY_SUFFIX] ?? [];
         dump('configure', $this->groupNamesBySuffix);
     }
--- a/rules/autodiscovery/src/Rector/FileSystem/MoveServicesBySuffixToDirectoryRector.php
+++ b/rules/autodiscovery/src/Rector/FileSystem/MoveServicesBySuffixToDirectoryRector.php
@@ -64,6 +64,7 @@ PHP
 
     public function refactor(SmartFileInfo $smartFileInfo): void
     {
+        dump('refactor', $this->groupNamesBySuffix);die;
         $nodes = $this->parseFileInfoToNodes($smartFileInfo);
 
         $class = $this->betterNodeFinder->findFirstInstanceOf($nodes, Class_::class);
@@ -77,6 +78,7 @@ PHP
     public function configure(array $configuration): void
     {
         $this->groupNamesBySuffix = $configuration[self::GROUP_NAMES_BY_SUFFIX] ?? [];
+        dump('configure', $this->groupNamesBySuffix);
     }
 
     /**
--- a/rector-ci.php
+++ b/rector-ci.php
@@ -3,6 +3,7 @@
 declare(strict_types=1);
 
 use PHPUnit\Framework\TestCase;
+use Rector\Autodiscovery\Rector\FileSystem\MoveServicesBySuffixToDirectoryRector;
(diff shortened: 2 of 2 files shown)


mssimi:
Tested on separate project as well, it does not work.

TomasVotruba:
Seems like it's unrelated to the rule, but rather to `configure` call in general. Configuration is not passes to the method. I'm I right?

TomasVotruba:
It might be config mergin issue. The rule is registered somewhere else, and empty `configure` is "called" first.

mssimi:
I am not in hurry, I was also thinking about de-fluent rector, but I have not found a way to exclude/include single rector just for one dir. I have only found a way to exclude dirs for all rectors and how to exclude file by annotation inside.

mssimi:
You don't understand the example. When rector is started, dump inside configure method return config data, that mean, config was loaded at start of rector process, after that rector runs and call configure method again and after that refactor method, this time both dump show empty array.

rules/autodiscovery/src/Rector/FileSystem/MoveServicesBySuffixToDirectoryRector.php
```diff
@@ -77,6 +78,7 @@ public function refactor(SmartFileInfo $smartFileInfo): void
     public function configure(array $configuration): void
     {
         $this->groupNamesBySuffix = $configuration[self::GROUP_NAMES_BY_SUFFIX] ?? [];
```
TomasVotruba, line 80:
```suggestion
		dump($configuration);
        $this->groupNamesBySuffix = $configuration[self::GROUP_NAMES_BY_SUFFIX] ?? [];
```

mssimi:
I don't know rector process, but configuration method is called twice during process once it has data once it does not.

TomasVotruba:
Great!

If you click on `configure()` method of the `ConfigurableRectorInterface` interface in PHPStorm, it shows 2 places.

I suspect one of these 2 calls has a bug:

- https://github.com/rectorphp/rector/blob/2bd1eac7053aeeb6502e38a3d686620c78ad98f3/packages/file-system-rector/src/FileSystemFileProcessor.php#L60-L63

- https://github.com/rectorphp/rector/blob/2bd1eac7053aeeb6502e38a3d686620c78ad98f3/src/PhpParser/NodeTraverser/RectorNodeTraverser.php#L103

mssimi:
> Getting closer...
> 
> If you click on `configure()` method of the `ConfigurableRectorInterface` interface in PHPStorm, it shows 2 places.
> 
> I suspect one of these 2 calls has a bug:
> 
> * https://github.com/rectorphp/rector/blob/2bd1eac7053aeeb6502e38a3d686620c78ad98f3/packages/file-system-rector/src/FileSystemFileProcessor.php#L60-L63
> * https://github.com/rectorphp/rector/blob/2bd1eac7053aeeb6502e38a3d686620c78ad98f3/src/PhpParser/NodeTraverser/RectorNodeTraverser.php#L103

mssimi:
Miss click closed, sorry :)

I guess it will be  `rector/packages/file-system-rector/src/FileSystemFileProcessor.php` as this occurs for `AbstractFileMovingFileSystemRector`

mssimi:
> Getting closer...
> 
> If you click on `configure()` method of the `ConfigurableRectorInterface` interface in PHPStorm, it shows 2 places.
> 
> I suspect one of these 2 calls has a bug:
> 
> * https://github.com/rectorphp/rector/blob/2bd1eac7053aeeb6502e38a3d686620c78ad98f3/packages/file-system-rector/src/FileSystemFileProcessor.php#L60-L63
> * https://github.com/rectorphp/rector/blob/2bd1eac7053aeeb6502e38a3d686620c78ad98f3/src/PhpParser/NodeTraverser/RectorNodeTraverser.php#L103

`enabledRectorsProvider` is always empty, it's method `addEnabledRector` is never called

Issue #3742, opened 2020-07-20
assertEquals() to assertSame()


dereuromark:
# Feature Request

I didn't see such a rector task in 
https://github.com/rectorphp/rector/tree/master/rules/phpunit/src/Rector/SpecificMethod
yet

Given the popularity of https://github.com/sebastianbergmann/phpunit-documentation-english/issues/3 (and linked refs) it seems it would be very useful to have a task that fixes asserts here to the stricter and typesafe comparison where possible.

I would recommend:
- Scalar values only for now (int, string, ...)

e.g.
```diff
- $this->assertEquals(1, $count);
+ $this->assertSame(1, $count);
```
to make sure it isn't float 1.0 or trueish.

I am not sure about float. AssertSameBoolNullToSpecificMethodRector takes care of bool/null it seems.

PS: Is rector able to also introspect variables, so knowing if $var is scalar and would apply for the above change as well, not just literal values?

PR #3884, opened 2020-08-03
[Order] Add test for StmtOrder

dobryy:


rules/order/tests/StmtOrderTest.php
```diff
@@ -66,6 +66,6 @@ public function testReorderClassStmtsByOldToNewKeys(): void
         $expectedClass->stmts[] = new Property(Class_::MODIFIER_PRIVATE, [new PropertyProperty('name')]);
         $expectedClass->stmts[] = new Property(Class_::MODIFIER_PRIVATE, [new PropertyProperty('price')]);
         $expectedClass->stmts[] = new Property(Class_::MODIFIER_PRIVATE, [new PropertyProperty('service')]);
-        $this->assertEquals($expectedClass->stmts, $actualClass->stmts);
+        $this->assertSame($expectedClass->stmts, $actualClass->stmts);
```
dobryy, line 69: @TomasVotruba what if I'd like to use assertEquals here? :)

dobryy: I actually wonder that we have such an issue https://github.com/rectorphp/rector/issues/3742 as it looks like there is a rule which replaces assertEquals to assertSame

TomasVotruba: > it looks like there is a rule

Where is it?

rules/order/tests/StmtOrderTest.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        $class = new Class_('ClassUnderTest');
+        $class->stmts[] = new Property(Class_::MODIFIER_PRIVATE, [new PropertyProperty('name')]);
+        $class->stmts[] = new Property(Class_::MODIFIER_PRIVATE, [new PropertyProperty('service')]);
+        $class->stmts[] = new Property(Class_::MODIFIER_PRIVATE, [new PropertyProperty('price')]);
+        $actualClass = $this->stmtOrder->reorderClassStmtsByOldToNewKeys($class, self::OLD_TO_NEW_KEYS);
+        $expectedClass = new Class_('ExpectedClass');
+        $expectedClass->stmts[] = new Property(Class_::MODIFIER_PRIVATE, [new PropertyProperty('name')]);
+        $expectedClass->stmts[] = new Property(Class_::MODIFIER_PRIVATE, [new PropertyProperty('price')]);
+        $expectedClass->stmts[] = new Property(Class_::MODIFIER_PRIVATE, [new PropertyProperty('service')]);
```
TomasVotruba, line 68: Better extract these to 2 methods, it's a lot of stuff going on here

rules/order/tests/StmtOrderTest.php
TomasVotruba: I never know which is which :D we use the strict one. It should work here too

rules/order/tests/StmtOrderTest.php
```diff
@@ -0,0 +1,71 @@
...
+    public function testCreateOldToNewKeys(): void
+    {
+        $actual = $this->stmtOrder->createOldToNewKeys(self::DESIRED_STMT_ORDER, self::CURRENT_STMT_ORDER);
+        $expected = [
+            0 => 1,
+            1 => 2,
+            2 => 0,
+        ];
+        $this->assertSame($expected, $actual);
+    }
```
TomasVotruba, line 56: Could you add more test fixture than just 1? 

It will help us in the future, when it breaks.
Like keeping the order untouched and inversed order

TomasVotruba: Btw, what do you think about data providers?

rules/order/tests/StmtOrderTest.php
dobryy: > > it looks like there is a rule
> 
> Where is it?

I've just checked and didn't find any... I wanted to use `assertEquals` because I'd like to compare two different instances of the object but CI changed it to `assertSame`

dobryy: > I never know which is which :D we use the strict one. It should work here too

I see but in this particular case `assertEquals` suites better. In case we are so strict I'll have to create more asserts and to compare method names. Also, it seems that it is not that easy to resolve the method names in this kind of tests.

TomasVotruba: I see, the new nodes are created manually and thus are different.

dobryy: Check it out https://github.com/rectorphp/rector/pull/3884/commits/12df4706265f126a21552cb2d056a27adeb6f087

rules/order/tests/StmtOrderTest.php
```diff
@@ -80,7 +87,19 @@ public function testReorderClassStmtsByOldToNewKeys(): void
         $class = $this->getTestClassNode();
         $actualClass = $this->stmtOrder->reorderClassStmtsByOldToNewKeys($class, self::OLD_TO_NEW_KEYS);
         $expectedClass = $this->getExpectedClassNode();
-        $this->assertSame($expectedClass->stmts, $actualClass->stmts);
```
TomasVotruba: I think this is good enoug. Checking node subproperties will only make test complicated

dobryy: Agree

TomasVotruba: I mean to keep this:

```php
$this->assertSame($expectedClass->stmts, $actualClass->stmts);
```

And remove the name node comparing

dobryy: Sorry, misunderstood what you meant. It doesn't work because objects inside the `stmts` are not the same, they are just equal :D
Even though they are the instances of the same class they are different objects and reference different locations in memory. (I don't understand it into detail so maybe, I'm writting bullshit :D)
```php
$this->assertSame(new \stdClass(), new \stdClass());
```
For example, the assert above will fail.

dobryy: If we'll decide to leave it as it is now we can get rid of `$expectedClass` and use plain string names in assertion, eg:
```php
        $this->assertSame(
            'price',
            $this->nodeNameResolver->getName($actualClass->stmts[1])
        );
```

TomasVotruba: I'm sorry, I didn't get how this test works and I'm used to strict testing, because if it was not strict if failed many months later and shoot me in the head :D

Let's work with current version

TomasVotruba:
Thank you :+1:

PR #3889, opened 2020-08-03
[Order] Make OrderPrivateMethodsByUseRector process file in one run u…

dobryy:
…sing while loop

### DIFF
--- a/rules/order/src/Rector/Class_/OrderPrivateMethodsByUseRector.php
+++ b/rules/order/src/Rector/Class_/OrderPrivateMethodsByUseRector.php
@@ -121,6 +121,7 @@ PHP
 
         return $node;
     }
+
     private function getPrivateMethodCallOrderAndClassMethods(Class_ $class): array
     {
         return [$this->getLocalPrivateMethodCallOrder($class), $this->resolvePrivateClassMethods($class)];
--- a/rules/order/src/Rector/Class_/OrderPrivateMethodsByUseRector.php
+++ b/rules/order/src/Rector/Class_/OrderPrivateMethodsByUseRector.php
@@ -121,6 +121,10 @@ PHP
 
         return $node;
     }
+    private function getPrivateMethodCallOrderAndClassMethods(Class_ $class): array
+    {
+        return [$this->getLocalPrivateMethodCallOrder($class), $this->resolvePrivateClassMethods($class)];
+    }
 
     /**
      * @return string[]
@@ -180,10 +184,5 @@ PHP
         }
 
         return $privateClassMethods;
-    }
-
-    private function getPrivateMethodCallOrderAndClassMethods(Class_ $node): array
-    {
-        return [$this->getLocalPrivateMethodCallOrder($node), $this->resolvePrivateClassMethods($node)];
     }
 }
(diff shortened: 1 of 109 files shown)


rules/order/src/Rector/Class_/OrderPrivateMethodsByUseRector.php
```diff
@@ -86,32 +86,53 @@ public function getNodeTypes(): array
...
-        if (count($desiredClassMethodOrder) !== count($desiredClassMethodOrder)) {
+        if (count($desiredPrivateMethodCallOrder) !== count($privateClassMethodsByKey)) {
             return null;
         }
 
-        $oldToNewKeys = $this->stmtOrder->createOldToNewKeys($desiredClassMethodOrder, $privateClassMethodsByKey);
+        $attempt = 0;
+        while (array_values($desiredPrivateMethodCallOrder) !== array_values($privateClassMethodsByKey)) {
+            $attempt++;
+            if ($attempt >= 5) {
```
dobryy, line 106: @TomasVotruba do you think we need a setting here? Or should it be enough to use constant?

rules/order/src/Rector/Class_/OrderPrivateMethodsByUseRector.php
```diff
@@ -86,32 +86,53 @@ public function getNodeTypes(): array
...
+                break;
```
dobryy, line 107: @TomasVotruba We discussed that it might be a good idea to throw an Exception. It there any existing suitable Exception that I can throw here?

rules/order/src/Rector/Class_/OrderPrivateMethodsByUseRector.php
TomasVotruba: > Or should it be enough to use constant?

:+1:  Just so it's obvious there is something configurable in the class

rules/order/src/Rector/Class_/OrderPrivateMethodsByUseRector.php
TomasVotruba: We use generic `ShouldNotHappenException` here.

dobryy:
@TomasVotruba rebased on master. If it looks fine it can be merged.

TomasVotruba:
After this merge, the CI started to be super slow, that Travis stopped:

https://github.com/rectorphp/rector/pull/3916/checks?check_run_id=951533356

image


I've disabled till it's fixed, so we have passing CI. Could you check what takes so long?

It might be good to compare all rules 1 by 1, their time in total. That could narrow which rule is so slow.

TomasVotruba:
Hm, I tested locally and these rule add big performance hit. But the problem is still there :+1: https://github.com/rectorphp/rector/pull/3916

dobryy:
Hm... I'll check.

TomasVotruba:
It might be something else... like dependency or architecture change in last 3 PRs. I haven't got very far yesterday

dobryy:
I ran it locally on master branch this morning. Passed without any timeouts...

TomasVotruba:
What the time run with/without order set?

dobryy:
The execution time with and without the Order set, I'd say, is kinda the same, 2:12 - 2:30 minutes.
At the moment, there are plenty of warnings in console because of ConsistentPregDelimiterRector.php, I guess, it could have some impact on execution time

TomasVotruba:
I see. So the problem will be probably in the PR. Could you review it?

https://github.com/rectorphp/rector/pull/3916


> are plenty of warnings in console because of ConsistentPregDelimiterRector.php

I'll check it

dobryy:
> I see. So the problem will be probably in the PR. Could you review the PR?
> 
> #3916

Hm... works too slow on this branch. I'll take a look on it.

PR #3896, opened 2020-08-04
FunctionTo* correction renames, add PHPStan rule to check correct category of rule

### LINKED ISSUE #3864: PHPStan rule to check that Namespace of Rector rules respect the Node type
Also, this should be covered by local PHPStan rule first. That way it's handled by CI code reviews and weak sposts are easily discovered

_Originally posted by @TomasVotruba in https://github.com/rectorphp/rector/issues/3696#issuecomment-661415300_

TomasVotruba:
Closes #3864


utils/phpstan-extensions/config/phpstan-extensions.neon
```diff
@@ -1,4 +1,9 @@
 services:
+    # prepared for Igor
+#    -
+#        class: Rector\PHPStanExtensions\Rule\RequireRectorCategoryByGetNodeTypesRule
+#        tags: [phpstan.rules.rule]
+
```
TomasVotruba, line 6: There you go @dobryy

PR #3903, opened 2020-08-04
RenamePropertyToMatchTypeRector failing test

### LINKED ISSUE #3897: RenamePropertyToMatchTypeRector issue
# Bug Report RenamePropertyToMatchTypeRector

Rector version: v0.7.61
Installed as composer dependency

**Vendor interface**

```php
namespace Some\Vendor\Namespace;

interface SomethingInterface 
{
    public function doSomething(Request $req): void;
}
```

**Before RenamePropertyToMatchTypeRector**

```php
namespace App\Service;

use Some\Vendor\Namespace\SomethingInterface;

final class Something implements SomethingInterface 
{
    public function doSomething(Request $req): void {
        // ...
    }
}
```

**After RenamePropertyToMatchTypeRector**

```php
namespace App\Service;

use Some\Vendor\Namespace\SomethingInterface;

final class Something implements SomethingInterface 
{
    public function doSomething(Request $request): void {
        // ...
    }
}
```

**Expected Behaviour**

Keep it as is. RenamePropertyToMatchTypeRector must not rename properties of methods that are implementation of vendor interface, as it will break code.

mssimi:
Closes #3897

### DIFF
--- a/rules/naming/tests/Rector/Class_/RenamePropertyToMatchTypeRector/Fixture/keep_implements_vendor.php.inc
+++ b/rules/naming/tests/Rector/Class_/RenamePropertyToMatchTypeRector/Fixture/keep_implements_vendor.php.inc
@@ -1,0 +1,14 @@
+<?php
+
+namespace Rector\Naming\Tests\Rector\Class_\RenamePropertyToMatchTypeRector\Fixture;
+
+use Rector\Naming\Tests\Rector\Class_\RenamePropertyToMatchTypeRector\Source\SingleSomeClass;
+use Rector\Naming\Tests\Rector\Class_\RenamePropertyToMatchTypeRector\Source\SomeVendorInterface;
+
+class KeepImplementsVendor implements SomeVendorInterface
+{
+    public function something(SingleSomeClass $class): void
+    {
+        get_class($class);
+    }
+}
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thank you for the test. The output is valid working code.

Could you provide the 3v4l.com code sample that demonstrates the fatal error?

TomasVotruba:
Maybe the test fixture would benefit from RunnableInterface

mssimi:
> Thank you for the test. The output is valid working code at first sight.
> 
> Could you provide the 3v4l.org code sample that demonstrates the fatal error?

How can I simulate vendor package on 3v4l.org?

I really don't understand what else you want from me, **SomeVendorInterface** is in vendor and when I implement interface from vendor in my own code base, it's methods params cannot be changed.

@dobryy Understands it, maybe he can explain better I am giving it up.

TomasVotruba:
> How can I simulate vendor package on 3v4l.org?

No need to simulate that. Here is your code: https://3v4l.org/3kcB2
It shows the correct expected output.

<br>

I basically need to see error you have after Rector changes the code.

TomasVotruba:
Works with interface too: https://3v4l.org/u4gqW

mssimi:
Right, here is code https://3v4l.org/LAmoY

I am really sorry, this is embarrassing, but I didn't know variable names can be changed in class where interface method is implemented. It certainly works and I was the one the is wrong and I am very sorry about it.

mssimi:
My apologize, I was wrong on this one.

TomasVotruba:
That's why I was so confused :D 
No troubles! 

Yet you discovered another bug, that Rename**Property** should not touch **Param**.

I'll look into it, thanks

PR #3904, opened 2020-08-04
Check that Recipe node types are imported

dobryy:


packages/rector-generator/src/ValueObject/Configuration.php
```diff
@@ -208,4 +208,16 @@ private function setName(string $name): void
 
         $this->name = $name;
     }
+
+    private function setNodeTypes(array $nodeTypes): void
+    {
+        foreach ($nodeTypes as $nodeType) {
+            if (! class_exists($nodeType)) {
+                $message = sprintf('Node type "%s" doesn\'t exist or not imported in Rector recipe', $nodeType);
```
TomasVotruba, line 216:
```suggestion
                $message = sprintf('Node type "%s" does not exist or not imported in Rector recipe', $nodeType);
```

Escaping quotes in strings is very hard to read, error prone and not neccessary

dobryy:
> Just curious, are there some Node interfaces, that can be used?

They all implement PhpParser\Node interface. Not sure how I can use this in this particular case because I don't have an instance of the class, I have only a class name thus can't check it using `$nodeType instanceOf NodeInterface`.

TomasVotruba:
> I don't have an instance of the class, I have only a class name thus can't check it using $nodeType instanceOf NodeInterface.

What about `is_a(..., ..., true)`?

TomasVotruba:
Looks good to me.
Can I merge it?

dobryy:
> Looks good to me.
> Can I merge it?

Yes, you can.

Issue #3696, opened 2020-07-13
[DX] Rename namespaces of Rector rules to respect node, instead of domain


TomasVotruba:
In the past, it was unclear how to name Rector rule catgories.

## Current Problem

See *HttpKernel*:

```yaml
services:
    Rector\Symfony\Rector\HttpKernel\GetRequestRector: null
```

<br>

Now the standard is more clear, use node type - it makes debugging and grouping rules much easier. Also open-space for re-use of similar logic :

See *MethodCall*:

```yaml
services:
     Rector\Symfony\Rector\MethodCall\ReadOnlyOptionToAttributeRector: null
```

## The Goal

Having 2 standards is confusing, so we need to rename former to latter

```diff
 services:
-    Rector\Symfony\Rector\HttpKernel\GetRequestRector: null
+    Rector\Symfony\Rector\MethodCall\GetRequestRector: null
```

That includes:

- rename rule namespace
- rename tests + fixtures namespace
- update sets with rule

TomasVotruba:
Also, this should be covered by local PHPStan rule first. That way it's handled by CI code reviews and weak sposts are easily discovered

dobryy:
There is a rule that processes two node types, eg. https://github.com/rectorphp/rector/blob/128bcab866a7b085a912813a69c544aa5758f8f2/rules/symfony/src/Rector/HttpKernel/GetRequestRector.php#L82

How this should be handled?

TomasVotruba:
What do you recommend?

dobryy:
I had only one idea... to split it. Nothing else is popping out in my brain :)

TomasVotruba:
I see. I'm afraid that might lead to overly fractured architecture. One rule in X classes.
Overcomplicated roles on the other hand, are checked by complexity rules in phpstan.


Current convention is use one of them, e.g. ClassMethod

dobryy:
What about using Stmt namespace for such cases? 
Or... combined namespaces ClassMethodMethoCall? The last sounds crazy :D

TomasVotruba:
> What about using Stmt namespace for such cases?

That seems too generic and makes discovery of rule very hard.

> ClassMethodMethoCall

What about `ClassMethodMethoCallStaticCallFuncCallArg` ? :laughing:

dobryy:
[...]

PR #3911, opened 2020-08-05
[DX] Rename namespaces of Rector rules to respect node, instead of domain

dobryy:
https://github.com/rectorphp/rector/issues/3696


TomasVotruba:
:+1:

dobryy:
@TomasVotruba Symfony related (maybe something else) rules are ready. Let's merge it and I'll continue with the next PR.

PS 64 rules left.

TomasVotruba:
I see :+1:

PR #3916, opened 2020-08-05
[SOLID] skip reference-write functions

TomasVotruba:


rules/solid/src/Guard/VariableToConstantGuard.php
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        if (function_exists($name)) {
+            $reflectionFunction = new ReflectionFunction($name);
+            foreach ($reflectionFunction->getParameters() as $parameter) {
+                if ($parameter->isPassedByReference()) {
+                    return false;
+                }
+            }
+        }
```
TomasVotruba, line 45: @Aerendir This might solve the reference functions ;)

PR #3918, opened 2020-08-06
Fix set path was not found

### LINKED ISSUE #3915: New SetList paths completely broke if using bamarni/composer-bin-plugin
# Bug Report

| Rector version | v0.7.62              |

I'm using `bamarni/composer-bin-plugin`.

it installs Rector in `vendor-bin/phpstan/vendor/rector`.

The file `Rector\Set\ValueObject\SetList` uses fixed paths:

```php
...
/**
     * @var string
     */
    public const ACTION_INJECTION_TO_CONSTRUCTOR_INJECTION = __DIR__ . '/../../config/set/action-injection-to-constructor-injection.php';
...
```

This, obviously, breaks the paths as they are not found and Rector cannot run.

# Suggested solution

I really have no idea 😅

# How to reproduce the issue

```console
composer rem rector/rector && composer req bamarni/composer-bin-plugin && composer bin rector req rector/rector && vendor/bin/rector process --dry-run
```

(this command should work: I'm writing it on the fly, without testing it)

zingimmick:
Only these four sets have different paths, ignore me if there's any reason for that.

Close #3915

### DIFF
--- a/packages/set/src/ValueObject/SetList.php
+++ b/packages/set/src/ValueObject/SetList.php
@@ -14,7 +14,7 @@ final class SetList
     /**
      * @var string
      */
-    public const ACTION_INJECTION_TO_CONSTRUCTOR_INJECTION = __DIR__ . '/../../config/set/action-injection-to-constructor-injection.php';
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Thank you :+1:

Issue #3891, opened 2020-08-03
[Order] Order Properties by Visibility 


dobryy:
# Feature Request

## Diff 

```diff
-private $privateProperty;
-protected $protectedProperty;
-public $publicProperty;
+public $publicProperty;
+protected $protectedProperty;
+private $privateProperty;
```

PR #3929, opened 2020-08-07
[Order] Order properties by visibility

dobryy:
Closes https://github.com/rectorphp/rector/issues/3891


dobryy:
@TomasVotruba if it looks fine, let's merge it and as a next PR, I'll do some refactoring of Order set.

PR #3933, opened 2020-08-08
[Order] Visibility rules refactoring

dobryy:


rules/order/src/Rector/ClassLike/OrderMethodsByVisibilityRector.php
```diff
@@ -100,63 +98,28 @@ public function getNodeTypes(): array
...
-            $classMethods[$classMethodName]['abstract'] = $classStmt->isAbstract();
-            $classMethods[$classMethodName]['final'] = $classStmt->isFinal();
-            $classMethods[$classMethodName]['static'] = $classStmt->isStatic();
-            $classMethods[$classMethodName][self::POSITION] = $position;
-        }
-
-        $sortedMethods = $this->getMethodsSortedByVisibility($classMethods);
-        $methodsInPreferredOrder = $this->getMethodsInPreferredOrder($sortedMethods);
+        $currentMethodsOrder = $this->stmtOrder->getStmtsOfTypeOrder($node, ClassMethod::class);
+        $methodsInDesiredOrder = $this->getMethodsInDesiredOrder($node);
```
TomasVotruba, line 106: :+1:

rules/order/src/Rector/ClassLike/OrderMethodsByVisibilityRector.php
```diff
@@ -100,63 +98,28 @@ public function getNodeTypes(): array
...
-
-        return array_keys($classMethods);
+        $classMethods = $this->stmtVisibilitySorter->sortMethods($classLike);
+        $classMethods = array_keys($classMethods);
+
+        return $this->applyPreferredPosition($classMethods);
     }
 
-    private function getMethodsInPreferredOrder(array $sortedMethods): array
+    private function applyPreferredPosition(array $classMethods): array
```
TomasVotruba, line 121: `array` would deserve types

I'm going to re-enable this in PHPStan, so it's covered by CI

TomasVotruba:
Just the `array` types. The rest looks good to me

rules/order/src/Rector/ClassLike/OrderMethodsByVisibilityRector.php
TomasVotruba: https://github.com/rectorphp/rector/pull/3936

dobryy:
> Just the `array` types. The rest looks good to me

Ready.
array<int,string> - when index is not continuous value from 0 to N.
string[] - when the array is an indexed array with the continuous index value. 

Let me know if it makes sense :)

TomasVotruba:
All good, thanks

PR #3940, opened 2020-08-11
[Doctrine] service entity repo combo

mssimi:


TomasVotruba:
Follow up in https://github.com/rectorphp/rector/pull/3945

PR #3957, opened 2020-08-14
Fix sample configuration generation

hxv:


hxv:
Sample in rule is fine - turns out problem was in printing samples for documentation, and it affects more rules. Fixed that and generated fresh docs, hope it's fine.

TomasVotruba:
Thank you so much :+1:

PR #3959, opened 2020-08-14
Clarify paths.

dereuromark:
What do you think about this?

It is recommended to clarify paths with the slash, as in most OS a file can be anything without an ext as well.
Even if it works without, I think the docs are clearer.

The PSR-4 map is also recommended to have the clarity of trailing slash in place.


TomasVotruba:
I assume this is known to particular OS users. Like upper/lower case for Windows and Linux, `PHP_EOL` differences or `\`, `/` separators for file paths. Adding all these to all documentations would only create clutter and make Rector Documentation complicated to read.

dereuromark:
The PSR4 part is still true, though.

TomasVotruba:
If it causes bugs on specific OS, it should be validated/required by composer then

dereuromark:
The normalized version is also documented and advised to be used.
Both are valid, but what I PRd is the recommended form of writing it.

TomasVotruba:
On what OS does this cause bug?

How does the bug exactly look like while using Rector?

dereuromark:
Its not a bug, it is the normalized and documented form what is supposed to be put into the composer path section.

PR #3968, opened 2020-08-16
[Order] Order first level class statements

dobryy:


TomasVotruba:
Looks good to me, thanks :+1:

PR #3969, opened 2020-08-16
[Order] Move ClassLike and Class_ rules into correct namespaces

dobryy:
Sorted the ClassLike and Class_ rules into correct namespaces.

Constants could exist only in regular and abstract classes, that is why these rules are in Class_ namesapce. The same for interface implementation.

The rules in ClassLike could be applied for regular, abstract class and trait that is why they are in ClassLike namespace.


TomasVotruba:
Looks fine, thanks :+1:

PR #3983, opened 2020-08-18
[DX] Architecture rules respect namespace

dobryy:


dobryy:
@TomasVotruba Ready, but there is some issue with some composer library... that is why the pipeline fails...

TomasVotruba:
It's because https://github.com/nikic/PHP-Parser/commit/aaee038b912e567780949787d5fe1977be11a778

Lockdown in `composer.json` to lower php-parser should help

dobryy:
@TomasVotruba, ready for merging.

TomasVotruba:
Thank you!

So clean :)


Rector code starts to smell like just washed laundry :lab_coat:  in the sunrise :white_flower:

Issue #3981, opened 2020-08-18
Calling a static method on a class-string variable


olivernybroe:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | v0.7.53         |
| Installed as   | composer dependency                   |

Rector crashes when calling a static method on a class-string variable

## Minimal PHP Code Causing Issue
https://getrector.org/demo/88314d95-fdc2-4dff-a286-9fa577a029ca

And change the setList to `SetList::PHP_70`

## Expected Behaviour
It should give some output

TomasVotruba:
Thanks for reporting :+1: 

Could you add failing test case to responsible rule?

PR #3986, opened 2020-08-19
Add failing test case for #3981

olivernybroe:
Here is the pr with a failing testcase for #3981 as requested 👍 

Hope I did it correctly.


TomasVotruba:
Thanks :+1:

PR #4032, opened 2020-08-26
[Type Declaration] return array type doc

dobryy:


TomasVotruba:
The test fixtures look all ok :+1:

TomasVotruba:
Could you:

- fix coding standard
- disable the rule in `rector-ci.php` for now

so we can merge it? Then we'll have less troubles to focus on

TomasVotruba:
Ah, there are broken tests too... that needs to be fixed first.

dobryy:
> Ah, there are broken tests too... that needs to be fixed first.

And that is the biggest problem :D I made Rector related issues fixed but some original fixtures don't work. I'll continue with it later today.

packages/phpstan-static-type-mapper/src/TypeMapper/ArrayTypeMapper.php
```diff
@@ -122,24 +121,7 @@ private function isGenericArrayCandidate(ArrayType $arrayType): bool
...
-
-        foreach ($keysArrayType->getValueTypes() as $key => $keyType) {
-            if (! $keyType instanceof ConstantIntegerType) {
-                return true;
-            }
-
-            if ($key !== $keyType->getValue()) {
-                return true;
-            }
-        }
```
TomasVotruba, line 140: Removing this now causes to add `int` key type, when not explicitly used

```php
/**
 * @var array<int, string>  
 */
private const VALUE = [
   'something'
];
```

:x: 

<br>

It should be:

```php
/**
 * @var string[]
 */
private const VALUE = [
   'something'
];
```

:heavy_check_mark:

TomasVotruba:
> And that is the biggest problem :D I made Rector related issues fixed but some original fixtures don't work. I'll continue with it later today.

Maybe next time would be better to add many small PR that passes, so we have finished and merged work. Making many changes with broken CI is huge productivity killer.

PR #4038, opened 2020-08-27
Performance improvement: use [] comparison check instead of count($var) for check array has data when possible

samsonasik:
Signed-off-by: Abdul Malik Ikhsan <samsonasik@gmail.com>

### DIFF
--- a/src/Application/RectorApplication.php
+++ b/src/Application/RectorApplication.php
@@ -115,11 +115,11 @@ final class RectorApplication
      */
     public function runOnFileInfos(array $phpFileInfos): void
     {
-        $fileCount = count($phpFileInfos);
-        if ($fileCount === 0) {
+        if (! $phpFileInfos) {
             return;
         }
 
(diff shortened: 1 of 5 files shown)


TomasVotruba:
CI needs passing.

What is the performance relative improvement on how big project?

samsonasik:
I updated with `[]` comparison check. Well, mostly micro optimization, other thing, it make consistent with others, searched with `[]` comparison in the `src` directory.

TomasVotruba:
We need data to do performance optimizations. Otherwise it's just guessing and random changes in the code, that might only add extra manual work for contributors, without any proven added value. 

To make it work, there would have to be person that does PRs like these after every new line of code is added.

**If you can provide data and make a Rector rule that do these change automatically**, I'd be happy to merge it.

TomasVotruba:
Closing for now, since the Rector rule is needed. 
Feel free to re-open new PR with one. 

Thanks for understanding

PR #4043, opened 2020-08-27
[Type Declaration] Fix unwrapping of multiple union types

dobryy:


dobryy:
@TomasVotruba, I created one new test fixture with a fix but then I realized that it is not that related to Array types but rather Union types. So it is possible that it should be fixed in another way, I have to implement logic to skip such cases by the rule.
Without this fix the result for this fixture will be:
`* @return \PHPStan\Reflection\MethodReflection|null`

TomasVotruba (approved):

TomasVotruba:
This looks correct, thanks :+1:

PR #4044, opened 2020-08-27
[Type Declaration] The smallest possible fixture fixed ReturnUuid

dobryy:
@TomasVotruba, as agreed, creating pull-request with the smallest possible fixture being fixed, to let you take a look :)


TomasVotruba:
Thanks for simple test :+1: , it allowed me to focus narrowly on the solution

PR #4055, opened 2020-08-28
Fix typo in method name that breaks all tests

dobryy:
Sneaky @TomasVotruba! :D


dobryy:
Hmmmm, looks like there are more issues than one typo...

TomasVotruba:
Thanks!


There was typo in the original method actually, so I released fix in Symplify. Just didn't get to coding since to fix it here too.

PR #4054, opened 2020-08-28
[Type Declaration] Create new unit test for ArrayTypeMapper with fixes

dobryy:

PR #4058, opened 2020-08-29
Fix array type

TomasVotruba:


TomasVotruba:
@dobryy Took 2 hours of hard-code hacking but CI is back now :) thanks for the test cases, they were very helpful!

packages/attribute-aware-php-doc/src/Ast/Type/AttributeAwareArrayTypeNode.php
```diff
@@ -6,26 +6,72 @@
...
+
+        if ($attributeAwareUnionTypeNode->isWrappedWithBrackets()) {
+            return $attributeAwareUnionTypeNode . '[]';
+        }
+
+        foreach ($attributeAwareUnionTypeNode->types as $unionedType) {
+            $unionedTypes[] = $unionedType . '[]';
+        }
+
+        return implode('|', $unionedTypes);
```
TomasVotruba, line 75: This was the reason for normalized, to avoid these custom printers for union/array mixes :)

packages/phpstan-static-type-mapper/tests/TypeMapper/ArrayTypeMapperTest.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Rector\PHPStanStaticTypeMapper\TypeMapper;
+namespace Rector\PHPStanStaticTypeMapper\Tests\TypeMapper;
```
TomasVotruba, line 5: @dobrry The namespace was incorrect here in your original PR and it caused few troubles. I guess PHPStan should tell you.

Each test has to be in `/Tests/` namespace. Maybe we should cover it with PHPStan rule?

dobryy: Weird, but I do remember that I added the Test in the namespace. I checked how it looks like in other packages to make it the same. Maybe I screw it somehow and some point. Before pushing the changes I ran the PHPStan and I had only one issue that I see only on my local environment.

It is clear that the tests should have their own namespace but PHPStan could be helpful to avoid such problems.

packages/attribute-aware-php-doc/src/Ast/Type/AttributeAwareArrayTypeNode.php
dobryy: But if I understand it correctly... now we have to handle them in order to generate correct docblock. 

All this type related part of the Rector looks way too complicated :) a bit less complicated after all the investigations I made but anyway...

TomasVotruba: Yes, by removing normalization on one place, now this part becomes more complicated :)

packages/phpstan-static-type-mapper/tests/TypeMapper/ArrayTypeMapperTest.php
TomasVotruba: That happens
https://github.com/rectorphp/rector/pull/4054/files#diff-12672d608506733ec39087a8afdea004

PR #4069, opened 2020-08-30
[Order] Add abstract class, check for changed order and move category to Class_, to prevent minus types

TomasVotruba:


rules/order/src/Rector/Class_/OrderMethodsByVisibilityRector.php
```diff
@@ -90,23 +76,24 @@ private function privateFunctionName();
...
-
         $currentMethodsOrder = $this->stmtOrder->getStmtsOfTypeOrder($node, ClassMethod::class);
         $methodsInDesiredOrder = $this->getMethodsInDesiredOrder($node);
 
         $oldToNewKeys = $this->stmtOrder->createOldToNewKeys($methodsInDesiredOrder, $currentMethodsOrder);
 
+        // nothing to re-order
+        if (! $this->hasOrderChanged($oldToNewKeys)) {
+            return null;
+        }
```
TomasVotruba, line 95: @dobryy Just few tips for next Rector rules.

This check prevents false positives on applied Rector rules:

image


If node is returned, the rule says "I've changed this code". In this case, the order remained the same, so it was false positive. 

So `null` should be returned instead.

rules/order/src/Rector/Class_/OrderPrivateMethodsByUseRector.php
```diff
@@ -87,18 +76,14 @@ private function call2()
...
     /**
-     * @param ClassLike $node
+     * @param Class_|Trait_ $node
      */
     public function refactor(Node $node): ?Node
     {
-        if ($node instanceof Interface_) {
-            return null;
-        }
-
```
TomasVotruba, line 101: @dobryy Tip for simpler code with types. 

In "minus typing" it's sometimes impossible to see what types are expected.

Here it was "all but Interface_..." how many nodes are checked? It's like `!($yes || $maybe)` conditions.

Here it's better to explicitly add types `Class_|Trait_` without "minus typing". This way IDE and PHPStan and human can read it better.

rules/order/src/Rector/Class_/OrderMethodsByVisibilityRector.php
```diff
@@ -42,22 +44,6 @@ final class OrderMethodsByVisibilityRector extends AbstractRector
...
-    /**
-     * @var StmtVisibilitySorter
-     */
-    private $stmtVisibilitySorter;
-
-    public function __construct(StmtOrder $stmtOrder, StmtVisibilitySorter $stmtVisibilitySorter)
-    {
-        $this->stmtOrder = $stmtOrder;
-        $this->stmtVisibilitySorter = $stmtVisibilitySorter;
-    }
```
TomasVotruba, line 59: @dobryy These 2 services are repeated over and over again, so it makes sense to decouple `Abstract...Rector`, that has them both.

Now it's easy to add common method like  `hasOrderChanged()` ;)

TomasVotruba (commented):
@dobryy I've refactored "order" rules a bit, so I though it could be useful to share tips for better Rector rules in the future :)

rules/order/src/Rector/Class_/OrderMethodsByVisibilityRector.php
dobryy: Usually I don't use Abstract classes when it's all about DI, if this is how you do it in Rector, I don't mind to follow this rule (PHPStan? :D). Now it makes sense to use Abstract class because there is new method that is required everywhere.

rules/order/src/Rector/Class_/OrderMethodsByVisibilityRector.php
dobryy: Good to know, thank you!

rules/order/src/Rector/Class_/OrderMethodsByVisibilityRector.php
TomasVotruba: Sure :) If you can distile PHPStan rule from this, go for it :+1: 

It's usually about common services/methods/logics. Same goes for PHPUnitAbstractRector.

rules/order/src/StmtVisibilitySorter.php
```diff
@@ -86,42 +72,24 @@ public function sortMethods(ClassLike $classLike): array
...
-                    $secondArray['static'],
-                    $firstArray['abstract'],
-                    $secondArray['final'],
-                    $secondArray[self::POSITION],
-                ];
-            }
-        );
-
-        return $classMethods;
+        return $this->sortByRanksAndGetNames($classMethodsRankeables);
```
TomasVotruba, line 83: @dobryy One last trick that I just learn to use (PHPStan rules in progress ;)): value objects beat arrays.

Here all the sorting arrays with ranks could be refactored to value objects.

I can see the common piece to all rules is "rank" (any rank).

So `RankeableInterface` can be used to sort them all and `sortByRanksAndGetNames()` can handle all the sorting stuff.

TomasVotruba:
Btw, great job with test quality :+1: 

I was able to do refactoring with any frustration :)

PR #4076, opened 2020-08-30
[Naming] Rename the rule to make it correspond with what rule does

dobryy:
@TomasVotruba 
After these changes https://github.com/rectorphp/rector/commit/8b831a75b85ae32a4a64ef39db0c441698a3a76e, it looks like this rule is no longer doing what it was expected originally to do, thus it should be renamed to `RenameVariableToMatchTypeRector` because now it resolves the variable name from method name but primarily from type. This PR is **only** about renaming.

Also, we no longer have fixtures that test the original logic for getting the name from the method name because all methods in fixtures were changed to return objects. One more issue is that this rule will fully skip methods that return array or mixed types but in such cases, we can use the logic that will return the name that could be extracted from the method name (the part after `get`). 

Let me know if the issues above are valid and I'll fix them in the next PR.


TomasVotruba:
The name could be changed for better.

Under `RenameVariableToMatchTypeRector` it can be relate to:

- param type

```php
public function some(SomeType $value)
{
    $value  ...
}
```

Not the case.


- assigned property type


```php
private someType $value;

public function some(SomeType $value)
{
    $value = $this->value;
}
```

Not the case either.

<br>

Could you think of better name?

dobryy:
> The name could be changed for better.
> 
> Under `RenameVariableToMatchTypeRector` it can be related to:
> 
> * param-infered type
> 
> ```
> public function some(SomeType $value)
> {
>     $value ...
> }
> ```
> 
> Not the case.
> 
> * assigned property-infered type
> 
> ```
> private SomeType $value;
> 
> public function some()
> {
>     $value = $this->value;
> }
> ```
> 
> Not the case either.
> 
> 
> Could you think of better name?

You're right... what about `RenameVariableToMatchMethodCallReturnType`?

TomasVotruba:
Like it :+1:

dobryy:
Ready for merging

TomasVotruba:
Thanks

dobryy:
What about other issues I described? Are they worth fixing?

PR #4089, opened 2020-08-31
[CI] Add type-declaration set

TomasVotruba:
Follow up to #4088


packages/static-type-mapper/src/PHPStan/NameScopeFactory.php
```diff
@@ -47,7 +47,10 @@ private function resolveUseNamesByAlias(array $useNodes): array
                     throw new ShouldNotHappenException();
                 }
 
-                $useNamesByAlias[$aliasName] = $useName;
+                // uses must be lowercase, as PHPStan lowercases it
+                $lowercasedAliasName = strtolower($aliasName);
+
+                $useNamesByAlias[$lowercasedAliasName] = $useName;
```
TomasVotruba, line 53: @dobryy This is my favorite WTF. PHPStan only uses imports in lowercase, so this was skipped:

```php
namespace App;

use Something\SomeMatch;


echo SomeMatch::class; // PHPStan though it's: "App\SomeMatch"
```

PR #4091, opened 2020-08-31
[DX] Add rule to check not "Tests" namespace outside "tests" directory

dobryy:


dobryy:
@TomasVotruba 
There're no such cases in the code. I created one to show that new rule works. 
Also the test is failing because the test fixture is in the `tests` directory of the `phpstan-extensions` util.

dobryy:
I can't come up with some solution to the problem

utils/phpstan-extensions/src/Rule/CheckNotTestsNamespaceOutsideTestsDirectoryRule.php
```diff
@@ -0,0 +1,47 @@
...
+            return [];
+        }
+
+        if (
+            in_array('Tests', $node->name->parts, true) &&
+            ! strstr($scope->getFileDescription(), '/tests/')
+        ) {
+            $errorMessage = sprintf(self::ERROR_MESSAGE, $node->name->toString(), $scope->getFileDescription());
+            return [$errorMessage];
+        }
```
TomasVotruba, line 43: Could you refactor this to early return, so it's better readable?

utils/phpstan-extensions/config/phpstan-extensions.neon
```diff
@@ -11,6 +11,10 @@ services:
          class: Rector\PHPStanExtensions\Rule\KeepRectorNamespaceForRectorRule
          tags: [phpstan.rules.rule]
 
+    -
+         class: Rector\PHPStanExtensions\Rule\CheckNotTestsNamespaceOutsideTestsDirectoryRule
+         tags: [phpstan.rules.rule]
```
TomasVotruba, line 16: :+1:

dobryy:
At the moment, I have no idea how to make it smarter :)
If it fine to leave it as it is I'd do some fine-tuning for if condition and remove the `TestNameSpace` and it is ready for merging.

TomasVotruba:
> Also the test is failing because the test fixture is in the tests directory of the phpstan-extensions util.

Make it pass without no error, as the location correct (there is namespace Tests and location tests)

TomasVotruba:
Looks good, thanks :+1:

utils/phpstan-extensions/tests/Rule/CheckNotTestsNamespaceOutsideTestsDirectoryRule/Fixture/tests/TestsNamespaceOutsideTestsDirectoryClass.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+
+namespace Rector\PHPStanExtensions\Tests\Rule\CheckNotTestsNamespaceOutsideTestsDirectoryRule\Fixture\tests;
```
TomasVotruba, line 4: I just see it, this should be 

```diff
-namespace Rector\PHPStanExtensions\Tests\Rule\CheckNotTestsNamespaceOutsideTestsDirectoryRule\Fixture\tests;
+namespace Rector\PHPStanExtensions\Tests\Rule\CheckNotTestsNamespaceOutsideTestsDirectoryRule\Fixture\Tests;
```

TomasVotruba: Without that, the class is not autoloaded.

PR #4092, opened 2020-08-31
Replace types with container access from fe72e003e. fixes #4090

### LINKED ISSUE #4090: *GetToConstructorInjectionRector rectors lost default configuration
# Bug Report

| Rector version | current dev-master              |
| Installed as   | any                      |

The default configurations have been lost for (at least) the rectors:
* GetToConstructorInjectionRector
* ContainerGetToConstructorInjectionRector

In commit fe72e003e, when the classes were made configurable, the `$getMethodAwareTypes` & `$containerAwareParentTypes`, previously set in the __construct() method of the rectors were not added to the configuration file config/set/symfony-constructor-injection.php (or set as default properties, which has happened with some other rector upgrades).

The similar configuration for config/set/easy-admin-bundle20.php was upgraded in fe72e003e, and originally set in commit bc0d9b367 (in file config/set/easy-corp/easy-admin-bundle20.php).

Other rectors that may have also lost configuration from default __constructor parameters (from commit fe72e003e):

* rules/order/src/Rector/ClassMethod/OrderConstructorDependenciesByTypeAlphabeticallyRector.php  
* rules/php74/src/Rector/Function_/ReservedFnFunctionRector.php

alister:
The properties `$containerAwareParentTypes` & `$getMethodAwareTypes` are set as property defaults, so that if they were explicitly enabled (without using config/set/symfony-constructor-injection.php), they would still work as expected without additional setup needed.

Other 'ContainerAware' classes may be useful to add, such as `\Symfony\Component\DependencyInjection\ContainerAwareTrait` & '\Symfony\Component\DependencyInjection\ContainerAwareInterface'.

### DIFF
--- a/rules/symfony/src/Rector/MethodCall/ContainerGetToConstructorInjectionRector.php
+++ b/rules/symfony/src/Rector/MethodCall/ContainerGetToConstructorInjectionRector.php
@@ -26,7 +26,10 @@ final class ContainerGetToConstructorInjectionRector extends AbstractToConstruct
     /**
      * @var string[]
      */
-    private $containerAwareParentTypes = [];
+    private $containerAwareParentTypes = [
+        'Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand',
+        'Symfony\Bundle\FrameworkBundle\Controller\Controller',
+    ];
 
     public function getDefinition(): RectorDefinition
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thanks :+1:

PR #4095, opened 2020-09-01
Fixes #4013 ConsoleExecuteReturnIntRector ignores already type casted variable

### LINKED ISSUE #4013: ConsoleExecuteReturnIntRector ignores already type casted variable
# Bug Report

| Rector version | v0.7.65              |

https://getrector.org/demo/9a18901a-b9d7-4545-97b9-53c5313cd65e

Bugged class: `\Rector\Symfony\Rector\ClassMethod\ConsoleExecuteReturnIntRector`

This is the result:

```diff
{
 		$initialized = parent::execute($input, $output);
         if (0 !== $initialized) {
-            return (int) $initialized;
+            return (int) (int) $initialized;
         }
 		
 		return 0;
 	}
 }
```

samsonasik:
Fixes #4013

### DIFF
--- a/rules/symfony/src/Rector/ClassMethod/ConsoleExecuteReturnIntRector.php
+++ b/rules/symfony/src/Rector/ClassMethod/ConsoleExecuteReturnIntRector.php
@@ -113,7 +113,6 @@ PHP
             }
 
             if ($node->expr instanceof Int_) {
-                $hasReturn = true;
                 return null;
             }
 
--- a/rules/symfony/src/Rector/ClassMethod/ConsoleExecuteReturnIntRector.php
+++ b/rules/symfony/src/Rector/ClassMethod/ConsoleExecuteReturnIntRector.php
@@ -112,6 +112,11 @@ PHP
                 return null;
             }
 
+            if ($node->expr instanceof Int_) {
+                $hasReturn = true;
+                return null;
+            }
+
             // is there return without nesting?
             if ($this->areNodesEqual($node->getAttribute(AttributeKey::PARENT_NODE), $classMethod)) {
                 $hasReturn = true;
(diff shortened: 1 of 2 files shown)


rules/symfony/tests/Rector/ClassMethod/ConsoleExecuteReturnIntRector/Fixture/skip-already-typehinted.php.inc
```diff
@@ -0,0 +1,65 @@
...
+        $initialized = parent::execute($input, $output);
+        if (0 !== $initialized) {
+            return (int) $initialized;
+        }
+
+        return 0;
+    }
+}
+
+?>
```
TomasVotruba, line 65: In case the file should be untouched (no changes), this whole part is not needed

samsonasik: fixed.

rules/symfony/src/Rector/ClassMethod/ConsoleExecuteReturnIntRector.php
```diff
@@ -112,6 +112,11 @@ private function addReturn0ToMethod(ClassMethod $classMethod): void
                 return null;
             }
 
+            if ($node->expr instanceof Int_) {
+                $hasReturn = true;
```
TomasVotruba, line 116: I think this value is not need, since it's used bellow to return `$node`/null.

What do you think?

samsonasik: yes, updated.

samsonasik:
How about rename fixture file to `skip-return-already-typecasted.php.inc` ?

TomasVotruba:
This is good enough, thank you :+1:

PR #4098, opened 2020-09-01
[PHPUnit] Add self call fixtures for AssertEqualsToSame rule

peter279k:
# Changed log

- There're two approaches about calling PHPUnit assertion methods.
One is about the `$this` and another is about the `self`.
- The ` AssertEqualsToSameRector` class should add fixtures about `self` method call.


TomasVotruba:
Thanks :+1: 

Could you add fix too?

peter279k:
> Thanks 
> 
> Could you add fix too?

Thanks. It's fixed now :).

TomasVotruba:
Thanks :+1:

PR #4099, opened 2020-09-01
Configurable downgrade

### LINKED ISSUE #4094: Proposal: allow to not add docblocks when downgrading code (property types, type unions)
# Feature Request

When downgrading the typed properties ([DowngradeTypedPropertyRector](https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#downgradetypedpropertyrector)) and union types (#4062), allow to configure the rules to simply remove the code, and avoid adding the docblocks.

Why? Because I'm using these rules to compile code for production, so that:

- I develop with PHP 7.4 (and 8.0 later on)
- I deploy to PHP 7.2

For production, I do not need the docblocks. However, they have an inconvenient side effect: they shift the line numbers of the code. Then, if in production I have a bug, the log saying it happens on line 30, I can't just go to my code in development and see what is there on line 30. But if the docblocks are not generated, then line 30 in production is line 30 in development.

leoloso:
Fixes #4094 

Please check the following:

- Option called `ADD_DOC_BLOCK`
- Default value: `true`
- To configure it, I've added in `rector.php`:

```php
// Don't output the docBlocks when removing typed properties
$services->set(DowngradeTypedPropertyRector::class)
    ->call('configure', [[
        DowngradeTypedPropertyRector::ADD_DOC_BLOCK => false,
    ]]);
```

Is this right? (Btw, it took me some time to find out about how to configure rectors, this info is not in the README... maybe you want to add a section to explain?)

- The tests have 2 configuration values (`true` and `false`). I couldn't find an example of a same test for different configurations. So I kept the original test (`DowngradeTypedPropertyRector`), and duplicated it under as `NoDocBlockDowngradeTypedPropertyRector\DowngradeTypedPropertyRector`. Is this ok?

### DIFF
--- a/rules/downgrade/src/Rector/Property/DowngradeTypedPropertyRector.php
+++ b/rules/downgrade/src/Rector/Property/DowngradeTypedPropertyRector.php
@@ -6,12 +6,12 @@ namespace Rector\Downgrade\Rector\Property;
 
 use PhpParser\Node;
 use PhpParser\Node\Stmt\Property;
-use Rector\Core\Rector\AbstractRector;
-use Rector\Core\RectorDefinition\CodeSample;
-use Rector\NodeTypeResolver\Node\AttributeKey;
-use Rector\Core\RectorDefinition\RectorDefinition;
 use Rector\BetterPhpDocParser\PhpDocInfo\PhpDocInfo;
 use Rector\Core\Contract\Rector\ConfigurableRectorInterface;
+use Rector\Core\Rector\AbstractRector;
+use Rector\Core\RectorDefinition\CodeSample;
+use Rector\Core\RectorDefinition\RectorDefinition;
+use Rector\NodeTypeResolver\Node\AttributeKey;
 
 /**
  * @see \Rector\Downgrade\Tests\Rector\Property\DowngradeTypedPropertyRector\DowngradeTypedPropertyRectorTest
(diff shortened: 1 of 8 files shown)


TomasVotruba (approved):

leoloso:
@TomasVotruba One test is failing, with message:

```
PHP Fatal error:  Cannot declare class Rector\Downgrade\Tests\Rector\Property\DowngradeTypedPropertyRector\Fixture\ClassNameClass, because the name is already in use in /tmp/_temp_fixture_easy_testing/input_42f32a6e0f67cda8388e_classname_type.php on line 5
```

I imagine it's because of the test under `NoDocBlockDowngradeTypedPropertyRector`, but I don't know how to fix it

TomasVotruba:
Hi, the feature look great :+1: 

You're right with the documentation. Could I humbly ask you to add it with this PR? I'm too deep to write it easily and lightly for Rector begginner. I'm affraid I would overcomplicated it and it not bring much value.
You seem better to candidate, as you've just learned it.

TomasVotruba:
> I imagine it's because of the test under NoDocBlockDowngradeTypedPropertyRector, but I don't know how to fix it

Test fixture class must be unique.

leoloso:
Alright, fixed now, please check it out

leoloso:
> Could I humbly ask you to add it with this PR?

Alright, will do it tomorrow

TomasVotruba:
Looks great, thank you :+1:

PR #4117, opened 2020-09-03
FixMissing Ruleset does not throw SetNotFoundException 

julianpollmann:


TomasVotruba:
I see, thank you :+1:

PR #4118, opened 2020-09-03
Abstract downgrade type + Implementation for object type

### LINKED ISSUE #4108: Proposal: [Downgrade] PHP 8.0 mixed type
# Feature Request

PHP 8.0 will add the [`mixed` type](https://stitcher.io/blog/new-in-php-8#new-mixed-type-rfc). It would be cool to be able to remove it, as to compile PHP 8.0 code to lower versions.
 
## Diff 

```diff
-function foo(mixed $bar): mixed {}
+/**
+ * @param mixed $bar
+ * @return mixed
+ */
+function foo($bar) {}
```

leoloso:
Fixes #4108 (well, 99% of it, read below)

The original proposal is to downgrade the `mixed` type. In this PR, I did something better: 

- I implemented it as an abstract class, so that any type can be downgraded
    - `AbstractDowngradeParamTypeDeclarationRector`
    - `AbstractDowngradeReturnTypeDeclarationRector`
- I implemented it for `object`:
    - `DowngradeParamObjectTypeDeclarationRector`
    - `DowngradeReturnObjectTypeDeclarationRector`

With these downgrade rules, I can already:

- Code using PHP 7.4
- Deploy it to PHP 7.1 (yay!)

I haven't implemented it for the `mixed` type yet, since PHP 8.0 is still a few months away. But it's a copy/paste of the `object` type implementations, so be welcome to do it at any time.

I've also added the 2 new rectors to the downgrade set, and I made all rules in this set be executed only if the feature is not supported by the intended PHP version (eg: `object` is removed if target PHP is 7.1, not for 7.2 upwards).

### DIFF
--- a/config/set/downgrade.php
+++ b/config/set/downgrade.php
@@ -2,11 +2,17 @@
 
 declare(strict_types=1);
 
+use Rector\Downgrade\Rector\FunctionLike\DowngradeParamObjectTypeDeclarationRector;
+use Rector\Downgrade\Rector\FunctionLike\DowngradeReturnObjectTypeDeclarationRector;
 use Rector\Downgrade\Rector\Property\DowngradeTypedPropertyRector;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 
 return static function (ContainerConfigurator $containerConfigurator): void {
     $services = $containerConfigurator->services();
+
+    $services->set(DowngradeParamObjectTypeDeclarationRector::class);
+
+    $services->set(DowngradeReturnObjectTypeDeclarationRector::class);
 
     $services->set(DowngradeTypedPropertyRector::class);
 };
--- a/rules/downgrade/src/Rector/DowngradeRectorInterface.php
+++ b/rules/downgrade/src/Rector/DowngradeRectorInterface.php
@@ -4,10 +4,10 @@ declare(strict_types=1);
 
 namespace Rector\Downgrade\Rector;
 
-trait DowngradeRectorTrait
+interface DowngradeRectorInterface
 {
     /**
      * Run the rector only when the feature is not supported
      */
-    abstract protected function getPhpVersionFeature(): string;
+    function getPhpVersionFeature(): string;
 }
--- a/rules/downgrade/src/Rector/DowngradeRectorInterface.php
+++ b/rules/downgrade/src/Rector/DowngradeRectorInterface.php
@@ -9,5 +9,5 @@ interface DowngradeRectorInterface
     /**
      * Run the rector only when the feature is not supported
      */
-    function getPhpVersionFeature(): string;
+    public function getPhpVersionFeature(): string;
 }
--- a/rules/downgrade/src/Rector/DowngradeRectorTrait.php
+++ b/rules/downgrade/src/Rector/DowngradeRectorTrait.php
@@ -1,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Downgrade\Rector;
+
+trait DowngradeRectorTrait
+{
+    /**
+     * Run the rector only when the feature is not supported
+     */
+    abstract protected function getPhpVersionFeature(): string;
+}
--- a/rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
+++ b/rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
@@ -13,14 +13,12 @@ use PhpParser\Node\Stmt\ClassMethod;
 use PhpParser\Node\Stmt\Function_;
 use Rector\BetterPhpDocParser\PhpDocInfo\PhpDocInfo;
 use Rector\Core\Contract\Rector\ConfigurableRectorInterface;
-use Rector\Downgrade\Rector\DowngradeRectorTrait;
+use Rector\Downgrade\Rector\DowngradeRectorInterface;
 use Rector\NodeTypeResolver\Node\AttributeKey;
 use Rector\TypeDeclaration\Rector\FunctionLike\AbstractTypeDeclarationRector;
 
-abstract class AbstractDowngradeParamTypeDeclarationRector extends AbstractTypeDeclarationRector implements ConfigurableRectorInterface
+abstract class AbstractDowngradeParamTypeDeclarationRector extends AbstractTypeDeclarationRector implements ConfigurableRectorInterface, DowngradeRectorInterface
 {
-    use DowngradeRectorTrait;
-
     /**
      * @var string
      */
--- a/rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
+++ b/rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
@@ -13,7 +13,7 @@ use PhpParser\Node\Stmt\ClassMethod;
 use PhpParser\Node\Stmt\Function_;
 use Rector\BetterPhpDocParser\PhpDocInfo\PhpDocInfo;
 use Rector\Core\Contract\Rector\ConfigurableRectorInterface;
-use Rector\Downgrade\Rector\DowngradeRectorInterface;
+use Rector\Downgrade\Contract\Rector\DowngradeRectorInterface;
 use Rector\NodeTypeResolver\Node\AttributeKey;
 use Rector\TypeDeclaration\Rector\FunctionLike\AbstractTypeDeclarationRector;
 
--- a/rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
+++ b/rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
@@ -12,9 +12,10 @@ use PhpParser\Node\Stmt\Function_;
 use PhpParser\Node\Stmt\ClassMethod;
 use Rector\NodeTypeResolver\Node\AttributeKey;
 use Rector\Downgrade\Rector\DowngradeRectorTrait;
+use Rector\Core\Contract\Rector\ConfigurableRectorInterface;
 use Rector\TypeDeclaration\Rector\FunctionLike\AbstractTypeDeclarationRector;
 
-abstract class AbstractDowngradeParamTypeDeclarationRector extends AbstractTypeDeclarationRector
+abstract class AbstractDowngradeParamTypeDeclarationRector extends AbstractTypeDeclarationRector implements ConfigurableRectorInterface
 {
     use DowngradeRectorTrait;
 
--- a/rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
+++ b/rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
@@ -111,7 +111,15 @@ abstract class AbstractDowngradeParamTypeDeclarationRector extends AbstractTypeD
         }
 
         // If it is the NullableType, extract the name from its inner type
-        $typeName = $isNullableType ? $this->getName($param->type->type) : $this->getName($param->type);
+        if ($isNullableType) {
+            /**
+             * @var NullableType
+             */
+            $nullableType = $param->type;
+            $typeName = $this->getName($nullableType->type);
+        } else {
+            $typeName = $this->getName($param->type);
+        }
 
         // Check it is the type to be removed
         return $typeName !== $this->getParamTypeName();
--- a/rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
+++ b/rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
@@ -111,7 +111,7 @@ abstract class AbstractDowngradeParamTypeDeclarationRector extends AbstractTypeD
         }
 
         // If it is the NullableType, extract the name from its inner type
-        $typeName = $isNullableType ? $param->type->type->name : $param->type->name;
+        $typeName = $isNullableType ? $this->getName($param->type->type) : $this->getName($param->type);
 
         // Check it is the type to be removed
         return $typeName !== $this->getParamTypeName();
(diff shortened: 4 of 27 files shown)


leoloso:
phpstan is giving a few errors, and with my beginner knowledge of Rector I'm unable to fix them, so I need help.

Errors:

```

 rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php:86
 - '#Parameter \#1 \$node of method Rector\\StaticTypeMapper\\StaticTypeMapper\:\:mapPhpParserNodePHPStanType\(\) expects PhpParser\\Node, PhpParser\\Node\\Identifier\|PhpParser\\Node\\Name\|PhpParser\\Node\\NullableType\|PhpParser\\Node\\UnionType\|null given#'

 rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php:87
 - '#Parameter \#3 \$paramName of method Rector\\BetterPhpDocParser\\PhpDocInfo\\PhpDocInfo\:\:changeParamType\(\) expects string, PhpParser\\Node\\Expr\|string given#'

 rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php:114
 - '#Access to an undefined property PhpParser\\Node\\Identifier\|PhpParser\\Node\\Name\:\:\$name#'

 rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php:114
 - '#Access to an undefined property PhpParser\\Node\\Identifier\|PhpParser\\Node\\Name\|PhpParser\\Node\\NullableType\|PhpParser\\Node\\UnionType\:\:\$name#'

 rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php:114
 - '#Access to an undefined property PhpParser\\Node\\Identifier\|PhpParser\\Node\\Name\|PhpParser\\Node\\NullableType\|PhpParser\\Node\\UnionType\:\:\$type#'

 rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeReturnTypeDeclarationRector.php:52
 - '#Parameter \#1 \$node of method Rector\\StaticTypeMapper\\StaticTypeMapper\:\:mapPhpParserNodePHPStanType\(\) expects PhpParser\\Node, PhpParser\\Node\\Identifier\|PhpParser\\Node\\Name\|PhpParser\\Node\\NullableType\|PhpParser\\Node\\UnionType\|null given#'

 rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeReturnTypeDeclarationRector.php:87
 - '#Access to an undefined property PhpParser\\Node\\Identifier\|PhpParser\\Node\\Name\:\:\$name#'

 rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeReturnTypeDeclarationRector.php:87
 - '#Access to an undefined property PhpParser\\Node\\Identifier\|PhpParser\\Node\\Name\|PhpParser\\Node\\NullableType\|PhpParser\\Node\\UnionType\:\:\$name#'

 rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeReturnTypeDeclarationRector.php:87
 - '#Access to an undefined property PhpParser\\Node\\Identifier\|PhpParser\\Node\\Name\|PhpParser\\Node\\NullableType\|PhpParser\\Node\\UnionType\:\:\$type#'
```

rules/downgrade/src/Rector/DowngradeRectorTrait.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Rector\Downgrade\Rector;
+
+trait DowngradeRectorTrait
+{
+    /**
+     * Run the rector only when the feature is not supported
+     */
+    abstract protected function getPhpVersionFeature(): string;
+}
```
TomasVotruba, line 13: We're trying to avoid traits (except for dependency injection decoupling of `AbstractRector` class), as they're very hard to declutter. 

This should be `interface` instead, wich clear contract.

config/set/downgrade.php
```diff
@@ -2,11 +2,17 @@
...
+use Rector\Downgrade\Rector\FunctionLike\DowngradeReturnObjectTypeDeclarationRector;
 use Rector\Downgrade\Rector\Property\DowngradeTypedPropertyRector;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 
 return static function (ContainerConfigurator $containerConfigurator): void {
     $services = $containerConfigurator->services();
 
+    $services->set(DowngradeParamObjectTypeDeclarationRector::class);
+
+    $services->set(DowngradeReturnObjectTypeDeclarationRector::class);
```
TomasVotruba, line 15: I wonder, why keeping them separate? How is it useful to run just one without the other?

rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
```diff
@@ -0,0 +1,119 @@
...
+
+        if ($this->addDocBlock) {
+            $node = $functionLike;
+            /** @var PhpDocInfo|null $phpDocInfo */
+            $phpDocInfo = $node->getAttribute(AttributeKey::PHP_DOC_INFO);
+            if ($phpDocInfo === null) {
+                $phpDocInfo = $this->phpDocInfoFactory->createEmpty($node);
+            }
+
+            $type = $this->staticTypeMapper->mapPhpParserNodePHPStanType($param->type);
```
TomasVotruba, line 86:
```bash
 rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php:86
 - '#Parameter \#1 \$node of method Rector\\StaticTypeMapper\\StaticTypeMapper\:\:mapPhpParserNodePHPStanType\(\) expects PhpParser\\Node, PhpParser\\Node\\Identifier\|PhpParser\\Node\\Name\|PhpParser\\Node\\NullableType\|PhpParser\\Node\\UnionType\|null given#'
```

It means, the `$param->type` can be null here, which would lead to crash

rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
```diff
@@ -0,0 +1,119 @@
...
+            $phpDocInfo->changeParamType($type, $param, $param->var->name);
```
TomasVotruba, line 87:
```
 rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php:87
 - '#Parameter \#3 \$paramName of method Rector\\BetterPhpDocParser\\PhpDocInfo\\PhpDocInfo\:\:changeParamType\(\) expects string, PhpParser\\Node\\Expr\|string given#'
```

This means, that `$param->var->name` can als oreturn `Expr`, which would crash.
The best way is to use `getName()` method to handle this internally:

```suggestion

            /** @var string */   
            $paramName = $param->var->name;
            $phpDocInfo->changeParamType($type, $param, $paramName);
```

rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        // It can either be the type, or the nullable type (eg: ?object)
+        $isNullableType = $param->type instanceof NullableType;
+        if (! ($param->type instanceof Identifier || $isNullableType)) {
+            return true;
+        }
+
+        // If it is the NullableType, extract the name from its inner type
+        $typeName = $isNullableType ? $param->type->type->name : $param->type->name;
```
TomasVotruba, line 114: Similar to above.
```suggestion
        $typeName = $isNullableType ? $this->getName($param->type->type) : $this->getName($param->type);
```

Or possibly `getValue()` instead, as types do not have names

TomasVotruba (commented):
Hi, thanks so much for this work.

I've added few comments and tips, how to fix PHPStan errors. It seems 2 kinds of errors over and over. I hope it gives you guidance to solve it.

Let me know if you need any more help.

config/set/downgrade.php
leoloso: Hmmmm, that's a good point. I guess they are always executed together.

But even if that's the case, I wonder if there is value in keeping them separate:

- Each rule can have its own, more specific tests (eg: the test `multiple_params.php.inc` doesn't make sense for return types)
- It makes code more SOLID

If this makes sense, then you could possibly have them shipped through a set `downgrade-object`, so that they are still always executed together, and then have sets execute not only rules but also other sets, so that set `downgrade` executes set `downgrade-object`. Is this possible?

Otherwise, I'll combine the 2 rules into 1: `DowngradeObjectTypeDeclarationRector`.

Please let me know what you prefer

rules/downgrade/src/Rector/DowngradeRectorTrait.php
leoloso: Alright, replaced as `DowngradeRectorInterface`

rules/downgrade/src/Rector/DowngradeRectorInterface.php
```diff
@@ -0,0 +1,13 @@
...
+declare(strict_types=1);
+
+namespace Rector\Downgrade\Rector;
+
+interface DowngradeRectorInterface
+{
+    /**
+     * Run the rector only when the feature is not supported
+     */
+    function getPhpVersionFeature(): string;
```
samsonasik, line 12: `public` modifier should be added

leoloso: Oops, fixed now

rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
leoloso: Fixed, checking param is not `null` before logic

rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
leoloso: Fixed

rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
leoloso: Fixed

leoloso:
Question: should implementing `DowngradeRectorInterface` be enforced on all rules under the `Downgrade` namespace? `ShouldNotHappenException`?

config/set/downgrade.php
leoloso: Actually, yes, there is a use case: In PHP 8.0, `static` is a return type, but not a param type: https://stitcher.io/blog/new-in-php-8#new-static-return-type-rfc

leoloso: Keeping it decoupled, the rule to remove return type `static` can use the same abstract class as for return types `object` and `mixed`

rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
```diff
@@ -0,0 +1,126 @@
...
+
+    public function configure(array $configuration): void
+    {
+        $this->addDocBlock = $configuration[self::ADD_DOC_BLOCK] ?? true;
+    }
+
+    /**
+     * Name of the type to remove
+     */
+    abstract protected function getParamTypeName(): string;
```
TomasVotruba, line 60: This should not be needed now

rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeReturnTypeDeclarationRector.php
```diff
@@ -0,0 +1,98 @@
...
+
+    public function configure(array $configuration): void
+    {
+        $this->addDocBlock = $configuration[self::ADD_DOC_BLOCK] ?? true;
+    }
+
+    /**
+     * Name of the type to remove
+     */
+    abstract protected function getReturnTypeName(): string;
```
TomasVotruba, line 69: Should be extracted to interface. Abstract methods are hiding contracts and hard to filter by

config/set/downgrade.php
TomasVotruba: Thanks for detailed explanation. Let's go with your approach. I believe it will bring more value on additional type rules

leoloso: Which approach? What is already implemented, so I don't do anything? Or create a set `downgrade-object` and execute it within set `downgrade`?

rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
leoloso: Done. Extracted to `DowngradeTypeRectorInterface`, renamed to function `getTypeNameToRemove`

leoloso: (This is the same solution as below)

rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeReturnTypeDeclarationRector.php
leoloso: Done. Extracted to `DowngradeTypeRectorInterface`, renamed to function `getTypeNameToRemove`

config/set/downgrade.php
TomasVotruba: > What is already implemented, so I don't do anything?

:+1:

TomasVotruba:
> Question: should implementing DowngradeRectorInterface

That's a good idea. Could you make it in next PR, just so we have it separated as inspiration for other rules?

TomasVotruba:
For me it's ready to merge. Do you want to add somethign or shall I merge it?

leoloso:
Please merge it! :)

TomasVotruba:
Let's :ship:  it 

Thank you :)

leoloso:
@TomasVotruba if you agree, I'll now create separate issues to keep downgrading:

- PHP 8.0 - `mixed` type 
- PHP 8.0 - `static` return type 
- PHP 7.4 - Null coalescing assignment operator 
- PHP 7.4 - Arrow functions

The first 2 I can easily implement in 5 min, but I guess we should wait until PHP 8.0 is out (eg: it's not stable for the CI yet). 

The last 2, I could possibly attempt to tackle them, but not anytime soon because I checked some similar rectors for reference and I don't fully understand them, and I'm a bit busy at the time.

What do you think?

TomasVotruba:
Let's start with first 2. 
The CI could be extended in next PR, we'll need it for PHP 8.0 anyway.

After we merge them, I could help you navigate null coalesing. That will be matter of adding one `null` node.

leoloso:
Alright, can. But I'd like to already create the last 2 issues, so in my project I can link to them for future reference, as "coming soon" features

TomasVotruba:
Sure

PR #4119, opened 2020-09-03
Correct various mysql_* to mysqli_* conversion bugs

jjthiessen:
This change addresses the following:

* Remove additional arguments (beyond host, username, and password) when converting `mysql_connect` and `mysql_pconnect`. The remaining arguments do not match up between the mysql and mysqli versions.
* Don't swap arguments on `mysql_fetch_row` — both it and the mysqli version take exactly one argument.
* Convert `mysql_fetch_field` into `mysqli_fetch_field` (no argument case) or `mysqli_fetch_field_direct` (one argument case). The pre-existing code adds a static loop for running `mysqli_fetch_field` five times. It is unclear why.
* `mysql_num_fields` should map to `mysqli_num_fields`; not `mysqli_field_count`.
* The pre-existing handling of `mysql_query` would convert `$result = mysql_query($sql);`into `$result = mysqli_query($result, $sql);`; and `$link = mysql_connect($host, $user, $pass); $result = mysql_query($sql);`into `$link = mysqli_connect($host, $user, $pass); $result = mysqli_query($result, $link, $sql);`. The updated heuristics seems to handle these and similar cases better (most of the time).
* A number of mysql functions took an optional link parameter. The corresponding mysqli functions take mandatory link parameters. This change set handles those functions in the link adding rule (`MysqlQueryMysqlErrorWithLinkRector`) instead of using the argument pass though function renaming approach.
* Add more test content to the `mysql-to-mysqli` set. The test is not minimal, but was expanded to capture function renaming (not captured elsewhere) and interplay between the various rules (ensuring valid rule set sequencing and link/connection discovery/handling).

I have notably not renamed `MysqlQueryMysqlErrorWithLinkRector` as I wasn't sure whether I should introduce a breaking change for this, or not. The class name `MysqlQueryMysqlErrorWithLinkRector` is (after this change), a bit of a misnomer as it handles more cases now, but it functionally does the same thing (it adds seemingly missing mandatory link arguments).

### DIFF
--- a/rules/mysql-to-mysqli/src/Rector/FuncCall/MysqlQueryMysqlErrorWithLinkRector.php
+++ b/rules/mysql-to-mysqli/src/Rector/FuncCall/MysqlQueryMysqlErrorWithLinkRector.php
@@ -10,7 +10,7 @@ use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\Assign;
 use PhpParser\Node\Expr\FuncCall;
 use PhpParser\Node\Name;
-use PHPStan\Type\ObjectType;
+use PHPStan\Type\ResourceType;
 use PHPStan\Type\UnionType;
 use Rector\Core\Rector\AbstractRector;
 use Rector\Core\RectorDefinition\CodeSample;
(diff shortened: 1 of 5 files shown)


rules/mysql-to-mysqli/src/Rector/FuncCall/MysqlQueryMysqlErrorWithLinkRector.php
```diff
@@ -82,44 +98,64 @@ public function getNodeTypes(): array
...
-        }
-
         foreach (self::FUNCTION_RENAME_MAP as $oldFunction => $newFunction) {
             if (! $this->isName($node, $oldFunction)) {
                 continue;
             }
 
+            if (
+                count($node->args) === 0
+                || ! $this->isProbablyMysql($node->args[0]->value)
```
jjthiessen, line 108: An alternative approach to this would be to partition the handled functions by the new function arity, then add a link/connection parameter whenever the number of args applied is less than what's required.

That may handle the case of externally initialised link arguments better (where PHPStan doesn't seem to ascertain the type correctly, and assumes that the link is not a resource or `mysqli` object (leading to too many parameters being provided (similar to the pre-existing code))).

TomasVotruba:
Amazing work :+1::
Thank you

BC breaks are allowed, since we're in 0.x version

TomasVotruba (approved):

PR #4132, opened 2020-09-05
Commented out vendor locking

### LINKED ISSUE #4131: [Downgrade] Inconsistent removing of params because of locking
# Bug Report

| Rector version | v0.8.6             |

Running `DowngradeParamObjectTypeDeclarationRector` does not remove the `object` param when the class implements an interface, but it does remove it from the interface, hence there is a mismatch (the rector fails [here](https://github.com/getpop/component-model/blob/3585c08ac7298ad4f58b3a2972780db484ad7c46/src/TypeResolvers/AbstractTypeResolver.php#L1300)).

The reason is [this code in `AbstractDowngradeParamTypeDeclarationRector`](https://github.com/rectorphp/rector/blob/71682d2b52edd8d0f36fe822e077c7f05a8311f4/rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php#L92):

```php
if ($this->vendorLockResolver->isClassMethodParamLockedIn($functionLike, $position)) {
    return true;
}
```

Does locking make sense in the context of downgrading? I'd say:

- If downgrading only one file, it makes sense
- If downgrading the whole project (as I am doing), it does not

So, what is the best solution?

- Remove this code?
- Add configuration `ENABLE_LOCKING`, read it in the several `isVendorLocked` functions, and set it to `false` in rector.php?
- Other?

The same solution should also be implemented for `AbstractDowngradeReturnTypeDeclarationRector`, which currently does not implement locking.

leoloso:
Fixes #4131 

Possible solution: It removes the code calling `isVendorLock`.

### DIFF
--- a/rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
+++ b/rules/downgrade/src/Rector/FunctionLike/AbstractDowngradeParamTypeDeclarationRector.php
@@ -87,7 +87,7 @@ abstract class AbstractDowngradeParamTypeDeclarationRector extends AbstractTypeD
         $param->type = null;
     }
 
-    private function shouldSkipParam(Param $param, FunctionLike $functionLike, int $position): bool
+    private function shouldSkipParam(Param $param): bool
     {
         if ($param->variadic) {
             return true;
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Thank you. I don't use it myself, so can't have a opinion about it.
I trust your practical experience with this rule and that you choose the most practical approach.

:+1:

PR #4136, opened 2020-09-07
Fix example: exclude directory in config

obstschale:
The line looks wrong. I am not sure if the PR is correct or if a trailing slash is required.


TomasVotruba:
Thank you, the core correct is good

PR #4135, opened 2020-09-07
Downgrade null coalescing operator

leoloso:
Fixes #4124

PR #4137, opened 2020-09-07
Downgrade arrow functions

### LINKED ISSUE #4125: [Downgrade] PHP 7.4 - Arrow functions
# Feature Request

Replace PHP 7.4's [arrow functions](https://www.php.net/manual/en/functions.arrow.php) with its equivalent pre-7.4 code.
 
## Diff 

```diff
$factor = 10;
-$nums = array_map(fn($n) => $n * $factor, [1, 2, 3, 4]);
+$nums = array_map(
+  function ($n) use ($factor) {
+    return $n * $factor;
+  },
+  [1, 2, 3, 4]
+);
```

leoloso:
Fixes #4125 

Please notice that the new rule `ArrowFunctionToAnonymousFunctionRector` reuses the code from `CreateFunctionToAnonymousFunctionRector` to obtain the variables to declare as `use(...)`.

Then, I first extracted the code from this class into `AbstractConvertToAnonymousFunctionRector`, and inherited from it. This new abstract class is placed on a new namespace `Rector\AbstractRector\Rector`, since it is used by `CreateFunctionToAnonymousFunctionRector` (under `Php72`) and `ArrowFunctionToAnonymousFunctionRector` (under `Downgrade`).

Let me know if this namespace should be something else.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,6 +64,7 @@
     },
     "autoload": {
         "psr-4": {
+            "Rector\\AbstractRector\\": "rules/abstract-rector/src",
             "Rector\\Architecture\\": "rules/architecture/src",
             "Rector\\AnonymousClass\\": "packages/anonymous-class/src",
             "Rector\\PostRector\\": "packages/post-rector/src",
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,6 @@
     },
     "autoload": {
         "psr-4": {
-            "Rector\\AbstractRector\\": "rules/abstract-rector/src",
             "Rector\\Architecture\\": "rules/architecture/src",
             "Rector\\AnonymousClass\\": "packages/anonymous-class/src",
             "Rector\\PostRector\\": "packages/post-rector/src",
--- a/rules/abstract-rector/src/Rector/AbstractConvertToAnonymousFunctionRector.php
+++ b/rules/abstract-rector/src/Rector/AbstractConvertToAnonymousFunctionRector.php
@@ -1,0 +1,108 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\AbstractRector\Rector;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\Assign;
+use PhpParser\Node\Expr\Closure;
+use PhpParser\Node\Expr\ClosureUse;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Param;
+use PhpParser\Node\Stmt;
+use PhpParser\Node\Stmt\Expression;
+use Rector\Core\Rector\AbstractRector;
+use Rector\NodeTypeResolver\Node\AttributeKey;
+
+/**
+ * @see https://www.php.net/functions.anonymous
+ */
+abstract class AbstractConvertToAnonymousFunctionRector extends AbstractRector
+{
+    public function refactor(Node $node): ?Node
+    {
+        if ($this->shouldSkip($node)) {
+            return null;
+        }
+
+        $parameters = $this->getParameters($node);
+        $body = $this->getBody($node);
+        $useVariables = $this->resolveUseVariables($body, $parameters);
+
+        $anonymousFunctionNode = new Closure();
+
+        foreach ($parameters as $parameter) {
+            $anonymousFunctionNode->params[] = new Param($parameter);
+        }
+
+        if ($body !== []) {
+            $anonymousFunctionNode->stmts = $body;
+        }
+
+        foreach ($useVariables as $useVariable) {
+            $anonymousFunctionNode->uses[] = new ClosureUse($useVariable);
+        }
+
+        return $anonymousFunctionNode;
+    }
+
+    abstract protected function shouldSkip(Node $node): bool;
+
+    /**
+     * @return Variable[]
+     */
+    abstract protected function getParameters(Node $node): array;
+
+    /**
+     * @return Expression[]|Stmt[]
+     */
+    abstract protected function getBody(Node $node): array;
+
+    /**
+     * @param Node[] $nodes
+     * @param Variable[] $paramNodes
+     * @return Variable[]
+     */
+    private function resolveUseVariables(array $nodes, array $paramNodes): array
+    {
+        $paramNames = [];
+        foreach ($paramNodes as $paramNode) {
+            $paramNames[] = $this->getName($paramNode);
+        }
+
+        $variableNodes = $this->betterNodeFinder->findInstanceOf($nodes, Variable::class);
+
+        /** @var Variable[] $filteredVariables */
+        $filteredVariables = [];
+        $alreadyAssignedVariables = [];
+        foreach ($variableNodes as $variableNode) {
+            // "$this" is allowed
+            if ($this->isName($variableNode, 'this')) {
+                continue;
+            }
+
+            $variableName = $this->getName($variableNode);
+            if ($variableName === null) {
+                continue;
+            }
+
+            if (in_array($variableName, $paramNames, true)) {
+                continue;
+            }
+
+            $parentNode = $variableNode->getAttribute(AttributeKey::PARENT_NODE);
+            if ($parentNode instanceof Assign) {
+                $alreadyAssignedVariables[] = $variableName;
+            }
+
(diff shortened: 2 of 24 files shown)


leoloso:
I must still check that the return type is preserved, working on it now

composer.json
```diff
@@ -64,6 +64,7 @@
     },
     "autoload": {
         "psr-4": {
+            "Rector\\AbstractRector\\": "rules/abstract-rector/src",
```
TomasVotruba, line 67: We tend to put abstract classes near to their first/most use.
E.g. `AbstractPHPUnitRector` as at `rules/phpunit` package etc. Could you move it to the place it's used?

rules/abstract-rector/src/Rector/AbstractConvertToAnonymousFunctionRector.php
```diff
@@ -0,0 +1,108 @@
...
+
+        $parameters = $this->getParameters($node);
+        $body = $this->getBody($node);
+        $useVariables = $this->resolveUseVariables($body, $parameters);
+
+        $anonymousFunctionNode = new Closure();
+
+        foreach ($parameters as $parameter) {
+            $anonymousFunctionNode->params[] = new Param($parameter);
+        }
```
TomasVotruba, line 37: What about direct assign of all parameters?

rules/abstract-rector/src/Rector/AbstractConvertToAnonymousFunctionRector.php
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * @return Variable[]
+     */
+    abstract protected function getParameters(Node $node): array;
+
+    /**
+     * @return Expression[]|Stmt[]
+     */
+    abstract protected function getBody(Node $node): array;
```
TomasVotruba, line 60: Interface over abstract method

rules/abstract-rector/src/Rector/AbstractConvertToAnonymousFunctionRector.php
```diff
@@ -0,0 +1,108 @@
...
+                $alreadyAssignedVariables[] = $variableName;
+            }
+
+            if ($this->isNames($variableNode, $alreadyAssignedVariables)) {
+                continue;
+            }
+
+            $filteredVariables[$variableName] = $variableNode;
+        }
+
```
TomasVotruba, line 105: What does this foreach do? Could you extract it to own method, so the name tells us?

rules/downgrade/tests/Rector/ArrowFunction/ArrowFunctionToAnonymousFunctionRector/Fixture/no_use.php.inc
```diff
@@ -0,0 +1,31 @@
...
+
+class WithNoUseClass
+{
+    public function duplicate()
+    {
+        $numbers = [3, 5, 7];
+        return array_map(function (int $number) {
+            return $number * 2;
+        }, $numbers);
+    }
```
TomasVotruba, line 28: Beautiful tests 😍

composer.json
leoloso: But it is used in 2 places: by `CreateFunctionToAnonymousFunctionRector` (under `Php72`) and `ArrowFunctionToAnonymousFunctionRector` (under `Downgrade`)... that's why I put it on yet another place...

Tell me exactly where to put it, and I'll do it

rules/abstract-rector/src/Rector/AbstractConvertToAnonymousFunctionRector.php
leoloso: Right, I'll check it out

rules/abstract-rector/src/Rector/AbstractConvertToAnonymousFunctionRector.php
leoloso: Ok, I'll work on it

rules/abstract-rector/src/Rector/AbstractConvertToAnonymousFunctionRector.php
leoloso: Ok done

rules/abstract-rector/src/Rector/AbstractConvertToAnonymousFunctionRector.php
leoloso: Fixed

composer.json
TomasVotruba: Which of those 2 you consider first?

rules/abstract-rector/src/Rector/AbstractConvertToAnonymousFunctionRector.php
leoloso: This foreach is calculating the variables that will go under `use`:
```php
function ($var1) use ($var2) {
```

For instance, if when parsing the function it finds variable `$foo`, then it checks that it's not a parameter to the function, and that it was not first assigned in the function, then the variable must've been defined outside and must go to `use`.

So I think there's no need to extract it, since it's already doing `resolveUseVariables`

leoloso:
@TomasVotruba phpstan is now failing (in GitHub, not in my computer), and I have no idea why, so I need your help:

```

  utils/documentation-generator/src/Command/DumpNodesCommand.php:190
  - '#Cognitive complexity for "Rector\\Utils\\DocumentationGenerator\\Command\\DumpNodesCommand\:\:execute\(\)" is 105, keep it under 8#'

  rules/php70/src/EregToPcreTransformer.php:98
  - '#Cognitive complexity for "Rector\\Php70\\EregToPcreTransformer\:\:_ere2pcre\(\)" is 33, keep it under 8#'

  rules/privatization/src/Rector/Class_/ChangeLocalPropertyToVariableRector.php:128
  - '#Cognitive complexity for "Rector\\Privatization\\Rector\\Class_\\ChangeLocalPropertyToVariableRector\:\:collectPropertyFetchByMethods\(\)" is 9, keep it under 8#'
```

composer.json
leoloso: `php72`, since the `downgrade` one applies to PHP 7.4

leoloso:
The same problem is also happening in the other PR https://github.com/rectorphp/rector/pull/4135! And the change there was minimal. So the issue must not be related to my code

composer.json
TomasVotruba: Let's go with `php72` then

TomasVotruba:
I'll handle the phpstan later, don't worry about it

leoloso:
Alright, I fixed everything now :)

TomasVotruba:
Thank you 👍 


Btw, what is your timezone and timespan for open-source? 
I want to plan to be at your disposal, if you need help

leoloso:
I'm based in Kuala Lumpur, Malaysia, GMT +8. But don't worry about me, I can also wait until the following day!

My plan is to keep working on new rules tomorrow: I'll implement the downgrade type union, `static` and `mixed` type rules for PHP 8, and, if you agree, I'll create new issues for additional downgrades (from PHP 7.3, 7.4 and 8.0) to be implemented some time in the future, and give myself a break.

Then, I'll keep working on downgrading my WordPress plugin from PHP 7.4 to 7.1 these few days, and then promote this solution among the WordPress community (since it's the [hot topic right now](https://make.wordpress.org/core/2020/08/24/proposal-dropping-support-for-old-php-versions-via-a-fixed-schedule/)). I'll also write an article about it, which will be really cool. It should be published in a few weeks, I'll let you know :)

TomasVotruba:
> I'm based in Kuala Lumpur, Malaysia, GMT +8

Cool, that's one of 2 countries I've visited in Asia. K-L :+1: 

> if you agree, I'll create new issues for additional downgrades 

Please do :+1: 

> Then, I'll keep working on downgrading my WordPress plugin from PHP 7.4 to 7.1 these few days

That's great work that you're doing. If this set is complete, which IMO is matter of 1-3 months of work, then any project can code in PHP 8 and deploy in PHP 7.1-7.4 without any such discussions.

Best of both worlds :)

PR #4138, opened 2020-09-07
apply run ecs check-heredoc-nowdoc command to packages and rules directory

samsonasik:
Run `ecs` command for `HeredocNowdocPHPCodeFormatter` to `rules` and `packages` directory:

```
packages/easy-coding-standard/bin/ecs check-heredoc-nowdoc ../rector/packages
packages/easy-coding-standard/bin/ecs check-heredoc-nowdoc ../rector/rules
```

by https://github.com/symplify/symplify/pull/2137

### DIFF
--- a/rules/phpunit/src/Rector/MethodCall/DelegateExceptionArgumentsRector.php
+++ b/rules/phpunit/src/Rector/MethodCall/DelegateExceptionArgumentsRector.php
@@ -32,8 +32,8 @@ final class DelegateExceptionArgumentsRector extends AbstractPHPUnitRector
             [
                 new CodeSample(
                     '$this->setExpectedException(Exception::class, "Message", "CODE");',
-                    <<<PHP
-{$this->setExpectedException}(Throwable::class);
+                    <<<'PHP'
+{$this->setExpectedException}(Exception::class);
 {$this->expectExceptionMessage}('Message');
 {$this->expectExceptionCode}('CODE');
 PHP
--- a/rules/phpunit/src/Rector/MethodCall/DelegateExceptionArgumentsRector.php
+++ b/rules/phpunit/src/Rector/MethodCall/DelegateExceptionArgumentsRector.php
@@ -33,9 +33,9 @@ final class DelegateExceptionArgumentsRector extends AbstractPHPUnitRector
                 new CodeSample(
                     '$this->setExpectedException(Exception::class, "Message", "CODE");',
                     <<<PHP
-$this->setExpectedException(Throwable::class);
-$this->expectExceptionMessage('Message');
-$this->expectExceptionCode('CODE');
+{$this->setExpectedException}(Throwable::class);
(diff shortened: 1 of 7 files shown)


TomasVotruba:
Thanks for this :+1: 

I wonder how can we get rid of `strict_types`, because they often seem as unnecesary clutter.

We could have standalone `ecs.php`, but that seems as extra work. If we apply "keep the same as before" approach, it might be hard to add these types.

What do you think?

samsonasik:
what about add option like `--no-strict-declaration` option?

TomasVotruba:
Sounds good :+1:

samsonasik:
I've added `--no-strict-types-declaration` option at https://github.com/symplify/symplify/pull/2141 and run on rules and packages directory:

```
packages/easy-coding-standard/bin/ecs check-heredoc-nowdoc ../rector/packages --no-strict-types-declaration
packages/easy-coding-standard/bin/ecs check-heredoc-nowdoc ../rector/rules --no-strict-types-declaration
```

rules/phpunit/src/Rector/MethodCall/DelegateExceptionArgumentsRector.php
```diff
@@ -33,9 +33,9 @@ public function getDefinition(): RectorDefinition
                 new CodeSample(
                     '$this->setExpectedException(Exception::class, "Message", "CODE");',
                     <<<'PHP'
-$this->setExpectedException(Exception::class);
-$this->expectExceptionMessage("Message");
-$this->expectExceptionCode("CODE");
+$this->setExpectedException(Throwable::class);
```
samsonasik, line 36: I am not sure why the ecs change `Exception::class` to `Throwable::class` on this, @TomasVotruba could you verify?

TomasVotruba: There is a rule that does this change as part of upgrade to PHP 7.0.

Could you update the line above also to heredoc?
That way the change would be in both snippets and correct

samsonasik: I've updated to heredoc

samsonasik: Do you think `Throwable::class` need to be reverted to `Exception::class` ?

samsonasik: I reverted to nowdoc and changed back to `Exception::class`

samsonasik: Now all green, I reverted to nowdoc as phpstan complain about it.

TomasVotruba:
Thank you :+1:

PR #4144, opened 2020-09-08
Try add changes detected to markdown check

samsonasik:


samsonasik:
Got error:

```

Started: bash /home/runner/work/_actions/stefanzweifel/git-auto-commit-action/v4.1.6/entrypoint.sh
INPUT_REPOSITORY value: .
INPUT_BRANCH value: master
error: pathspec 'master' did not match any file(s) known to git
Error: Invalid status code: 1
    at ChildProcess.<anonymous> (/home/runner/work/_actions/stefanzweifel/git-auto-commit-action/v4.1.6/index.js:17:19)
    at ChildProcess.emit (events.js:210:5)
    at maybeClose (internal/child_process.js:1021:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:283:5) {
  code: 1
}
Error: Invalid status code: 1
    at ChildProcess.<anonymous> (/home/runner/work/_actions/stefanzweifel/git-auto-commit-action/v4.1.6/index.js:17:19)
    at ChildProcess.emit (events.js:210:5)
    at maybeClose (internal/child_process.js:1021:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:283:5)
```

.github/workflows/check_markdown.yml
```diff
@@ -1,7 +1,7 @@
 name: Check markdown
 
 on:
-    push:
+    pull_request:
         branches:
             - master
```
TomasVotruba, line 6: If you look at other workflows, there should be working format of this

TomasVotruba: Also, in `rector_ci.yaml`, there is a snippet that already handles autocommits.

On the other hand, maybe it's not needed to autocommit on everydocs change, as it might be annoying to contributors to always pull or force-push on bot commits.

What do you think?

samsonasik: Do you think I need to copy/modify step at rector_ci.yaml ? https://github.com/rectorphp/rector/blob/6508aedd1321f5905c1d2728c9e1579071fd885a/.github/workflows/rector_ci.yaml#L45-L80 

The check-markdown command will keep the file if already follow coding standard, so it only commit when not follow coding standard.

TomasVotruba: > so it only commit when not follow coding standard.

That's it the issue. If someone wants to contribute a new rule, they'd have to handle this fix manually or pull the commits commits from a bot. That is a big surprise and requires extra git skill.

I wonder, maybe it could be part of `composer.json` scripts and not CI.

```json
{
    "scripts": {
        "check-cs": "vendor/bin/ecs check --ansi",
        "fix-cs": [
            "vendor/bin/ecs check --fix --ansi",
            "bin/clear_posts_trailing_whitespaces.sh"
        ],
    }
}
```

TomasVotruba: We had same troubles with Generated Documentation. It's not neccesary on every contributions and only bothers new contributors who just want to add that single rule.

So we decided to put that workflow into inactive and handle this manaully from time to time. It's not perfect, but developer experience has higher priority.

samsonasik: ok, I will try add to composer script

TomasVotruba: Thanks

PR #4147, opened 2020-09-08
Organize downgrades by PHP version

### LINKED ISSUE #4143: Proposal: Add the PHP namespace to the downgrade rules
# Feature Request

As much as there's namespace `Php71`, `Php72`, etc, to migrate to new features, I think it makes sense to add such namespaces below `Downgrade`, to indicate from what version we are downgrading from.

This makes it easier to add them into the set, since the downgrading must be executed in order:

- First all PHP 8.0 rules
- Then all PHP 7.4 rules
- Then all PHP 7.3 rules
- ...

This is because `??=` in 7.4 downgrades to `??` from 7.3, which can itself be downgraded to `isset` in 5.6.

In addition, a great side effect of adding the namespace is that, I imagine, when listing all rules in the rules overview `.md`, they with all be ordered all together by PHP version.
 
## Diff 

```diff
-namespace Rector\Downgrade\Rector\Property;
+namespace Rector\Downgrade\Php74\Rector\Property;

final class DowngradeTypedPropertyRector
```

leoloso:
Fixes #4143


TomasVotruba:
Thank you :+1:

Issue #4062, opened 2020-08-29
[Downgrade] PHP 8.0 union types to doc types


TomasVotruba:
See https://github.com/rectorphp/rector/issues/4051

leoloso:
Hey @TomasVotruba if you want, I can already handle this issue, because I can use the same abstract classes from the other issues I just implemented:

To downgrade type unions, I can just inherit from `AbstractDowngradeParamTypeDeclarationRector` and `AbstractDowngradeReturnTypeDeclarationRector`, and set `shouldSkip` as `!($property->type instanceof UnionType)`.

In addition, I can extract class `AbstractDowngradeTypedPropertyRector` from `DowngradeTypedPropertyRector`, and from there create `DowngradeTypeUnionTypedPropertyRector`.

I've seen you've done plenty of changes in the associated PR, including adding PHP 8.0 to the matrix. We could then decouple this issue into 2 issues:

- Adding support for PHP 8.0
- Downgrading the type union

Then, we could have this issue only handle the first part "Adding support for PHP 8.0" (removing file `DowngradeUnionTypeToDocBlockRector` from the PR), and possibly you can already merge it to master.

And then, I can create a new issue for "Downgrading the type union", and submit a PR with the 3 new rules:

- `DowngradeTypeUnionTypedPropertyRector`
- `DowngradeParamTypeUnionDeclarationRector`
- `DowngradeReturnTypeUnionDeclarationRector`

What do you think?

TomasVotruba:
Hi @leoloso,

thanks for detailed plan. I love how you prepare a path and fill my possible questions with answers :+1: 

Please, go for it

leoloso:
Great! So please take care of converting this issue and PR to "Adding support for PHP 8.0". After it's merged, I can then submit PRs with PHP 8 code

TomasVotruba:
This issue can remain as it is, since we often close them very quickly.
Just close it with union PRs.

leoloso:
I'm not concerned with this issue, but with PR #4082 since it integrates PHP 8.0 to the CI, so I need that merged before I can submit another PR with PHP 8 code...

Or you mean that I should also work on that same branch, `downgrade-union`? There are conflicting files there, and I wouldn't be able to solve that.

TomasVotruba:
That is not neccessary as it can take weeks or months to merge.
There are many 3rd party blockres.

See https://github.com/rectorphp/rector/runs/1051205204

[...]

PR #4150, opened 2020-09-08
Downgrade union type

leoloso:
Implements 3 rules to downgrade union types (for property types, params and returns), as described in https://github.com/rectorphp/rector/issues/4062#issuecomment-687712666


TomasVotruba (approved):

leoloso:
I believe it works, but in practice we haven't run the PHP 8.0 tests yet

TomasVotruba:
Thank you

PR #4153, opened 2020-09-08
[SymfonyPHPConfig] Add monorepo split for value objects function

TomasVotruba:
## Why?

So people can use `Rector\SymfonyPhpConfig\inline_value_objects` function without requiring whole Rector

## Todo

- [ ] follow up with post on https://getrector.org/blog


TomasVotruba:
# UX improvement in missing Rule on Run

## Before

uggly


## Now

now

PR #4162, opened 2020-09-08
[Downgrade PHP 8.0] mixed and static types

leoloso:
Remove `mixed` and `static` types:

```php
public function foo(mixed $anything): mixed
{
}

public function bar(): static
{
}
```


TomasVotruba:
:tada: 

Clicking on merge button was never so much fun :-1:

leoloso:
Once again, I believe it works, but never run the PHP 8.0 tests

TomasVotruba:
Understand. No need to report it under your work anymore :+1: 

We'll only see it, when we have PHP 8.0 here

PR #4172, opened 2020-09-08
Added downgrade rules

leoloso:
I had forgotten to add the 3 new rules to the `downgrade` set in PR #4162

### DIFF
--- a/config/set/downgrade.php
+++ b/config/set/downgrade.php
@@ -7,6 +7,9 @@ use Rector\DowngradePhp72\Rector\FunctionLike\DowngradeReturnObjectTypeDeclarati
 use Rector\DowngradePhp74\Rector\ArrowFunction\ArrowFunctionToAnonymousFunctionRector;
 use Rector\DowngradePhp74\Rector\Coalesce\DowngradeNullCoalescingOperatorRector;
 use Rector\DowngradePhp74\Rector\Property\DowngradeTypedPropertyRector;
+use Rector\DowngradePhp80\Rector\FunctionLike\DowngradeParamMixedTypeDeclarationRector;
+use Rector\DowngradePhp80\Rector\FunctionLike\DowngradeReturnMixedTypeDeclarationRector;
+use Rector\DowngradePhp80\Rector\FunctionLike\DowngradeReturnStaticTypeDeclarationRector;
 use Rector\DowngradePhp80\Rector\FunctionLike\DowngradeUnionTypeParamDeclarationRector;
 use Rector\DowngradePhp80\Rector\FunctionLike\DowngradeUnionTypeReturnDeclarationRector;
 use Rector\DowngradePhp80\Rector\Property\DowngradeUnionTypeTypedPropertyRector;
@@ -23,6 +26,9 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     $services->set(DowngradeUnionTypeTypedPropertyRector::class);
     $services->set(DowngradeUnionTypeReturnDeclarationRector::class);
     $services->set(DowngradeUnionTypeParamDeclarationRector::class);
+    $services->set(DowngradeParamMixedTypeDeclarationRector::class);
+    $services->set(DowngradeReturnMixedTypeDeclarationRector::class);
+    $services->set(DowngradeReturnStaticTypeDeclarationRector::class);
     // PHP 7.4
     $services->set(DowngradeTypedPropertyRector::class);
     $services->set(ArrowFunctionToAnonymousFunctionRector::class);


leoloso:
Btw, I finished adding all the downgrade issues, so I'm now done 🎉

Btw2, I also added Symfony's polyfill dependencies to my plugin, so now it supports all [these PHP features from PHP 7.1 to 8.0](https://github.com/GraphQLAPI/graphql-api-for-wp#php-versions). It looks so cool!

config/set/downgrade.php
```diff
@@ -7,6 +7,9 @@
 use Rector\DowngradePhp74\Rector\ArrowFunction\ArrowFunctionToAnonymousFunctionRector;
 use Rector\DowngradePhp74\Rector\Coalesce\DowngradeNullCoalescingOperatorRector;
 use Rector\DowngradePhp74\Rector\Property\DowngradeTypedPropertyRector;
+use Rector\DowngradePhp80\Rector\FunctionLike\DowngradeParamMixedTypeDeclarationRector;
+use Rector\DowngradePhp80\Rector\FunctionLike\DowngradeReturnMixedTypeDeclarationRector;
+use Rector\DowngradePhp80\Rector\FunctionLike\DowngradeReturnStaticTypeDeclarationRector;
```
TomasVotruba, line 12: I just noticed, this downgrade set should be split by version, e.g.

- `downgrade-php80.php`
- `downgrade-php74.php`

etc. As other configs are, e.g. `php74.php`, `php80.php` etc

leoloso: I think there's no need, because all downgrade rules are implementing interface `DowngradeRectorInterface` with function `getPhpVersionFeature`, and only if the target PHP is supported then the downgrade is executed

leoloso: So, basically you have 2 options to choose from:

1. Keep it as it is, defining what downgrades to perform with:

```php
  SetList::DOWNGRADE,
]);

$parameters->set(Option::PHP_VERSION_FEATURES, '7.1');
```

2. Have to specify all the sets to execute: 

```php
  SetList::DOWNGRADE_PHP80,
  SetList::DOWNGRADE_PHP74,
  SetList::DOWNGRADE_PHP73,
  SetList::DOWNGRADE_PHP72,
]);
```

If you choose the 2nd one, I guess `DowngradeRectorInterface` should be removed

leoloso: But with the 2nd option, you're allowing the developer to introduce bugs! Downgrading must happen strictly from higher to lower PHP versions, so this can fail:

```php
  SetList::DOWNGRADE_PHP74,
  SetList::DOWNGRADE_PHP80,
  SetList::DOWNGRADE_PHP73,
  SetList::DOWNGRADE_PHP72,
]);
```

leoloso: Ok done. CI is failing, I'll need you to check on that.

I haven't removed `DowngradeRectorInterface`, so right now the rules in each downgrade set are only executed when `Option::PHP_VERSION_FEATURES` is lower than the downgrade PHP version. I should remove it, right? Please confirm if to remove it or keep it.

leoloso: > We can address the possible bug in standalone issue. Please make one that describes the problem with reproducible steps

Done: #4176

leoloso: Btw, do you plan to create a new tag soon? Then I can already include it in my writeup

leoloso: Btw, would you like to create projects (maybe one per PHP version) to manage all the downgrade issues?

TomasVotruba: > should remove it, right? Please confirm if to remove it or keep it.

Please do. Sets are now the way to go

TomasVotruba: > Btw, do you plan to create a new tag soon?

I can, if you need

TomasVotruba: > Btw, would you like to create projects (maybe one per PHP version) to manage all the downgrade issues?

No need to, we try to keep is simple here. 
Maybe one meta-issue would be better, to prevent spam in issue tracker. As we have for PHP 8.0 - https://github.com/rectorphp/rector/issues/3127

TomasVotruba:
I tried to answer all your questions. Let me know if I missed something :)

config/set/downgrade.php
leoloso: > Please do. Sets are now the way to go

Ok, I'll do it now

leoloso: > I can, if you need

Yes. I'll submit the PR now to remove `DowngradeRectorInterface`, then please tag it 😄

TomasVotruba: Allright :) I'll tag today after few more bugfixes... in 4-10 hours

PR #4174, opened 2020-09-08
[DX] Configurable rule must have configure code sample, and vice versa

dobryy:


dobryy:
@TomasVotruba is there any example of how to check that parent classes implement (or not) some interface?

TomasVotruba:
What about $scope->getClassReflection?

dobryy:
> What about $scope->getClassReflection?

`nul` :-\

TomasVotruba:
Ah yea, there is this wird problem with PHPStan that ClassReflection is only there, when inside the Class_.

Can you get class name from `$node` somehow? I'd try `dump($node)`

dobryy:
> Ah yea, there is this wird problem with PHPStan that ClassReflection is only there, when inside the Class_.
> 
> Can you get class name from `$node` somehow? I'd try `dump($node)`.
> I just always dump all arguments of the method and look for something useful :D

Believe me or not :D but I dumped it, also tried to check it using xdebugger before asking. I use `Class_` node for the new rule.

TomasVotruba:
That's weird. I think I use the `namespaceName` from Class_ in other rules. Maybe something changed. I'll look into it :)

TomasVotruba:
Looking at other PHPStan rules we have, the `KeepRectorNamespaceForRectorRule` contains the FQN name

TomasVotruba:
Ah. I see. So how can one get interfaces with just a class name?

TomasVotruba:
I misunderstood your question, I'm sorry

dobryy:
> Ah. I see. So how can one get interfaces with just a class name?
In the new rule, I use `$class->implements` but it doesn't return Interfaces that are implemented in the parent class.

TomasVotruba:
I see. So all the interface some class implements?

I tried to google it: https://www.google.com/search?sxsrf=ALeKk00-QPQDOxbxbayo2aFJKkFdL1MAiQ%3A1599643362790&ei=4p5YX5jtL_qBhbIPhbC48Aw&q=php+all+class+interface&oq=php+all+class+interface&gs_lcp=CgZwc3ktYWIQAzIICAAQCBANEB4yCAgAEAgQDRAeMggIABAIEA0QHjoICCEQFhAdEB46BwghEAoQoAE6BAgjECc6DQgjECcQiwMQpgMQqAM6CAgAELEDEIMBOgUIABCxAzoCCC46BAgAEEM6DggAELEDEIsDEKgDEKcDOgQILhBDOgIIADoHCAAQsQMQQzoFCAAQywE6CAgAEBYQChAeOgYIABAWEB5Q3HBYxZ4BYLGfAWgCcAB4AYABlQKIAd0tkgEHMC4yMS4xMJgBAKABAaABAqoBB2d3cy13aXq4AQLAAQE&sclient=psy-ab&ved=0ahUKEwiY5p_z39vrAhX6QEEAHQUYDs4Q4dUDCA0&uact=5

This looks the function I use for it:
https://www.php.net/manual/en/function.class-implements.php

dobryy:
@TomasVotruba. can you please check rules on which PHPStan is failing at the moment. There are two issues:
1. It looks like only `TYPES_TO_MATCH` is configurable and it is used only in `AbstractConfigurableMatchTypeRector::isMatchedType` method which is unused. So it looks like the rules are not configurable.
2. The second issue is that `AbstractFluentChainMethodCallRector` implements `ConfigurableRectorInterface` but at the same time there is no implementation of `configure` method in `AbstractFluentChainMethodCallRector` or in any child class. It is implemented in the parent abstract class `AbstractConfigurableMatchTypeRector`. I hear you "Rector rule?" :D

dobryy:
Something wrong with tests, even on `master` branch...

dobryy:
I missed one usage of `TYPES_TO_MATCH` constant in guzzle set... but anyway `isMatchedType` is unused. Can I remove `TYPES_TO_MATCH` from guzzle set configuration? hm...

TomasVotruba:
One of the files has broken syntax:

```bash
ParseError: syntax error, unexpected end of file, expecting variable (T_VARIABLE) or heredoc end (T_END_HEREDOC) or ${ (T_DOLLAR_OPEN_CURLY_BRACES) or {$ (T_CURLY_OPEN)
```

TomasVotruba:
> missed one usage of TYPES_TO_MATCH constant in guzzle set... but anyway isMatchedType is unused. Can I remove TYPES_TO_MATCH from guzzle set configuration? hm...

That should remain there, as it only limits to few specific classes:


image

TomasVotruba:
As for Syntax issues, I've added linter PHP to CI, so we can see what file has wrong syntax quickly:
https://github.com/rectorphp/rector/pull/4188/files

TomasVotruba:
This is how to use clear cache in PHSPtan:

```bash
vendor/bin/phpstan clear-result-cache
```

TomasVotruba:
See changes in HERE/NOW doc in PHP 7.3
https://laravel-news.com/flexible-heredoc-and-nowdoc-coming-to-php-7-3

dobryy:
Let's merge it! :)

TomasVotruba:
I forgot to add automerge tag... Now it's merged :). Thanks!

TomasVotruba:
The HERE/NOW doc will be covered soon by coding standard :), I've just added 2 rules:


https://github.com/symplify/symplify/pull/2156

PR #4175, opened 2020-09-09
Split downgrade sets

leoloso:
As requested in https://github.com/rectorphp/rector/pull/4172#discussion_r485041789

### DIFF
--- a/config/set/downgrade-php72.php
+++ b/config/set/downgrade-php72.php
@@ -1,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+use Rector\DowngradePhp72\Rector\FunctionLike\DowngradeParamObjectTypeDeclarationRector;
+use Rector\DowngradePhp72\Rector\FunctionLike\DowngradeReturnObjectTypeDeclarationRector;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
+    $services = $containerConfigurator->services();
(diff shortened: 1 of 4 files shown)


leoloso:
@TomasVotruba please check the error in the CI, I have no clue what that is:

```
 -
     message: '#Use explicit interface contract or a service over unclear abstract methods#'
     paths:
         - rules/generic/src/Rector/AbstractIsAbleRector.php
         - rules/nette/src/Rector/Identical/AbstractWithFunctionToNetteUtilsStringsRector.php
         - src/Testing/PHPUnit/AbstractGenericRectorTestCase.php
         - src/Testing/PHPUnit/AbstractNodeVisitorTestCase.php
```

Btw, I'm still unable to run `composer phpstan`

TomasVotruba:
> I have no clue what that is

I wish PHPStan would show rule name, so people can understand it.

It's because this new rules was added to symplify set: https://github.com/symplify/symplify/pull/2144

The goal here is to extract behaviors from hidden `abstract` method to interface contract: https://github.com/symplify/symplify/pull/2144/files#diff-11c3b9432cf183a58cfe8cb83c2a2564

If you'd like to try to handle it, go for it. If not, I'll refactor later :+1:

TomasVotruba:
I'll merge the sets as it they are, since the PHPStan error is unrelated and unnecessary blocker here.

Thanks

leoloso:
> If you'd like to try to handle it, go for it. If not, I'll refactor later 👍

Yep, you do it please 😛

PR #4177, opened 2020-09-09
Fixed invalid method name in code sample.

bkonetzny:
The rule doesn't change the method name from index to execute, it only fixes the return type and the TypeHint.


TomasVotruba:
Wow, good spot! How did you noticed this?

Thank you :+1:

PR #4179, opened 2020-09-09
Remove `DowngradeRectorInterface`

leoloso:
As explained here: https://github.com/rectorphp/rector/pull/4172#discussion_r485289483


TomasVotruba:
Thanks :+1:

TomasVotruba:
I see there is lot of upstream mergin. 

I think you deserve close access after all the work you're contributing :clap: 
I've invited you to collaborators, so you don't have to deal with that maintenance issues.

leoloso:
CI error is the same as before, so you can still merge this PR 😁

PR #4182, opened 2020-09-10
Fixes #3789 add skip Option to exclude files by rule

### LINKED ISSUE #3789: [Configuration] Add "skip" to easily excluded undesired files by rule
Basically copy of https://github.com/symplify/easy-coding-standard#ignore-what-you-cant-fix

samsonasik:


samsonasik:
@TomasVotruba could you give information what parts need to modified/added after the addition of new constant to `Option` class? Thank you.

TomasVotruba:
Sure.

This is the `Skipper` services in ECS: https://github.com/symplify/symplify/blob/master/packages/easy-coding-standard/src/Skipper.php
It decides whether file + rule should be skipped or not. The `shouldSkipCheckerAndFile()` is the important one for us here.

Where the service decides? In ECS it's in the file proccessor on each file + rule iteration:
https://github.com/symplify/symplify/blob/b740d246a81cb0c4311b4284db8ec36e87cd7444/packages/easy-coding-standard/packages/fixer-runner/src/Application/FixerFileProcessor.php#L150-L152

In Rector it's could be added in `AbstractRector` here - https://github.com/rectorphp/rector/blob/44f383038f1d5041ffb89bb3ccfcd97ade44af64/src/Rector/AbstractRector.php#L193

Like this: https://github.com/rectorphp/rector/pull/4184/files

samsonasik:
Current update, functional for path as collection of files:

```php
        Rector\CodeQuality\Rector\Array_\CallableThisArrayToAnonymousFunctionRector::class => [
           __DIR__ . '/skip/Foo.php',
        ],
        Rector\CodeQuality\Rector\Array_\ArrayThisCallToThisMethodCallRector::class => [
            __DIR__ . '/skip/Foo.php',
        ]
     ]);
```

samsonasik:
Now functional with both directories and files registered:

```php
        Rector\CodeQuality\Rector\Array_\CallableThisArrayToAnonymousFunctionRector::class => [
           __DIR__ . '/skip/',
           __DIR__ . '/foo/bar/baz.php',
        ],
        Rector\CodeQuality\Rector\Array_\ArrayThisCallToThisMethodCallRector::class => [
            __DIR__ . '/skip/',
            __DIR__ . '/foo/bar/baz.php',
        ]
     ]);
```

samsonasik:
All checks have passed 🎉 . I've updated README.md for its usage. @TomasVotruba it is ready for review

src/Skip/Skipper.php
```diff
@@ -0,0 +1,67 @@
...
+
+        foreach ($locations as $location) {
+            if (is_dir($location)) {
+                $finder = new Finder();
+                $finder->files()->in($location)->name('*.php');
+
+                if ($this->isFoundInDirectory($finder, $filePathName)) {
+                    return true;
+                }
+            }
```
TomasVotruba, line 47: This method could take a big performance hit on every directory + file + node.

Could you think of better way to check file path is in some directory path?

TomasVotruba (commented):
Nice job! I've added my review

src/Skip/Skipper.php
samsonasik: we may can do with : 

```php
$target = $location . DIRECTORY_SEPARATOR . (new SplFileInfo($filePathName))->getFileName();
if (is_file($target == $filePathName)) {
    return true;
}
```

However, this will only work for same structure, not recursive deep inside directory. Do you think that is enough?

samsonasik: I've updated code to cache found files in previous rules to static property, if exists, no need to re-search in same directory.

src/Skip/Skipper.php
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        return false;
+    }
+
+    private function isFoundInDirectory(string $directory, string $filePathName): bool
+    {
+        $directory = rtrim($directory, '/') . '/';
+        if (! array_key_exists($directory, self::$filesInDirectory)) {
```
samsonasik, line 60: here, we check if `$directory` is a key of `self::$filesInDirectory`

src/Skip/Skipper.php
```diff
@@ -0,0 +1,71 @@
...
+            $finder = new Finder();
+            $finder->files()->in($directory)->name('*.php');
+
+            foreach ($finder as $file) {
+                self::$filesInDirectory[$directory][] = $file->getRealPath();
```
samsonasik, line 65: add files in directory based on `$directory` as a key so in next iteration can be check if exists, no need to re-search the files inside it.

samsonasik:
All checks have passed again 🎉 . Performance has been optimized. @TomasVotruba it is ready for review again.

src/Skip/Skipper.php
TomasVotruba: What about same mechanisms as in ECS? Just match regex?

samsonasik: I've updated with match regex, and updated the readme for its usage.

samsonasik:
All checks have passed 🎉 . I've updated with match regex like in the ECS, @TomasVotruba it is ready for review again.

src/Skip/Skipper.php
TomasVotruba: Thank you for quick update

TomasVotruba:
Thank you :+1:

PR #4192, opened 2020-09-11
Downgrade PHP7.1 features: nullable types and void return type

leoloso:
Downgraded PHP 7.1 features:

- [Nullable types](https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types)
- [Void return type](https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.void-functions)


leoloso:
Btw, I tried to create a branch in rectorphp, but I got error: 

```
ERROR: Permission to rectorphp/rector.git denied to leoloso.
fatal: Could not read from remote repository.
```

Being a collaborator to rectorphp/rector does not allow me to create branches?

So I created it in my fork. I also made a new fork from scratch, as to get rid of all those annoying merges, which were always out of sync.

TomasVotruba:
I tried to promote your access. Could you check now?

rules/downgrade-php71/tests/Rector/FunctionLike/DowngradeNullableTypeReturnDeclarationRector/DowngradeNullableTypeReturnDeclarationRectorTest.php
```diff
@@ -0,0 +1,40 @@
...
+
+use Iterator;
+use Rector\Core\Testing\PHPUnit\AbstractRectorTestCase;
+use Rector\Core\ValueObject\PhpVersionFeature;
+use Rector\DowngradePhp71\Rector\FunctionLike\DowngradeNullableTypeReturnDeclarationRector;
+use Symplify\SmartFileSystem\SmartFileInfo;
+
+/**
+ * @requires PHP >= 7.1
+ */
```
TomasVotruba, line 15: This is not neccessary, since min PHP runnign on Rector is 7.2

leoloso:
> I tried to promote your access. Could you check now?

Yes, now it works, thanks.

TomasVotruba:
Great :) Thank you  :+1:

src/ValueObject/PhpVersionFeature.php
```diff
@@ -161,6 +161,16 @@ final class PhpVersionFeature
...
 
+    /**
+     * @var string
+     */
+    public const BEFORE_VOID_RETURN_TYPE = '7.0';
+
+    /**
+     * @var string
+     */
+    public const BEFORE_NULLABLE_TYPE = '7.0';
```
TomasVotruba, line 172: I wonder, I have no idea why I choose string type for PHP Versions. Maybe based on `composer.json`

I wonder, is there a better type to handle PHP versions here?

E:g.

- `70400`
- 7.4 (float)

<br>

This way here could re-used another numeric value:

```php
public const BEFORE_NULLABLE_TYPE = NULLABLE_TYPE - 0.1;

// 0.1 could be also constant
public const BEFORE_NULLABLE_TYPE = NULLABLE_TYPE - MINOR_BUMP; 
```


What do you think?

leoloso: `70400` <= I like it, it simplifies the comparison, and you can directly check against `PHP_VERSION_ID`.

Better doing `-1` instead of `-0.1` then it's always an `int`:

```php
public const BEFORE_NULLABLE_TYPE = NULLABLE_TYPE - 1;
```

TomasVotruba: Even better :+1: I'll make na issue

Issue #4086, opened 2020-08-31
[DX] Add Rule that checks that Rule + value object have same starts


TomasVotruba:
![image](https://user-images.githubusercontent.com/924196/91703711-e93f5200-eb7a-11ea-9198-ae741e063e9d.png)

Here should be:

- ClassConstantRenameRector
- ClassConstantRename

PR #4202, opened 2020-09-11
[DX] Add Rule that checks that Rule + value object have same starts

dobryy:
Closes: https://github.com/rectorphp/rector/issues/4086


dobryy:
@TomasVotruba can you please check the PHPStan rule, if it is okay, I'll fix PHPStan issues.

utils/phpstan-extensions/src/Rule/RectorRuleAndValueObjectHaveSameStartsRule.php
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * @see \Rector\PHPStanExtensions\Tests\Rule\RectorRuleAndValueObjectHaveSameStartsRule\RectorRuleAndValueObjectHaveSameStartsRuleTest
+ */
+final class RectorRuleAndValueObjectHaveSameStartsRule implements Rule
+{
+    /**
+     * @var string
+     */
+    public const ERROR = 'Value Object class name (%s) is incorrect. The correct class name is "%s".';
```
TomasVotruba, line 27:
```suggestion
    public const ERROR = 'Value Object class name "%s" is incorrect. The correct class name is "%s".';
```

TomasVotruba: For consistent output

TomasVotruba:
Looks good :+1: 

There is one edge case, when this should be skipped.

`RenameMethodRector` accepts 2 value objects, there is check for interface: https://github.com/rectorphp/rector/blob/42f377c8251ea23a1f8b6dc46c3488d1b333afc9/rules/renaming/src/Rector/MethodCall/RenameMethodRector.php#L35-L38

dobryy:
> Looks good 
> 
> There is one edge case, when this should be skipped.
> 
> `RenameMethodRector` accepts 2 value objects, there is check for interface:
> 
> https://github.com/rectorphp/rector/blob/42f377c8251ea23a1f8b6dc46c3488d1b333afc9/rules/renaming/src/Rector/MethodCall/RenameMethodRector.php#L35-L38

I see, I see, I wanted to handle this situation as well but then decided to do not complicate the logic in hope that we do not have such cases :D

TomasVotruba:
It's just 1 case, so I think we can just skip it now :)

dobryy:
> It's just 1 case, so I think we can just skip it now :)

Is there such a case configured for some set? If yes, do you mean to skip it using PHPStan ignore, or make rule skip such cases?

dobryy:
I see, there is such a case in `symfony41` set

TomasVotruba:
Ignore it, if the value object implements and interface

dobryy:
> Ignore it, if the value object implements and interface

After this change, PHPStan shows 66 errors. Before the change, it was 101 errors.

dobryy:
@TomasVotruba I have a feeling that sometimes not Value Object but Rector rule should be renamed :)
Each Value Object was renamed in the separate commit. Can you please check if it makes sense more to rename Rector rule rather than Value Object for some cases?

TomasVotruba:
> I have a feeling that sometimes not Value Object but Rector rule should be renamed :)

I see the changes and agree. Could you update it with the cases you have feeling it should be the rule renamed?

TomasVotruba:
I can also merge this as it is, so we have less code to review :) then you can follow up with PR with Rule + Value object rename. Ok?

dobryy:
Merge it, please :)

TomasVotruba:
Kaboom :boom:

config/set/symfony32.php
```diff
@@ -13,7 +13,7 @@
     $services->set(ArgumentAdderRector::class)
         ->call('configure', [[
             ArgumentAdderRector::ADDED_ARGUMENTS => inline_value_objects([
-                new AddedArgument(
+                new ArgumentAdder(
                     'Symfony\Component\DependencyInjection\ContainerBuilder',
                     'addCompilerPass',
                     2,
```
dobryy, line 19: AddArgumentRector

config/set/symfony42.php
```diff
@@ -164,44 +164,64 @@
...
+                    false,
+                    null
+                ),
+                new ArgumentDefaultValueReplacer(
+                    'Symfony\Component\HttpFoundation\Cookie',
+                    '__construct',
+                    8,
+                    null,
+                    'lax'
+                ),
```
dobryy, line 210: ReplaceArgumentDefaultValueRector and ReplaceArgumentDefaultValue

config/set/twig-underscore-to-namespace.php
```diff
@@ -172,7 +172,7 @@
     $services->set(PseudoNamespaceToNamespaceRector::class)
         ->call('configure', [[
             PseudoNamespaceToNamespaceRector::NAMESPACE_PREFIXES_WITH_EXCLUDED_CLASSES => inline_value_objects([
-                new NamespacePrefixWithExcludedClasses('Twig_'),
+                new PseudoNamespaceToNamespace('Twig_'),
             ]),
         ]]);
 };
```
dobryy, line 178: ChangePseudoNamespaceToNamespaceRector and ChangePseudoNamespaceToNamespace

dobryy (commented):
@TomasVotruba I checked this PR once more time and it seems like there is only one Rector rule that should definitely renamed rather than Value Object - `PropertyToMethodRector` should be renamed to `PropertyToMethodCallRector` but I commented some places that could be unified.

Haven't you thought about some naming convention for the rector rule? It looks like usually rules are doing action likes these:
rename
make
change
replace
remove
add
swap
wrap
move
split

So... maybe we may require one of these words at the beginning of the rule name or we may require -er form of the word (renamer, remover, mover, splitter, etc...) before Rector word.

Also sometimes in the rule names, we have Arg, sometimes Argument. This is something we can also unify.

Let me know what do you think.

config/set/cakephp36.php
```diff
@@ -26,9 +26,9 @@
...
         ->call('configure', [[
             PropertyToMethodRector::PROPERTIES_TO_METHOD_CALLS => inline_value_objects([
-                new PropertyToMethodCall('Cake\Controller\Controller', 'name', 'getName', 'setName'),
-                new PropertyToMethodCall('Cake\Controller\Controller', 'plugin', 'getPlugin', 'setPlugin'),
-                new PropertyToMethodCall('Cake\Form\Form', 'validator', 'getValidator', 'setValidator'),
+                new PropertyToMethod('Cake\Controller\Controller', 'name', 'getName', 'setName'),
+                new PropertyToMethod('Cake\Controller\Controller', 'plugin', 'getPlugin', 'setPlugin'),
+                new PropertyToMethod('Cake\Form\Form', 'validator', 'getValidator', 'setValidator'),
             ]),
         ]]);
```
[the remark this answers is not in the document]
TomasVotruba: This coudl be renamed even better, to respect node naming.
As Property is not PropertyFetch.

This should be renamed to `PropertyFetchToMethodCallRector`

```diff
-PropertyToMethodCallRector
+PropertyFetchToMethodCallRector
```

TomasVotruba:
> Haven't you thought about some naming convention for the rector rule?

Not yet, but it seems Rector is matured enough to have one.
Could you make that as new issue?

Ideally with PHPStan rule pseudo-algorithm in mind, that would check this.

Issue #4105, opened 2020-09-02
[DX] PHPStan rule to check getNodeTypes() return classes of `PhpParser\Node`


TomasVotruba:
We had this check on runtime, but it's too late to check that.
PHPStan rule is much better for it:

![image](https://user-images.githubusercontent.com/924196/91979988-e1241580-ed26-11ea-9306-3c6c65824241.png)

PR #4210, opened 2020-09-12
[DX] PHPStan rule to check getNodeTypes() return classes of `PhpParser\Node`

dobryy:
Closes: #4105

### DIFF
--- a/utils/phpstan-extensions/src/Rule/CheckGetNodeTypesReturnPhpParserNodeRule.php
+++ b/utils/phpstan-extensions/src/Rule/CheckGetNodeTypesReturnPhpParserNodeRule.php
@@ -1,0 +1,113 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\PHPStanExtensions\Rule;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Name\FullyQualified;
+use PhpParser\Node\Stmt\ClassMethod;
+use PhpParser\Node\Stmt\Return_;
+use PhpParser\NodeFinder;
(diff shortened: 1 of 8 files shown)


dobryy:
@TomasVotruba 
Seems like we have no such issues, that is why I've added the failing case https://github.com/rectorphp/rector/pull/4210/commits/e59dff8dc8af4c599c54ae5bf58ba5909b806afe.
Once you'll find that rule is okay, I'll fix it.

utils/phpstan-extensions/src/Rule/CheckGetNodeTypesReturnPhpParserNodeRule.php
```diff
@@ -0,0 +1,113 @@
...
+
+        $nodeFinder = new NodeFinder();
+        /** @var Return_[] $returns */
+        $returns = $nodeFinder->findInstanceOf($classMethod, Return_::class);
+
+        if (count($returns) === 1) {
+            return false;
+        }
+
+        throw new ShouldNotHappenException('More than one return statement exist');
```
dobryy, line 77: @TomasVotruba, at the moment, we don't have such cases. Do you think I have to add handling of such a situation into this rule or some separate rule?

TomasVotruba: I think exception here is the best solution :+1:

TomasVotruba:
> Once you'll find that rule is okay, I'll fix it.

Very nice, thank you :+1:

TomasVotruba (approved):

dobryy:
Can be merged.

PR #4214, opened 2020-09-13
Fixes #3939 Add IssetOnPropertyObjectToPropertyExists rule

### LINKED ISSUE #3939: [CodeQuality] Isset on property object to property_exists()
Similar to `ArrayKeyExistsOnPropertyRector`

image

samsonasik:
@TomasVotruba this is for #3939. There are some limitation for it, that can cause a bug:

1. like we discussed at https://github.com/rectorphp/rector/issues/3939#issuecomment-685766717 on property has value.
2. multiple passed value currently not detected yet, eg : `isset($this->var, $this->var2);`

I am thinking about how about make : 

```php
isset($this, 'x');
```

to 

```php
property_exists('classname', 'x') && $this->x !== null;
```

What do you thing?


TomasVotruba:
> I am thinking about how about make :

It makes sense :+1: . I would never though of such case. 

Go for it

samsonasik:
@TomasVotruba could you give hint how to apply `BooleanAnd` for value !== null ? For example, I tried like the following:

```php
            return new BooleanAnd(
                new FuncCall(new Name('property_exists'), [new Arg($object), new Arg(new String_($property))]),
                new NotIdentical(new String_('A'), new String_('B'))
            );
```

and got the following output:

```php
property_exists($obj, 'x') && 'A' !== 'B';
```

What I want is to make it like the following: 

```php
property_exists($obj, 'x') && $obj->x !== null;
```

How to replace 'A' with `$obj->x` and make 'B' to null ?

TomasVotruba:
```diff
-new String_('A'),
+new PropertyFetch(new Variable('...'), 'propertyName
```

See https://github.com/rectorphp/rector/blob/master/docs/nodes_overview.md#phpparsernodeexprpropertyfetch

---

For null, there is helper function:

```diff
-new String_('B')
+$this->createNull();
```

See https://github.com/rectorphp/rector/blob/master/docs/nodes_overview.md#phpparsernodeexprconstfetch

samsonasik:
All checks have passed 🎉 . It is implemented. 

Current limitation is it currently can't change multiple passed parameters in isset, eg:

```php
isset($this->x, $this->y);
```

So, it will make only (expected vs actual):

```diff
-        property_exists($this, 'x') && $this->x !== null && property_exists($this, 'y') && $this->y !== null;;
+        property_exists($this, 'x') && $this->x !== null;
```

So, the `y` property is not checked.

@TomasVotruba ready for review

samsonasik:
I've added handle skip multiple parameters inside `isset()`, so the following :

```php
isset($this->x, $this->y);
```

will be skipped. 

All checks have passed 🎉 @TomasVotruba ready for review.

samsonasik:
@TomasVotruba I just pushed addition of `?>` and end of line in fixture tests. The Github actions failures (cs and phpstan) seems unrelated as it previously just works:


Screen Shot 2020-09-15 at 12 03 04 PM
Screen Shot 2020-09-15 at 12 03 19 PM

TomasVotruba:
It's usually new rules I add to Symplify set - **you can skip those**. Just no to forget to apply them here too.
It would be nice to know what rule caused the error in the error formatter, so you know you can skip "Symplify\*" rules.

samsonasik:
I've rebased. I've applied:

- change 'PHP' nowdoc to 'CODE_SAMPLE' nowdoc
- ignore phpstan new errors

All checks have passed 🎉 @TomasVotruba ready for review.

TomasVotruba:
I've added temporary fix of PHPStan on `master` https://github.com/rectorphp/rector/pull/4224

samsonasik:
I've rebased latest master. All checks have passed 🎉 @TomasVotruba ready for review.

rules/code-quality/tests/Rector/Isset_/IssetOnPropertyObjectToPropertyExistsRector/Fixture/property_after_instantiation.php.inc
```diff
@@ -0,0 +1,33 @@
...
+{
+    public $x;
+    public $y;
+}
+
+$obj = new SomeClass2();
+isset($obj->x);
+isset($obj->y);
+isset($obj->x) && isset($obj->y);
+
```
TomasVotruba, line 15: These snippets should be wrapped in class method call, like the one above. This file is included, so this part would get executed and cause unintended changes.

rules/code-quality/tests/Rector/Isset_/IssetOnPropertyObjectToPropertyExistsRector/Fixture/skip_multiple_properties.php.inc
```diff
@@ -0,0 +1,14 @@
...
+class SomeClass3
+{
+    private $x;
+    private $y;
+
+    public function run()
+    {
+        isset($this->x, $this->y);
+    }
+}
```
TomasVotruba, line 14: Just curious, why is this case skipped?

rules/code-quality/tests/Rector/Isset_/IssetOnPropertyObjectToPropertyExistsRector/Fixture/property_after_instantiation.php.inc
samsonasik: fixed.

rules/code-quality/tests/Rector/Isset_/IssetOnPropertyObjectToPropertyExistsRector/Fixture/skip_multiple_properties.php.inc
samsonasik: If not skipped, previously, it will only change first parameter, so only changed to: 

```diff
+ property_exists($this, 'x') && $this->x !== null;
```

It may need to use `NodeVisitorAbstract` like in https://github.com/rectorphp/rector/blob/master/packages/node-type-resolver/src/NodeVisitor/StatementNodeVisitor.php but I've no idea how that works with `Isset_` yet. So the safest way currently is just skip it.

```php
            if ($previous && $previous->getAttribute(AttributeKey::PARENT_NODE) === $current) {
                continue;
            }

            if ($next && $next->getAttribute(AttributeKey::PARENT_NODE) === $current) {
                continue;
            }
```

@TomasVotruba do you have example for how to append multiple `BooleanAnd` expression?

TomasVotruba: I see. We can handle this in 2nd iteration of this rule.

I'll merge this as it is now.

TomasVotruba:
Thank you :+1:

Issue #4004, opened 2020-08-22
[Naming] Make bool property respect is/has/was naming


TomasVotruba:
```diff
-private $full = false;
+private $isFull = false;

 public function isFull()
 {
-    return $this->full;
+    return $this->isFull;
 }
```

PR #4215, opened 2020-09-13
[Naming] Make bool property respect is/has/was naming

dobryy:
Closes: https://github.com/rectorphp/rector/issues/4004

TODO:
- [x] Conflicts
- [x] Decouple copy-pasted logic
- [x] PHP 7.4 test
- [x] Improve property type detection
- [x] Apply Factory

### DIFF
--- a/rules/naming/src/Naming/PropertyNaming.php
+++ b/rules/naming/src/Naming/PropertyNaming.php
@@ -5,14 +5,20 @@ declare(strict_types=1);
 namespace Rector\Naming\Naming;
 
 use Nette\Utils\Strings;
+use PhpParser\Node\Stmt\ClassMethod;
+use PhpParser\Node\Stmt\Property;
+use PhpParser\Node\Stmt\Return_;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\StaticType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeWithClassName;
 use PHPStan\Type\UnionType;
+use Rector\Core\PhpParser\Node\BetterNodeFinder;
 use Rector\Naming\RectorNamingInflector;
 use Rector\Naming\ValueObject\ExpectedName;
 use Rector\NetteKdyby\Naming\VariableNaming;
+use Rector\NodeNameResolver\NodeNameResolver;
+use Rector\NodeTypeResolver\Node\AttributeKey;
 use Rector\PHPStan\Type\SelfObjectType;
 use Rector\PHPStan\Type\ShortenedObjectType;
 use Rector\PHPStanStaticTypeMapper\Utils\TypeUnwrapper;
@@ -36,6 +42,12 @@ final class PropertyNaming
     private const INTERFACE = 'Interface';
 
     /**
+     * @see https://regex101.com/r/RDhBNR/1
+     * @var string
+     */
+    private const PREFIXED_CLASS_METHODS_REGEX = '#^(is|are|was|were|has|have|had|can)[A-Z].+#';
+
+    /**
      * @var TypeUnwrapper
      */
     private $typeUnwrapper;
@@ -45,10 +57,26 @@ final class PropertyNaming
      */
     private $rectorNamingInflector;
 
-    public function __construct(TypeUnwrapper $typeUnwrapper, RectorNamingInflector $rectorNamingInflector)
-    {
+    /**
+     * @var BetterNodeFinder
+     */
+    private $betterNodeFinder;
+
+    /**
+     * @var NodeNameResolver
+     */
+    private $nodeNameResolver;
+
+    public function __construct(
+        TypeUnwrapper $typeUnwrapper,
+        RectorNamingInflector $rectorNamingInflector,
+        BetterNodeFinder $betterNodeFinder,
+        NodeNameResolver $nodeNameResolver
+    ) {
         $this->typeUnwrapper = $typeUnwrapper;
         $this->rectorNamingInflector = $rectorNamingInflector;
+        $this->betterNodeFinder = $betterNodeFinder;
+        $this->nodeNameResolver = $nodeNameResolver;
     }
 
     public function getExpectedNameFromMethodName(string $methodName): ?ExpectedName
@@ -129,6 +157,23 @@ final class PropertyNaming
         $pascalCaseName = str_replace('_', '', ucwords($underscoreName, '_'));
 
         return lcfirst($pascalCaseName);
+    }
+
+    public function getExpectedNameFromBooleanPropertyType(Property $property): ?string
+    {
+        $prefixedClassMethods = $this->getPrefixedClassMethods($property);
+        if ($prefixedClassMethods === []) {
+            return null;
+        }
+
+        $classMethods = $this->filterClassMethodsWithPropertyFetchReturnOnly($prefixedClassMethods, $property);
+
+        if (count($classMethods) !== 1) {
+            return null;
+        }
+
+        $classMethod = reset($classMethods);
+        return $this->nodeNameResolver->getName($classMethod);
     }
 
     private function getClassName(TypeWithClassName $typeWithClassName): string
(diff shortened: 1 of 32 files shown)


rules/naming/tests/Rector/Property/MakeBoolPropertyRespectIsHasWasMethodNamingRector/Fixture/skip_method_name_after_prefix_not_matches_property_name.php.inc
```diff
@@ -0,0 +1,19 @@
...
+    private $full = false;
+
+    public function isSomething()
+    {
+        return $this->full;
+    }
+
+}
+
+?>
```
dobryy, line 19: @TomasVotruba, before finishing I came up with this test fixture. Does this rule meant to rename properties in such cases? When there is nothing similar in naming but the method return the property value.

rules/naming/src/Rector/Property/MakeBoolPropertyRespectIsHasWasMethodNamingRector.php
```diff
@@ -0,0 +1,216 @@
...
+        $phpDocInfo = $property->getAttribute(AttributeKey::PHP_DOC_INFO);
+        if ($phpDocInfo === null) {
+            return false;
+        }
+
+        if (! $phpDocInfo->getVarType() instanceof BooleanType) {
+            return false;
+        }
+
+        return true;
```
dobryy, line 148: @TomasVotruba Is there any better way to figure out the property type? I guess, that current logic doesn't support PHP 7.4 property type. Also, I believe that it is possible to detect property type by value, not sure whether it is a good idea

dobryy:
This rule is not yet finished. I created PR to ask some related questions.

rules/naming/src/Rector/Property/MakeBoolPropertyRespectIsHasWasMethodNamingRector.php
TomasVotruba: What about `$this->getStaticType()/getObjectType()` methods?

rules/naming/tests/Rector/Property/MakeBoolPropertyRespectIsHasWasMethodNamingRector/Fixture/skip_method_name_after_prefix_not_matches_property_name.php.inc
TomasVotruba: What do you think most people expect to happen in this case?

src/PhpParser/Node/BetterNodeFinder.php
```diff
@@ -167,7 +167,7 @@ public function hasVariableOfName($nodes, string $name): bool
     /**
      * @param Node|Node[] $nodes
      */
-    public function findClassMethodOfName($nodes, string $name): ?ClassMethod
+    private function findClassMethodOfName($nodes, string $name): ?ClassMethod
```
dobryy: This method is not used outside of class but I'd like to keep it public for further usage. @TomasVotruba, is there any way to skip it?

rules/naming/src/Rector/Property/MakeBoolPropertyRespectIsHasWasMethodNamingRector.php
dobryy: If I remember correctly, I tried both and it didn't work but I'll give it a try tomorrow.

rules/naming/tests/Rector/Property/MakeBoolPropertyRespectIsHasWasMethodNamingRector/Fixture/skip_method_name_after_prefix_not_matches_property_name.php.inc
dobryy: I wish I could know... I wouldn't ask if I knew the answer to this question :)
Maybe it is better to rename it in such cases, even though I'll need to re-write the rule to handle this case :D

src/PhpParser/Node/BetterNodeFinder.php
TomasVotruba: Why make it `public` if it's not used?

rules/naming/tests/Rector/Property/MakeBoolPropertyRespectIsHasWasMethodNamingRector/Fixture/skip_method_name_after_prefix_not_matches_property_name.php.inc
TomasVotruba: I don't know either. I'd have to ask everyone too :) I just don't know much about rule, since I haven't made it.

> Maybe it is better to rename it in such cases, even though I'll need to re-write the rule to handle this case :D

That's what I'd expect. I also try to test it on Rector to get more data, if such assumption is right. It usually helps to decide better

rules/naming/src/Rector/Property/MakeBoolPropertyRespectIsHasWasMethodNamingRector.php
TomasVotruba: I'd check other rules, that also use `$this->getStaticType()/getObjectType()` on `$property`.
There might be some tip/trick

rules/naming/tests/Rector/Property/MakeBoolPropertyRespectIsHasWasMethodNamingRector/Fixture/skip_method_name_after_prefix_not_matches_property_name.php.inc
dobryy: Nothing special after this change in Rector itself but since we'd both expect that the rule will rename the property in such cases, I'm leaving it as it is now.

src/PhpParser/Node/BetterNodeFinder.php
dobryy: After some changes, it looks like we don't need both of the added methods that is why I removed them. Nevertheless, I consider such classes as some "helper" libs that have some API. Maybe some methods are not used at the moment even inside the class but they can be helpful for someone else.

I know your answer and more or less agree with you :D

rules/naming/src/Rector/Property/MakeBoolPropertyRespectIsHasWasMethodNamingRector.php
```diff
@@ -0,0 +1,200 @@
...
+
+        $classLike = $property->getAttribute(AttributeKey::CLASS_NODE);
+        if ($classLike === null) {
+            return [];
+        }
+
+        $classMethods = $this->betterNodeFinder->findInstanceOf($classLike, ClassMethod::class);
+        return array_filter($classMethods, function (ClassMethod $classMethod): bool {
+            $classMethodName = $this->getName($classMethod);
+            return Strings::match($classMethodName, '#^(is|was|has)[A-Z].+#') !== null;
```
mssimi, line 125: Maybe "can" too?

dobryy: I was also thinking about "are", 'were", "have" and 'had'. Maybe it is better to use some configuration param for it...

@TomasVotruba, what do you think?

TomasVotruba: No need to configure it now, it would be too much detailed.
I think regex should be enough.

dobryy:
This PR still remains in progress :)

dobryy:
@TomasVotruba can you please advise what is wrong with CI right now?

dobryy:
Ready for review

TomasVotruba:
> @TomasVotruba can you please advise what is wrong with CI right now?

It's in the `details` link. The Rector is failing. Not sure why it's not fixing it

dobryy:
> > @TomasVotruba can you please advise what is wrong with CI right now?
> 
> It's in the `details` link. The Rector is failing. Not sure why it's not fixing it.
> 
> Manual update is needed now.
 
I checked it before but I was wondering why it is not fixing. I had the same problem with my previous PR :(

TomasVotruba:
> I had the same problem with my previous PR :(

Me too. Not sure when that changed :/

rules/naming/src/Naming/PropertyNaming.php
```diff
@@ -243,6 +282,65 @@ private function removeInterfaceSuffixPrefix(string $className, string $shortNam
...
+
+        $classLike = $property->getAttribute(AttributeKey::CLASS_NODE);
+        if ($classLike === null) {
+            return [];
+        }
+
+        $classMethods = $this->betterNodeFinder->findInstanceOf($classLike, ClassMethod::class);
+        return array_filter($classMethods, function (ClassMethod $classMethod): bool {
+            $classMethodName = $this->nodeNameResolver->getName($classMethod);
+            return Strings::match($classMethodName, '#^(is|are|was|were|has|have|had|can)[A-Z].+#') !== null;
```
TomasVotruba, line 303: Could you extract the 2nd arguments as regex constant with https://regex101.com/ link? Just so we have it covered.

It's in `phpstan.neon`, but ignored now. I'll make sure it's not ignored :)

TomasVotruba:
I've added 1 comment. After you resolve it, it's ready to merge

rules/naming/src/Naming/PropertyNaming.php
dobryy: Ready

Issue #4218, opened 2020-09-14
Rector doesn't work when installed using --prefer-lowest


tmotyl:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.8 (master)              |
| Installed as   | git clone                     |

Rector tests are red when installed with `composer --prefer-lowest`, as rector uses API which is available in newest packaes. 
One example is symfony/dependency injection. Recent rector allow `symfony/dependency-injection: ^4.4.8|^5.0.6` and

`Symfony\Component\DependencyInjection\Loader\Configurator\inline_service` is available only since Symfony 5.1 (before the method has different name if I read it correctly the 5.1 changelog: https://github.com/symfony/dependency-injection/blob/34eefcfff4ff4e628f6571f4460b0825e5e7e47b/CHANGELOG.md#510)

So the test is failing with
```
2) Rector\MysqlToMysqli\Tests\SetTest::test with data set #0 (Symplify\SmartFileSystem\SmartFileInfo Object (...))
Error: Call to undefined function Symfony\Component\DependencyInjection\Loader\Configurator\inline_service()
```

## Minimal PHP Code Causing Issue

```
git clone git@github.com:rectorphp/rector.git
cd rector
composer update --prefer-lowest
./vendor/bin/phpunit 
```

## Expected Behaviour

Tests are green.
Rector is tested in CI (github actions ) also with lowest dependencies

TomasVotruba:
Thanks for very detailed report. 
With such a clear and narrow error message, I could find the fix in matter of 1 minute.

Does this help?
https://github.com/rectorphp/rector/pull/4219

tmotyl:
Hi
@TomasVotruba it did help, however there is (I believe) similar issue with other function:

115) Rector\TypeDeclaration\Tests\Rector\FunctionLike\ReturnTypeDeclarationRector\ReturnTypeDeclarationRectorTest::test with data set #8 (Symplify\SmartFileSystem\SmartFileInfo Object (...))
BadMethodCallException: No such method: isAOf

I think the method isAOf comes from "webmozart/assert": "^1.0", however it was introduced in webmozart/assert 1.7:

![image](https://user-images.githubusercontent.com/515397/93098754-82a45300-f6a7-11ea-8d7a-c32b0cd71d63.png)

TomasVotruba:
I think @staabm solved this in
https://github.com/rectorphp/rector/pull/4220

[...]

PR #4220, opened 2020-09-14
Github Actions: add a job which tests the lowest supported versions

staabm:
in https://github.com/rectorphp/rector/issues/4218 there was identified, that rector cannot be installed properly with lowest dependency resolution.

this PR adds automated coverage to prevent regressing the lower version boundary

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -12,7 +12,10 @@ jobs:
         strategy:
             matrix:
                 php: ['7.2', '7.3', '7.4']
-                dependencies: [lowest, locked]
+                dependencies: [locked]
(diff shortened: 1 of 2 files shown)


staabm:
with this PR we can see that there are even more problems when running with `--prefer-lowest`.

either the lower bound should be raised, or the errors fixed.

TomasVotruba:
Could you include fix as well?
CI is now failing

staabm:
ohh .. this was easier then expected.

I think we are good to go @TomasVotruba :-)

TomasVotruba:
You know more than you think ;)

Thanks, purfect :)

Issue #4216, opened 2020-09-14
[Naming] Foreach over "data", renames to "datum"


TomasVotruba:
@dobryy Hi, we have a bug in `RenameForeachValueVariableToMatchMethodCallReturnTypeRector`
Could you fix it with priority? It's blocking for release.

![image](https://user-images.githubusercontent.com/924196/93058588-e3179e00-f66f-11ea-8689-33925b7b6f70.png)

## Now

```diff
-foreach ($vendorData->getPackagesData() as $packageData) {
+foreach ($vendorData->getPackagesData() as $packageDatum) {
```

## Expected 

(no change)

```diff
foreach ($vendorData->getPackagesData() as $packageData) {
```

As some nouns are always plural, the plural "s" has to be on previous word:

```
packagesData
packageData
```

PR #4221, opened 2020-09-14
[Naming] Foreach over "data", renames to "datum"

dobryy:
Closes: https://github.com/rectorphp/rector/issues/4216

### DIFF
--- a/rules/naming/src/Naming/PropertyNaming.php
+++ b/rules/naming/src/Naming/PropertyNaming.php
@@ -21,6 +21,7 @@ use Rector\PHPStanStaticTypeMapper\Utils\TypeUnwrapper;
  * @deprecated
  * @todo merge with very similar logic in
  * @see VariableNaming
+ * @see \Rector\Naming\Tests\Naming\PropertyNamingTest
  */
 final class PropertyNaming
 {
@@ -60,7 +61,7 @@ final class PropertyNaming
 
         $originalName = lcfirst($matches[1]);
 
-        return new ExpectedName($originalName, $this->inflector->singularize($originalName));
+        return new ExpectedName($originalName, $this->singularize($originalName));
     }
 
     public function getExpectedNameFromType(Type $type): ?ExpectedName
@@ -103,7 +104,7 @@ final class PropertyNaming
 
         // prolong too short generic names with one namespace up
         $originalName = $this->prolongIfTooShort($shortClassName, $className);
-        return new ExpectedName($originalName, $this->inflector->singularize($originalName));
+        return new ExpectedName($originalName, $this->singularize($originalName));
     }
 
     /**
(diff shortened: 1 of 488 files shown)


rules/naming/src/Naming/PropertyNaming.php
```diff
@@ -44,10 +45,10 @@ final class PropertyNaming
      */
     private $inflector;
 
-    public function __construct(TypeUnwrapper $typeUnwrapper, Inflector $inflector)
+    public function __construct(TypeUnwrapper $typeUnwrapper, RectorNamingInflector $rectorNamingInflector)
     {
         $this->typeUnwrapper = $typeUnwrapper;
-        $this->inflector = $inflector;
+        $this->inflector = $rectorNamingInflector->build();
```
dobryy, line 51: @TomasVotruba, I didn't find (quickly) any better and at the same time easy way to decouple the logic for the creation of customized Inflector. If it looks way too bad from your point of you, I can invest more time into it.

rules/naming/src/RectorNamingInflector.php
```diff
@@ -0,0 +1,31 @@
...
+
+final class RectorNamingInflector
+{
+    public function build(): Inflector
+    {
+        // @see https://www.doctrine-project.org/projects/doctrine-inflector/en/2.0/index.html
+        return InflectorFactory::createForLanguage(Language::ENGLISH)
+            ->withSingularRules(
+                new Ruleset(
+                    new Transformations(new Transformation(new Pattern('/^packagesData$/i'), 'packageData')),
```
dobryy, line 25: BTW, these exceptions could be a rule setting in the future :)

dobryy: or... Set setting.

TomasVotruba: Could you add test for this class? Just to confirm its doing the job right

TomasVotruba: Could you cover " anythingData" in tests too?

dobryy: In order to make "anythingData" work, I had to change the logic.

dobryy:
@TomasVotruba, ready for review.

dobryy:
BTW, I guess that PHPStan issues are caused by some package update and I don't feel like a boy scout when it comes to 300+ errors :)
I also tried to lock versions of some composer packages but it didn't help.

TomasVotruba:
> BTW, I guess that PHPStan issues are caused by some package update and I don't feel like a boy scout when it comes to 300+ errors :)

I agree. It's more like "a superman scout" ;)

It's usually new rules I add to Symplify set. Just no to forget to apply them here too.
It would be nice to know what rule caused the error in the error formatter, so you know you can skip "Symplify\*" rules.

TomasVotruba:
The feature itself looks good, thanks for more detailed test  :+1:

dobryy:
> > BTW, I guess that PHPStan issues are caused by some package update and I don't feel like a boy scout when it comes to 300+ errors :)
> 
> I agree. It's more like "a superman scout" ;)
> 
> It's usually new rules I add to Symplify set. Just no to forget to apply them here too.
> It would be nice to know what rule caused the error in the error formatter, so you know you can skip "Symplify*" rules.

I noticed the error message of the new rule, is it a common approach that you use in projects? I mean do use factories in all cases instead of plain object creation?

TomasVotruba:
This project is special, so all `*/Node/*` classes should be excluded.
But otherwise the repeated object creation should be decoupled to services. 

Same way logic should not be in Controller/Command/EventDispatcher, but it own service that are only called there.

Issue #4071, opened 2020-08-30
[EarlyReturn] Change if && to early return


TomasVotruba:
![image](https://user-images.githubusercontent.com/924196/91659453-e89db180-eacf-11ea-8f25-4997e98646c0.png)

TomasVotruba:
![image](https://user-images.githubusercontent.com/924196/91876339-df9e1300-ec7c-11ea-8b04-6a9816c4cd11.png)

↓

![Uploading image.png…]()

TomasVotruba:
![rector_to_early_return](https://user-images.githubusercontent.com/924196/91888122-0c0d5b80-ec8c-11ea-8784-94e6ebe9123a.png)

dobryy:
Should be a new early-return set? I've noticed that we already have some early return rules in SOLID set.

TomasVotruba:
I'd go with SOLID then

samsonasik:
It seems already implemented at https://github.com/rectorphp/rector/pull/4395 . I am closing it.

PR #4231, opened 2020-09-16
Fixes #4229 Add property camel case naming rule

### LINKED ISSUE #4229: [Naming] Add property camel case naming rule
```diff
-private $some_property;
+private $someProperty;


// ...

-$this->some_property
+$this->someProperty
```

samsonasik:


samsonasik:

samsonasik:
I just tried to use fixture that property has initial value, and it removed : 

```diff
-    private $someProperty = 1;
+    private $someProperty;
```

Need to check if there is a way to keep existing value if initial property has value.

samsonasik:
All checks have passed again 🎉 . I've added handle to keep existing inital value of property if any. 

@TomasVotruba it is ready for review.

TomasVotruba (commented):
Thank you!

What about property that is in 3rd party code?
It should not be renamed.

TomasVotruba:
Ah, I just checked and we already have this rule: https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#underscoretocamelcasepropertynamerector

My bad, I'm sorry :facepalm: 

Could you check it's the expected behaviour?

samsonasik:
`UnderscoreToCamelCasePropertyNameRector` seems already handle it, I'm closing it.

TomasVotruba:
Thanks for checking

dobryy:
> Ah, I just checked and we already have this rule: https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#underscoretocamelcasepropertynamerector
> 
> My bad, I'm sorry 
> 
> Could you check it's the expected behaviour?

I'm wondering that this rule renames `public` and `protected` properties, also it doesn't solve any type of conflict. Am I missing something?

TomasVotruba:
> I'm wondering that this rule renames public and protected properties, also it doesn't solve any type of conflict. Am I missing something?

No, you're not. The feature is missing :) would you handle it?

> Also... don't we want to move this rule to the Naming set?

Good idea, go for it :+1:

dobryy:
Ok, I'll check it.
At the moment, I'm finishing the decoupling of PropertyRenamer from one rule where everything is handled pretty well. So it shouldn't be a problem to apply for this rule as well.

dobryy:
It looks like I won't get to https://github.com/rectorphp/rector/issues/4071 this week then :) or I'll get to it at the very end of the week.

dobryy:
One more rule that doesn't handle anything https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#renamepropertyrector

TomasVotruba:
> One more rule that doesn't handle anything

What do you mean? It's used here:

https://github.com/rectorphp/rector/blob/3442308bba6cfbb4d5e1a508dd8fe9c897a71419/config/set/cakephp34.php#L38-L43

dobryy:
Sorry, I meant it also doesn't handle conflicts and could cause problems with public and protected properties but maybe for this particular rule it is okay because it is up to the developer what he is renaming.

TomasVotruba:
I see. Could you create standalone issue so we have this knowledge one place?

PR #4237, opened 2020-09-16
Add Proccessed file Caching

mssimi:


rector-ci.php
```diff
@@ -61,4 +61,5 @@
 
     # so Rector code is still PHP 7.2 compatible
     $parameters->set(Option::PHP_VERSION_FEATURES, '7.2');
+    $parameters->set(Option::ENABLE_CACHE, true);
```
mssimi, line 64: @TomasVotruba Not sure here, we can turn that on, but dead code is being skipped.

TomasVotruba: Sure we can :) 
I'd make it `false` by default before merging.

How about warning people in command line about the `dead code` being skipped on cache?

mssimi: There already is warning, but it just says number of rectors skipped, do you want to warn with list of rectors being skipped?

mssimi: It is false by default even if I just remove it, should I rather keep it with false or just remove it?

TomasVotruba: Ah, great :)  I confused this with internal config. All good then :+1:

TomasVotruba:
Thank you :+1:

PR #4241, opened 2020-09-16
[Docs] update docs and type check

TomasVotruba:


utils/node-documentation-generator/snippet/new_property.php.inc
```diff
@@ -0,0 +1,10 @@
+<?php
+
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\Property;
+use PhpParser\Node\Stmt\PropertyProperty;
+
+$propertyProperty = new PropertyProperty(new Identifier('propertyName'));
+
+return new Property(Class_::MODIFIER_PUBLIC, [$propertyProperty]);
```
TomasVotruba, line 10: @mssimi So the goal is to create a 1 or 2 snippets per node. The smaller code the better, but also so it's practical.

Each snippet has to return the node. That's how it gets to documentation :)


This one return `PhpParser\Node\Stmt\Property` so it will be bounded in docs to `PhpParser\Node\Stmt\Property;`

The `*php.inc` suffix is needed, so PHPStan and ECS don't care about this file.

TomasVotruba: The best would be to start with the most common nodes like:

- `ClassMethod`
- `ClassConst`
- `Property`
- `ConstFetch`
- `PropertyFetch`
- `ClassConstFetch`

...

PR #4242, opened 2020-09-16
[Naming] Decouple property renaming into PropertyRenamer

dobryy:


rules/naming/src/Rector/Class_/RenamePropertyToMatchTypeRector.php
```diff
@@ -123,69 +120,38 @@ private function refactorClassProperties(ClassLike $classLike): void
...
-
             $expectedName = $this->expectedNameResolver->resolveForPropertyIfNotYet($property);
             if ($expectedName === null) {
                 continue;
             }
 
-            if ($this->shouldSkipProperty($property, $oldName, $expectedName, $conflictingPropertyNames)) {
+            $currentName = $this->getName($property);
+            $propertyType = $this->getObjectType($property);
+            $propertyClassLike = $property->getAttribute(AttributeKey::CLASS_NODE);
```
dobryy, line 130: Looks here is a good place to use Factory, maybe some logic could be implemented in Value Object's getters.
I'd prefer to implement Factory when I'll apply PrepoertyRenamer and ValueObject for other rules as a separate PR.

rules/naming/src/PropertyRenamer.php
```diff
@@ -0,0 +1,71 @@
...
+        $this->breakingVariableRenameGuard = $breakingVariableRenameGuard;
+    }
+
+    public function rename(PropertyRename $propertyRename): void
+    {
+        if ($this->breakingVariableRenameGuard->shouldSkipProperty($propertyRename)) {
+            return;
+        }
+
+        $onlyPropertyProperty = $propertyRename->getProperty()->props[0];
```
dobryy, line 46: One more candite to be added into ProperyRename Value Object.

rules/naming/src/Guard/BreakingVariableRenameGuard.php
```diff
@@ -118,10 +120,13 @@ public function shouldSkipVariable(
...
-    public function shouldSkipProperty(Property $property, string $currentName): bool
+    public function shouldSkipProperty(PropertyRename $propertyRename): bool
     {
-        $propertyType = $this->nodeTypeResolver->resolve($property);
-        return $this->isDateTimeAtNamingConvention($propertyType, $currentName);
+        if ($this->isRamseyUuidInterface($propertyRename->getProperty())) {
+            return true;
+        }
+
+        return $this->isDateTimeAtNamingConvention($propertyRename->getProperty());
```
dobryy, line 129: I wanted to pass ValueObject into these methods as well, it makes, perfectly, sense for `isDateTimeAtNamingConvention` but to make this I need to make some additional changes in RenameParamToMatchTypeRector.php.

If it makes to you (to pass ValueObect to these methods) I'd prefer to do this in separate PR.

rules/naming/src/Rector/Class_/RenamePropertyToMatchTypeRector.php
TomasVotruba: > I'd prefer to implement Factory when I'll apply PrepoertyRenamer and ValueObject for other rules as a separate PR.

:+1:

rules/naming/src/Guard/BreakingVariableRenameGuard.php
TomasVotruba: > If it makes to you (to pass ValueObect to these methods) I'd prefer to do this in separate PR.

Sure :+1: 

Yet it should be somehow standardize in naming-related method. I was actually surprised, that property name resolved once into object and once into string.

dobryy:
@TomasVotruba I decoupled the property renaming logic but without conflict resolution for now. I wanted to keep the PR as small as possible.

Let me know if it looks fine to you. I already tried this change in the new akeBoolPropertyRespectIsHasWasMethodNamingRector and it works like a charm.

rules/naming/src/PropertyRenamer.php
```diff
@@ -0,0 +1,71 @@
...
+use Rector\Core\PhpParser\NodeTraverser\CallableNodeTraverser;
+use Rector\Core\Rector\AbstractRector\NameResolverTrait;
+use Rector\Core\Rector\AbstractRector\NodeTypeResolverTrait;
+use Rector\Naming\Guard\BreakingVariableRenameGuard;
+use Rector\Naming\ValueObject\PropertyRename;
+
+final class PropertyRenamer
+{
+    use NameResolverTrait;
+    use NodeTypeResolverTrait;
```
TomasVotruba, line 20: These should be used only in `abstract class AbstractRector`.
Anywhere else DI should be used

rules/naming/src/ValueObject/PropertyRename.php
```diff
@@ -0,0 +1,76 @@
...
+
+    public function getPropertyType(): Type
+    {
+        return $this->propertyType;
+    }
+
+    public function getClassLike(): ?ClassLike
+    {
+        return $this->classLike;
+    }
```
TomasVotruba, line 75: In what case the property can be without Class_?

TomasVotruba (commented):
Looks good, I left my comments

rules/naming/src/PropertyRenamer.php
dobryy: Yeah, that is something I want to ask about, one of them is not even used btw...
This because of `$this->isLocalPropertyFetchNamed`, it is present only in trait and it seems like there is no something similar in `NodeNameResolver`.

rules/naming/src/ValueObject/PropertyRename.php
dobryy: `$property->getAttribute(AttributeKey::CLASS_NODE)` can return null and in case, for some reason, it will be null it will skip renaming in PropertyFetch 
https://github.com/rectorphp/rector/pull/4242/files#diff-397302fe4eb22d04e3248d9f60054cbcR53-R55

Hm... I'll make it non-nullable and add docblock that it is just a ClassLike...

rules/naming/src/PropertyRenamer.php
TomasVotruba: What about move `isLocalPropertyFetchNamed()` to `NodeNameResolver`?

rules/naming/src/ValueObject/PropertyRename.php
TomasVotruba: If it's not null on purpose, maybe it should be skipped ealier before full rename.

Also, throwing `ShouldNotHappenException` makes sense here.

TomasVotruba: > make it non-nullable

:+1:

rules/naming/src/PropertyRenamer.php
dobryy: Is it okay in your opinion to move this method to `NodeNameResolver`?

TomasVotruba: Yes. Basically traits should only delegate, never make their own method. It creates crap design.

That was just my laziness and absence of PHPStan rule :)

rules/naming/src/Rector/Class_/RenamePropertyToMatchTypeRector.php
TomasVotruba: We could add those factoriies to `ValueObjectFactory` namespace :+1:

rules/naming/src/PropertyRenamer.php
dobryy: Done

rules/naming/src/ValueObject/PropertyRename.php
dobryy: Done

dobryy:
@TomasVotruba is there any scout that is already solving new PHPStan issues? :)

TomasVotruba:
If so, it would be here :) https://github.com/rectorphp/rector/pulls

Do I understand it correctly, this is ready to merge?

dobryy:
> Looks good, I left my comments

Ready, can you please take a look and merge the PR?

TomasVotruba:
Thank you :+1:

PR #4253, opened 2020-09-18
[Naming] Decouple conflict resolution

dobryy:


rules/naming/src/PropertyRenamer.php
```diff
@@ -40,15 +40,17 @@ public function __construct(
...
         if ($this->breakingVariableRenameGuard->shouldSkipProperty($propertyRename)) {
-            return;
+            return false;
         }
 
         $onlyPropertyProperty = $propertyRename->getProperty()->props[0];
         $onlyPropertyProperty->name = new VarLikeIdentifier($propertyRename->getExpectedName());
         $this->renamePropertyFetchesInClass($propertyRename);
+
+        return true;
```
TomasVotruba, line 53: Return book on non bool method cannot be seen from name. Any idea how to decouple this?

dobryy: I can't say I like this approach but check this commit out 61cb7e4b6483c3acdc672542adf78807abce55e4

TomasVotruba: That create space for bugs, since the value is set once `true` for all following cases.

This part is tricky. I wonder if the `RenamePropertyToMatchTypeRector` rule could work with subscribe to `Property` and `PropertyFetch` instead.

rules/naming/src/PropertyRenamer.php
```diff
@@ -42,13 +47,22 @@ public function __construct(
 
     public function rename(PropertyRename $propertyRename): void
```
TomasVotruba, line 48: Maybe a conventions that we use in the rest of code could be applied here.

- If the `Property` is renamed, it should return it.
- If not, the `null`

What do you think?

rules/naming/src/PropertyRenamer.php
dobryy: At the beginning of the method, there is `$this->isRenamed = false` to prevent this kind of bugs but anyway, some sort of bugs could appear.

rules/naming/src/PropertyRenamer.php
dobryy: Yeah, makes sense! My brain was in the trap of old logic in the rule.

dobryy: Ready

rules/naming/src/ValueObject/PropertyRename.php
```diff
@@ -8,7 +8,7 @@
 use PhpParser\Node\Stmt\Property;
 use PHPStan\Type\Type;
 
-final class PropertyRename extends Property
+final class PropertyRename
```
TomasVotruba, line 11: Ha, I haven't noticed this. What was it for?

TomasVotruba:
Thank you :+1:

rules/naming/src/ValueObject/PropertyRename.php
dobryy: It was a surprise for me as well... I somehow by mistake extended Property...

Issue #4239, opened 2020-09-16
Not recognized syntax: $model


Aerendir:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | v0.8.6              |
| Installed as   | composer dependency                      |

> Could not process "tests/Model/ModelSmokeTest.php" file, due to:                                               
>         "Pick more specific node than "PhpParser\Node\Expr\MethodCall", e.g. "$node->name"                             
>
> Caused in "tests/Model/ModelSmokeTest.php" file:88 on code "$method->getName()"                                
>
> Look at                                                                                                        
>         vendor-bin/phpstan/vendor/rector/rector/rules/php70/src/Rector/FuncCall/NonVariableToVariableOnFunctionCallRect
>         or.php:76". 

## Minimal PHP Code Causing Issue

Unrecognized syntax:

```php
$model->{$method->getName()}();
```

https://getrector.org/demo/cfe97677-72df-43c9-b250-e831e4e65900

## Expected Behaviour

The syntax is recognized.

## Workaround

Use a more verbose syntax:

```php
$method = $method->getName();
$model->$method();
```

TomasVotruba:
Thanks for reporting.

Could you make the demo minimal code that fails? It should be round 2-3 lines that causes this.
It's hard to read now.

mssimi:
@TomasVotruba I will take it, how can I assign myself?

TomasVotruba:
:+1: 

I usually use "Control + F" on the page I'm on.

![image](https://user-images.githubusercontent.com/924196/93532326-180e4400-f941-11ea-811c-2f934295a165.png)

TomasVotruba:
Btw, could you finish the test of cache invalidation first? Just so we have just one thign at a time

mssimi:
Ok, I see it now, I had no right. Let's call about cache later.

Aerendir:
> Could you make the demo minimal code that fails? It should be round 2-3 lines that causes this.
It's hard to read now.

@TomasVotruba , done: now should be more readable...

PR #4254, opened 2020-09-18
node name resolver skip on identifier

mssimi:
Closes: #4239

### DIFF
--- a/rules/php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/keep_curly_braces.php.inc
+++ b/rules/php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/keep_curly_braces.php.inc
@@ -2,12 +2,12 @@
 
 namespace Rector\Php70\Tests\Rector\FuncCall\NonVariableToVariableOnFunctionCallRector\Fixture;
 
-final class ModelSmokeTest
+final class KeepCurlyBraces
 {
     /** @var array|null */
     private $models;
 
-    public function testGetters(): void
+    public function run(): void
     {
         foreach ($this->models as $model) {
             $methods = (new \ReflectionClass($model))->getMethods();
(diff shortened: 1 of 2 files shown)


rules/php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/keep_curly_braces.php.inc
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Rector\Php70\Tests\Rector\FuncCall\NonVariableToVariableOnFunctionCallRector\Fixture;
+
+final class ModelSmokeTest
+{
+    /** @var array|null */
+    private $models;
+
+    public function testGetters(): void
```
TomasVotruba, line 10: This looked as test at first sight. To me a while to found out it's a fixture :D

Better use dummy name `run()` as in other tests

rules/php70/tests/Rector/FuncCall/NonVariableToVariableOnFunctionCallRector/Fixture/keep_curly_braces.php.inc
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Rector\Php70\Tests\Rector\FuncCall\NonVariableToVariableOnFunctionCallRector\Fixture;
+
+final class ModelSmokeTest
```
TomasVotruba, line 5: The class should be named with same words as in file.

Hm... I know it's not much obvious.
 Could you make a CI rule for that?

mssimi: I will make CI rule in new task.

TomasVotruba:
Thanks :+1:

PR #4255, opened 2020-09-18
[Naming] Factory to create PropertyRename

dobryy:


TomasVotruba (changes requested):

rules/naming/src/ValueObjectFactory/PropertyRenameFactory.php
```diff
@@ -0,0 +1,70 @@
...
+
+use PhpParser\Node\Stmt\Property;
+use Rector\Core\Exception\ShouldNotHappenException;
+use Rector\Naming\Naming\ExpectedNameResolver;
+use Rector\Naming\ValueObject\PropertyRename;
+use Rector\NodeNameResolver\NodeNameResolver;
+use Rector\NodeTypeResolver\Node\AttributeKey;
+use Rector\NodeTypeResolver\NodeTypeResolver;
+
+final class PropertyRenameFactory
```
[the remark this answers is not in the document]
dobryy: I've just tried to create something simple but for this, I have to create Property with all nested nodes (PropertyProperty, VarLikeIndentifier, blah blah blah), parent ClassLike node, and PhpDocInfo which looks even more complicated with all its constructor params :)

That is why I'd like to ask whether we have some examples of how to load the fixture. parse it using PhpParser and then to use the PhpParser nodes in the test? Or maybe you can advise some better approach.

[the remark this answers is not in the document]
TomasVotruba: That's a good question!

You could re-use this service:
https://github.com/rectorphp/rector/blob/e0cb8c228b472809c5f793deb557c3b5505ac1f0/packages/file-system-rector/src/Parser/FileInfoParser.php#L30-L48

Then use `$betterNodeFinder->findInstanceOf()` to get your node.

rules/naming/src/ValueObject/PropertyRename.php
```diff
@@ -26,27 +26,27 @@ final class PropertyRename
     private $property;
 
     /**
-     * @var Type
+     * @var ClassLike
      */
-    private $propertyType;
```
dobryy, line 31: PropertyType was removed because it is unused at the moment. It will be used in `BreakingVariableRenameGuard::isRamseyUuidInterface` and `BreakingVariableRenameGuard::isDateTimeAtNamingConvention` once I'll get  Param renaming.
For reference https://github.com/rectorphp/rector/pull/4242#discussion_r489739962

rules/naming/tests/ValueObjectFactory/PropertyRenameFactory/PropertyRenameFactoryTest.php
```diff
@@ -0,0 +1,79 @@
...
+    public function testCreate(Property $property, array $expectedPropertyData): void
+    {
+        /** @var PropertyRename $actualPropertyRename */
+        $actualPropertyRename = $this->propertyRenameFactory->create($property);
+
+        $this->assertSame($property, $actualPropertyRename->getProperty());
+        $this->assertSame($expectedPropertyData['expectedName'], $actualPropertyRename->getExpectedName());
+        $this->assertSame($expectedPropertyData['currentName'], $actualPropertyRename->getCurrentName());
+        $this->assertInstanceOf(PropertyProperty::class, $actualPropertyRename->getPropertyProperty());
+        $this->assertInstanceOf(ClassLike::class, $actualPropertyRename->getClassLike());
```
dobryy, line 60: To be honest, I'm not sure how it could be tested better. I'm opened to propositions :)

rules/naming/tests/ValueObjectFactory/PropertyRenameFactory/PropertyRenameFactoryTest.php
```diff
@@ -0,0 +1,79 @@
...
+        ];
+    }
+
+    private function getPropertyFromFile(string $string): ?Node
+    {
+        $smartFileInfo = new SmartFileInfo($string);
+        $nodes = $this->fileInfoParser->parseFileInfoToNodesAndDecorate($smartFileInfo);
+        return $this->betterNodeFinder->findFirstInstanceOf($nodes, Property::class);
+    }
+}
```
dobryy, line 79: What is the correct approach here. To specify that the return type is `Property` and to introduce a new variable with docblock or current version is okay?

rules/naming/src/ValueObjectFactory/PropertyRenameFactory.php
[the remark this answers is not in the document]
dobryy: Ready and I added some comments/questions.

dobryy:
F*ck, again scout is needed :D

rules/naming/tests/ValueObjectFactory/PropertyRenameFactory/PropertyRenameFactoryTest.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public function createDataProvider(): Iterator
+    {
+        yield [
+            $this->getPropertyFromFile(__DIR__ . '/Fixture/SomeClass.php'), [
+                'expectedName' => 'eliteManager',
+                'currentName' => 'eventManager',
+            ],
+        ];
```
TomasVotruba, line 70: This is a way to go :+1: 

Maybe instead of array, the direct parameter can be used

```php
yield [
    $this->getPropertyFromFile(__DIR__ . '/Fixture/SomeClass.php'),
    'eliteManager',
    'eventManager',
]
```

rules/naming/tests/ValueObjectFactory/PropertyRenameFactory/PropertyRenameFactoryTest.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    private function getPropertyFromFile(string $string): ?Node
+    {
+        $smartFileInfo = new SmartFileInfo($string);
+        $nodes = $this->fileInfoParser->parseFileInfoToNodesAndDecorate($smartFileInfo);
+        return $this->betterNodeFinder->findFirstInstanceOf($nodes, Property::class);
```
TomasVotruba, line 77: :+1:

rules/naming/tests/ValueObjectFactory/PropertyRenameFactory/PropertyRenameFactoryTest.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        yield [
+            $this->getPropertyFromFile(__DIR__ . '/Fixture/SomeClass.php'), [
+                'expectedName' => 'eliteManager',
+                'currentName' => 'eventManager',
+            ],
+        ];
+    }
+
+    private function getPropertyFromFile(string $string): ?Node
```
TomasVotruba, line 73: Here could be `?Property`, or even better `Property`, to cache bugs in case of not found `Property` node.

rules/naming/tests/ValueObjectFactory/PropertyRenameFactory/PropertyRenameFactoryTest.php
dobryy: So according to your comment, it is better to introduce the variable with docblock :)

rules/naming/tests/ValueObjectFactory/PropertyRenameFactory/PropertyRenameFactoryTest.php
dobryy: At first, I expected that there'll be more elements in the array. I'll change it to direct params.

dobryy:
@TomasVotruba, CI expects that under Fixture folder files have extension `*.php.inc` but when I'm changing it the tests don't work with no errors, just saying no tests found. That is why I renamed it to FixtureBasic, I found this 'approach' here https://github.com/rectorphp/rector/tree/master/packages/better-php-doc-parser/tests/PhpDocInfo/PhpDocInfoPrinter

dobryy:
@TomasVotruba let's merge it if it is okay in your opinion?

rules/naming/tests/ValueObjectFactory/PropertyRenameFactory/PropertyRenameFactoryTest.php
```diff
@@ -33,10 +32,8 @@ final class PropertyRenameFactoryTest extends AbstractKernelTestCase
      */
     private $betterNodeFinder;
 
-    public function __construct(?string $name = null, array $data = [], $dataName = '')
+    protected function setUp(): void
```
TomasVotruba, line 35: @dobryy I guess why you used `__construct`... 
Yet `__construct` should never by used in TestCase
PHPStan rule for this is waiting - https://github.com/symplify/symplify/issues/2225

Follow bellow...

rules/naming/tests/ValueObjectFactory/PropertyRenameFactory/PropertyRenameFactoryTest.php
```diff
@@ -45,31 +42,36 @@ public function __construct(?string $name = null, array $data = [], $dataName =
...
         $this->assertSame($currentName, $actualPropertyRename->getCurrentName());
-        $this->assertInstanceOf(PropertyProperty::class, $actualPropertyRename->getPropertyProperty());
-        $this->assertInstanceOf(ClassLike::class, $actualPropertyRename->getClassLike());
     }
 
-    public function createDataProvider(): Iterator
+    public function provideData(): Iterator
     {
-        yield [$this->getPropertyFromFile(__DIR__ . '/FixtureBasic/SomeClass.php'), 'eliteManager', 'eventManager'];
+        yield [new SmartFileInfo(__DIR__ . '/Fixture/SomeClass.php.inc'), 'eliteManager', 'eventManager'];
```
TomasVotruba, line 62: @dobryy Here is the root problem. Data provider should be just simple bare data.

The `getPropertyFromFile()` calls a sevices, that supposed to be set up in `setUp()` method. But that is `null`, right?

The problem is order of these method:

- first `dataProvider()` is called
- then `setUp()` - only now service actually exist
- then `test*()` method

A PHPStan rule proposed here https://github.com/symplify/symplify/issues/2226

dobryy: I see, I see, you're right before I created `getPropertyFromFile()` I used `setUp()`. I didn't want to put this logic into the test itself.

composer.json
```diff
@@ -254,6 +254,7 @@
             "rules/generic/tests/Rector/Name/PseudoNamespaceToNamespaceRector/Source"
         ],
         "files": [
+            "rules/naming/tests/ValueObjectFactory/PropertyRenameFactory/Fixture/SomeClass.php.inc",
```
dobryy, line 257: 👍🏼

rules/naming/tests/ValueObjectFactory/PropertyRenameFactory/PropertyRenameFactoryTest.php
TomasVotruba: Yea, this is crap in PHPUnit :/. I also expect `setUp()` being run before any other code.

TomasVotruba:
Thanks, merged :+1:

PR #4256, opened 2020-09-18
changed files detector test

mssimi:


packages/caching/src/Cache/Adapter/FilesystemAdapterFactory.php
```diff
@@ -5,17 +5,29 @@
...
+    }
+
     public function create(): FilesystemAdapter
     {
         return new FilesystemAdapter(
             // unique per project
             Strings::webalize(getcwd()),
             0,
-            sys_get_temp_dir() . '/_rector_cached_files'
+            sys_get_temp_dir() . '/' . $this->parameterProvider->provideParameter(Option::CACHE_DIR)
```
TomasVotruba, line 30: :+1:

packages/caching/tests/ChangedFilesDetectorTest.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    private function getSmartFileInfo(): SmartFileInfo
+    {
+        return new SmartFileInfo($this->filePathToTest());
+    }
+
+    private function filePathToTest(): string
+    {
+        return __DIR__ . '/ChangedFilesDetectorTest.php';
```
TomasVotruba, line 74: Better have a standalone fixture file, as it can change and be worked with.

packages/caching/tests/ChangedFilesDetectorTest.php
```diff
@@ -0,0 +1,76 @@
...
+    {
+        $smartFileInfo = $this->getSmartFileInfo();
+        $dependantFile = $this->filePathToTest();
+
+        $this->changedFilesDetector->addFileWithDependencies($smartFileInfo, [$dependantFile]);
+        $dependantSmartFileInfos = $this->changedFilesDetector->getDependentFileInfos($smartFileInfo);
+
+        $this->assertCount(1, $dependantSmartFileInfos);
+        $this->assertSame($dependantFile, $dependantSmartFileInfos[0]->getPathname());
+    }
```
TomasVotruba, line 60: There 3 methods are very difficult to read for me.
How can we extend it with more case in the future? 

What about using 1 method and 1 data provider?

packages/caching/src/ChangedFilesDetector.php
```diff
@@ -11,6 +11,7 @@
 
 /**
  * Inspired by https://github.com/symplify/symplify/pull/90/files#diff-72041b2e1029a08930e13d79d298ef11
+ * @see \Rector\Caching\Tests\ChangedFilesDetectorTest
```
TomasVotruba, line 14: :+1:

packages/caching/config/config.php
```diff
@@ -20,6 +20,7 @@
     $parameters = $containerConfigurator->parameters();
 
     $parameters->set(Option::ENABLE_CACHE, false);
+    $parameters->set(Option::CACHE_DIR, '_rector_cached_files');
```
TomasVotruba, line 23: This looks like the root directory will contain the cache directory.

TomasVotruba (commented):
Thanks for PR. I've added few notes

packages/caching/config/config.php
mssimi: Better now? Anyone can choose destination now.

packages/caching/tests/ChangedFilesDetectorTest.php
mssimi: I don't understand, data provider is used when you want to run same test with multiple different sets of params, this is not the cases here. Can you make example how you want it to look like?

packages/caching/config/config.php
```diff
@@ -20,7 +20,7 @@
     $parameters = $containerConfigurator->parameters();
 
     $parameters->set(Option::ENABLE_CACHE, false);
-    $parameters->set(Option::CACHE_DIR, '_rector_cached_files');
+    $parameters->set(Option::CACHE_DIR, sys_get_temp_dir() . '/_rector_cached_files');
```
TomasVotruba, line 23: :+1:

packages/caching/tests/Detector/ChangedFilesDetectorTest.php
```diff
@@ -71,6 +71,6 @@ private function getSmartFileInfo(): SmartFileInfo
 
     private function filePathToTest(): string
     {
-        return __DIR__ . '/ChangedFilesDetectorTest.php';
+        return __DIR__ . '/Source/file.php';
```
TomasVotruba, line 74: :+1:

packages/caching/tests/ChangedFilesDetectorTest.php
TomasVotruba: >  data provider is used when you want to run same test with multiple different sets of params

Exactly!

This is what I have in mind:

```php
public function test(array $inputFiles, array $changedFiles, int $expectedInvalidateCount)
{
    // ...
}

public function provideData()
{
    
    yield [[`A.php`, `B.php`, `C.php`], [], 0];

    // A is alone
    yield [[`A.php`, `B.php`, `C.php`], ['A.php'], 1];

    // C extends B, so changed B, also invalidated C
    yield [[`A.php`, `B.php`, `C.php`], ['B.php'], 2];
}
```

mssimi: I still don't understand, this is just changed files detector it is not responsible for invalidation of dependent files.

mssimi: Changed.

TomasVotruba:
Thank you!

Failing CI is related to my other changes, no to this PR

PR #4259, opened 2020-09-18
node docs

mssimi:


docs/rector_rules_overview.md
```diff
@@ -4861,7 +4861,7 @@ Remove null coalescing operator ??=
 ### `DowngradeTypedPropertyRector`
 
 - class: [`Rector\DowngradePhp74\Rector\Property\DowngradeTypedPropertyRector`](/rules/downgrade-php74/src/Rector/Property/DowngradeTypedPropertyRector.php)
-- [test fixtures](/rules/downgrade-php74/tests/Rector/Property/DowngradeTypedPropertyRector/Fixture)
+- [test fixtures](/rules/downgrade-php74/tests/Rector/Property/NoDocBlockDowngradeTypedPropertyRector/Fixture)
```
mssimi, line 4864: @TomasVotruba I don't know why is this still happening.

TomasVotruba: I think it's a path matching bug in `dump-rectors` command.

The end path of both files is the same, so the slash need to be added to regular expression or path matching or whatever is used there to find a test case

mssimi: I can, it's order issue.

mssimi: Here: #4260

mssimi:
I have added some snippets, I am taking break for now. Will continue with more later, but this can get merged imo.

TomasVotruba:
Looks good, mergin first wave :+1: 

Btw, we usually prefix the PR titles with `[type]`, e.g. `[Docs]`. Then we generate changelog from titles, also descriptive titles are helpful for searching of old PRs for debugging.

This one would be named e.g.:

`[Docs] Add new node samples`

PR #4260, opened 2020-09-18
fix rector docs generator sorting issue

mssimi:
Fixes docs generating in random order.


TomasVotruba:
Thanks :+1: 

Could you add test for matchin of `TypedProperty*Rector and `TypedDowngradeProperty*Rector`? I would help us to spot this bugs faster.

PR #4264, opened 2020-09-21
Update php-parser and phpstan to latest version

samsonasik:
Update php-parser and phpstan to latest version


TomasVotruba:
This will require lot of fine-tuning in php-parser internals.

I'll handle it

Issue #4168, opened 2020-09-08
[Downgrade PHP 7.3] Trailing commas in function calls


leoloso:
# Feature Request

Downgrade new feature from PHP 7.3 to its equivalent PHP 7.2 code
 
## Diff 

```diff
$compacted = compact(
    'posts',
-    'units',
+    'units'
);
```

quisse:
After some digging I can conclude that this is not possible using php-parser because php-parser isn't aware of the trailing comma AFAIK.
Any help would be appreciated.

TomasVotruba:
Closed by https://github.com/rectorphp/rector/pull/4273,

merged in https://github.com/rectorphp/rector/pull/4796

PR #4273, opened 2020-09-22
[Downgrade PHP 7.3] Trailing commas in function calls

quisse:
https://github.com/rectorphp/rector/issues/4168

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "jean85/pretty-package-versions": "^1.5.1",
         "nette/robot-loader": "^3.2",
         "nette/utils": "^3.1",
-        "nikic/php-parser": "4.10.2",
+        "nikic/php-parser": "4.10.3",
         "phpstan/phpdoc-parser": "^0.4.9",
         "phpstan/phpstan": "^0.12.52",
         "phpstan/phpstan-phpunit": "^0.12.16",
--- a/config/set/downgrade-php73.php
+++ b/config/set/downgrade-php73.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 use Rector\Downgrade\Rector\LNumber\ChangePhpVersionInPlatformCheckRector;
 use Rector\DowngradePhp73\Rector\List_\DowngradeListReferenceAssignmentRector;
+use Rector\DowngradePhp73\Rector\FuncCall\DowngradeTrailingCommasInFunctionCallsRector;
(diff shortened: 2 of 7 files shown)


quisse:
After some digging I can conclude that this is not possible using php-parser because php-parser isn't aware of the trailing comma AFAIK.
Any help would be appreciated.

TomasVotruba:
Same here. Maybe php-cs-fixer is better for this job

quisse:
So is including php-cs-fixer as a dependency an option? It would be nice if rectorphp makes downgrading every php version possible.

quisse:
Triggering regeneration removes the trailing commas. The only side effect I'm aware of is that when having a method call inline in a string in a method call, the curly brackets are doubled.
```
    {
        if($node->args){
            $node->setAttribute(AttributeKey::ORIGINAL_NODE,null);
        }
        return $node;
    }
```

Makes 
```
        $this->setOnClick("[Zip ID: {$modelid}]  {$e->getMessage($modelId,)}");
```
results in 
```
        $this->setOnClick("[Zip ID: {$modelid}]  {{$e->getMessage($modelId)}}");
```
A possible workaround could be
```
    {
        if($node->args){
            if(! $node->getAttribute(AttributeKey::PARENT_NODE) instanceof Node\Scalar\Encapsed){
                $node->setAttribute(AttributeKey::ORIGINAL_NODE,null);
            }
        }
        return $node;
    }
```
But the trailing comma isn't removed then.

TomasVotruba:
Removing original node usually lead to reprint without format. Then it depends on default printer setting.

What could work is do is hack around `BetterStandardPrinter` and remove comma there.

quisse:
With this one I am waiting for https://github.com/nikic/PHP-Parser/pull/732.
That issue causes that `"[Zip ID: {$modelid}]  {$e->getMessage($modelId)}"` is rewritten as `"[Zip ID: {$modelid}]  {{$e->getMessage($modelId)}}"`

quisse:
@TomasVotruba updating nikic/phpparser doesn't seem sufficient, do I need to wait for phpstan to include the latest phpparser or am I missing something?

TomasVotruba:
I'll look into this

TomasVotruba:
These 2 last details are ok. Do not change the code meaning and solves the comma.

Merged in: https://github.com/rectorphp/rector/pull/4796

Thank you :+1:

PR #4279, opened 2020-09-23
[PHPStan] Enable regex constant rule

TomasVotruba:


TomasVotruba:
@dobryy This will now require constant reference in any use of `match()` regex patterns.
It took me few hours to move ~100 strings to constants, but now it's done ;)

PR #4280, opened 2020-09-23
[ci] show rector ci fail

TomasVotruba:


TomasVotruba:
cc @dobryy And we're back :) I've accidentaly added `--dry-run`, but then there were no files to fix

PR #4281, opened 2020-09-23
[DOCS] Switch DumpRectorCommand from console to string printer

TomasVotruba:


composer.json
```diff
@@ -304,7 +304,7 @@
             "vendor/bin/changelog-linker cleanup --ansi"
         ],
         "docs": [
-            "bin/rector dump-rectors > docs/rector_rules_overview.md",
+            "bin/rector dump-rectors --output-file docs/rector_rules_overview.md --ansi",
```
TomasVotruba, line 307: @samsonasik This is the new updated version to generated rules overview, btw

samsonasik: thank you 👍

Issue #4275, opened 2020-09-22
[Naming] Move UnderscoreToCamelCasePropertyNameRector to Naming Set and Apply PropertyRenamer


dobryy:
Follow-up https://github.com/rectorphp/rector/pull/4231#issuecomment-693597482

PR #4283, opened 2020-09-23
[Naming] Make UnderscoreToCamelCasePropertyNameRector work with PropertyRenamer

dobryy:
Closes: https://github.com/rectorphp/rector/issues/4275


dobryy:
@TomasVotruba, I'll call you to discuss this madness :D

rules/naming/src/ExpectedNameResolver/AbstractExpectedNameResolver.php
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Rector\Naming\ExpectedNameResolver;
+
+use Nette\Utils\Strings;
+use PhpParser\Node\Stmt\Property;
+use Rector\Naming\Naming\PropertyNaming;
+use Rector\NodeNameResolver\NodeNameResolver;
+use Rector\NodeTypeResolver\NodeTypeResolver;
+
+abstract class AbstractExpectedNameResolver implements ExpectedNameResolverInterface
```
dobryy, line 13: For now AbstractExpectedNameResolver, but it is very possible that will be renamed to AbstractPropertyExpectedNameResolver when Param, Variable, etc will be refactored to new approach

rules/naming/src/Rector/Property/UnderscoreToCamelCasePropertyNameRector.php
```diff
@@ -0,0 +1,111 @@
...
+    public function refactor(Node $node): ?Node
+    {
+        $nodeName = $this->getName($node);
+        if ($nodeName === null) {
+            return null;
+        }
+
+        if (! Strings::contains($nodeName, '_')) {
+            return null;
+        }
```
dobryy, line 98: How do you think, should I move such logic into the new Guard class and apply it somehow only for this particular Rule?

rules/naming/src/PropertyRenamer/AbstractPropertyRenamer.php
```diff
@@ -0,0 +1,140 @@
...
+
+        if ($this->propertyRenameGuard->shouldSkip($propertyRename, [
+            $this->notPrivatePropertyGuard,
+            $this->conflictingPropertyNameGuard,
+            $this->ramseyUuidInterfaceGuard,
+            $this->dateTimeAtNamingConventionGuard,
+            $this->hasMagicGetSetGuard,
+        ])) {
+            return null;
+        }
```
dobryy, line 101: Don't you think it might be a good idea to decouple this logic to `shouldSkip` method in this class so that it could be possible to override the guards in particular renamers when needed? For example, for UnderscoreCamelCasePropertyRenamer we can add an additional guard which will skip renaming logic if there is no `_` in the name. 
Another possibility is to add the ability to pass additional guards directly from the rule itself.

dobryy (commented):
@TomasVotruba sorry for another huge PR. It is ready for the first round of review. 

I'm uncertain in some class names, structure, also some Interfaces might be redundant but I believe it will be more clear once I'll start decoupling Param renaming when I'll start bool param renaming task.

rules/naming/src/PropertyRenamer/AbstractPropertyRenamer.php
```diff
@@ -0,0 +1,140 @@
...
+{
+    /**
+     * @var RenameGuardInterface
+     */
+    protected $propertyRenameGuard;
+
+    /**
+     * @var GuardInterface
+     */
+    protected $conflictingPropertyNameGuard;
```
TomasVotruba, line 34: These `protected` properties usually means hidden dependency on sub-class, but we can handle that later

rules/naming/src/PropertyRenamer/AbstractPropertyRenamer.php
TomasVotruba: Yes, there is similar case for some type resolver. It allows to exclude one or more type resolvers based on type.

TomasVotruba:
Let's shipt it :)

rules/naming/src/PropertyRenamer/AbstractPropertyRenamer.php
dobryy: By "Yes" you mean the first option with overriding or to pass additional guards directly as a method param? I bet second because I saw something like this somewhere.

rules/naming/src/PropertyRenamer/AbstractPropertyRenamer.php
dobryy: These are used in all child classes that is why they are in the parent class. Do you mean even though they are in all child classes they should be moved to each of them?

rules/naming/src/PropertyRenamer/AbstractPropertyRenamer.php
TomasVotruba: I prefer visual lists with numbers, if there are more options :D

2nď I guess, I'd have to see it

rules/naming/src/PropertyRenamer/AbstractPropertyRenamer.php
TomasVotruba: That's an assumption. There could be a child class with nullable property.

This would actually break in PHP 7.4 typed properties.

The correct way is to use contract via interface method and change protected property/method/constant to a contract guaranteed value

PR #4284, opened 2020-09-24
Add UnderscoreToCamelCaseLocalVariableNameRector to only change local variable name

samsonasik:
The `UnderscoreToCamelCaseVariableNameRector` change both parameter and local variable methods. Here, I added `UnderscoreToCamelCaseLocalVariableNameRector` that the ability is only change local variable name, not parameter name, so the following code:

```php
final class SomeClass
{
    public function run($a_b)
    {
        $some_value = $a_b;
    }
}
```

will be changed to :

```php
final class SomeClass
{
    public function run($a_b)
    {
        $someValue = $a_b;
    }
}
```

Only local `$some_value` that changed to `$someValue` variable.


samsonasik:
@TomasVotruba this is can be usefull when developer need to only change local variable, not parameter/arg name, as the application may have a documentation separately that define parameter name, so developer only want to change local variable inside class methods.

For running `bin/rector dump-nodes > docs/nodes_overview.md`, It seems cause emptying rule example at `docs/rector_rules_overview.md` ( I run composer update before), so I reverted the change of it.

All checks have passed 🎉  it is ready for review.

rules/coding-style/tests/Rector/Variable/UnderscoreToCamelCaseLocalVariableNameRector/temp/first_underscore.php.inc
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Rector\CodingStyle\Tests\Rector\Variable\UnderscoreToCamelCaseLocalVariableNameRector\Fixture;
+
+final class FirstUnderscore
+{
+    public function run($_a)
+    {
+        $_some_value = $_a;
+    }
```
TomasVotruba, line 10: Could you also add a fixture that tests rename of following use of the variable?

```diff
 public function run($_a)
 {
-    $_some_value = $_a;
+    $someValue = $_a;

-    return $_some_value;
+    return $someValue;
 }
```

TomasVotruba:
> @TomasVotruba this is can be usefull when developer need to only change local variable, not parameter/arg name, as the application may have a documentation separately that define parameter name, so developer only want to change local variable inside class methods.

It makes sense, thanks :+1: 

> For running bin/rector dump-rectors > docs/rector_rules_overview.md, It seems cause emptying rule example at docs/rector_rules_overview.md ( I run composer update before), so I reverted the change of it.

I need to update this as I've refactored `dump-rectors` command

MGatner:
This is looking promising, and we'd be eager to have it on a few projects. Thanks @samsonasik

rules/coding-style/tests/Rector/Variable/UnderscoreToCamelCaseLocalVariableNameRector/temp/first_underscore.php.inc
samsonasik: implemented.

samsonasik: I've added `Fixture/param_assign_return.php.inc` for that.

samsonasik:

TomasVotruba (approved):

TomasVotruba:
Thank you

PR #4293, opened 2020-09-24
Fixes #4290 wrong skip assignment not from parameter in UnderscoreToCamelCaseLocalVariableNameRector

### LINKED ISSUE #4290: UnderscoreToCamelCaseLocalVariableNameRector bug: wrong skip when assignment not from parameter
# Bug Report

| Rector version | dev-master             |
| Installed as   | prefixed Rector PHAR                      |

## Minimal PHP Code 

The following code still not changed correctly:

```php
   public function run()

     {

         $_a = 'a';

         $_some_value = $_a;

         $this->run($_some_value);

     }
```

## Expected Behaviour

All changed correctly.

I will check it.

samsonasik:


samsonasik:

samsonasik:
I found another bug, when using expression:

```php
final class InsideExpression
{
    public function run($_a)
    {
        if ($_a) {
            $_some_value = $_a;
        }
    }
}
```

`$_a` in ` if ($_a) {` expression should be skipped.

samsonasik:
I've added handling for inside expression

samsonasik:
I found another bug, the following should be skipped:

```php
final class AssignToClassProperty
{
    public function run($a_b)
    {
        $this->a_b = $a_b;
    }
}
```

samsonasik:
I've added handling assignment to property fetch

samsonasik:
I found another bug, the following should be skipped:

```php
final class FromExpressionWithCheckValueComparison
{
    public function run($a_b)
    {
        if ($a_b === false) {

        }
    }
}
```

samsonasik:
I've added handling comparison check inside expression

samsonasik:
I found another bug, the following should be skipped:

```php
final class SkipPassFunctionParameterToClosure
{
    public function run(string $a_b, string $c_d)
	{
		function ($a_b) use ($c_d) {

		};
	}
}
```

samsonasik:
I've added handling ClosureUse parameter

samsonasik:
I think it is ready

TomasVotruba:
Looks good, thank you :+1:

PR #4297, opened 2020-09-25
[renaming] Fixes #4295 add \ prefix on FQ method call

### LINKED ISSUE #4295: RenameStaticMethodRector produces relative namespace (missing slash at the beginning)
# Bug Report

| Rector version | e.g. v0.8.10             |
| Installed as   | composer dependency             |

The code produced with RenameStaticMethodRector has relative namespace instead of absolute (the slash is missing at the beginning of the class namespace, which make the code invalid).

See here:
https://github.com/rectorphp/rector/blob/master/rules/renaming/tests/Rector/StaticCall/RenameStaticMethodRector/Fixture/fixture2.php.inc

the code produced is:
```
$result = Nette\Bridges\FormsLatte\Runtime::renderFormBegin($someArgs);
```
but should be 
```
$result = \Nette\Bridges\FormsLatte\Runtime::renderFormBegin($someArgs);
```

We have similar case in typo3-rector 
see rector:
https://github.com/sabbelasichon/typo3-rector/blob/master/config/typo3-104.php#L20

And fixture (which is green now, but should be red):
https://github.com/sabbelasichon/typo3-rector/blob/master/tests/Rector/Core/Utility/Fixture/use-environment-api2.php.inc#L9

line 9 should be `\TYPO3\CMS\Core\Core\Environment::isRunningOnCgiServer();`

samsonasik:


samsonasik:

rules/renaming/src/ValueObject/RenameStaticMethod.php
```diff
@@ -46,7 +46,7 @@ public function getOldMethod(): string
 
     public function getNewClass(): string
     {
-        return $this->newClass;
+        return '\\' . $this->newClass;
```
TomasVotruba, line 49: I think this should be fixed with `FullyQualified` in the place it's used

samsonasik: implemented.

rules/renaming/src/Rector/StaticCall/RenameStaticMethodRector.php
```diff
@@ -96,7 +96,7 @@ private function rename(StaticCall $staticCall, RenameStaticMethod $renameStatic
         $staticCall->name = new Identifier($renameStaticMethod->getNewMethod());
 
         if ($renameStaticMethod->hasClassChanged()) {
-            $staticCall->class = new Name($renameStaticMethod->getNewClass());
+            $staticCall->class = new Name('\\' . $renameStaticMethod->getNewClass());
```
TomasVotruba, line 99: The `FullyQualified` has the same effect, just internally adds `\\` to it's node.
It's also better to use it, as then further analysis knows its FQN

```suggestion
            $staticCall->class = new FullyQualified($renameStaticMethod->getNewClass());
```

samsonasik: implemented.

TomasVotruba:
Thank you :+1:

PR #4298, opened 2020-09-26
[Code Quality] Fixes #4286 Skip ArrayKeyExistsTernaryThenValueToCoalescingRector when else is a not null

### LINKED ISSUE #4286: ArrayKeyExistsTernaryThenValueToCoalescingRector misbehaves
# Bug Report

| Rector version | e.g. 0.8.7            |

`ArrayKeyExistsTernaryThenValueToCoalescingRector` is applied where it should not.
The rectoreed code is not the same as before.

https://getrector.org/demo/a047e490-98e3-4796-bcbf-d6461a480f4b

## Expected Behaviour

`DemoFile->run()` should return `null` instead of `['value' => null]`.

samsonasik:


samsonasik:

rules/code-quality/src/Rector/Ternary/ArrayKeyExistsTernaryThenValueToCoalescingRector.php
```diff
@@ -76,6 +77,10 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
+        if ($node->else instanceof Variable) {
+            return null;
+        }
```
TomasVotruba, line 82: I think that if `$node->else` is anything but null, it should be skipped.

Could you add also test fixture where else is string or false?

samsonasik: I will try.

TomasVotruba: Ah. Seeing this, the else can be anything and it's correct.
I got confused, sorry.


What exactly is the bug here?

TomasVotruba:
Ref #4286 (the id in title is not clickable on phone)

rules/code-quality/src/Rector/Ternary/ArrayKeyExistsTernaryThenValueToCoalescingRector.php
samsonasik: I think when else is a variable, we should skip that, not sure about false or string or other value..., /cc @infabo for verification

samsonasik: I've added handle else is not null to be skipped.

samsonasik: The following stil not checked: 

```php
return array_key_exists('value', $v) ? $v['value'] : 'null';
```

I will re-check.

samsonasik: I've added handling for string "null"

samsonasik:

samsonasik:
The `node->if` need to check null value ?

samsonasik:
I think it is dangerous to assume that `value` never be changed in `$v = ['value' => null];`, for example the `null` value is a dynamic from parameter/db.

TomasVotruba:
> I think it is dangerous to assume that value never be changed in $v = ['value' => null];, for example the null value is a dynamic from parameter/db.

I think you're right. That leads me to a questoin, are there some cases when this rule brings a value?

I can only think of 1 - when the else value is actually a `null`. Only that way the result is the same in reliable way:

https://3v4l.org/JcGTS

rules/code-quality/src/Rector/Ternary/ArrayKeyExistsTernaryThenValueToCoalescingRector.php
```diff
@@ -76,6 +77,10 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
+        if ($this->getValue($node->else) !== 'null' || $node->else instanceof String_) {
```
TomasVotruba, line 80: Btw, you can use helper method `$this->isNull($node->else)` here.

rules/code-quality/tests/Rector/Ternary/ArrayKeyExistsTernaryThenValueToCoalescingRector/Fixture/skip_else_string_null.php.inc
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Rector\CodeQuality\Tests\Rector\Ternary\ArrayKeyExistsTernaryThenValueToCoalescingRector\Fixture;
+
+class SkipElseStringNullValue
+{
+    public function run()
+    {
+        $v = ['value' => null];
+        return array_key_exists('value', $v) ? $v['value'] : 'null';
+    }
```
TomasVotruba, line 11: :+1:

rules/code-quality/src/Rector/Ternary/ArrayKeyExistsTernaryThenValueToCoalescingRector.php
samsonasik: I tried `$this->isNull`, that not work for this use case.

samsonasik:
I've added more fixture for $v["key"] has value and else is null/not

samsonasik:
I've added  fixture for if is null, else is null .

rules/code-quality/src/Rector/Ternary/ArrayKeyExistsTernaryThenValueToCoalescingRector.php
TomasVotruba: You mean like this?

```php
if (! $this->isNull($node->else)) {
    return null;
}
```

samsonasik: thank you, implemented.

samsonasik:
I think it is ready.

TomasVotruba:
Agreed, thank you :+1:

Issue #4276, opened 2020-09-22
[Generic] RenamePropertyRector doesn't handle conflicts and could cause problems with public and protected properties


dobryy:
Follow-up https://github.com/rectorphp/rector/pull/4231#issuecomment-693667850

dobryy:
@TomasVotruba Isn't it better to move this rule to Renaming set? If I do understand the idea correctly it doesn't belong to Naming.

TomasVotruba:
Yes :+1:

TomasVotruba:
Closing until this hypothetical case is reported in practise.

PR #4303, opened 2020-09-27
[Renaming] Move RenamePropertyRector to appropriate set

dobryy:
According to comments from https://github.com/rectorphp/rector/issues/4276 as a first PR moving the rule to appropriate Set.


TomasVotruba:
Thank you :+1:

Issue #3718, opened 2020-07-15
Argument 1 passed to PhpParser\PrettyPrinterAbstract::pModifiers()  must be of the type int, null given


bramdevries:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | `^0.7.20`              |
| Installed as   | Docker image                      |

I have set up the following alias to run `rector` through a docker container in the current directory:

```bash
function rector {
  docker run --rm -v $(pwd):/project rector/rector:0.7.48 process /project/app --config /project/rector.yaml --autoload-file /project/vendor/autoload.php
}

$ rector
```

Once I run this I see a lot of notices and errors appear: 

```bash
Notice: Undefined property: PhpParser\Node\Param::$flags in /project/vendor/nikic/php-parser/lib/PhpParser/PrettyPrinter/Standard.php on line 21

...

[ERROR] Could not process
         "../project/app/Projects/Processing/Traits/KnowsProcessors.php" file,
         due to:
         "Argument 1 passed to PhpParser\PrettyPrinterAbstract::pModifiers()
         must be of the type int, null given, called in
         /project/vendor/nikic/php-parser/lib/PhpParser/PrettyPrinter/Standard.p
         hp on line 21".
```

This issue seems to have been reported before in https://github.com/rectorphp/rector/issues/3640 but was mentioned to be fixed, which is why I opened a new one. I tried running various versions and the errors are present in all versions starting from 0.7.20.

TomasVotruba:
This probably won't be resolved until PHPStan upgrade to php-parser 4.6

bramdevries:
Just to clarify: Is this something caused by the code in my project? It seems very strange to me that all published releases from the past 3 months are unusable due to this bug.

TomasVotruba:
Nope. This is due to conflicting dependencies of Rector, PHPStan, they're using php-parser in different version.

bramdevries:
From what I can tell the [`0.7.48`](https://github.com/rectorphp/rector/blob/v0.7.48/composer.json#L21) release of Rector is locked to `nikic/php-parser:4.5` and the [`0.12.32`](https://github.com/phpstan/phpstan-src/blob/0.12.32/composer.json#L22) release of PHPStan is using [`^4.5`](https://semver.mwl.be/#!?package=nikic%2Fphp-parser&version=%5E4.5.0&minimum-stability=stable) which would already result in the `4.6` release being installed.


[...]

Issue #3643, opened 2020-07-03
Incompatibility with Nikic PHp Parser 4.6


Aerendir:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.7.41              |
| Installed as   | composer dependency                      |

## Minimal PHP Code Causing Issue

Run `composer update` and be sure the version 4.6.0 of `nikic/php-parser`.

Then a series of errors will be thrown:

```
PHP Notice:  Undefined property: PhpParser\Node\Param::$flags in /path/to/app/vendor/nikic/php-parser/lib/PhpParser/PrettyPrinter/Standard.php on line 21
Notice: Undefined property: PhpParser\Node\Param::$flags in /path/to/app/vendor/nikic/php-parser/lib/PhpParser/PrettyPrinter/Standard.php on line 21
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to PhpParser\PrettyPrinterAbstract::pModifiers() must be of the type integer, null given, called in /path/to/app/vendor/nikic/php-parser/lib/PhpParser/PrettyPrinter/Standard.php on line 21 and defined in /path/to/app/vendor/nikic/php-parser/lib/PhpParser/PrettyPrinterAbstract.php:1035
Stack trace:
#0 /path/to/app/vendor/nikic/php-parser/lib/PhpParser/PrettyPrinter/Standard.php(21): PhpParser\PrettyPrinterAbstract->pModifiers(NULL)
#1 /path/to/app/vendor/nikic/php-parser/lib/PhpParser/PrettyPrinterAbstract.php(521): PhpParser\PrettyPrinter\Standard->pParam(Object(PhpParser\Node\Param))
#2 /path/to/app/vendor/nikic/php-parser/lib/ in /path/to/app/vendor/nikic/php-parser/lib/PhpParser/PrettyPrinterAbstract.php on line 1035
```

# Suggested fix

In the Rector's `composer.json`, mark as incompatible `nikic/php-parser: 4.6.0`, waiting for a fix by Nikic:

```
"conflict": {
        "nikic/php-parser": "4.6.0"
    }
```

# Workaround

To work around this issue in your own apps and libraries, add this to the `composer.json` file:

```
"conflict": {
        "nikic/php-parser": "4.6.0"
    }
```

This will force Composer to not download the version `4.6.0` that will break the use of Rector.

TomasVotruba:
[...]

Issue #3537, opened 2020-06-17
Version conflict with phpstan


Synchro:
# Bug Report

Rector won't install with the current (recent) release of phpstan. I see that the composer file has a hard limit on phpstan's version:

```
"phpstan/phpstan": "^0.12.25,<=0.12.28",
```

phpstan's current release is [0.12.29](https://github.com/phpstan/phpstan/releases), which breaks this dependency. I also note that this is the only hard version limit in the composer.json file, but I can't find any reference to why it's there.

```sh
 composer require --dev rector/rector 
    1/1:        http://repo.packagist.org/p/provider-latest$7e6effa0ea22a07a7b98316c96d8b9669c697a90d2f0037b34af23533c3f0d56.json
    Finished: success: 1, skipped: 0, failure: 0, total: 1
Using version ^0.7.34 for rector/rector
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for rector/rector ^0.7.34 -> satisfiable by rector/rector[v0.7.34].
    - rector/rector v0.7.34 requires phpstan/phpstan ^0.12.25,<=0.12.28 -> satisfiable by phpstan/phpstan[0.12.25, 0.12.26, 0.12.27, 0.12.28] but these conflict with your requirements or minimum-stability.


Installation failed, reverting ./composer.json to its original content.
```

TomasVotruba:
That's not a bug, bit lock due to BC breaks.
PHPStan needs to be upgraded in Rector and new reports fixed.

PR #4304, opened 2020-09-27
Support latest php-parser

nexxai:
Update composer.json to pin to latest version of php-parser

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "migrify/php-config-printer": "^0.3.45",
         "nette/robot-loader": "^3.2",
         "nette/utils": "^3.1",
-        "nikic/php-parser": "4.10.1",
+        "nikic/php-parser": "4.10.2",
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Thank you :+1:

sandermarechal:
Why is the version pinned to the patch version begin with? It makes installing rector a hassle. I have php-parser 4.10.2 in my project. I cannot install rector at the moment because composer won't downgrade. Why not use php-parser 4.10.* or even ^4.10?

TomasVotruba:
Because of constant BC breaks that happen with Rector, php-parser and phpstan trio.

See e.g.
https://github.com/nikic/PHP-Parser/issues/714

TomasVotruba:
Also https://github.com/rectorphp/rector/issues/3718, 
https://github.com/rectorphp/rector/issues/3643, https://github.com/rectorphp/rector/issues/3537

PR #4305, opened 2020-09-28
[coding-style] make UnderscoreToCamelCaseVariableNameRector update @param docblock as well

samsonasik:


rules/coding-style/src/Rector/Variable/UnderscoreToCamelCaseVariableNameRector.php
```diff
@@ -87,7 +96,53 @@ public function refactor(Node $node): ?Node
...
+            $docComment->getText(),
+            sprintf(self::PARAM_NAME_REGEX, $variableName),
+            function ($match) use ($camelCaseName): string {
+                $match['paramName'] = $camelCaseName;
+                return $match['paramPrefix'] . $match['paramName'];
+            }
+        );
+
+        if ($parentNode instanceof ClassMethod) {
+            $variable->getAttribute(AttributeKey::METHOD_NODE)->setDocComment(new Doc($newdocComment));
```
samsonasik, line 142: I tried to implement upate `@param` when updating to camel case, but this is not works: 

```bash
--- Expected
+++ Actual
@@ @@
 {
        /**
      * @param string $x
-     * @param string $yZ
+     * @param string $y_z
      */
     public function run(string $x, string $yZ)
```

I printed the `$newdocComment` and it shows printed correctly: 

```bash
....^ """
/**\n
     * @param string $x\n
     * @param string $yZ\n
     */
"""
```

I tried both `$parentNode` and use `$variable` to get method node, both not works. @TomasVotruba do you have some hints for it? Thank you.

TomasVotruba: There is a rule that change param name in docblock too:
https://github.com/rectorphp/rector/blob/master/rules/naming/tests/Rector/ClassMethod/RenameParamToMatchTypeRector/Fixture/update_doc_name.php.inc

This might help:
https://github.com/rectorphp/rector/blob/219069f29f2d8e4ed1fe954d3747ea86b2a153cf/rules/naming/src/Rector/ClassMethod/RenameParamToMatchTypeRector.php#L122-L123

samsonasik: It works, thank you, it is implemented.

samsonasik:

TomasVotruba:
Nice job, thank you :+1:

dobryy:
Not sure what is Coding Style set for but shouldn't we home this rule to the Naming set? :)

TomasVotruba:
Sure. Could you update it?

dobryy:
Who is always changing `move` to `home` in my messages? I had this typos couple of times already :D buggy brain...
Doing it...

TomasVotruba:
Home is where people move, so it makese sense :D

PR #4312, opened 2020-09-28
[Naming] Move UnderscoreToCamelCaseVariableNameRector to Naming set

dobryy:


config/set/coding-style.php
```diff
@@ -69,6 +68,5 @@
 
     $services->set(CamelCaseFunctionNamingToUnderscoreRector::class);
     $services->set(SplitGroupedUseImportsRector::class);
-    $services->set(UnderscoreToCamelCaseVariableNameRector::class);
```
dobryy, line 72: Hmmmm..., @TomasVotruba should I keep it in both Sets?

TomasVotruba: That would be confusing I guess. Better have it unique.

dobryy: Ready for merging then.

Issue #4240, opened 2020-09-16
[Naming] Add param naming convention related to bool params


TomasVotruba:
To prevent these:

![image](https://user-images.githubusercontent.com/924196/93376442-e4a4ba00-f859-11ea-901f-8031d90e3895.png)

Follow up to https://github.com/rectorphp/rector/pull/4215

## Example

```diff
-public function someMethod(bool $name)
+public function someMethod(bool $isName)
 {
-   $this->isName = $name;
+   $this->isName = $isName;
 }
```

We make an assumption choice on is/has/was, but it's the only thing we can do.

TomasVotruba:
@dobryy You might enjoy this :)

dobryy:
```diff
-bool $enableCache,
+bool $isCacheEnabled,
```
Are you kidding me? :D This case is damn tricky!

dobryy:
Do not forget about https://github.com/rectorphp/rector/pull/4242#discussion_r489739962

TomasVotruba:
Not priority ATM

PR #4335, opened 2020-09-30
[Naming] Apply ParamRenamer without conflict resolution for Naming set

dobryy:
Related to #4240

### DIFF
--- a/rules/naming/src/ExpectedNameResolver/AbstractExpectedNameResolver.php
+++ b/rules/naming/src/ExpectedNameResolver/AbstractExpectedNameResolver.php
@@ -5,8 +5,7 @@ declare(strict_types=1);
 namespace Rector\Naming\ExpectedNameResolver;
 
 use Nette\Utils\Strings;
-use PhpParser\Node\Stmt\Property;
-use Rector\Naming\Naming\PropertyNaming;
+use PhpParser\Node;
 use Rector\NodeNameResolver\NodeNameResolver;
 use Rector\NodeTypeResolver\NodeTypeResolver;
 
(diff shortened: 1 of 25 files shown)


rules/naming/src/ExpectedNameResolver/AbstractExpectedNameResolver.php
```diff
@@ -17,40 +16,32 @@ abstract class AbstractExpectedNameResolver implements ExpectedNameResolverInter
...
-    ) {
+    public function __construct(NodeNameResolver $nodeNameResolver, NodeTypeResolver $nodeTypeResolver)
+    {
         $this->nodeNameResolver = $nodeNameResolver;
         $this->nodeTypeResolver = $nodeTypeResolver;
-        $this->propertyNaming = $propertyNaming;
     }
 
-    public function resolveIfNotYet(Property $property): ?string
+    public function resolveIfNotYet(Node $node): ?string
```
TomasVotruba, line 30: Here should be some Node type specification in `@param`

```suggestion
    /**
     * @name Property|Variable|...
     */
    public function resolveIfNotYet(Node $node): ?string
```

So the scope is narrow and it's easier to debug later. Now it can be anything, even `Switch_` :D 

Also these `Property|Variable|...` are [PHP 8-ready union types](https://wiki.php.net/rfc/union_types_v2). So we can upgrade Rector this year to:

```php
    public function resolveIfNotYet(Property|Variable Node $node): ?string
```

rules/naming/src/Rector/ClassMethod/RenameParamToMatchTypeRector.php
```diff
@@ -99,8 +105,6 @@ public function getNodeTypes(): array
      */
     public function refactor(Node $node): ?Node
     {
-        $this->hasChanged = false;
-
```
TomasVotruba, line 103: Why is this removed, when its still used bellow?

rules/naming/src/ValueObject/RenamePropertyValueObjectInterface.php
```diff
@@ -0,0 +1,20 @@
...
+interface RenamePropertyValueObjectInterface extends RenameValueObjectInterface
+{
+    public function getClassLike(): ClassLike;
+
+    public function getClassLikeName(): string;
+
+    public function getProperty(): Property;
+
+    public function getPropertyProperty(): PropertyProperty;
+}
```
TomasVotruba, line 20: :+1:

TomasVotruba (commented):
I've added few comment :)

rules/naming/src/Rector/ClassMethod/RenameParamToMatchTypeRector.php
dobryy: Oooops, it was accidently removed, my bad..

rules/naming/src/ExpectedNameResolver/AbstractExpectedNameResolver.php
dobryy: For now it is just Property and Param.

dobryy:
> I've added few comment :)

Ready :)

rules/naming/src/ExpectedNameResolver/AbstractExpectedNameResolver.php
TomasVotruba: These should be there.

TomasVotruba: I think also in the contract, to make the whole family of contract + implementer type aware

dobryy: Now, it should be okay.
Interface and abstract class have Property|Param.
Name resolvers have appropriate Node type(s) depending on the Node type(s) they handle.

TomasVotruba: Thank you

PR #4339, opened 2020-10-01
Fixes #4337 : Make Option::SKIP works again with ParameterProvider

### LINKED ISSUE #4337: Option::SKIP no longer works to exclude file by rule
# Bug Report

| Rector version | latest             |
| Installed as   | composer dependency / prefixed Rector PHAR                      |

It seems Option::SKIP no longer works to exclude file by rule, eg:

```php
                Rector\Naming\Rector\Variable\UnderscoreToCamelCaseVariableNameRector::class => [
			__DIR__ . '/A/ClassExcludedByRule.php'
		],
	]);
```

Printed `$skip` in `Rector\Core\Skip\Skipper` now got empty array.

## Expected Behaviour

Should works.

samsonasik:


TomasVotruba:
Thank you, this is super

phpstan.neon
```diff
@@ -512,10 +513,209 @@ parameters:
...
+
+        -
+            message: '#Do not use factory in constructor#'
+            paths:
+                - src/Testing/PHPUnit/Runnable/ClassLikeNamesSuffixer.php # 27
+                - src/Testing/PHPUnit/Runnable/RunnableClassFinder.php # 29
+                - utils/node-documentation-generator/src/ValueObject/NodeInfo.php # 29
+
+        -
+            message: '#Method with @required need to be named autowire\+class name#'
```
TomasVotruba, line 631: @samsonasik This seems like a lot of work. We'll need a Rector rule for this.
Could you handle it?

samsonasik: I will try

PR #4340, opened 2020-10-01
[Docs] Add missing php opening tag

dobryy:
This is an easy fix. We have one more rule with extra file content and it has opening php tag.

 We might add PHPStan rule that will check that this tag is present or... make doc generator check whether it is present or not. If not it can add it.

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -234,14 +234,11 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     $services = $containerConfigurator->services();
 
     $services->set(MoveValueObjectsToValueObjectDirectoryRector::class)
-        ->call(
-            'configure',
-            [[
-                MoveValueObjectsToValueObjectDirectoryRector::TYPES => ['ValueObjectInterfaceClassName'],
-                MoveValueObjectsToValueObjectDirectoryRector::SUFFIXES => ['Search'],
-                MoveValueObjectsToValueObjectDirectoryRector::ENABLE_VALUE_OBJECT_GUESSING => true,
-            ]]
-        );
+        ->call('configure', [[
+            MoveValueObjectsToValueObjectDirectoryRector::TYPES => ['ValueObjectInterfaceClassName'],
+            MoveValueObjectsToValueObjectDirectoryRector::SUFFIXES => ['Search'],
+            MoveValueObjectsToValueObjectDirectoryRector::ENABLE_VALUE_OBJECT_GUESSING => true,
+        ]]);
 };
 ```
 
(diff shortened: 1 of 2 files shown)


phpstan.neon
```diff
@@ -10,7 +10,7 @@ includes:
 services:
     # require constant in argument position
     -
-        class: Symplify\CodingStandard\Rules\ForceMethodCallArgumentConstantRule
+        class: Symplify\CodingStandard\Rules\RequireMethodCallArgumentConstantRule
```
dobryy, line 13: Sneaky @TomasVotruba :D

packages/better-php-doc-parser/src/PhpDocInfo/PhpDocInfoFactory.php
```diff
@@ -90,7 +90,8 @@ public function createFromNode(Node $node): ?PhpDocInfo
             $tokens = [];
             $phpDocNode = new AttributeAwarePhpDocNode([]);
         } else {
-            $content = $node->getDocComment()->getText();
+            $content = $node->getDocComment()
+                ->getText();
```
dobryy, line 94: @TomasVotruba , these kind of fixes by `fix-cs` look weird to me... I don't like :)
I could agree that it makes sense when the line is too long, or to make line break after the second chained method. :)

packages/better-php-doc-parser/src/PhpDocNodeFactory/MultiPhpDocNodeFactory.php
```diff
@@ -61,16 +61,13 @@ public function getTagValueNodeClassesToAnnotationClasses(): array
             IdTagValueNode::class => 'Doctrine\ORM\Mapping\Id',
             GeneratedValueTagValueNode::class => 'Doctrine\ORM\Mapping\GeneratedValue',
             JoinColumnTagValueNode::class => 'Doctrine\ORM\Mapping\JoinColumn',
-
```
dobryy, line 64: @TomasVotruba removing one empty line also doesn't make sense :) It makes sense to replace multiple empty lines by one. Sometimes it is handy to group lines somehow.

TomasVotruba: What ECS rule is causing this?

phpstan.neon
TomasVotruba: This should be fixed in `master`. Have you rebased? :)

packages/better-php-doc-parser/src/PhpDocInfo/PhpDocInfoFactory.php
TomasVotruba: I see. The rule of 1 method chain call per line is more consistent and logical to read.

Same way 1 item per array item line.

```php
[
    'one' => 'two',
    'one' => 'two',
    'one' => 'two',
]
```

packages/better-php-doc-parser/src/PhpDocNodeFactory/MultiPhpDocNodeFactory.php
dobryy: I didn't notice it. We will find it later

dobryy:
@TomasVotruba Ready for merging

PR #4344, opened 2020-10-03
[SOLID] Change if && to early return

dobryy:


dobryy:
Just checking out very basic logic...

rules/nette/src/TemplatePropertyAssignCollector.php
```diff
@@ -100,11 +100,13 @@ private function collectVariableFromAssign(Assign $assign): void
 
             $this->nodesToRemove[] = $assign;
         }
-
-        // $x = $this->template
```
TomasVotruba, line 104: This comment should be kept

rules/solid/src/Rector/If_/ChangeAndIfToEarlyReturnRector.php
```diff
@@ -0,0 +1,227 @@
...
+        $invertedRightCondition = $this->conditionInverter->createInvertedCondition($expr->right);
+
+        $firstIf = new If_($invertedLeftCondition);
+        $firstIf->stmts = [new Return_()];
+        $secondIf = new If_($invertedRightCondition);
+        $secondIf->stmts = [new Return_()];
+
+        $this->addNodeAfterNode($firstIf, $node);
+        $this->addNodeAfterNode($secondIf, $node);
+        $this->addNodeAfterNode($ifReturn, $node);
```
TomasVotruba, line 125: These is also `addNodesAfterNode()` method

rules/solid/tests/Rector/If_/ChangeAndIfToEarlyReturnRector/Fixture/equals.php.inc
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Rector\SOLID\Tests\Rector\If_\ChangeAndIfToEarlyReturnRector\Fixture;
+
+class EqualsClass
+{
+    public function canDrive(Car $car)
+    {
+        if ($car->hasWheels === true && $car->hasFuel === true) {
```
TomasVotruba, line 9: Could you also add test for 3 items?

```php
$value === true && $value === true && $value === true 
```

dobryy: At the moment, the case with 3 items is skipped, there is a fixture for this. More than two items will make rule a bit more complicated I think. I wanted to keep the PR as small as possible.

rules/solid/tests/Rector/If_/ChangeAndIfToEarlyReturnRector/Fixture/skip_more_than_two_conditions.php.inc
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Rector\SOLID\Tests\Rector\If_\ChangeAndIfToEarlyReturnRector\Fixture;
+
+class SkipMoreThanTwoConditionClass
+{
+    public function canDrive(Car $car)
+    {
+        if ($car->hasWheels && $car->hasFuel && $car->hasEngine) {
```
dobryy, line 9: @TomasVotruba This situation can be handled in the next PR.

rules/solid/tests/Rector/If_/ChangeAndIfToEarlyReturnRector/Fixture/skip_non_first_level_if.php.inc
```diff
@@ -0,0 +1,21 @@
...
+    {
+        if ($car->wheelsCount() > 2) {
+            if ($car->hasWheels && $car->hasFuel) {
+                $this->canDrive = true;
+            }
+
+            return;
+        }
+
+        $this->canDrive = true;
```
dobryy, line 17: @TomasVotruba This situation can be handled in the next PR.

rules/solid/tests/Rector/If_/ChangeAndIfToEarlyReturnRector/Fixture/skip_if_in_foreach.php.inc
```diff
@@ -0,0 +1,24 @@
...
+    private function refactorStmts(array $nodes): void
+    {
+        foreach ($nodes as $node) {
+            if ($node instanceof Name && $node instanceof Identifier) {
+                $this->processNameOrIdentifier($node);
+            }
+
+            return;
+        }
+
```
dobryy, line 19: @TomasVotruba This situation looks crazy but also can be handled in the next PR.

rules/solid/tests/Rector/If_/ChangeAndIfToEarlyReturnRector/Fixture/equals.php.inc
TomasVotruba: I see

TomasVotruba (changes requested):
I've added few comments

dobryy:
> I've added few comments

I've implemented changes as a new commit https://github.com/rectorphp/rector/pull/4344/commits/38047f43579ff5916c14f97fbd122002bb0464b9

dobryy:
@TomasVotruba Pipeline is finally green.

TomasVotruba:
Thank you

ERuban:
@dobryy @TomasVotruba hey, I've just a one question - why it is a SOLID rule?

TomasVotruba:
From all the sets there are now, SOLID and early return are closest.
What set would you suggest instead?

ERuban:
@TomasVotruba So, I would suggest `CodingStyle` set for this rule - seems it will be better here, just because it is really more about accepted conventions in teams/companies, imho.
Also, why do you think it is closest to SOLID rules?

ERuban:
Hm, it will be really nice to move it to a standalone set. 
Or, maybe, to CodeQuality set :)

TomasVotruba:
Then, next month we a question, why it's part of CodeQuality set :D 

Btw, is current situation causing any undesired behavior of Rector in your code or is it rather academic curiosity?

TomasVotruba:
There is an issue here: https://github.com/rectorphp/rector/issues/4464

ERuban:
So, ok, standalone set looks ok, imo. 
Yeap, it is just academic curiosity (we haven't such rule in our team). 

Btw, I've some issues after updating Rector from the older version, but will ask you separately about it. ;)

PR #4347, opened 2020-10-03
Add missing abstract keyword

dobryy:


dobryy:
@TomasVotruba wondering that PHPStan didn't inform about this issue...

samsonasik:
There is symplify's phpstan rules for that https://github.com/symplify/symplify/blob/c39ce87feca25581f3f3c9a3f897860927d780b7/packages/coding-standard/src/Rules/CheckRequiredAbstractKeywordForClassNameStartWithAbstractRule.php

TomasVotruba:
@dobryy Is the rule loaded in our set here?

dobryy:
> @dobryy Is the `CheckRequiredAbstractKeywordForClassNameStartWithAbstractRule` rule loaded in our `phpstan.neon`?

Looks like nope... enabling it...

dobryy:
> @dobryy Is the `CheckRequiredAbstractKeywordForClassNameStartWithAbstractRule` rule loaded in our `phpstan.neon`?

I added it the rule to phpstan.neon but it doesn't work for some reason (I remove the `absrtact` keyword in `AbstractPropertyConflictingNameGuard`). What am I doing wrong?

TomasVotruba:
> What am I doing wrong?

What in the rule does not work?

dobryy:
> > What am I doing wrong?
> 
> What in the rule does not work? Try debugging it

It doesn't say any errors even though there is one in the code.

Okay, I'll check it.

phpstan.neon
```diff
@@ -724,3 +723,58 @@ parameters:
...
+                - packages/post-rector/src/Rector/AbstractRector/NodeCommandersTrait.php # 71
+                - packages/reporting/src/Rector/AbstractRector/NodeReportCollectorTrait.php # 24
+                - packages/reporting/src/Rector/AbstractRector/NodeReportCollectorTrait.php # 24
+                - rules/doctrine/src/AbstractRector/DoctrineTrait.php # 23
+                - rules/doctrine/src/AbstractRector/DoctrineTrait.php # 23
+                - rules/doctrine/src/AbstractRector/DoctrineTrait.php # 23
+                - src/Rector/AbstractRector/BetterStandardPrinterTrait.php # 37
+                - src/Rector/AbstractRector/BetterStandardPrinterTrait.php # 37
+                - src/Rector/AbstractRector/CallableNodeTraverserTrait.php # 24
+                - src/Rector/AbstractRector/CallableNodeTraverserTrait.php # 24
```
dobryy, line 757: I ignored new PHPStan errors. Should I create new issue for it? (I don't feel like a scout today :))

dobryy:
I the end there no need to add this rule to `phpstan.neon` it was enough to delete the error from `ignoreErrors` section

phpstan.neon
TomasVotruba: We should have this covered in CI already

dobryy: I thought it is covered but CI didn't fix it for some reason....

TomasVotruba: You probably haven't rebased on `master`

I removed it with no PHPStan report:
https://github.com/rectorphp/rector/pull/4354

TomasVotruba: So you were right, it was a bug. My bad

PR #4354, opened 2020-10-04
[static] replace string attirbutes with contants + fix require rule

TomasVotruba:


TomasVotruba:
@dobryy Hm, the PHPStan rule might be broken :thinking: I'll check it

TomasVotruba:
Here is the fix PR in @symplify https://github.com/symplify/symplify/pull/2317

The bug was in class/trait naming. In Rector/php-parser, when you're working with trait, it is reated to trait.
In PHPStan, if you're in a trait, you're actually in a class. So all the trait method names wanted to be named `autowireAbstractRector` instead of `autowire<trait-name>`.

Thus the bug.

PR #4357, opened 2020-10-05
Downgrade PHP7.4 strip_tags array arg

### LINKED ISSUE #4167: [Downgrade PHP 7.4] strip_tags accepting arrays
# Feature Request

Downgrade new feature from PHP 7.4 to its equivalent PHP 7.3 code
 
## Diff 

```diff
-strip_tags($string, ['a', 'p'])
+strip_tags($string, '<a><p>')
```

leoloso:
Fixes #4167


leoloso:
@TomasVotruba I'm getting this error:

```
[ERROR] The                                                                    
         "rules/downgrade-php74/tests/Rector/FuncCall/DowngradeStripTagsCallWith
         ArrayRector/Fixture/skip_strings.php.inc" file has same content before 
         and after. Remove the 2nd half of the file.  
```

That is because I added a test to check that nothing has changed: `strip_tags('foo', '<a>')` is still `strip_tags('foo', '<a>')`. Should I remove these tests? (I hope not, there's value in them!) Or is there some other way to set them up?

TomasVotruba:
> Remove the 2nd half of the file.

That means the content after `-----`

```diff
 a
------
-a
```

This is to make it clear there is no change at all.

leoloso:
@TomasVotruba PR ready, please check if everything is alright, and be welcome to merge :)

rules/downgrade-php74/src/Rector/FuncCall/DowngradeStripTagsCallWithArrayRector.php
```diff
@@ -0,0 +1,184 @@
...
+    {
+        return new Ternary(
+            new BooleanAnd(
+                new NotIdentical($expr, $this->createNull()),
+                new FuncCall(new Name('is_array'), [new Arg($expr)])
+            ),
+            $this->getConvertArrayToStringFuncCall($expr),
+            $expr
+        );
+    }
```
TomasVotruba, line 183: These should be named like `createXFromY()` with more narrow param/return types

```diff
-private function getIfArrayConvertArrayToStringFuncCall(Expr $expr): Expr
+private function createFuncCallFromArray(Array_ $array): FuncCall
```

- It helps with better static control
- The `create*()` method name makes clear the node is completely new, not just modified

leoloso: Ok, please check if the new function names are OK. I can't narrow the param type from `Expr` since it can receive many types, so I made them explicit in the phpdoc

TomasVotruba: Looks good :+1:

TomasVotruba (approved):

TomasVotruba:
Thank you

PR #4361, opened 2020-10-05
[Downgrade PHP 7.4] Numeric Literal Separator Rector

### LINKED ISSUE #4166: [Downgrade PHP 7.4] Numeric Literal Separator
# Feature Request

Downgrade new feature from PHP 7.4 to its equivalent PHP 7.3 code. [RFC](https://wiki.php.net/rfc/numeric_literal_separator)
 
## Diff 

```diff
-$number = 107_925_284.88;
+$number = 107925284.88;
```

leoloso:
Fixes #4166

### DIFF
--- a/config/set/downgrade-php74.php
+++ b/config/set/downgrade-php74.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 use Rector\DowngradePhp74\Rector\ArrowFunction\ArrowFunctionToAnonymousFunctionRector;
 use Rector\DowngradePhp74\Rector\Coalesce\DowngradeNullCoalescingOperatorRector;
+use Rector\DowngradePhp74\Rector\LNumber\DowngradeNumericLiteralSeparatorRector;
 use Rector\DowngradePhp74\Rector\Property\DowngradeTypedPropertyRector;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 
@@ -12,4 +13,5 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     $services->set(DowngradeTypedPropertyRector::class);
     $services->set(ArrowFunctionToAnonymousFunctionRector::class);
     $services->set(DowngradeNullCoalescingOperatorRector::class);
+    $services->set(DowngradeNumericLiteralSeparatorRector::class);
 };
--- a/rules/downgrade-php74/src/Rector/LNumber/DowngradeNumericLiteralSeparatorRector.php
+++ b/rules/downgrade-php74/src/Rector/LNumber/DowngradeNumericLiteralSeparatorRector.php
@@ -1,0 +1,82 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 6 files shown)


leoloso:
@TomasVotruba I need your input here.

The way I found to remove the `_` is by casting the number to string:

```php
$node->value = (string) $node->value;
```

I have one issue though: since the number is already parsed, I can't tell if it's in scientific notation, like `10e+3`. I'd like to not refactor these numbers (they can't have `_`), but right now I don't see how, so the test is not succeeding here:

```diff
-        $scientific_notation = 10e+3;
+        $scientific_notation = 10000;
```

I see 2 alternatives:

- Keep it as it is, transforming `10e+3` to `10000` since for downgrading it works too
- Introduce attribute `IS_SCIENTIFIC_NOTATION` in `LNumber`

What do you think?

leoloso:
Btw, I don't think this works, since the number is already transformed: https://github.com/rectorphp/rector/blob/master/rules/php74/src/Rector/LNumber/AddLiteralSeparatorToNumberRector.php#L118-L121

TomasVotruba:
image

This would be hell to search for in history in case of bug in the rule :D 

We have convention of `[Set] Add RuleName`

TomasVotruba:
> 10e+3

Is that required for downgrade to lower PHP? If not, I'd skip it

leoloso:
> This is be hell to search for in pull-request history in case of bug in the rule :D

Oops... I hadn't paid attention, fixed now

> Is that required for downgrade to lower PHP? If not, I'd skip it

Not needed, I'll skip it then

leoloso:
Btw, do you have a policy of adding the PR/issue URLs as @see docblock on the rector class? I haven't done so yet, I can complete them

TomasVotruba:
No nee to. That's actually covered by Github blame and link to origin of Rule:

image

TomasVotruba:
> Oops... I hadn't paid attention, fixed now

Thanks! I wish we had a Github bot for that that would autocomplete the name, but didn't have time to investigate if such Github bot exists

leoloso:
I'm done now, please review it and if it's ok, merge it! 😃

TomasVotruba:
Looks good. Nice job with simple test cases :+1:

TomasVotruba (approved):

TomasVotruba:
Thank you

leoloso:
Btw, don't forget this, if you want I can create an issue:

> Btw, I don't think this works, since the number is already transformed: https://github.com/rectorphp/rector/blob/master/rules/php74/src/Rector/LNumber/AddLiteralSeparatorToNumberRector.php#L118-L121

TomasVotruba:
What issue you mean?

leoloso:
In `AddLiteralSeparatorToNumberRector` there is this code:

```php
        // e+/e-
        if (Strings::match($numericValueAsString, '#e#i')) {
            return true;
        }
```

If I'm right, that condition will never be true, because `$numericValueAsString` will already be the number `10000`, and not its notation `10e+3` which is lost after the number is parsed

Issue #4333, opened 2020-09-30
Replacing iterator_to_array with unpack could result in Cannot unpack array with string keys


gisostallenberg:
# Bug Report

To reproduce (PHP 7.4):
```php
<?php

function getGenerator($key): \Generator {
    yield $key => 'value';
    yield 'value';
}

var_dump(iterator_to_array(getGenerator(0)));
var_dump(iterator_to_array(getGenerator('key')));
var_dump([...getGenerator(0)]);
var_dump([...getGenerator('key')]);

```

```txt
array(2) {
  [0]=>
  string(5) "value"
  [1]=>
  string(5) "value"
}
array(2) {
  ["key"]=>
  string(5) "value"
  [0]=>
  string(5) "value"
}
array(2) {
  [0]=>
  string(5) "value"
  [1]=>
  string(5) "value"
}
PHP Fatal error:  Uncaught Error: Cannot unpack Traversable with string keys in Standard input code:11
Stack trace:
#0 {main}
  thrown in Standard input code on line 11
```
I'm not sure you can detect if the generator contains string keys, but otherwise this is adding potential bugs.

According to an answer in https://stackoverflow.com/questions/40663687/cannot-unpack-array-with-string-keys I suppose the unpacking will work in PHP 8

TomasVotruba:
Hi,

thanks for reporting. This is definitely a bug.

Could you add this failing fixture to responsbile rule into `/Fixture` directory via pull-request? So we have it covered :+1:

gisostallenberg:
You mean 'add a file' in `rules/php74/tests/Rector/FuncCall/ArraySpreadInsteadOfArrayMergeRector/Fixture/`?

gisostallenberg:
I'm not really sure what to do to help you 😄

TomasVotruba:
> You mean 'add a file' in

Yes ;)

gisostallenberg:
@TomasVotruba I'm really not sure how to help, I'm sorry.
I would add somehting like this, right?:
`rules/php74/tests/Rector/FuncCall/ArraySpreadInsteadOfArrayMergeRector/Fixture/skip_string_keys_from_generator.php.inc`
```php
<?php

namespace Rector\Php74\Tests\Rector\FuncCall\ArraySpreadInsteadOfArrayMergeRector\Fixture;

function x(): \Generator
{
    yield 'key' => 'value';
}

class SkipStringKeysFromGenerator
{
    public function run()
    {
[...]

PR #4362, opened 2020-10-06
Add a fixture for generators with keys

gisostallenberg:
See https://github.com/rectorphp/rector/issues/4333
This adds a fixture that shows what not to be fixed, a `\Generator` that has string keys.


gisostallenberg:
The actual issue still needs to be fixed though.

TomasVotruba:
Hi,

thanks for the fixture. Could you include the fix as well?

gisostallenberg:
Yeah, sure. Can you point me at some code where to start?

TomasVotruba:
Great!

I think we need to check for string key type of the value in `iterator_to_array()`.

That can be done here https://github.com/rectorphp/rector/blob/master/rules/php74/src/Rector/FuncCall/ArraySpreadInsteadOfArrayMergeRector.php#L128 with `getStaticType()` method:

```php
$type = $this->getStaticType($value);
dump($type);

// check for array/iterator key type, if string, `return null`
```

That's my rough guess

TomasVotruba:
Resolved by https://github.com/rectorphp/rector/pull/4415

`iterator_to_array()` is now skipped completelly

PR #4364, opened 2020-10-06
[docs] How to add Rector rule test

TomasVotruba:
First improvement of Rector documentation :) 

@leoloso Could you review?


docs/how_to_add_test_for_rector_rule.md
```diff
@@ -0,0 +1,119 @@
...
+
+`FinalizeClassesWithoutChildrenRectorTest.php` (test file)
+
+Right here:
+
+![Rule Test Case](/docs/images/docs_rule_test_case.png)
+
+## 4. Add Change or No-Change fixture
+
+Next to test case, there is `/Fixture` directory, there we create our test fixture file, e.g. `add_final.php.inc`. The `.php.inc` is there on purpose, so the file is hidden from coding standard tools and static analysis.
```
leoloso, line 59: You may want to say that a test can have many fixture tests, not just one.

Btw, I find sentence `"there we create our test fixture file"` a bit confusing myself: what does "fixture" mean? Is it a word used everywhere concerning testing, or just for Rector? Is it the same as "test cases"?

I've seen that many tests are called `fixture.php.inc`, is that supposed to be the main one? Or there's no need even to call it `fixture.php.inc` (say, when there's only 1 test)?

docs/how_to_add_test_for_rector_rule.md
```diff
@@ -0,0 +1,119 @@
...
+## 2. Detect the Minimal File
+
+Usually the Rector diff output is long and contains many other errors related to other rules. It's a mess, we can't use that for test fixture. We need to find **1 responsible line**.
+
+The best way is to copy the file to local code, e.g. `app/SomeFile.php` a put only the broken line there.
+
+In our case, all we need is:
+
+```php
+class StaticEasyPrefixer
```
leoloso, line 26: I think it should be mentioned that this is because `FinalizeClassesWithoutChildrenRector` adds `final` to the class, otherwise it's not clear...

leoloso: I mean, the image can be seen in the image, but for newbies it's still a bit not clear, and I guess you're targetting newbies with this doc

docs/how_to_add_test_for_rector_rule.md
```diff
@@ -0,0 +1,119 @@
...
+
+### 2. The Code Should not be Touched
+
+```bash
+<code before>
+```
+
+<br>
+
+In this perticular case, the code should change - `final` should be added, so the test fixture would look like this:
```
leoloso, line 79: perticular => particular

docs/how_to_add_test_for_rector_rule.md
```diff
@@ -0,0 +1,119 @@
...
+vendor/bin/phpunit rules/solid/tests/Rector/Class_/FinalizeClassesWithoutChildrenRector/FinalizeClassesWithoutChildrenRectorTest.php
+```
+
+To run only the single test fixture, add `--filter test#X`, where X is the order number of fixture.
+
+```bash
+vendor/bin/phpunit rules/solid/tests/Rector/Class_/FinalizeClassesWithoutChildrenRector/FinalizeClassesWithoutChildrenRectorTest.php --filter test#4
+```
+
+If PHPUnit fails, you've successfully added a test case! :)
```
leoloso, line 119: Assuming you're creating the test before implementing the rule, which is not always the case

leoloso (commented):
I think it's good, but can be improved

docs/how_to_add_test_for_rector_rule.md
TomasVotruba: Thanks for feedback, I'll send updated version to address your comments.

docs/how_to_add_test_for_rector_rule.md
TomasVotruba: This is manual for contributors to add test fixture to a broken Rule. So the rule is always implemented.

Real use case is issue like this one: https://github.com/rectorphp/rector/issues/4333
The idea is to paste link to this post, so the issue author know how to add this test case to Rector repository.

TomasVotruba:
Ready for round 2 :)

docs/how_to_add_test_for_rector_rule.md
leoloso: > This is manual for contributors to add test fixture to a broken Rule. So the rule is always implemented.

Alright. However, there are many pieces of information here, useful to everyone, also to developers creating the test at the same time as the rule, not only for demonstrating a broken rule. For instance, I didn't know about `---` the other day, or what a fixture meant. So you can perfectly link to this doc from the `README`, under some header `Creating test for your rules`, targeting everyone

leoloso (commented):
I'm happy with the changes so far. Just check if to make this doc more generic, not just for broken rules. If you decide not to, then I'll send my approval

docs/how_to_add_test_for_rector_rule.md
```diff
@@ -0,0 +1,125 @@
...
+
+### 1. The Code Should Change
+
+```bash
+<code before>
+-----
+<code after>
+```
+
+### 2. The Code Should not Be Touched
```
samsonasik, line 77:
```suggestion
### 2. The Code Should Be Skipped
```

TomasVotruba:
> I'm happy with the changes so far. Just check if to make this doc more generic, not just for broken rules. If you decide not to, then I'll send my approval

Let's ship it

Issue #4363, opened 2020-10-06
PHP 7.4 property with direct assignment and union type is wrongly fixed to typed property


gisostallenberg:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.8.15 |
| Installed as   | composer dependency |

While creating typed properties a property with an assignment but a doc-block with a union type still gets converted to a typed property. After this the property can no longer be assigned another type. Maybe `int` is no problem, but it does this as well with classes.

```diff
--- Original
+++ New
@@ -3,8 +3,5 @@
 namespace App;

 class Foo {
-    /**
-     * @var bool|int
-     */
-    private $bar = 0;
+    private int $bar = 0;
 }
```

```diff
--- Original
+++ New
@@ -3,8 +3,5 @@
 namespace App;

 class Foo {
-    /**
-     * @var bool|int
-     */
-    private $bar = false;
+    private bool $bar = false;
 }
```
```diff
--- Original
+++ New
@@ -3,8 +3,5 @@
 namespace App;

 class Foo {
-    /**
-     * @var bool|int
-     */
-    private $bar = true;
+    private bool $bar = true;
 }
```

## Expected Behaviour
I would expect rector to respect the docblock for PHP 7.4 and add a union type in PHP 8.

TomasVotruba:
Hi,

thanks for reporting. Could you add a failing fixture to that rule?

PR #4366, opened 2020-10-06
Do not make a union type into a typed property

gisostallenberg:
See https://github.com/rectorphp/rector/issues/4363


TomasVotruba (commented):
Thanks!

The docblock type is actually just one of places to detect type. Also setters, getters, default values are taken into an account.
See 
https://github.com/rectorphp/rector/blob/master/rules/php74/src/Rector/Property/TypedPropertyRector.php#L133

To make it really union type, there should be 2 setters, e.g. one setting bool, one setting interger.

Could you add those?

TomasVotruba:
Any ETA on this?

TomasVotruba:
I did huge refactoring over the weekend, with typed properties and their value resolving

Should be covered by https://github.com/rectorphp/rector/pull/4394

PR #4367, opened 2020-10-07
[PHP74] Do not transform callable|null into typed property

antograssiot:
Before
```php
 final class ClassWithProperty
 {
    /**
     * @var callable|null
     */
    private $shouldBeSkippedAlso;
 }
```

would have been transformed to 
```php
 final class ClassWithProperty
 {
    private ?callable $shouldBeSkippedAlso = null;
 }
```

Which result in invalid PHP as explained in the  [RFC](https://wiki.php.net/rfc/typed_properties_v2)

I'm not familiar with the project, this is my first contribution so you might know a less dirty way to fix this.
Thanks by advance for your consideration

### DIFF
--- a/rules/php74/src/Rector/Property/TypedPropertyRector.php
+++ b/rules/php74/src/Rector/Property/TypedPropertyRector.php
@@ -177,10 +177,6 @@ CODE_SAMPLE
      */
     private function shouldSkipNonClassLikeType(Node $node): bool
     {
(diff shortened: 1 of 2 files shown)


TomasVotruba (approved):

TomasVotruba:
Hi, this looks good 👍

TomasVotruba:
Thank you

Issue #4316, opened 2020-09-29
Proposal: Create exceptions `DowngradeNotImplementedException` and `DowngradeNotPossibleException`


leoloso:
# Feature Request

There are features from new PHP version that cannot be downgraded to a previous version, such as #4317, so we could throw `DowngradeNotPossibleException` whenever finding these features in the code.

There could also be features that can be downgraded, but we haven't done it yet (eg: because we disregarded that use case).

For instance, the downgrade rule from #4314 may be applied to a still-unsupported use case. Then, the rule could fail with throwing `DowngradeNotImplementedException`, and showing an error message asking the developer to add an issue for this new use case in the Rector repo.

If after analyzing the use case, we come to the conclusion that it cannot be downgraded, then we could throw a `DowngradeNotPossibleException`, explaining why this feature cannot be supported by previous PHP versions. This may well happen with annotations from PHP 8.0 (#4156)

Several exceptions could happen in a single run, so we could extract the message from all of them and show them all together once the run has ended.

TomasVotruba:
This work for PHPStan rules. Better make standalone project

leoloso:
Why PHPStan? This is purely about downgrading, not about validation of code.

`DowngradeNotPossibleException` I think is needed, as I explained in https://github.com/rectorphp/rector/issues/4324#issuecomment-738023936

Concerning `DowngradeNotImplementedException`, it would allow to implement certain downgrade rules in a partial manner. For instance, annotations cannot be naturally downgraded, but the annotation for deprecation can. So we can at least support that bit, and throw an error for everything else.

These are enabling features. I think you should consider reopening this

TomasVotruba:
Because PHPStan is read-only and Rector write-only. This would mix features of both.

We can get back to it, after the known downgrade paths are covered.

leoloso:
But if you close it, it will be out of sight, and we'll forget about it. Why don't we keep it open, and maybe add a label "low priority"? Or otherwise, if you want it close, what about adding a new column "Review later" on [the downgrades project](https://github.com/rectorphp/rector/projects/3) and place it there?

TomasVotruba:
Low priority is basically closed, to prevent issues overflow. That's what we have now.
We should get under 30 active issues, so it's clear what is important and what not. Also for contributors.

PR #4371, opened 2020-10-08
[Downgrade PHP 7.3] List reference assignment

### LINKED ISSUE #4315: [Downgrade PHP 7.3] `list()` and array destructuring reference assignment
# Feature Request

Downgrade new feature from PHP 7.3 to its equivalent PHP 7.2 code. [RFC with explanation](https://wiki.php.net/rfc/list_reference_assignment)

This feature must be handled with care. Check in the RFC all the nuances, and if there is any instance where downgrading is not possible, identify that case in the rule and throw a `DowngradeNotPossibleException`.
 
## Diff 

```diff
$array = [1, 2];
-list($a, &$b) = $array;
+$a = $array[0];
+$b =& $array[1];

/* Note; []= syntax works the same, so the following is functionally equivalent to the example */
-[$a, &$b] = $array;
+$a = $array[0];
+$b =& $array[1];
```

leoloso:
Fixes #4315

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -149,6 +149,7 @@
             "Rector\\NetteCodeQuality\\": "rules/nette-code-quality/src",
             "Rector\\DowngradePhp71\\": "rules/downgrade-php71/src",
             "Rector\\DowngradePhp72\\": "rules/downgrade-php72/src",
+            "Rector\\DowngradePhp73\\": "rules/downgrade-php73/src",
             "Rector\\DowngradePhp74\\": "rules/downgrade-php74/src",
             "Rector\\DowngradePhp80\\": "rules/downgrade-php80/src",
             "Rector\\SymfonyPhpConfig\\": "rules/symfony-php-config/src"
@@ -239,6 +240,7 @@
             "Rector\\NetteCodeQuality\\Tests\\": "rules/nette-code-quality/tests",
             "Rector\\DowngradePhp71\\Tests\\": "rules/downgrade-php71/tests",
             "Rector\\DowngradePhp72\\Tests\\": "rules/downgrade-php72/tests",
+            "Rector\\DowngradePhp73\\Tests\\": "rules/downgrade-php73/tests",
             "Rector\\DowngradePhp74\\Tests\\": "rules/downgrade-php74/tests",
             "Rector\\DowngradePhp80\\Tests\\": "rules/downgrade-php80/tests",
             "Rector\\SymfonyPhpConfig\\Tests\\": "rules/symfony-php-config/tests",
--- a/config/set/downgrade-php73.php
+++ b/config/set/downgrade-php73.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
+    $services = $containerConfigurator->services();
+};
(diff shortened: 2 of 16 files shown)


leoloso:
@TomasVotruba I need your input here.

I'm checking if this use case is downgradable:

```php
$array = [[1, 2], [3, 4]];
foreach ($array as list(&$a, $b)) {
    $a = 7;
}
var_dump($array)
/*
array(2) {
  [0]=>
  array(2) {
    [0]=>
    int(7)
    [1]=>
    int(2)
  }
  [1]=>
  array(2) {
    [0]=>
    &int(7)
    [1]=>
    int(4)
  }
}
*/
```

According to the [RFC](https://wiki.php.net/rfc/list_reference_assignment) this is not downgradable:

> The predominant advantage of adding support for this is that it allows you to use reference assignment for multiple variables at once, **which is not currently possible**. [...] the advantage here is that **you can reference assign some, but not all of the variables in list()**.

After running this PHP code, the reference exists on `$array[1][0]` since `$a` is still in scope after the `foreach()`. This part **cannot be reproduced in PHP 7.2** 

If we didn't care about the references, but only about the final values, this is doable:

```php
$array = [[1, 2], [3, 4]];
foreach ($array as &$item) {
    $item[0] = 7;
}
dump($array);
/*
[0]=>
  array(2) {
    [0]=>
    int(7)
    [1]=>
    int(2)
  }
  [1]=>
  &array(2) {
    [0]=>
    int(7)
    [1]=>
    int(4)
  }
}
*/
```

Notice how the reference at the end points to the 2nd item in the array, and not to the 1st element of that item.

Hence, the final references cannot be recreated, but the values can.

So, what to do? I see 2 options:

- Downgrade it, explaining this situation clearly in the documentation
- Throw a `DowngradeNotPossibleException` (#4316)

Or maybe a bit more complex:

- Downgrade it, unless variable `$a` is still referenced after running the code, if so throw `DowngradeNotPossibleException`

What do you think?

leoloso:
I thought a better alternative:

- Allow to downgrade the code, if enabled by a configuration entry `DANGEROUSLY_DOWNGRADE_LIST_REFERENCE_ASSIGNMENT_IN_FOREACH`
- By default, it is disabled
- If disabled, and that code is found, throw a `DangerousDowngradePossibleByConfigurationOnlyException`, and there it shows the error message explaining that values are recreated, but references are not

leoloso:
Or maybe we could create a new set `DANGEROUS_DOWNGRADE_PHP73`, and extract this as a separate rule `DangerousDowngradeListReferenceAssignmentInsideForeachRector`

leoloso:
Otherwise this rule is ready. So I'll remove it from draft, and you can decide if to already ship it

TomasVotruba:
I think we can go for low hanging fruit now and test it in the wild. Just not to waste time on problem that might have 1 person from million.

If it will cause troubles, we can improve it :)

TomasVotruba:
Thank you :+1:

leoloso:
Ok, but we need to discuss this as a concept. This is just one example of a functionality that cannot be downgraded without side-effects. I actually think that having a set `DANGEROUS_DOWNGRADE_PHP73` is a pretty good idea. We'll also have `DANGEROUS_DOWNGRADE_PHP74` and `DANGEROUS_DOWNGRADE_PHP80` too, for instance for `WeakMap` and others.

We can talk about it on our next meeting.

Btw, if you can generate a tag that would be cool, so I can already publicize this new set

TomasVotruba:
Set names makes sense :+1: 

I'll release it today

PR #4374, opened 2020-10-09
Fixes config/set link in readme

samsonasik:
The link previously got 404 page. I think root `/config/set` link is enough since there are multiple set in per-each category, eg: symfony (symfony-autowire.php, symfony-codequality.php, etc).


TomasVotruba:
Thank you

PR #4375, opened 2020-10-09
Downgrade php74 array spread

leoloso:
Fixes #4165

Issue #4165, opened 2020-09-08
[Downgrade PHP 7.4] Array spread operator


leoloso:
# Feature Request

Downgrade new feature from PHP 7.4 to its equivalent PHP 7.3 code. [RFC](https://wiki.php.net/rfc/spread_operator_for_array)

## Diff 

```diff
$parts = ['apple', 'pear'];
-$fruits = ['banana', 'orange', ...$parts, 'watermelon'];
+$fruits = array_merge(['banana', 'orange'], $parts, ['watermelon']);
```

Watch out the [edge case where the input is empty](https://stitcher.io/blog/new-in-php-74#array_merge-without-arguments-upgrading):

```diff
-$merged = array_merge(...$arrayOfArrays);
+$merged = array_merge([], []);
```

leoloso:
From the [RFC](https://wiki.php.net/rfc/spread_operator_for_array#advantages_over_array_merge):

> `array_merge` only supports array, while spread operator also supports objects implementing `Traversable`.

So the spread can either be an array or an iterable, and it must end up always being an array. We do that with `is_array($parts) ? $parts : iterator_to_array($parts)`:

```diff
$parts = ['apple', 'pear'];
-$fruits = ['banana', 'orange', ...$parts, 'watermelon'];
+$fruits = array_merge(['banana', 'orange'], is_array($parts) ? $parts : iterator_to_array($parts), ['watermelon']);
```

leoloso:
We can also spread a method call:

```php
function getArr() {
  return ['a', 'b'];
}
$arr6 = [...getArr(), 'c']; //['a', 'b', 'c']
```

We can't execute it twice inside the ternary, as it could have side effects (such as setting some global flag). So we first extract its execution into a variable, and then use the variable inside `array_merge`:

```php
$unpacked = getArr();
$arr6 = array_merge([is_array($unpacked) ? $unpacked : iterator_to_array($unpacked), ['c']]; //['a', 'b', 'c']
```

PR #4380, opened 2020-10-09
[Downgrade PHP 7.4] Array spread - don't use ternary for arrays, or PHPStan complains

leoloso:
New issue came up related to #4165 (after implementing PR #4375)

I run PHPStan to validate that the downgraded code works in the target PHP version. 

After downgrading the code, I might get this:

```php
$parts = ['apple', 'pear'];
$fruits = array_merge(['banana', 'orange'], is_array($parts) ? $parts : iterator_to_array($parts), ['watermelon']);
```

And PHPStan throws an error over this code:

```
Else branch is unreachable because ternary operator condition is always true.
```

Bleh.

So I fixed it! Now, when we know the type is `ConstantArrayType`, it prints this:

```php
$parts = ['apple', 'pear'];
$fruits = array_merge(['banana', 'orange'], $parts, ['watermelon']);
```

Much nicer! 🥳 

---

But then, when it's not `ConstantArrayType`, I don't know what it is, so it may still be an array or an iterator. So right now this code:

```php
$arr7 = [...new ArrayIterator(['a', 'b', 'c'])];
```

Becomes:

```php
$item0Unpacked = new ArrayIterator(['a', 'b', 'c']);
$arr7 = array_merge(is_array($item0Unpacked) ? $item0Unpacked : iterator_to_array($item0Unpacked));
```

And now PHPStan complains with:

```
Call to function is_array() with ArrayIterator<int, string> will always evaluate to false.
```

Bleh.

So I had an idea: let's print `/** @phpstan-ignore-next-line */` before that line, so PHPStan ignores it! I want to produce this output:

```php
$item0Unpacked = new ArrayIterator(['a', 'b', 'c']);
/** @phpstan-ignore-next-line */
$arr7 = array_merge(is_array($item0Unpacked) ? $item0Unpacked : iterator_to_array($item0Unpacked));
```

So I added this code in `DowngradeArraySpreadRector`:

```php
$commentableNode = $this->commentableNodeResolver->resolve($newNode);
$commentableNode->setAttribute(AttributeKey::COMMENTS, [new Comment('/** @phpstan-ignore-next-line */')]);
```

But it doesn't work, the comment is not printed in the output. I also tried with this:

```php
$this->addComment($newNode, '/** @phpstan-ignore-next-line */');
```

It doesn't work.

What's going on? Can add a comment?

---

Btw, tests do not have `/** @phpstan-ignore-next-line */` yet, that's why they are succeeding


leoloso:
> Can add a comment?

I think no need anymore 🥳 

I'm now inferring the type always, and if I know if it is `array` or `Traversable` then no need for the ternary. This is much cleaner code!

TomasVotruba:
Nice work :+1:

TomasVotruba (approved):

PR #4395, opened 2020-10-11
[SOLID] Change if && to early return (process nested ifs)

dobryy:
Follow-up: https://github.com/rectorphp/rector/pull/4344#discussion_r502641921 and https://github.com/rectorphp/rector/pull/4344#discussion_r502642814
Related to: #4071


rules/sensio/src/TypeDeclaration/ReturnTypeDeclarationUpdater.php
```diff
@@ -86,9 +86,13 @@ private function updatePhp(ClassMethod $classMethod, string $className): void
...
+            if ($returnTypeName === null) {
                 return;
             }
+            if (!is_a($returnTypeName, $className, true)) {
+                return;
+            }
+            return;
         }
 
         $classMethod->returnType = new FullyQualified($className);
```
dobryy: @TomasVotruba this looks like an edge case that has two options to be solved:
1. Skip `if` if it contains `return;` (void) statement
2. Process if and replace `if`'s statements with `continue;` statement

Let me know what is the preferred solution in your opinion.

TomasVotruba: I think 1. is right in this case.
What do you think?

dobryy: For this particular rule option #1 makes more sense to me as well.

dobryy:
@TomasVotruba If it looks fine to you, this PR is ready for merging.

TomasVotruba (approved):

TomasVotruba:
Looks good to me, thank you

PR #4407, opened 2020-10-14
Add a tag value node for Doctrine Embedded tags

matthiasnoback:
Although I'm not submitting an actual rule that uses this, at least it makes the list of supported Doctrine annotations more complete. I also needed this one for a project I'm working on now.

When running the tests on my machine PHPStan produces 89 errors but I guess they have to do with the PHP runtime I'm using locally.


TomasVotruba:
This looks very good, thank you :+1:

TomasVotruba:
I'ved rebased it to remove all PHPStan error

and fixed getter name  in https://github.com/rectorphp/rector/pull/4424

matthiasnoback:
Thanks for merging!

PR #4408, opened 2020-10-14
Downgrade Rector to PHP 7.1 - Use same signature for `prettyPrintFile` as in PrettyPrinterAbstract

leoloso:
Param `$stmts` in function `prettyPrintFile` is of type `array` in `PhpParser\PrettyPrinterAbstract`, but had no type in Rector's `BetterStandardPrinter`. So I added the type, to make it work under PHP 7.1, or it throws this warning:

```
PHP Warning:  Declaration of Rector\Core\PhpParser\Printer\BetterStandardPrinter::prettyPrintFile($stmts): string should be compatible with PhpParser\PrettyPrinterAbstract::prettyPrintFile(array $stmts): string in /user/Temporary/graphql-api-php71/vendor/rector/rector/src/PhpParser/Printer/BetterStandardPrinter.php on line 34
```

Also check the docblock changed: it supported `Node|null` before, now not anymore.

@TomasVotruba Please check if this is right


leoloso:
Well, the test is not passing:

```
TypeError: Argument 1 passed to Rector\Core\PhpParser\Printer\BetterStandardPrinter::prettyPrintFile() must be of the type array, object given, called in /home/runner/work/rector/rector/rules/removing-static/src/Printer/FactoryClassPrinter.php on line 57
```

So there's more work to do...

TomasVotruba:
> Well, the test is not passing:

It's better to run tests locally before commiting to save time on tis

leoloso:
> It's better to run tests locally before commiting to save time on this

Yeah, sorry about that, I had imagined it would work directly. I'll check all those calls, try to fix them using the right signature

TomasVotruba:
Thank you :+1: 

I didn't realized the method signature can be changed that much :) 

I've added issue for PHPStan rule that will prevent this in the future: https://github.com/symplify/symplify/issues/2373

TomasVotruba (approved):

leoloso:
This is not a bug, it's a feature from PHP 7.2:

https://www.php.net/manual/en/migration72.new-features.php#migration72.new-features.param-type-widening

That's why this error only emerged when trying out Rector in PHP 7.1

TomasVotruba:
So are traits, but they overcomplicate simple code :)
We should avoid ambiguous code/issues like these

PR #4410, opened 2020-10-14
[SOLID] Add MoveVariableDeclarationNearReferenceRector [needs more work before use]

TomasVotruba:


rules/solid/src/Rector/Variable/MoveVariableDeclarationNearReferenceRector.php
```diff
@@ -90,27 +98,36 @@ public function refactor(Node $node): ?Node
...
     {
-        return $this->betterNodeFinder->findFirst(
+        $desiredVariableName = $this->getName($desiredVariable);
+
+        $firstVariableUsage = $this->betterNodeFinder->findFirst(
             (array) $parentScope->getStmts(),
-            function (Node $node) use ($variable, $assign): bool {
-                return $this->isVariable($node) &&
-                    ! $this->isOriginalAssign($node, $assign) &&
-                    $this->betterStandardPrinter->areNodesEqual($node, $variable);
```
TomasVotruba: I'm not sure where the problem was. Maybe here. This is very hard to read for me, so I used early-returns. It might have fixed the issue.

TomasVotruba: I tested to use `areNodesEqual()` instead of `isName()`, but it had no effect.

rules/solid/tests/Rector/Variable/MoveVariableDeclarationNearReferenceRector/Fixture/variable_in_closure.php.inc
```diff
@@ -16,7 +16,7 @@ function variableInClosure()
 
 function variableInClosure()
 {
-    $closure = function () {
+    return function () {
```
TomasVotruba: This is just for testing purposes, it had no effect. The original test case is covered here:
https://github.com/rectorphp/rector/pull/4410/commits/97a3004237cf923373bceb0da3d6ad749ce13bc2#diff-7cabfedb3086541a6f5995aed5be677cc009cb92d09cd5560ede872dc4f67947

TomasVotruba (commented):
Ha, got more information :)

So the `$closure =` is the problem for some reason.
`return ` works correctly

packages/post-rector/src/Collector/NodesToAddCollector.php
```diff
@@ -103,7 +103,7 @@ public function clearNodesToAddBefore(Node $node): void
 
     private function resolveNearestExpressionPosition(Node $node): string
     {
-        if ($node instanceof Expression) {
+        if ($node instanceof Expression || $node instanceof Stmt) {
```
TomasVotruba, line 106: @afilina This was the goled spot :laughing:

afilina: Thanks for looking into it! There's so much going on under the hood that it's intimidating :)

TomasVotruba: Don't worry, take your time. It took me about 6 month to start to understand php-parser and AST.
3 more months to make first usable version Rector around it.

It might take to write 10-15 Rector rules to get some confidence working with it. But then, nothing is imposible :) everything is matter of time

afilina:
Should I close my copy of the PR to avoid confusion? We can merge this one as soon as the closure case is fixed and then worry about the `variable_with_usage_in_multiple_levels` a bit later.

TomasVotruba:
> @TomasVotruba Should I close my copy of the PR to avoid confusion? We can merge this one as soon as the closure case is fixed and then worry about the variable_with_usage_in_multiple_levels a bit later.

That would be great!

Could you hook into this PR, ignore the fixture (e.g. add `_skip` suffix) and make the rest of CI pass?

I can merge it right away then

TomasVotruba:
Merging as it is to prevent from further stale.

<br>

There are some false positives that need to excluded:

- https://github.com/rectorphp/rector/commit/fe8298b10ac890a25d9e8078e9fa135ab8bbc51d
- https://github.com/rectorphp/rector/commit/5a986acdf3304d8a390ff4ce87ff7ffcf1279583
- https://github.com/rectorphp/rector/commit/ff9548f5e6406d05c5e7327b83367e145e57c3a4

rules/code-quality/src/Naming/MethodCallToVariableNameResolver.php
```diff
@@ -63,14 +63,13 @@ private function getVariableName(MethodCall $methodCall, string $methodCallVarNa
...
         }
-
-        $fallbackVarName = $this->getFallbackVarName($methodCallVarName, $methodCallName);
         $argValue = $methodCall->args[0]->value;
         if ($argValue instanceof ClassConstFetch && $argValue->name instanceof Identifier) {
             return $this->getClassConstFetchVarName($argValue, $methodCallName);
         }
 
         if ($argValue instanceof String_) {
+            $fallbackVarName = $this->getFallbackVarName($methodCallVarName, $methodCallName);
```
samsonasik, line 72: @TomasVotruba this moving variable is a bug, as it used after `if` at last.

TomasVotruba: Thanks for the info.

The rule should not run at all, as there are many cases to be covered.
I'll try to found out why it's being run

PR #4415, opened 2020-10-15
Issue #4413: Do not refactor iterator_to_array.

### LINKED ISSUE #4413: ArraySpreadInsteadOfArrayMergeRector should not convert iterator_to_array to spread operator
# Bug Report

| Rector version | 0.8.24              |
| Related issues | #4333 |

I'm writing a lazy PHP collection library working with iterator and generators. This library is available for `php >= 7.1.3`.
I'd like to publish a new major release for `php > 7.4` and I used rector to convert the code and rewrite some parts, especially the closures.

After that, I tried to run the tests and many tests were failing because `iterator_to_array($var)` was converted to `[...$var]`.
While this seems to be valid most of the time, it can be wrong as well when `$var` is an iterator containing duplicated key values.

Context: https://3v4l.org/YSABX

Code:

```php
<?php

$generator = static function()
{
    yield 1 => 1;
    yield 2 => 2;
    yield 1 => 3;
    yield 2 => 4;
};

var_dump(iterator_to_array($generator()));
var_dump([...$generator()]);
```

## Expected Behaviour

I think in this case, while in most of the cases it will work, rector should not change `iterator_to_array()` to spread operator.
Adding a check prior converting would be too much of a hassle and not efficient.

Also, the spread operator cannot unpack with non-integer keys while `iterator_to_array()` does: https://3v4l.org/uECeu

drupol:
This PR:

* [x] Disable and remove code that trigger the refactoring of `iterator_to_array`. 
* [x] Update the test.

Fix #4413 #4333


rules/php74/tests/Rector/FuncCall/ArraySpreadInsteadOfArrayMergeRector/Fixture/iterator_to_array.php.inc
```diff
@@ -16,11 +16,11 @@ class IteratorToArray
...
 
-class IteratorToArray
+final class IteratorToArray
 {
     public function run($items)
     {
-        $values = [...$items];
+        $values = iterator_to_array($items);
     }
 }
```
samsonasik, line 25: The second code block start with `-----` can be removed entirely if code change is skipped. You may can also rename the fixture to `skip_iterator_to_array.php.inc`.

drupol: Thanks, fixed!

TomasVotruba (approved):

TomasVotruba:
Thank you

PR #4441, opened 2020-10-18
[CodeQuality] Register IssetOnPropertyObjectToPropertyExistsRector to code-quality config set

samsonasik:


samsonasik:

rules/code-quality/src/Rector/Foreach_/ForeachToInArrayRector.php
```diff
@@ -139,7 +139,7 @@ public function refactor(Node $node): ?Node
 
     private function shouldSkipForeach(Foreach_ $foreach): bool
     {
-        if (isset($foreach->keyVar)) {
+        if (property_exists($foreach, 'keyVar') && $foreach->keyVar !== null) {
```
TomasVotruba, line 142: The `$foreach->keyVar` always exists - https://github.com/nikic/PHP-Parser/blob/1d1bc8a36459d67acef508b952dad097960b110e/lib/PhpParser/Node/Stmt/Foreach_.php#L12



```suggestion
        if ($foreach->keyVar !== null) {
```

TomasVotruba: Could you update the `IssetOnPropertyObjectToPropertyExistsRector` so it handle this case too?

samsonasik: Changing that will make the following fixture code not changed as always exists as well:

```php
class SomeClass
{
    private $x;
    private $y;

    public function run()
    {
        isset($this->x);
        isset($this->y);
        isset($this->x) && isset($this->y);
    }
}
```

Do you have any suggestion?

TomasVotruba: I think such case should be changed like this:

```diff
-isset($this->x);
+$thix->x !== null;
```

samsonasik: ok, I will try

TomasVotruba: It's a rough guess... testing on Rector's code will tell more :)

samsonasik: implemented.

samsonasik: I think for object (no ThisType) that always has a property need to be changed to `$obj->var !== null` as well for consistency.

samsonasik: implemented.

samsonasik:

TomasVotruba:
Thank you :clap:

Issue #4396, opened 2020-10-12
Typed arrays become mixed[]


szepeviktor:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.8.19             |
| Installed as   | composer dev dependency                      |

Original https://travis-ci.com/github/szepeviktor/phpstan-higher-order-collections/builds/189364074#L771-L776

## Minimal PHP Code Causing Issue

```diff
     /**
-     * @var PropertyReflection[]
+     * @var mixed[]
      */
-    private array $reflections;
+    private $reflections;
```

## Expected Behaviour

```diff
-    private array $reflections;
+    private $reflections;
```

Could Rector leave the original type in the docblock?

TomasVotruba:
Hi, we need failing demo link, as the context code might be causing this.

szepeviktor:
There it is: https://travis-ci.com/github/szepeviktor/phpstan-higher-order-collections/builds/189364074#L771-L776

Original source code: https://github.com/Sustainabil-IT/phpstan-higher-order-collections/tree/main/src

TomasVotruba:
Thanks for quick reply

By demo link i ment https://getrector.org/demo

We need isolate dependencies as much as possible, to make a test fixture of it

szepeviktor:
I see! getrector.org demo!
I need time to reduce the code to a demo.

szepeviktor:
It is more complicated than I thought.
https://getrector.org/demo/1fa64649-f887-4b74-80ce-2239a3138dd1

Please teach me how to write code not to get mixed as property type.

szepeviktor:
Just a thought: that class is insantiated only here https://github.com/Sustainabil-IT/phpstan-higher-order-collections/blob/9b4fc26b485381ca789dcc9b018430df821688ef/src/Extensions/HigherOrderCollectionMethodExtension.php#L22-L29

The method filling that array (that later gets mixed type) uses generics https://github.com/Sustainabil-IT/phpstan-higher-order-collections/blob/9b4fc26b485381ca789dcc9b018430df821688ef/src/Extensions/BaseHigherOrderCollectionExtension.php#L26-L38
Maybe that??

ComiR:
I could not get a demo to run (f.e. https://getrector.org/demo/5a1487bc-2dff-4dad-a1fb-adf079bcd892#result), but in #4443 I have some examples which show the problem.

There probably should be further discussion on how to handle differing types.

[...]

PR #4443, opened 2020-10-18
Infer var type annotation only if none exists

ComiR:
solves #4396 

Caveat: It keeps *all* var type annotations, even if they conflict with the "real" type.


TomasVotruba:
Hi,

thanks for PR!

Could you decouple test fixture to its own file? It easier to review smaller scope and debug it in the future.

ComiR:
Breaking up the test cases even more would only make sense if they were handled differently. As of now, the only discriminator is the existence of a var type annotation.

Idk why the build/checks failed for this commit - no relevant changes to the passing commit before…

TomasVotruba:
The CI is failing. Could you rebase on `master`? That should help

ComiR:
The method `getVarTagValue()` was renamed to `getVarTagValueNode()` 😩

TomasVotruba:
Thank you :clap:

PR #4447, opened 2020-10-19
[CI] add demo for build php 7.1 version

TomasVotruba:


.github/workflows/build_php71.yaml
```diff
@@ -0,0 +1,31 @@
...
+            -   uses: actions/checkout@v2
+
+            # see https://github.com/shivammathur/setup-php
+            -   uses: shivammathur/setup-php@v1
+                with:
+                    php-version: 7.2
+                    coverage: none
+
+            -   run: coposer install --no-dev
+            -   run: vendor/bin/rector process src vendor --set downgrade-php71
```
TomasVotruba, line 23:
```bash
vendor/bin/rector process src/Core --set downgrade-php71 

# then 

vendor/bin/rector process src --set downgrade-php71 
```

leoloso:
@TomasVotruba do you have "implicit" rules, which must be executed always? Or otherwise is there a bug?

For instance, when on the rector project root I run this:

```
bin/rector process vendor/symfony/yaml/Exception/ExceptionInterface.php --set=downgrade-php72 --dry-run --ansi
```

it is executing a rule which does not belong to `DOWNGRADE_PHP72`, to import `Throwable` and replace `\Throwable`:

Screenshot 2020-10-26 at 3 35 37 PM


The name of this applied rule is not mentioned at the end of the run.

There is no need to run these implicit rules for the downgrade... Is there a way to avoid it?

leoloso:
@TomasVotruba since we're executing the downgrade sets purely through CLI, I think we need two more options:

- `exclude-path` <= can be declared several times
- `php-version-features` or `target-php-version`

Eg:

```
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector --set=downgrade-php72 --exclude-path=/Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/bin/* --exclude-path=/Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/ci/* --exclude-path=/Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/docs/* --exclude-path=/Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/tests/* --exclude-path=/Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/**/tests/* --exclude-path=/Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/packages/rector-generator/templates/* --target=php-version=7.1 --dry-run --ansi
```

What do you think?

leoloso:
The bash script to retrieve all packages to downgrade, version by version, is looking good.

It is doing this:

- Remove DEV dependencies
- Find out which dependencies must be downgraded for each PHP version
- Add DEV dependencies again
- Execute Rector on the source path for each package, using the specific `--set`

downgrade-bash

TomasVotruba:
> @TomasVotruba do you have "implicit" rules, which must be executed always? Or otherwise is there a bug?
> 

This is handled by:

```php
# rector.php
// ...

// ...
```

leoloso:
@TomasVotruba For testing, I want to run ALL sets, to make sure they all work, but `--set` accepts only 1 param, right?

So doing this will execute only the last one, right?:

```
bin/rector process src --dry-run \
  --set defluent \
  --set action-injection-to-constructor-injection \
  --set laravel-array-str-functions-to-static-call \
  --set cakephp30 \
  --set cakephp34 \
  --set cakephp35 \
  --set cakephp36 \
  --set cakephp37 \
  --set cakephp38 \
  --set cakephp40 \
  --set cakephp41 \
  --set cakephp-fluent-options \
  --set codeigniter-40 \
  --set code-quality \
  --set coding-style \
  --set coding-style-advanced \
  --set contributte-to-symfony \
  --set dead-classes \
  --set dead-code \
  --set doctrine25 \
  --set doctrine-behaviors-20 \
  --set doctrine-code-quality \
  --set doctrine-common-20 \
  --set doctrine-dbal-210 \
  --set doctrine-dbal-30 \
  --set doctrine-gedmo-to-knplabs \
  --set doctrine-id-to-uuid-step-1 \
  --set doctrine-id-to-uuid-step-2 \
  --set doctrine-id-to-uuid-step-3 \
  --set doctrine-id-to-uuid-step-4 \
  --set doctrine-id-to-uuid-step-5 \
  --set doctrine-repository-as-service \
  --set doctrine-services \
  --set easy-admin-bundle20 \
  --set framework-extra-bundle-40 \
  --set framework-extra-bundle-50 \
  --set gmagick_to_imagick \
  --set jms-decouple \
  --set kdyby-events-to-contributte-event-dispatcher \
  --set kdyby-to-symfony \
  --set kdyby-translator-to-contributte-translation \
  --set laravel50 \
  --set laravel51 \
  --set laravel52 \
  --set laravel53 \
  --set laravel54 \
  --set laravel55 \
  --set laravel56 \
  --set laravel57 \
  --set laravel58 \
  --set laravel60 \
  --set laravel-static-to-injection \
  --set mockista-to-mockery \
  --set monolog20 \
  --set mysql-to-mysqli \
  --set naming \
  --set nette-30 \
  --set nette-30-param-types \
  --set nette-30-return-types \
  --set nette-30-dependency-injection \
  --set nette-code-quality \
  --set nette-control-to-symfony-controller \
  --set nette-forms-to-symfony \
  --set nette-30-param-types \
  --set nette-30-return-types \
  --set nette-tester-to-phpunit \
  --set nette-to-symfony \
  --set nette-utils-code-quality \
  --set order \
  --set performance \
  --set phalcon40 \
  --set phpexcel-to-phpspreadsheet \
  --set phpspec30 \
  --set phpspec40 \
  --set phpspec-to-phpunit \
  --set phpstan \
  --set phpunit80-dms \
  --set phpunit40 \
  --set phpunit50 \
  --set phpunit60 \
  --set phpunit70 \
  --set phpunit75 \
  --set phpunit80 \
  --set phpunit90 \
  --set phpunit91 \
  --set phpunit-code-quality \
  --set phpunit-exception \
  --set phpunit-injector \
  --set phpunit-mock \
  --set phpunit-specific-method \
  --set phpunit-yield-data-provider \
  --set php52 \
  --set php53 \
  --set php54 \
  --set php55 \
  --set php56 \
  --set php70 \
  --set php71 \
  --set php72 \
  --set php73 \
  --set php74 \
  --set php80 \
  --set php-code-sniffer30 \
  --set php-di-decouple \
  --set privatization \
  --set psr-4 \
  --set safe07 \
  --set solid \
  --set swiftmailer60 \
  --set symfony26 \
  --set symfony28 \
  --set symfony30 \
  --set symfony31 \
  --set symfony32 \
  --set symfony33 \
  --set symfony34 \
  --set symfony40 \
  --set symfony41 \
  --set symfony42 \
  --set symfony43 \
  --set symfony44 \
  --set symfony50 \
  --set symfony50-types \
  --set symfony-code-quality \
  --set symfony-constructor-injection \
  --set symfony-phpunit \
  --set symfony-autowire \
  --set twig112 \
  --set twig127 \
  --set twig134 \
  --set twig140 \
  --set twig20 \
  --set twig240 \
  --set twig-underscore-to-namespace \
  --set type-declaration \
  --set unwrap-compat
```

leoloso:
> @TomasVotruba For testing, I want to run ALL sets, to make sure they all work, but `--set` accepts only 1 param, right?

Well, I guess it's the same as executing `rector` for each set

TomasVotruba:
> @TomasVotruba For testing, I want to run ALL sets, to make sure they all work, but --set accepts only 1 param, right?

Just 1. It's better to use config for that.

Also, testing all sets is not neccessary, as main error will pop-up with just couple of sets. I'd go wit `PHP_x` only

TomasVotruba:
We should have multiple use of `--set x --set y` thrown an exception, with suggesion to use a config instead.

leoloso:
> @TomasVotruba since we're executing the downgrade sets purely through CLI, I think we need two more options:
> 
>     * `exclude-path` <= can be declared several times
> 
>     * `php-version-features` or `target-php-version`
> 
> 
> Eg:
> 
> ```
> bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector --set=downgrade-php72 --exclude-path=/Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/bin/* --exclude-path=/Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/ci/* --exclude-path=/Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/docs/* --exclude-path=/Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/tests/* --exclude-path=/Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/**/tests/* --exclude-path=/Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/packages/rector-generator/templates/* --target=php-version=7.1 --dry-run --ansi
> ```
> 
> What do you think?

Waiting for your feedback on this question

leoloso:
> Also, testing all sets is not neccessary, as main error will pop-up with just couple of sets. I'd go wit `PHP_x` only

Is that true? If using autoloading, then the classes in a set might not be loaded if not tested, including a class which might have some leftover PHP 7.2 code. Or they are loaded always?

TomasVotruba:
> `exclude-path` <= can be declared several time

This option is better used in PHP config, e.g `rector.php`.

> `php-version-features` or `target-php-version`

`target-php-version` is defined in sets themselvesd, e.g. DONWGRADE_PHP_71 has target in the name

TomasVotruba:
> Is that true? If using autoloading, then the classes in a set might not be loaded if not tested, including a class which might have some leftover PHP 7.2 code. Or they are loaded always?

I'm not sure what you mean by autoloading. The idea is to do smoke testing with major sets, not all of them. Take the low handing fruit, 80/20 rule.

leoloso:
> > `exclude-path` <= can be declared several time
> 
> This option is better used in PHP config, e.g `rector.php`.

But then, I'd need to generate a `rector-downgrade.php` on runtime! Because different packages might have different paths to exclude.

This is how I'm dealing with this now: every package can define its own set ([here](
https://github.com/rectorphp/rector/blob/263c5fbcacb371348c81e20888e9b2221617a1c9/.github/workflows/scripts/downgrade_packages.sh#L45)):

```sh
declare -A package_excludes=( \
    ["rector/rector"]="$(pwd)/.docker/*;$(pwd)/.github/*;$(pwd)/bin/*;$(pwd)/ci/*;$(pwd)/docs/*;$(pwd)/tests/*;$(pwd)/**/tests/*;$(pwd)/packages/rector-generator/templates/*" \
)
```

If no entry is declared for that package, only then [I exclude `**/tests/*`](https://github.com/rectorphp/rector/blob/263c5fbcacb371348c81e20888e9b2221617a1c9/.github/workflows/scripts/downgrade_packages.sh#L148):

```sh
    if [ -z $exclude ]
    then
        exclude="${path_to_downgrade}/**/tests/*"
    fi
```

For rector, excluding `/packages/rector-generator/templates/*` is mandatory, or the process fails.

leoloso:
Ok well, I can add a `rector-downgrade.php` containing all exclude paths for all packages. It won't be as nice, but it will work

leoloso:
To provide the exclude path, I've had to create 2 config files:

- `rector-downgrade-rector.php` <= For Rector
- `rector-downgrade-dependency.php` <= For its dependencies

The reason is that Rector needs to exclude `/vendor/`, but then no dependency can be processed using that config file.

For the dependencies, I made them all exclude `/tests/`. Next, I will run Rector on each of them, and see if they need additional excludes.

leoloso:
This is the list of Rectors for all dependencies to downgrade. I'll go one by one, fixing their possible errors:

```
Running set downgrade-php73 for package sebastian/diff on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/sebastian/diff
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/sebastian/diff --set=downgrade-php73 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package rector/rector on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector --set=downgrade-php72 --config=rector-downgrade-rector.php --dry-run --ansi

Running set downgrade-php72 for package doctrine/inflector on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/doctrine/inflector
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/doctrine/inflector --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package doctrine/lexer on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/doctrine/lexer
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/doctrine/lexer --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package migrify/migrify-kernel on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/migrify/migrify-kernel
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/migrify/migrify-kernel --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package migrify/php-config-printer on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/migrify/php-config-printer
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/migrify/php-config-printer --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package psr/event-dispatcher on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/psr/event-dispatcher
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/psr/event-dispatcher --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package sebastian/diff on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/sebastian/diff
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/sebastian/diff --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/cache on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/cache
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/cache --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/cache-contracts on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/cache-contracts
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/cache-contracts --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/config on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/config
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/config --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/console on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/console
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/console --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/dependency-injection on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/dependency-injection
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/dependency-injection --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/error-handler on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/error-handler
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/error-handler --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/event-dispatcher on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/event-dispatcher
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/event-dispatcher --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/event-dispatcher-contracts on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/event-dispatcher-contracts
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/event-dispatcher-contracts --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/filesystem on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/filesystem
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/filesystem --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/finder on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/finder
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/finder --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/http-client-contracts on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/http-client-contracts
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/http-client-contracts --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/http-foundation on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/http-foundation
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/http-foundation --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/http-kernel on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/http-kernel
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/http-kernel --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/process on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/process
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/process --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/service-contracts on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/service-contracts
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/service-contracts --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/string on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/string
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/string --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/var-dumper on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/var-dumper
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/var-dumper --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/var-exporter on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/var-exporter
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/var-exporter --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symfony/yaml on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/yaml
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symfony/yaml --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symplify/autowire-array-parameter on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/autowire-array-parameter
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/autowire-array-parameter --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symplify/composer-json-manipulator on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/composer-json-manipulator
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/composer-json-manipulator --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symplify/console-color-diff on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/console-color-diff
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/console-color-diff --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symplify/easy-testing on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/easy-testing
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/easy-testing --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symplify/package-builder on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/package-builder
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/package-builder --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symplify/set-config-resolver on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/set-config-resolver
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/set-config-resolver --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symplify/smart-file-system on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/smart-file-system
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/smart-file-system --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi

Running set downgrade-php72 for package symplify/symplify-kernel on path(s) /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/symplify-kernel
bin/rector process /Users/mac/GitRepos/GitHub/Libraries/rectorphp/rector/vendor/symplify/symplify-kernel --set=downgrade-php72 --config=rector-downgrade-dependency.php --dry-run --ansi
```

leoloso:
There is a challenge concerning the current error in the CI:

```
Running set downgrade-php72 for package sebastian/diff on path(s) /home/runner/work/rector/rector/vendor/sebastian/diff
PHP Fatal error:  Declaration of Symfony\Contracts\EventDispatcher\EventDispatcherInterface::dispatch(object $event, ?string $eventName = NULL): object must be compatible with Psr\EventDispatcher\EventDispatcherInterface::dispatch($event) in /home/runner/work/rector/rector/vendor/symfony/event-dispatcher-contracts/EventDispatcherInterface.php on line 30
Rector downgrade failed on set downgrade-php72 for package sebastian/diff
```

This error happens because dependencies are downgraded independently from each other, and the list of packages is provided not in the loading order, but alphabetically.

Then, `psr/event-dispatcher` is being downgraded before `sebastian/diff`, however it is the latter package that depends on the former one (through package `symfony/event-dispatcher-contracts`).

So, an interface from `psr/event-dispatcher` is modified, and then this causes an error when processing `sebastian/diff`.

There are 2 possible solutions:

1. Obtain the packages in the strict loading order: `sebastian/diff` => `symfony/event-dispatcher-contracts` => `psr/event-dispatcher`
2. Downgrade all dependencies together

I will attempt the first option first.

leoloso:
Composer doesn't give the option to retrieve packages by their loading order. So I will try this strategy:

1. For each package, retrieve the list of "ancestor" packages:

```bash
composer why psr/event-dispatcher -r | cut -d' ' -f1 | awk '!a[$0]++'
```

For `psr/event-dispatcher`, this command produces:

```
rector/rector
migrify/migrify-kernel
symplify/package-builder
migrify/php-config-printer
symplify/autowire-array-parameter
symplify/composer-json-manipulator
symplify/console-color-diff
symplify/easy-testing
symplify/symplify-kernel
symplify/set-config-resolver
symfony/http-kernel
symfony/event-dispatcher
symfony/event-dispatcher-contracts
```

2. Have a double loop to downgrade packages, asking if all the ancestors for the package have already been downgraded, or let it keep iterating until next loop

leoloso:
I tried the strategy I mentioned above, but there is a problem: there are circular references.

For instance:

```
symplify/autowire-array-parameter => symplify/package-builder
symplify/package-builder => symplify/autowire-array-parameter
```

So right now not all packages get downgraded, because I can't obtain the order of execution.

To solve this, I'll pre-generate a list with an order that works, and hardcode it into the script, to use it in case of circular references, so the script will have a mix of automatic and manual behavior.

Otherwise, I'd have to execute all packages together in a single rector call, but I believe we want to avoid that unless there's no working alternative.

leoloso:
There is an issue with Symfony: package `symfony/expression-language` is always required for DEV: [in `dependency-injection`](https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/DependencyInjection/composer.json#L28) and a couple more.

That seems like a bug to me. For instance, `Symfony\Component\DependencyInjection\ContainerBuilder` references `ExpressionLanguage`... or is `ContainerBuilder` not expected be used for PROD?

To solve it, because Rector makes use of `ContainerBuilder`, I added `symfony/expression-language` as a PROD dependency.

leoloso:
@TomasVotruba the script is working beautifully 🎉 (the testing is failing, but I haven't worked on that yet):

- It calculates automatically what dependencies must be downgraded, for each PHP version
- It calculates in what order they must be downgraded
- It executes Rector on them, one by one
- Whenever there is a cluster of circular references (`package1` <=> `package2`, or `package1` => `package2` => `package3` => `package1`), it executes Rector on them all together
- As soon as any package fails, the process exits

You can see the log here: https://github.com/rectorphp/rector/pull/4447/checks?check_run_id=1335979951#step:4:1

Because everything is automatic, this script can easily be used not just to downgrade Rector, but to downgrade any project! 🥳 

Now I'll take my time off. When I'm back in a few weeks, I will work on polishing the script a bit:

- Extract duplicated code into functions
- Make variables more coherent (now they use camelCase and snake_case)
- Pass rector-downgrade-rector.php by argument, instead of hardcoded

And then also work on the testing

leoloso:
Actually, the script could be added to the GitHub Actions marketplace, so everyone can use it. That would be awesome!

leoloso:
@TomasVotruba I'm doing a final development before I take the time off: I found out why couldn't execute the tests, which was related to the new Composer v2's platform checks, so I fixed that.

But when testing, I'm now getting a new error:

```
The "--set" option does not exist.
```

Have you removed `--set` from the CLI? I need that to execute the set... or is this a bug?

leoloso:
@TomasVotruba I currently need 2 rector files: [`rector-downgrade-dependency.php`](https://github.com/rectorphp/rector/blob/e554a0a322b175dcad91dca3482c0e2d2c9d58ee/rector-downgrade-dependency.php) and [`rector-downgrade-rector.php`](https://github.com/rectorphp/rector/blob/e554a0a322b175dcad91dca3482c0e2d2c9d58ee/rector-downgrade-rector.php). By removing `--set`, I'll need to create 8 files, i.e. these 2 for every PHP version to downgrade to (7.1, 7.2, 7.3 and 7.4), and the code in them is exactly the same, just changing 1 line: the set.

And then, when PHP 8.1 comes out, we need to create yet another pair. It will slowly become unmanageable.

Then, why not keeping `--set` in the CLI? I think removing it adds a lot of bureaucracy, duplicated code, etc

TomasVotruba:
Here is the way to generate PHP rector.php config: 
https://github.com/rectorphp/rector/blob/master/packages/testing/src/PHPUnit/AbstractRectorTestCase.php#L513

Here is README: https://github.com/symplify/php-config-printer#2-full-config

```php
$rectorPhpConfig = $this->yamlToPhpConverter->convertYamlArray([
     'parameters' => [
           'sets' => [...]
     ]
]);
```

leoloso:
The current issue in the CI:

```
PHP Fatal error:  Declaration of Symfony\Component\Console\Style\SymfonyStyle::ask(string $question, ?string $default = NULL, $validator = NULL) must be compatible with Symfony\Component\Console\Style\StyleInterface::ask(string $question, ?string $default = NULL, ?callable $validator = NULL) in /home/runner/work/rector/rector/vendor/symfony/console/Style/SymfonyStyle.php on line 36
```

is happening because parameter type widening is not downgraded yet to PHP 7.1 (#4754). I'll work on it next

leoloso:
To run the parameter type widening (#4754) rector, we may not be able to run the downgrade script on a package-by-package basis (as done currently):

https://github.com/rectorphp/rector/issues/4754#issuecomment-738031609

If that's the case, I'll work on the alternative: downgrading the whole project together (and see how long it takes)

leoloso:
@TomasVotruba I see a beautiful ✅ mark!

😀👏🙏🙌🏻

So the script is "almost" ready. These are the next steps:

## Check the testing

In addition to [running all the tests (on PHP 7.4, not 7.1)](https://github.com/rectorphp/rector/blob/ca67be80886b8e8c15f69aab391d2eee15fa7d38/.github%2Fworkflows%2Fbuild_php71.yaml#L36), I wanted to [run PHPStan on PHP 7.1](https://github.com/rectorphp/rector/blob/ca67be80886b8e8c15f69aab391d2eee15fa7d38/.github%2Fworkflows%2Fbuild_php71.yaml#L64), to make sure there's no 7.2+ code left behind:

```yaml
-   run: vendor/bin/phpstan analyse --ansi --error-format symplify --level 0
```

But we can't do that, because Rector was downgraded for PROD, not DEV, so `Symplify\PHPStanRules\Rules\ForbiddenMethodCallOnTypeRule` still has 7.3 code and makes PHPStan fail.

So currently I'm just [running Rector](https://github.com/rectorphp/rector/blob/ca67be80886b8e8c15f69aab391d2eee15fa7d38/.github%2Fworkflows%2Fbuild_php71.yaml#L61):

```yaml
-   run: bin/rector process src --dry-run
```

This works but it's not exhaustive, since it's just testing the code for the default set and nothing else.

So we must discuss if this is enough, or what else must be done.

## Make it faster

The script took 7m min. I'll modify it, adding the option to downgrade all dependencies in a single call (instead of a call per dependency), as to avoid calculating the dependency order, which takes a bit of time.

## Remove (replace) all dependencies from `composer.json`:

Since downgraded Rector will already ship with its own `vendor/`, we must remove all dependencies, but still have them loaded in the autoloader. I've seen you've done this on Rector Prefixed already:

https://github.com/rectorphp/rector-prefixed/blob/master/composer.json

How did you do it? Can I reuse any code?

And then, the downgraded Rector needs to require the Symfony Polyfill libraries:

- [Symfony Polyfill / Php72](https://github.com/symfony/polyfill-php72/)
- [Symfony Polyfill / Php73](https://github.com/symfony/polyfill-php73/)

The 7.2 one is mandatory, since Symfony Console uses `stream_isatty`, and the polyfill provides it.

## Change the PHP requirement to 7.1 in `composer.json`

Once the step above is done, I can do `composer require php` in the CI running PHP 7.1, to change the `require` to PHP 7.1.

## Fix hacks?

[This code](https://github.com/rectorphp/rector/blob/ca67be80886b8e8c15f69aab391d2eee15fa7d38/.github%2Fworkflows%2Fbuild_php71.yaml#L58) feels a bit hacky:

```yaml
# Rector without DEV needs phpstan.neon removed
-   run: rm phpstan.neon
```

Some better way to do it? Or let it be? 🎤🎵

leoloso:
I just finished adding all the testing:

- Run the full `vendor/bin/phpunit` on the downgraded code, on PHP 7.4 
- Execute`bin/rector` on some code with the default set, using the downgraded code, on PHP 7.1
- Run `vendor/bin/phpstan` on the downgraded code, on PHP 7.1

Everything passing! ✅

leoloso:
The script is now downgrading all dependencies in a single call, like this:

```bash
bin/rector process /home/runner/work/rectorphp/rector/vendor/sebastian/diff /home/runner/work/rectorphp/rector/vendor/symplify/autowire-array-parameter /home/runner/work/rectorphp/rector/vendor/symplify/composer-json-manipulator /home/runner/work/rectorphp/rector/vendor/symplify/console-color-diff /home/runner/work/rectorphp/rector/vendor/symplify/easy-testing /home/runner/work/rectorphp/rector/vendor/symplify/markdown-diff /home/runner/work/rectorphp/rector/vendor/symplify/package-builder /home/runner/work/rectorphp/rector/vendor/symplify/php-config-printer /home/runner/work/rectorphp/rector/vendor/symplify/rule-doc-generator /home/runner/work/rectorphp/rector/vendor/symplify/set-config-resolver /home/runner/work/rectorphp/rector/vendor/symplify/simple-php-doc-parser /home/runner/work/rectorphp/rector/vendor/symplify/skipper /home/runner/work/rectorphp/rector/vendor/symplify/smart-file-system /home/runner/work/rectorphp/rector/vendor/symplify/symfony-php-config /home/runner/work/rectorphp/rector/vendor/symplify/symplify-kernel /home/runner/work/rectorphp/rector/vendor/doctrine/inflector /home/runner/work/rectorphp/rector/vendor/doctrine/lexer /home/runner/work/rectorphp/rector/vendor/nette/utils /home/runner/work/rectorphp/rector/vendor/psr/event-dispatcher /home/runner/work/rectorphp/rector/vendor/symfony/cache /home/runner/work/rectorphp/rector/vendor/symfony/cache-contracts /home/runner/work/rectorphp/rector/vendor/symfony/config /home/runner/work/rectorphp/rector/vendor/symfony/console /home/runner/work/rectorphp/rector/vendor/symfony/dependency-injection /home/runner/work/rectorphp/rector/vendor/symfony/error-handler /home/runner/work/rectorphp/rector/vendor/symfony/event-dispatcher /home/runner/work/rectorphp/rector/vendor/symfony/event-dispatcher-contracts /home/runner/work/rectorphp/rector/vendor/symfony/filesystem /home/runner/work/rectorphp/rector/vendor/symfony/finder /home/runner/work/rectorphp/rector/vendor/symfony/http-client-contracts /home/runner/work/rectorphp/rector/vendor/symfony/http-foundation /home/runner/work/rectorphp/rector/vendor/symfony/http-kernel /home/runner/work/rectorphp/rector/vendor/symfony/service-contracts /home/runner/work/rectorphp/rector/vendor/symfony/string /home/runner/work/rectorphp/rector/vendor/symfony/var-dumper /home/runner/work/rectorphp/rector/vendor/symfony/var-exporter /home/runner/work/rectorphp/rector/vendor/symfony/yaml --config=ci/downgrade/rector-downgrade-dependency-to-php71.php --ansi
```

**Benefits:** The script takes 1 minute less (6 min instead of 7) because it doesn't need to calculate the execution order among dependencies.

**Drawbacks:** In this method it is more difficult to understand which dependency produces an error.

leoloso:
To execute all dependencies in a single call to Rector, there are 2 variables in [script `downgrade_packages.sh`](https://github.com/rectorphp/rector/blob/97f4d464ce0e7af6ce58f1f56e2d229c66a008c4/.github%2Fworkflows%2Fscripts%2Fdowngrade_packages.sh#L51) that must have some value:

- `GROUP_RECTOR_CONFIGS`
- `DOWNGRADE_DEPENDENCIES_TOGETHER`

Right now they are both in `true`, hardcoded in the script. But we could change this to provide it as a parameter, so that if the script fails and the developer doesn't understand where there error is, can then disable them, and run the script again executing Rector on each dependency, providing a more granular detail.

leoloso:
Next step: the script must remove the dependencies from `composer.json`, to not have Composer install them. They are not needed, since Rector will already ship with all its needed (downgraded) dependencies under its own `vendor/` subfolder.

Loading the Composer `autoload.php` is already handled within `bin/rector.php`:

```php
final class AutoloadIncluder
{
    public function includeCwdVendorAutoloadIfExists(): void
    {
        $this->loadIfExistsAndNotLoadedYet(getcwd() . '/vendor/autoload.php');
    }
    // ...
}
```

So we only need to remove the `require` and `require-dev` entries for all downgraded dependencies, and then add `require` again for the Symfony polyfill packages.

leoloso:
I did a simple solution: after the downgrade is done, I copy [`ci/downgrade/composer.json`](https://github.com/rectorphp/rector/blob/05f98183148ed54430d2439545d6aeb864325609/ci%2Fdowngrade%2Fcomposer.json) (containing no dependencies whatsoever) into `composer.json`.

leoloso:
I created a minimal [`ci/downgrade/phpstan.neon` config](https://github.com/rectorphp/rector/blob/e489673d2d53841cde0f0e5ff25520a6f7a132a1/ci%2Fdowngrade%2Fphpstan.neon#L37-L38). So there is no need to duplicate it and adapt it from the one in root, and satisfies `level 0` just to check that the code doesn't have any lingering PHP 7.2+ code.

Everything working! 🎉🎊🥳🍾👏🍻🥂🙌🏻🙏

Next step: deploy the downgraded code to the [Rector PHP 7.1 repo](https://github.com/rectorphp/rector-php71)

leoloso:
@TomasVotruba I definitely need your help for the next task, to deploy the code to the repo. Is it done with [Package Scoper](https://github.com/symplify/package-scoper)?

leoloso:
I made a last improvement: By grouping PHP versions together, the script can analyze library support for the target PHP version only, and the running time for the script went down to 5 minutes. I don't think it can go below that.

leoloso:
For consistency, the downgrade sets have been renamed and split into 2 groups:

1.  Grouping PHP versions together, such as `latest-to-php71` (runs everything between PHP 8.0 and 7.1)
2. Analyzing PHP versions independently, such as `php72-to-71` (must run several sets to downgrade from PHP 8.0 to 7.1)

The first group is used when [`GROUP_RECTOR_CONFIGS`](https://github.com/rectorphp/rector/blob/64afaa8ac976792adb6731cc3f7b640b7935b219/.github%2Fworkflows%2Fscripts%2Fdowngrade_packages.sh#L30) is not empty. This is the case right now, as hardcoded in the script.

In practice we'll only be using the first group for Rector. But since it's already coded, we can also offer using the 2nd group if providing a Rector service to downgrade other projects.

leoloso:
The downgraded code is being uploaded to the CI run as artifact `rector-downgraded-code`:

Screenshot_2020-12-19  ci  add demo for build php 7 1 version by TomasVotruba · Pull Request #4447 · rectorphp rector

This is useful for debugging. For instance, if the downgrade starts failing because some dependency introduced a bug, we can download the artifact and run it locally to find out which files must be excluded.

.github/workflows/build_php71.yaml
```diff
@@ -0,0 +1,77 @@
...
+            -   uses: shivammathur/setup-php@v2
+                with:
+                    php-version: 7.4
+                    coverage: none
+
+            # The Symfony Polyfill libraries are needed
+            -   run: composer require symfony/polyfill-php72 symfony/polyfill-php73 symfony/polyfill-php74 symfony/polyfill-php80 --no-install --no-update
+
+            # Execute the downgrade script
+            -   run: ./.github/workflows/scripts/downgrade_packages.sh 7.1
```
TomasVotruba, line 26: @note Create GitHub Action with Rector parameter?

.github/workflows/build_php71.yaml
```diff
@@ -0,0 +1,77 @@
...
+                with:
+                    name: rector-downgraded-code
+                    path: build/rector-php71.zip
+
+            # Run all the tests on PHP 7.4 (because JUnit needs PHP 7.3+),
+            # but using Rector downgraded to PHP 7.1
+            -   run: vendor/bin/phpunit
+
+            # Avoid Composer v2 platform checks (composer.json requires PHP 7.3+, but below we are running 7.1)
+            -   run: composer config platform-check false
```
TomasVotruba, line 43: :+1:

.github/workflows/build_php71.yaml
```diff
@@ -0,0 +1,77 @@
...
+            # Regenerating the autoload files will remove vendor/composer/platform_check.php, otherwise executing bin/rector on 7.1 fails
+            -   run: composer dumpautoload
+
+            # Remove DEV dependencies
+            -   run: composer install --no-dev
+
+            # Replace the composer.json file
+            -   run: cp ci/downgrade/composer.json ./composer.json
```
TomasVotruba, line 51: Override to remove dependencies, maybe put into scsoper sympfliy

ci/downgrade/phpstan.neon
```diff
@@ -0,0 +1,37 @@
...
+        - ../../vendor/symfony/string/Slugger/AsciiSlugger.php
+        - ../../vendor/symfony/cache/Adapter/CouchbaseBucketAdapter.php
+
+        # broken in PHPStan https://github.com/rectorphp/rector/runs/1305002460#step:5:56
+        - ../../utils/phpstan-extensions/*
+        - ../../packages/testing/src/PHPUnit/*.php
+
+    ignoreErrors:
+        # trait in trait call
+        - '#Call to an undefined method Rector\\PostRector\\Rector\\AbstractPostRector\:\:isObjectType\(\)#'
```
TomasVotruba, line 37: This should be autogenerated from `ci/downgrade/configuration.php`

TomasVotruba:
The code should be pushed https://github.com/rectorphp/rector-php71

TomasVotruba:
With Github Action: https://github.com/marketplace/actions/github-push

.github/workflows/build_php71.yaml
leoloso: Yeah! I'll work on that

.github/workflows/build_php71.yaml
leoloso: To confirm: you'd rather read the current `composer.json`, remove all entries in `require` using `ComposerJson`'s `setRequire` from Symplify, and save again?

Isn't the current solution much simpler? More since [the new `composer.json`](https://github.com/rectorphp/rector/blob/ci-build-php71/ci/downgrade/composer.json) is completely empty... (unless there's something from the current `composer.json` that we need to keep?)

ci/downgrade/phpstan.neon
leoloso: What do you mean? Have you already done that somewhere else, so I can reproduce it?

I copied that error from the [original `phpstan.neon`](https://github.com/rectorphp/rector/blob/master/phpstan.neon#L291) (all the other errors do not happen in this context, so could be removed)

leoloso:
I added the code to deploy to `rectorphp/rector-php71`, but it's failing because it [is not executing the `commit`](https://stackoverflow.com/a/7572252) because the repo is empty:

```
error: src refspec master does not match any
error: failed to push some refs to 'https://github.com/rectorphp/rector'
```

I tried to both add some content to `master`, and create a fork to push to, but I don't have permissions for either

.github/workflows/build_php71.yaml
leoloso: Do you want to create a new repo `rectorphp/rector-downgrade` to host the GitHub action?

leoloso:
Deploying to repo is working now 🎉 

The problem was that input `package-directory` was `.` (since the downgrade is performed on itself), but then it also copied the `.git` folder overriding the one from the target repo.

To fix it, I'm now copying all files (except `.git/`) to `rector-downgrade/` first.

This is a bit of a workaround. If I could exclude `.git` from being copied over, then using `.` as source will already work. Maybe adding an input `"exclude-files"` to monorepo-split-github-action would be a nicer solution?

Also, please notice that rector-php71 is on branch "main", not "master", so I had to do `branch: 'main'`

leoloso:
To test it, I have to use `"dev-main"` instead of `"dev-master"`, or it fails:

```
Root composer.json requires rector/rector-php71 dev-master, found rector/rector-php71[dev-main] but it does not match the constraint
```

Not good, right? Switch rector-php71 to branch `"master"`?

leoloso:
It's working beautifully:

downgrade-rector-php71-test

leoloso:
Workflow in target repo passing now. It's executing PHPStan, not PHPUnit, since we have no tests for PHP 7.1.

Screenshot_2020-12-30 rectorphp rector-php71

leoloso:
Added PHPUnit 7.5 tests now

TomasVotruba:
Let's :ship: it

PR #4448, opened 2020-10-19
[DeadCode] Add expr-names support to RemoveEmptyMethodCallRector

TomasVotruba:


rules/dead-code/src/Rector/MethodCall/RemoveEmptyMethodCallRector.php
```diff
@@ -86,66 +79,37 @@ public function refactor(Node $node): ?Node
...
         return $node;
     }
 
-    private function isNonEmptyMethod(Class_ $class, MethodCall $methodCall): bool
+    private function isEmptyMethod(Class_ $class, MethodCall $methodCall): bool
     {
-        /** @var Identifier $methodIdentifier */
-        $methodIdentifier = $methodCall->name;
-        /** @var ClassMethod|null $classMethod */
-        $classMethod = $class->getMethod((string) $methodIdentifier);
```
TomasVotruba, line 142: @samsonasik Just FYI, this actually causes a bug:

```php
(string) $methodIdentifier
```

In case property/method name is magic, there is not `Identifier` but an `Expr`.
That's why it's better to sue `$this->getName(...)` instead of `(string) $name`

Such magic calls look like these:

```php
$this->$magic()
$this->$magicProperty
```

samsonasik: @TomasVotruba your change make rector-ci error: 

```bash                                                                                
 [ERROR] Could not process                                                      
         "packages/node-type-resolver/src/NodeTypeResolver/ScalarTypeResolver.ph
         p" file, due to:                                                       
         "leaveNode() returned invalid value of type integer".                                                                                                  
```

TomasVotruba: Thanks, I plan to fix CI before merge

PR #4454, opened 2020-10-20
Check for property fetch type in DoctrineCollectionDoctype rector

pierredup:
If you have any static methods in a Doctrine entity, the `doctrine-code-quality` rector currently fails with the following error:

```
TypeError: Argument 1 passed to Rector\DoctrineCodeQuality\Rector\Property\ImproveDoctrineCollectionDocTypeInEntityRector::matchPropertyFetchToClassProperty() must be an instance of PhpParser\Node\Expr\PropertyFetch, instance of PhpParser\Node\Expr\StaticPropertyFetch given
```

This PR adds a check to skip if a node is not an instance of `PropertyFetch`

### DIFF
--- a/rules/doctrine-code-quality/src/Rector/Property/ImproveDoctrineCollectionDocTypeInEntityRector.php
+++ b/rules/doctrine-code-quality/src/Rector/Property/ImproveDoctrineCollectionDocTypeInEntityRector.php
@@ -208,6 +208,10 @@ CODE_SAMPLE
             return null;
         }
 
+        if (! $propertyFetches[0] instanceof PropertyFetch) {
+            return null;
+        }
+
         $property = $this->matchPropertyFetchToClassProperty($propertyFetches[0]);
         if ($property === null) {
             return null;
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Hi, thanks for the fixture and the fix!
Could you look at failing CI? We need it to pass before merge

pierredup:
The failure doesn't look related, it's a PHPStan failure

```
         message: '#Method or Static call in foreach is not allowed#'
         paths:
             - packages/better-php-doc-parser/tests/PhpDocParser/TagValueNodeReprint/TagValueNodeReprintTest.php # 42
             - packages/caching/src/FileSystem/DependencyResolver.php # 53
             - packages/node-type-resolver/src/NodeTypeCorrector/PregMatchTypeCorrector.php # 68
             - packages/node-type-resolver/src/PHPStan/Scope/PHPStanNodeScopeResolver.php # 200
             - rules/symfony/src/FormHelper/FormTypeStringToTypeProvider.php # 96
             - src/PhpParser/Node/Manipulator/ClassConstManipulator.php # 70
             - src/PhpParser/Node/Manipulator/PropertyManipulator.php # 170
             - src/Rector/AbstractRector/ComplexRemovalTrait.php # 220
             - src/Rector/AbstractRector/ComplexRemovalTrait.php # 220
```

Unless I'm missing something?

samsonasik:
@pierredup I think you can ignore the phpstan error by register the error message to phpstan.neon for now.

TomasVotruba:
@samsonasik Could you resolve these new PHPStan errors so we don't have to ignore them? Most of them are  related to `If` rule

samsonasik:
@TomasVotruba I will try, I think that use case of `$this->addNodebeforeNode()` that we discussed today.

TomasVotruba:
Thank you :+1:

PR #4457, opened 2020-10-20
[CodeQuality] Add MoveOutMethodCallInsideIfConditionRector

samsonasik:


samsonasik:
@TomasVotruba current progress: I tried for `If_`, it currently added new variable before if: 

```diff
         $obj = new self();
+        $objCondition = $obj->condition($arg);
 
         if ($obj->condition($arg) === 1) {

         }
```

But method call inside `if->cond` still not replaced.

samsonasik:
I think we can use `MethodCall` instead for node type to be able to replace to current node itself as if using `$if->cond`, we need to deep find the method call and then replace.

samsonasik:
Using `MethodCall` seems issue as well, as `->getAttribute('parent')` can be from inside if statement itself, so that will not only change inside `if->cond`

src/Console/Command/AbstractCommand.php
```diff
@@ -76,6 +77,8 @@ protected function initialize(InputInterface $input, OutputInterface $output): v
             // clear cache
             $this->changedFilesDetector->clear();
         }
+        $inputGetOption = $input->getOption(Option::OPTION_CLEAR_CACHE);
+        $inputGetOption = $input->getOption(Option::OPTION_CLEAR_CACHE);
```
samsonasik, line 81: repetitive `if (methodcall)` seems make addition double.

rules/doctrine/src/PhpDocParser/DoctrineDocBlockResolver.php
```diff
@@ -98,6 +98,8 @@ public function isDoctrineProperty(Property $property): bool
         if ($phpDocInfo === null) {
             return false;
         }
+        $phpDocInfoHasByType = $phpDocInfo->hasByType(ColumnTagValueNode::class);
+        $phpDocInfoHasByType = $phpDocInfo->hasByType(ColumnTagValueNode::class);
```
samsonasik, line 102: It is weird.

src/Console/Command/AbstractCommand.php
samsonasik: fixed.

rules/doctrine/src/PhpDocParser/DoctrineDocBlockResolver.php
samsonasik: fixed.

samsonasik:

rules/code-quality/src/Rector/Foreach_/SimplifyForeachToCoalescingRector.php
```diff
@@ -152,10 +152,11 @@ private function processForeachNodeWithReturnInside(Foreach_ $foreach, Return_ $
         } else {
             return null;
         }
+        $foreachGetAttribute = $foreach->getAttribute(AttributeKey::NEXT_NODE);
```
TomasVotruba, line 155: The variable name from `getAttribute()`, should be named after `AttributeKey::<NEXT_NODE>`.

Like this:

```php
$nextNode = $foreach->getAttribute(AttributeKey::NEXT_NODE);

$phpDocInfo = $foreach->getAttribute(AttributeKey::PHP_DOC_INFO);
```

etc.

packages/post-rector/src/NodeAnalyzer/NetteInjectDetector.php
```diff
@@ -45,8 +45,9 @@ private function isInjectPropertyAlreadyInTheClass(Class_ $class): bool
             if ($phpDocInfo === null) {
                 continue;
             }
+            $phpDocInfoGetTagsByName = $phpDocInfo->getTagsByName('inject');
 
-            if ($phpDocInfo->getTagsByName('inject') === []) {
```
TomasVotruba, line 49: In many cases, you can re-use logic from https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#renamevariabletomatchmethodcallreturntyperector

This service, to be specific:
https://github.com/rectorphp/rector/blob/5988584736f30bd3a43b2a65a0724d59d0504459/rules/naming/src/Rector/Assign/RenameVariableToMatchMethodCallReturnTypeRector.php#L155

<br>

The expected result here would be:

```bash
$phpDocTagNode = $phpDocInfo->getTagsByName('inject');

# or better
$injectPhpDocTagNode = $phpDocInfo->getTagsByName('inject');
```

rules/code-quality/src/Rector/Foreach_/SimplifyForeachToCoalescingRector.php
TomasVotruba: Btw, I recommend reverting the applied Rector commit (these changes above). So we work with original code

TomasVotruba (changes requested):

packages/post-rector/src/NodeAnalyzer/NetteInjectDetector.php
samsonasik: I will try.

rules/code-quality/src/Rector/Foreach_/SimplifyForeachToCoalescingRector.php
samsonasik: @TomasVotruba do you mean we don't need addition to rector config set ?

samsonasik:
I am updating to use `Rector\Naming\Naming\ExpectedNameResolver`, the `$this->expectedNameResolver->resolveForCall($methodCall)` got null:

Screen Shot 2020-10-21 at 2 02 55 AM

I will re-check again.

rules/code-quality/src/Rector/Foreach_/SimplifyForeachToCoalescingRector.php
TomasVotruba: We do. Just the Rector commits, e.g. [`1a7d320` (#4457)](https://github.com/rectorphp/rector/pull/4457/commits/1a7d320fe3835515c4988110ca38bf276d88af31)

should be reverted before applied new version rule.


If you change the rule now, then it will not update any code, because it's been moved from `if()`.

If you revert the Rector commit, then change the rule, then the Rule will be applied in it's new form.

samsonasik:
The `$this->expectedNameResolver->resolveForCall($methodCall)` seems not work when fallback check `count($expr->args) > 0` in ExpectedNameResolver class:

Screen Shot 2020-10-21 at 2 20 01 AM

It seems need fallback if that return null.

samsonasik:
@TomasVotruba I've reverted other changes, updated `variableName` to:

```php
$variableName = $this->expectedNameResolver->resolveForCall($methodCall) 
        ?? $methodCallVarName . ucfirst($methodCallName);
```

so make `$this->expectedNameResolver->resolveForCall($methodCall) ` as preferred, fallback to use `$methodCallVarName . ucfirst($methodCallName);`.

Please review again

samsonasik:
rebased.

samsonasik:
All checks have passed :tada: @TomasVotruba I think it is ready.

rules/code-quality/tests/Rector/If_/MoveOutMethodCallInsideIfConditionRector/Fixture/if_condition_method_call_with_arg_on_no_assignment.php.inc
```diff
@@ -0,0 +1,52 @@
...
+abstract class AbstractCommand extends Command
+{
+    public function run(InputInterface $input, OutputInterface $output): int
+    {
+        return 0;
+    }
+
+    protected function initialize(InputInterface $input, OutputInterface $output): void
+    {
+        $inputGetOption = $input->getOption(Option::OPTION_DEBUG);
```
TomasVotruba, line 45: This case should consider the part `::OPTION_DEBUG`, e.g.

```suggestion
        $optionDebug = $input->getOption(Option::OPTION_DEBUG);
```

TomasVotruba:
One more case with constants

rules/code-quality/tests/Rector/If_/MoveOutMethodCallInsideIfConditionRector/Fixture/if_condition_method_call_with_arg_on_no_assignment.php.inc
samsonasik: It seems not catched by `$this->expectedNameResolver->resolveForCall($methodCall)` like my comment at https://github.com/rectorphp/rector/pull/4457#issuecomment-713087908 , I wil re-check

TomasVotruba: Yea, I think this must be added extra

TomasVotruba: Yea, I think this must be added extra

samsonasik: It seems it because `$input->getOption()` is returning `PHPStan\Type\UnionType` :

```php
^ PHPStan\Type\UnionType^ {#80356
  -types: array:4 [
    0 => PHPStan\Type\ArrayType^ {#80347
      -keyType: PHPStan\Type\MixedType^ {#80352
        -isExplicitMixed: false
        -subtractedType: null
      }
      -itemType: PHPStan\Type\StringType^ {#80348}
    }
    1 => PHPStan\Type\BooleanType^ {#80354}
    2 => PHPStan\Type\StringType^ {#80353}
    3 => PHPStan\Type\NullType^ {#37016}
  ]
}
```

So, the :

```php
$this->propertyNaming->getExpectedNameFromType($returnedType)
```

returns null on calling:

```php
$this->typeUnwrapper->unwrapNullableType($type)
```

So, it it seems need the different handling.

samsonasik: implemented.

samsonasik:
phpstan issue seems unrelated to this code as it still not registered to config set yet.

samsonasik:

samsonasik:
rebased.

samsonasik:

TomasVotruba:
Looks good, let's test in practice.

Thank you

PR #4482, opened 2020-10-25
Map iterable in ScalarStringToTypeMapper

ComiR:
`iterable` return types were mapped to implicit `MixedType`s before. Now they will be handled like `array`, but with the corresponding `IterableType`.

### DIFF
--- a/packages/node-type-resolver/tests/StaticTypeMapper/StaticTypeMapperTest.php
+++ b/packages/node-type-resolver/tests/StaticTypeMapper/StaticTypeMapperTest.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace Rector\NodeTypeResolver\Tests\StaticTypeMapper;
 
 use Iterator;
+use PhpParser\Node;
+use PhpParser\Node\Identifier;
 use PhpParser\Node\Scalar\String_;
 use PHPStan\PhpDocParser\Ast\Type\ArrayTypeNode;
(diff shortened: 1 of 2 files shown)


ComiR:
Maybe this check could be removed: https://github.com/rectorphp/rector/blob/f37cd4cac367c3522497cf1c55fb3ee2f2d5a271/packages/static-type-mapper/src/PhpDocParser/IdentifierTypeMapper.php#L78-L80

TomasVotruba:
We'll need a failing test fixture here too.

TomasVotruba:
[Here is how to add it](https://github.com/rectorphp/rector/blob/master/docs/how_to_add_test_for_rector_rule.md)

ComiR:
For sth. like
```php
function foo(): iterable {}
```
,
```php
$returnTypeNode = $functionLike->getReturnType();
$this->staticTypeMapper->mapPhpParserNodePHPStanType($returnTypeNode);
```
would return `MixedType()` instead of the correct `IterableType()`.

TomasVotruba:
We'll need this in tests

ComiR:
The failing tests mention some kind of token - so I don't think they fail because of sth. in the code.

I hope this test is sufficient 😃

TomasVotruba:
The failing token is unrelated, I'm looking into it.

Thank you for the test case :+1: Very nice job

PR #4483, opened 2020-10-25
Return also explicitly mixed types in IdentifierTypeMapper

ComiR:
For explicit `MixedType`s it was trying to resolve a type but ultimately fail and return an implicit `MixedType`.

### DIFF
--- a/packages/node-type-resolver/tests/StaticTypeMapper/StaticTypeMapperTest.php
+++ b/packages/node-type-resolver/tests/StaticTypeMapper/StaticTypeMapperTest.php
@@ -54,6 +54,8 @@ final class StaticTypeMapperTest extends AbstractKernelTestCase
         ]);
 
         yield [$genericTypeNode, IterableType::class];
+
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Hi,

we need a failing test fixture to cover this change.
What exact rule broke? On what exact PHP code?

ComiR:
sth. like 
```php
/**
 * @return mixed[]
 */
function foo(): array {}
```
would return a `MixedType(false)` instead of a `MixedType(true)` when called like
```php
$returnTagValueNode = $phpDocInfo->getReturnTagValue();
$this->staticTypeMapper->mapPHPStanPhpDocTypeToPHPStanType($returnTagValueNode, $functionLike);
```

TomasVotruba:
Good! Could you add it to the tests?

ComiR:
The failing tests mention some kind of token - so I don't think they fail because of sth. in the code.

I hope this test is sufficient 😃

TomasVotruba:
Thank you :clap:

PR #4484, opened 2020-10-25
[ci] add ci-review job

TomasVotruba:
- [ci] add ci-review job
- add all
- [ci] add docs check


.github/workflows/rector_ci.yaml
```diff
@@ -45,43 +41,18 @@ jobs:
...
-            -   name: Commit CS changes
-                if: steps.cs-git-check.outputs.modified == 'true'
-                run: git commit -am "[cs] ${COMMIT_MESSAGE}"
-
-
-            -   name: Push changes
-                if: steps.rector-git-check.outputs.modified == 'true'
-                run: git push
-                # In case we want to fail this job when there are changed files, just add "exit 1"
-
```
TomasVotruba, line 87: @JanMikes I've discovered very cool Github Action, that handles all the commiting for us!
https://github.com/EndBug/add-and-commit

We're running today for:

- doctrine/annotation sync
- documntation generator
- monorepo sync
- Rector

And so far so good :)

JanMikes: Thx for showing me, it shortens the syntax a little 😄 let's see how it behaves in some edge cases that might have been covered by us but not in action and maybe scenarios like that are completely okay 👍

TomasVotruba: Indeed, it's testing time this week :)

I look for some other action, that will stop if there is no change in previous action. So the coding standard does not run on unchanged files

TomasVotruba: **Update**: this is not Github Action, but solves it :) 
 https://github.com/rectorphp/rector/pull/4487/files#diff-3f20cc93285dee3e6dc5886aa6f9c5f8fb9713b57d575f099947fc8d86153c0c

PR #4485, opened 2020-10-25
RemoveDoubleUnderscoreInMethodNameRector: Only match names starting with double underscores.

j2L4e:
The current implementation wrongly matches `mybadlynamed__method()` and cuts off 2 chars in the front `badlynamed__method()`.

This PR fixes that.

Actually even `'#^__#'` should do it. The captured group is never read as it seems.


samsonasik:
I cherry-picked your commit at https://github.com/rectorphp/rector/pull/4504 while adding regex link uri.

j2L4e:
Great, closing this one :+1:

PR #4492, opened 2020-10-27
Improve MoveOutMethodCallInsideIfConditionRector

TomasVotruba:
- skip rename
- add method call and variable name for MoveOutMethodCallInsideIfConditionRector
- decouple MethodCallToVariableNameResolver
- use trick with Scope if variable is already defined
- fixup! use trick with Scope if variable is already defined
- cover case with variable name and method call
- add decopule note future


TomasVotruba:
@samsonasik I see it's stuck now, after last ci-review push. I have the same error now, right? :)

samsonasik:
@TomasVotruba yes, it hang again, The hang happen when with new commit named "[ci-review] Rector Rectify" applied. Close - reopen PR actually fix that.

TomasVotruba:
I see, I tried to report it here: https://github.com/EndBug/add-and-commit/issues/32

I'll investigate previous Rector CI setup - https://github.com/rectorphp/rector/commit/9c91a60018053984015b2ab6ecf501b6295d27da#diff-3f20cc93285dee3e6dc5886aa6f9c5f8fb9713b57d575f099947fc8d86153c0c

TomasVotruba:
This might help, lets see how it goes:
https://github.com/rectorphp/rector/pull/4492/commits/0c6c794ac90e5aa192a2a88fd3c7f92fd05bb659

PR #4514, opened 2020-10-30
Inform about useless second part in tests fixture, after -----, if it is identical

TomasVotruba:
- embed skip fixture of same before/after
- [Testing] Add exception on identical second half of the fixture, that should not change
- [ci] use composer 2 for code_analysis


TomasVotruba:
@matthiasnoback Hi, I've added the control for useless `fixture.php.inc` second half content.

```diff
 before
------
-before
```

Now this will be reported right when the test is run :)

matthiasnoback:
Very nice!

PR #4517, opened 2020-10-30
Add a Doctrine DBAL 2.11 Rule Set

chrisguitarguy:
See https://github.com/doctrine/dbal/blob/master/UPGRADE.md#upgrade-to-211

Lots of changes in DBAL 2.11, but I don't think all of them can be handled by rector. I also fixed a few things in the 3.0 rule set.

Questions:

- the [MasterSlaveConnection -> PrimaryReadReplicaConnection](https://github.com/doctrine/dbal/blob/master/UPGRADE.md#deprecated-masterslaveconnection-use-primaryreadreplicaconnection) includes renaming some arguments in `DriverManager::getConnection`'s array. Are there pre-existing rectors that can handle something like that?
- Unsure on the order which these are executed in practice (in order defined?). Some classes are renamed and have methods that changed (`DriverException` for instance)
- Really questioning the wisdom of including the `fetchAll` -> `fetchAllAssociative` rename -- [upgrade guide](https://github.com/doctrine/dbal/blob/master/UPGRADE.md#deprecated-fetchmode-and-the-corresponding-methods) mentions that quite a few `fetchAll*` methods were added, but the conneciton docblocks say `fetchAll` returns associative arrays. Statement docblocks are not clear, but PDO itself defaults to fetch both associative and numeric. Hard to say if this should be included.

### DIFF
--- a/config/set/doctrine-dbal-30.php
+++ b/config/set/doctrine-dbal-30.php
@@ -18,7 +18,7 @@ return static function (ContainerConfigurator $containerConfigurator): void {
         ->call('configure', [[
             RenameMethodRector::METHOD_CALL_RENAMES => inline_value_objects([
                 new MethodCallRename(
-                    'DBAL\Platforms\AbstractPlatform',
+                    'Doctrine\DBAL\Platforms\AbstractPlatform',
                     'getVarcharTypeDeclarationSQL',
                     'getStringTypeDeclarationSQL'
                 ),
--- a/config/set/doctrine-dbal-30.php
+++ b/config/set/doctrine-dbal-30.php
@@ -6,6 +6,7 @@ use PHPStan\Type\VoidType;
 use Rector\Renaming\Rector\MethodCall\RenameMethodRector;
 use Rector\Renaming\ValueObject\MethodCallRename;
 use function Rector\SymfonyPhpConfig\inline_value_objects;
+use Rector\Renaming\Rector\Name\RenameClassRector;
 use Rector\TypeDeclaration\Rector\ClassMethod\AddReturnTypeDeclarationRector;
 use Rector\TypeDeclaration\ValueObject\AddReturnTypeDeclaration;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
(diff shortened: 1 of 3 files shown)


TomasVotruba:
Hi,

thanks for your contribution :+1: 

1. renaming args - I don't see any example in the markdown file, just lot of huge headlines. 
Could you share a diff example of what you mean?

2. the order is by file, then by node level, e.g. namespace first, then class, then class method etc., then by the order of Rector rules in config. But that if e.g. class method gets renamed, the old names is still used in one single Rector run. So renaming method/classes does not break next rules that depend on the old name.

3. Again, hard to guess what changed for me. Diff would help.

I'll merge current PR as it is, so we can keep it simple. Feel free to open a new one with addition changes.

<br>

Btw, I see there is dbal 4.0 released/planned? Do you think it would be useful to have upgrade set in Rector for 4.0 as well?

config/set/doctrine-dbal-30.php
```diff
@@ -18,7 +19,7 @@
         ->call('configure', [[
             RenameMethodRector::METHOD_CALL_RENAMES => inline_value_objects([
                 new MethodCallRename(
-                    'DBAL\Platforms\AbstractPlatform',
+                    'Doctrine\DBAL\Platforms\AbstractPlatform',
```
TomasVotruba, line 22: :+1:

TomasVotruba (approved):

chrisguitarguy:
> renaming args - I don't see any example in the markdown file, just lot of huge headlines.
Could you share a diff example of what you mean?

Not a diff, but here's the relevant bit of the docs:

Screen Shot 2020-11-01 at 5 59 07 AM

the array keys changed from `master` -> `primary`, etc.

> [RE: fetch*] Again, hard to guess what changed for me. Diff would help.

Take `Doctrine\DBAL\Statement`, in 2.10:

```diff
  $statement = $connection->executeQuery('SELECT ...');

- $rows = $statement->fetchAll(FetchMode::ASSOCIATIVE);
+ $rows = $statement->fetchAllAssociative();
```


> Btw, I see there is dbal 4.0 released/planned? Do you think it would be useful to have upgrade set in Rector for 4.0 as well?

I don't think 3.X is out yet, so they may just be planning deprecations/removals for 4.X

chrisguitarguy:
Actually, something like [this `RenameMethodCalledBasedOnParameter`](https://github.com/rectorphp/rector/blob/master/rules/cakephp/src/Rector/MethodCall/RenameMethodCallBasedOnParameterRector.php) in `cakephp` might work for the `Statement` issue.

TomasVotruba:
Thanks for sharing. I see, let's take it step by step.

In `DriverManager::getConnection()`, you need to creata Rector rule that listens to `Array_::class`.
Then foreach it's keys and rename those 3 keys. It a good first rule issue.

Is that something you'd like to contribute?

If so, here is [how you create a new rule](https://github.com/rectorphp/rector/blob/master/docs/create_own_rule.md). I can guide you on the pull-request.

PR #4524, opened 2020-10-31
rector.php.dist: added missing use for commented rule

uestla:
In the example of single rule registration there is a class which is unfortunatelly missing amongst use statements - simply uncommenting the line wouldn't work.


TomasVotruba (approved):

TomasVotruba:
Thank you

PR #4538, opened 2020-11-01
Changed docs to be slightly less ambiguous

PoDuck:
When learning about this, I was not sure that "src" was a path that needed changing, or if it were a standard command line argument that wasn't a path.  Given that ambiguity, I imagine that other people have that problem too.  I changed several documents to replace "src" with "{{src}}" and gave an explanation in a couple places in README.md about what was to be done with "{{src}}".  Hopefully that is enough to help.  There are places in other documents that I left things as they were, as the code didn't seem too ambiguous to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,13 @@
 
 Rector helps you with 2 areas - major code changes and in daily work.
 
--   Do you have a legacy code base? Do you want to have that latest version of PHP or your favorite framework?
-    → **Rector gets you there with instant upgrade**.
+- Do you have a legacy code base? Do you want to have that latest version of PHP or your favorite framework?
+ → **Rector gets you there with instant upgrade**.
 
  <br>
 
--   Do you have code quality you need, but struggle to keep it with new developers in your team? Do you wish to have code-reviews for each member of your team, but don't have time for it?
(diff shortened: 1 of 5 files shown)


TomasVotruba:
Hi,

thanks for contribution. Could you remove the space indent? It's hard to review what has really changed

The `{{ syntax }}` looks like Twig/template/Smarty.

The `/src` is the directory for source code by convention. So it does not make sense to replace it everywhere.
Yet, it should be shown right in the start once.

PoDuck:
I see what you are talking about with the spaces.  I guess my editor automatically changed tabs to spaces.  As for the {{ syntax }}, I guess I wanted to make it obvious, and that makes it obvious, but what would you suggest?

One other thing, this is my first PR ever, so I'm not sure how they work.  Do I just update the branch, or do I have to do another PR?  How does that work?

TomasVotruba:
From the hindsight, using path as argument is common knowledge in any other CLI tool. PHPUnit, PHP CS Fixer, PHP CodeSniffer etc., so it's rather the experience with CLI tools that is relevant here.

Also, the twig syntax would be very confusing.

For the reasons above I've decided to keep docs as it is.

Thank you for your work.

PR #4547, opened 2020-11-03
Fixes #4543 Skip Rector check on empty files

samsonasik:

Issue #4543, opened 2020-11-02
Rector crashes on empty files


JanWennrichPCSG:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | `0.8.47`              |
| Installed as   | composer dependency                      |

## Minimal PHP Code Causing Issue

If Rector processes a file which has an opening `<?php`-tag but contains nothing else, it crashes.

An example can be found here: https://getrector.org/demo/1db33def-fcef-4e1a-92cd-748d1bef14ee#result

## Expected Behaviour

Rector should continue with the next file.

TomasVotruba:
Hi Jan, thanks for reporting :+1: 


I'm curious, how does such code apprear in your code base?

JanWennrichPCSG:
> I'm curious, how does such code apprear in your code base?

I'm currently upgrading an about 15 years old customer project from PHP 5.2 to PHP 7.4.
Many (inexperienced) people worked on it, therefore it's a total mess :smile:

TomasVotruba:
I see. I wish you good luck and heathy nerves :D 

What would you expect to happen with such file? I can think of complete skip

JanWennrichPCSG:
> What would you expect to happen with such file? I can think of complete skip

A skip (and maybe a notice) would be what I suggested but I also can't think of anything else.

PR #4549, opened 2020-11-03
add empty file test case

samsonasik:
Test case for https://github.com/rectorphp/rector/issues/4543

Fixed in https://github.com/rectorphp/rector/pull/4547


samsonasik:
@TomasVotruba I think it is ready for merge.

TomasVotruba:
Thanks for the ping

PR #4553, opened 2020-11-04
Fixes #4499 Code duplication on interface_exists and trait_exists inside if condition

### LINKED ISSUE #4499: Code gets duplicated
# Bug Report

| Rector version | 0.8.44             |

When running a set (any set) on file `stubs/Symfony/Component/VarDumper/Test/VarDumperTestTrait.php`, its code gets duplicated:

```diff
if (trait_exists('Symfony\Component\VarDumper\Test\VarDumperTestTrait\VarDumperTestTrait')) {
     return;
+}if (trait_exists('Symfony\Component\VarDumper\Test\VarDumperTestTrait\VarDumperTestTrait')) {
+    return;
 }
```

It doesn't indicate which is the applied rule

https://getrector.org/demo/67559c76-b2e7-44bd-bca2-ec78a35d7f47#result

samsonasik:


tests/Issues/Issue4499/DoNotDuplicateCode/DoNotDuplicateCodeTest.php
```diff
@@ -0,0 +1,31 @@
...
+use Symplify\SmartFileSystem\SmartFileInfo;
+
+final class DoNotDuplicateCodeTest extends AbstractRectorTestCase
+{
+    /**
+     * @dataProvider provideData()
+     */
+    public function test(SmartFileInfo $fileInfo): void
+    {
+        @$this->doTestFileInfo($fileInfo);
```
samsonasik, line 17: this `@` is on purpose to avoid phpunit break on warning : 

```php
assert(): assert($itemStartPos >= 0 && $itemEndPos >= 0 && $itemStartPos >= $pos)
```

that came from `vendor/nikic/php-parser/lib/PhpParser/PrettyPrinterAbstract.php` on line 759

samsonasik:

samsonasik:
I've added handle no bracket `{}` after if condition.

samsonasik:

src/PhpParser/Printer/BetterStandardPrinter.php
```diff
@@ -505,6 +529,33 @@ private function detectTabOrSpaceIndentCharacter(array $stmts): void
         }
     }
 
+    private function cleanUpDuplicateContent(string $content): string
```
TomasVotruba, line 532: This might looks like a random piece of code. Reference to the problem would help understand.


Could you refer the issue here?

samsonasik: I've added  `@see https://github.com/rectorphp/rector/issues/4499` comment to reference issue.

samsonasik:

TomasVotruba:
Thanks!

It should be still reported on php-parser as an issue.
It might be solves there in a clean way

samsonasik:
The issue is happen when `if (interface_exists('InterfaceName')) {}` or `trait_exists('TraitName') { }` followed by `interface InterfaceName` or `trait TraitName`. 

@TomasVotruba could you handle report to php-parser ? Thank you.

TomasVotruba:
I'll look into the issue for php-parser. Could you make an issue for that here, so I don't forget?

Thank you

samsonasik:
@TomasVotruba Yes, I created issue https://github.com/rectorphp/rector/issues/4556 for it.

TomasVotruba:
Thanks

PR #4557, opened 2020-11-05
[PHP 8.0] Fixes #4527 Add NullsafeOperatorRector

### LINKED ISSUE #4527: [PHP 8.0] Add new rule for nullsafe calls and properties
- RFC: https://wiki.php.net/rfc/nullsafe_operator

## Inspiratoin

- https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#simplifyifnotnullreturnrector

samsonasik:


samsonasik:
The `IfManipulator::matchIfNotNullReturnValue()` and `IfManipulator::matchIfValueReturnValue()` seems will only works if there is return statement, if no return statement like in the rfc https://wiki.php.net/rfc/nullsafe_operator , the variable is not catched.

samsonasik:
@TomasVotruba It is functional for changing the following:

```diff
     public function f($o)
     {
-        $o2 = $o->mayFail1();
-        if ($o2 === null) {
-            return null;
-        }
-
-        return $o2->mayFail2();
+        return $o?->mayFail1()?->mayFail2();
     }
```

Currently, it only one level if, todo:

- make dynamic for many if null check
- handle no return statement like in the rfc https://wiki.php.net/rfc/nullsafe_operator

samsonasik:
I've added handling no return in last statement:

```diff
-        $o2 = $o->mayFail1();
-        if ($o2 === null) {
-            return null;
-        }

-        $o2->mayFail2();
+        $o?->mayFail1()?->mayFail2();
```

samsonasik:
I've added handling null safe on property fetch:

```diff
-        $o2 = $o->property1;
-        if ($o2 === null) {
-            return null;
-        }

-        return $o2->property2;
+        return $o?->property1?->property2;
```

samsonasik:
I've applied multiple ifs (currently only handle 2 steps ifs) like the following:

```diff
-        $o2 = $o->mayFail1();
-        if ($o2 === null) {
-            return null;
-        }

-        $mayFail2 = $o2->mayFail2();
-        if ($mayFail2 === null) {
-            return null;
-        }

-        return $mayFail2->mayFail3();
+        return $o?->mayFail1()?->mayFail2()?->mayFail3(); 
```

samsonasik:
Finally :tada: :tada: :tada: , it can do dynamic if null check like as many as we want:

```diff
-        $o2 = $o->mayFail1();
-        if ($o2 === null) {
-            return null;
-        }

-        $mayFail2 = $o2->mayFail2();
-        if ($mayFail2 === null) {
-            return null;
-        }

-        $mayFail3 = $mayFail2->mayFail3();
-        if ($mayFail3 === null) {
-            return null;
-        }

-        $mayFail4 = $mayFail3->mayFail4();
-        if ($mayFail4 === null) {
-            return null;
-        }

-        return $mayFail4->mayFail5();
+        return $o?->mayFail1()?->mayFail2()?->mayFail3()?->mayFail4()?->mayFail5();
```

samsonasik:
All checks have passed 🎉 🎉 🎉 . For not identical check `$var !== null`, it seems need a separate PR . @TomasVotruba it is ready for review

samsonasik:
I applied skips on various conditions:

-  no direct usage after if
-  no direct usage after if in prev or next statement
-  no direct if after assignment

samsonasik:
All checks have passed 🎉 I think it already cover most of use cases for `if ($var === null)` check. 

Not Identical use case can be in separate PR for improvement.

@TomasVotruba it is ready for review.

TomasVotruba:
Thank you :+1:

TomasVotruba:
Thank you :clap:

rules/php80/tests/Rector/If_/NullsafeOperatorRector/Fixture/fixture_next_is_property.php.inc
```diff
@@ -0,0 +1,32 @@
...
+-----
+<?php
+
+namespace Rector\Php80\Tests\Rector\If_\NullsafeOperatorRector\Fixture;
+
+class SomeClassNextProperty
+{
+    public function f($o)
+    {
+        return $o?->property1?->property2;
```
iluuu1994, line 28: This seems wrong. Why do we need `?->` for `property1`?

rules/php80/tests/Rector/If_/NullsafeOperatorRector/Fixture/fixture_no_return_last.php.inc
```diff
@@ -0,0 +1,32 @@
...
+-----
+<?php
+
+namespace Rector\Php80\Tests\Rector\If_\NullsafeOperatorRector\Fixture;
+
+class SomeClassNoReturnLast
+{
+    public function f($o)
+    {
+        $o?->mayFail1()?->mayFail2();
```
iluuu1994, line 28: Same here.

iluuu1994 (commented):
There also seems to be a missing test for this:

```php
if ($o) {
    return null;
}

return $o->property1->property2;

======

// No ?-> needed for property2 because the first ?-> short-circuits the expression
return $o?->property1->property2;
```

rules/php80/tests/Rector/If_/NullsafeOperatorRector/Fixture/fixture_next_is_property.php.inc
samsonasik: That's needed to check if that's null, return early, else that uses re-assign with new object and call other property by new new object, ref https://3v4l.org/eZGoF

rules/php80/tests/Rector/If_/NullsafeOperatorRector/Fixture/fixture_no_return_last.php.inc
samsonasik: That's needed to check if that's null, return early, else that uses re-assign with new object and call other method by new new object, ref https://3v4l.org/U9vkk

samsonasik:
`!==null` check need to be handled in separate PR as I noted at https://github.com/rectorphp/rector/pull/4557#issuecomment-723543223

rules/php80/tests/Rector/If_/NullsafeOperatorRector/Fixture/fixture_next_is_property.php.inc
iluuu1994: But in the original code we're calling `$o->property1` without checking for `null`, these two code snippets will not actually behave the same. One throws a warning (and possibly an exception in the future), the other doesn't.

https://3v4l.org/lGM7B

samsonasik: When user write `$o2 = $o->property1`,  it is user responsibility to know that `$property1` exists. On the fixture example, it is assigned to `$o2`, next `$o2` is check against null via `if` statement to return null early if its result is null, otherwise, call `property2` on by $o2 as new object.

Your example is invalid as $o is a null, not object.

iluuu1994: Correct me if I'm wrong, but Rector shouldn't alter the behavior of code (at least in this case).

```php
$o2 = $o->property1;
if ($o2 === null) {
    return null;
}
return $o2->property2;
=====
return $o?->property1?->property2;
```

Those two code snippets are not semantically equivalent. The former will warn when `$o` is `null` but the latter will. The code you actually want is `return $o->property1?->property2;` (allow `property1` to be null but not `$o`).

> Your example is invalid as $o is a null, not object.

That's kind of the point, `$o?->property1` will check if `$o` is `null` which it will never be. Thus in this case the normal `->` operator should be used. If for some reason `$o` would actually be `null` the latter would just ignore the `null` value. This might not be intended by the user.

samsonasik: No, this rule is interpreting what user wrote was already correct. So, `$o->property1` will returns 2 things: null, or object that has `property2`.

iluuu1994: Side note: I'm the author and implementer of the nullsafe RFC.

> So, $o->property1 will returns 2 things: null, or object that has property2.

Yes, but the point is that you don't need `?->` before `property1`, just `property2`. Note that the nullsafe RFC, `?->` will check for `null` on the left hand side of the operator, not the right hand side.

https://3v4l.org/fr6qQ

As you can see, by making `->` into a `?->` you'll convert an exception to a `null` value for something that the user might have never considered. This is inherently unsafe. This obviously can't happen if you type hint the parameter properly but that's not necessarily always the case (it isn't in the test case).

Anyway, let's see what @TomasVotruba thinks about this :slightly_smiling_face:

samsonasik: Ok, I got your point to remove `?->` in very start property/method call. Next `->` on position 2, 3, and so on will still needed. 

Probably another PR for improvement ;)

iluuu1994: @samsonasik Cool :slightly_smiling_face: Thank you for your contributions!

PR #4560, opened 2020-11-05
fix: PHP_Parser T_Match issue with php 7

dameert:
nikic/PHP-Parser is throwing an "Syntax error, unexpected T_MATCH:699" when we are running  /vendor/bin/rector process .
Because our project uses caxy/php-htmldiff which has a Match class. This class is a reserved keyword in PHP8 but should work in PHP 7.x .
 
This is a known issue: https://github.com/nikic/PHP-Parser/issues/690
This PR configures the Lexer to take into account the current PHP version to prevent this error.

Thank you for taking this fix into consideration.
- David


TomasVotruba:
Looks good, thank you :+1:

PR #4565, opened 2020-11-09
[PHP 8.0] Improve NullsafeOperatorRector : Skip no direct usage after if in next statement at last

samsonasik:
It previously cause false positive removal when there is a non-direct usage after if just before last statement. It handle it.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -3,6 +3,10 @@ includes:
     - vendor/phpstan/phpstan/conf/bleedingEdge.neon
     - vendor/symplify/phpstan-extensions/config/config.neon
 
+    # see https://github.com/symplify/coding-standard
+    - vendor/symplify/coding-standard/config/symplify-rules.neon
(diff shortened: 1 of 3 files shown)


phpstan.neon
```diff
@@ -3,6 +3,10 @@ includes:
     - vendor/phpstan/phpstan/conf/bleedingEdge.neon
     - vendor/symplify/phpstan-extensions/config/config.neon
 
+    # see https://github.com/symplify/coding-standard
+    - vendor/symplify/coding-standard/config/symplify-rules.neon
+    - vendor/symplify/coding-standard/packages/cognitive-complexity/config/cognitive-complexity-services.neon
```
samsonasik, line 8: temporary rollback register above rules as it seems the removal of rules rely to latest `dev-master` of symplify's coding-standard which rector uses released sympilfy's coding standard version.

samsonasik:
All checks have passed :tada: @TomasVotruba it is ready for review.

TomasVotruba:
Thank you

PR #4598, opened 2020-11-13
[DeadCode] Fixes #4472 Remove method call on $this 

### LINKED ISSUE #4472: [RemoveEmptyClassMethodRector] Removed method but not related calls
Hi there 👋 

Consider the following rector output:
Screen Shot 2020-10-21 at 4 07 35 pm

Makes sense because this method is empty, however, in the same class we were calling this method rector removed but this call hasn't been removed. So we end up having code calling a method which doesn't exist anymore.

- Should this be the responsibility of `RemoveEmptyClassMethodRector`?
- Or maybe there is already an existing rector for that and it's not in the DeadCode set?
- Or maybe it is in the DeadCode set but it runs before `RemoveEmptyClassMethodRector`?

samsonasik:


samsonasik:

TomasVotruba:
Thank you

PR #4644, opened 2020-11-18
[CI] Run Top 50 GitHub packages testing with PHP 7.x Rector sets

TomasVotruba:


.github/workflows/weekly_pull_requests.yaml
```diff
@@ -4,9 +4,17 @@ on:
     schedule:
         # https://crontab.guru/once-a-week
         -   cron: "0 0 * * 0"
+        # todo make a bit human-like random: https://stackoverflow.com/a/9051029/1348344
 
 jobs:
     weekly_pull_requests:
+<<<<<<< HEAD
```
orklah, line 11: Seems like a broken conflict here

PR #4659, opened 2020-11-19
[Solid] Enable MoveVariableDeclarationNearReferenceRector in solid set

samsonasik:


src/Configuration/Configuration.php
```diff
@@ -17,87 +17,87 @@
...
 
     /**
-     * @var bool
+     * @var bool @var bool
      */
     private $shouldClearCache = false;
 
     /**
-     * @var string
+     * @var string @var string
```
samsonasik: @TomasVotruba it seems weird, try running rectify locally and working ok: 

```bash
➜  rector git:(enable-move-variable-dec) bin/rector rectify --config rector-ci.php src/Configuration/Configuration.php --dry-run --clear-cache 
Rector dev-master@bd52969
Config file: rector-ci.php

 4/4 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

                                                                                                                        
 [OK] Rector is done!                                                                                                   
                                                                                                                        
```

samsonasik: @TomasVotruba I tried disable again the rule and it keep adding `@var` via Rector Rectify auto commit, see https://github.com/rectorphp/rector/pull/4659/commits/4408426fdd128dd5097c971c698a528fc70219c8 , do you have idea?

tests/Configuration/ComposerJsonParserTest.php
```diff
@@ -12,7 +12,7 @@
 final class ComposerJsonParserTest extends AbstractKernelTestCase
 {
     /**
-     * @dataProvider dataProvider @dataProvider dataProvider
+     * @dataProvider dataProvider @dataProvider dataProvider @dataProvider dataProvider @dataProvider dataProvider
```
samsonasik, line 15: @TomasVotruba The `[ci-review] Rector Rectify` auto commit keep duplicating docblock

.github/workflows/rector_ci.yaml
```diff
@@ -54,7 +54,7 @@ jobs:
 
             ## First run Rector - here can't be --dry-run !!! it would stop the job with it and not commit anything in the future
             -   run: bin/rector rectify ${{ matrix.directories }} --config rector-ci.php --ansi --no-progress-bar
-            -   run: vendor/bin/ecs check --match-git-diff --fix --ansi
+           # -   run: vendor/bin/ecs check --match-git-diff --fix --ansi
```
samsonasik, line 57: It seems the addition `@var` on `[ci-review] Rector Rectify` related with ecs command, trying disable it.

src/Configuration/Configuration.php
TomasVotruba: What ECS rule is causing this?

packages/attribute-aware-php-doc/src/Ast/PhpDoc/AttributeAwareDataProviderTagValueNode.php
```diff
@@ -13,7 +13,7 @@ final class AttributeAwareDataProviderTagValueNode implements PhpDocTagValueNode
     use AttributeTrait;
 
     /**
-     * @var string
+     * @var string @var string
```
samsonasik, line 16: weird, ` [ci-review] Rector Rectify ` keep duplicating docblock even after ecs check command disabled.

samsonasik:
@TomasVotruba I am closing this PR for now as the auto-commit " [ci-review] Rector Rectify " keep duplicating docblocks.

src/Configuration/Configuration.php
samsonasik: I don't know, I tried disable ecs check command and it keep duplicating, see https://github.com/rectorphp/rector/pull/4659/commits/46e051199d035a35351cc071a7ebfa1002dfa6a7

TomasVotruba:
Could you try to find this bug and fix it? It might be change in one of dependencies, possibly nikic/php-parser or phpstan/phpdoc-printer

samsonasik:
I will try

samsonasik:
Re-open, now get different error: 

```bash
360) Rector\CodingStyle\Tests\Rector\ClassMethod\YieldClassMethodToArrayClassMethodRector\YieldClassMethodToArrayClassMethodRectorTest::test with data set #1 (Symplify\SmartFileSystem\SmartFileInfo Object (...))
_HumbugBoxc5228e318540\Nette\DI\ServiceCreationException: Service '0275': Class #Class with base "Regex" name is already used in "MongoDB\\BSON\\Regex", "Symplify\\PHPStanRules\\ValueObject\\Regex"\. Use unique name to make classes easy to recognize# not found.
```

samsonasik:
pin symplify/phpstan-rules to #dd99959572b906d0fbaa456520b940ecb5426ce5 make All checks have passed, I will re-enable the rule.

tests/Configuration/ComposerJsonParserTest.php
```diff
@@ -12,7 +12,7 @@
 final class ComposerJsonParserTest extends AbstractKernelTestCase
 {
     /**
-     * @dataProvider dataProvider
+     * @dataProvider dataProvider @dataProvider dataProvider
```
samsonasik: It happen again. probably happen when `addNodeBeforeNode()` called.

packages/node-nesting-scope/src/NodeFinder/ScopeAwareNodeFinder.php
```diff
@@ -52,10 +52,13 @@ public function findParentType(Node $node, array $allowedTypes): ?Node
...
     public function findParent(Node $node, callable $callable, array $allowedTypes): ?Node
     {
+        $this->isBreakingNodeFoundFirst = false;
         $parentNestingBreakTypes = array_diff(ControlStructure::BREAKING_SCOPE_NODE_TYPES, $allowedTypes);
 
-        $this->isBreakingNodeFoundFirst = false;
-        $foundNode = $this->betterNodeFinder->findFirstPrevious($node, function (Node $node) use (
+        if ($this->isBreakingNodeFoundFirst) {
+            return null;
+        }
```
samsonasik, line 60: probably caused by it changes the behaviour of this class, that `$this->isBreakingNodeFoundFirst` called first, moved up to before return of Method Call.

samsonasik:
Finally, All checks have passed :tada: :tada: :tada: @TomasVotruba it is ready for review.

samsonasik:
rebased.

samsonasik:
rebased.

samsonasik:
All checks have passed 🎉  @TomasVotruba it is ready for review.

rules/solid/src/Rector/Assign/MoveVariableDeclarationNearReferenceRector.php
```diff
@@ -130,6 +130,10 @@ private function findFirstVariableUsageInScope(
     ): ?Variable {
         $desiredVariableName = $this->getName($desiredVariable);
 
+        if ($desiredVariableName === null) {
```
samsonasik, line 133: handle of `$this->getName()` can returns null

samsonasik:

samsonasik:
@TomasVotruba I tried in codebase with `__callStatic()` inside it, and it hang. I will check if I can reproduce it in failing fixture.

samsonasik:
I found another false positive when assignment already inside condition, and it is re-assign from function arg:

```php
	function foo(string $content = ''): string
	{
		if (true) {
			$content = 1;
		}

		echo $content . 'test';
	}
```

above should be skipped.

samsonasik:
use case first assignment is already inside condition handled

samsonasik:
It seems cause an issue, I will re-check

samsonasik:
@TomasVotruba It seems it is too complicated for use `Assign::class` for type node to be checked. 

I will create separate PR for use `Variable::class` for type node and let see if that more usable.

TomasVotruba:
It's very long PR now and I'm getting lost in it.
It would be great if we could merge this even in WIP state, e.g. with rule disabled in CI.

samsonasik:
@TomasVotruba I reverted the failing fixture and re-disable it from solid set. It is ready for merge.

samsonasik:
I am closing this as using Variable node type in #4670 is really more usable.

PR #4661, opened 2020-11-19
[FEATURE] Make NameImportingPostRector skippable 

sabbelasichon:
At the moment you can configure rector to skip certain rules for specific files. But it is not possible to skip these files for the PostRectors. This PR is a proof of concept to make also the PostRectors capable to skip these files. 
So if you put the following code in your rector configuration, the NameImportingPostRector is skipped for the defined file pattern.

```php
        NameImportingPostRector::class => [
            __DIR__ . '/packages/*/ext_emconf.php',
        ],
    ]);
```

**Background information**
In the context of TYPO3 this is needed for a lot of old legacy configuration files which gets merged in a huge cached configuration file. If every configuration file would import the used classes we would end in multiple use statements for the same class, so we get an PHP Fatal Error [...]the name is already in use.

At the moment i provide no tests for the PR. This is bad. If you can point me in the right direction where the tests for the PostRectors are currently handled i will add tests too.

### DIFF
--- a/packages/post-rector/src/Application/PostFileProcessor.php
+++ b/packages/post-rector/src/Application/PostFileProcessor.php
@@ -8,6 +8,7 @@ use PhpParser\Node;
 use PhpParser\NodeTraverser;
 use Rector\Core\Exception\ShouldNotHappenException;
 use Rector\PostRector\Contract\Rector\PostRectorInterface;
+use Symplify\SmartFileSystem\SmartFileInfo;
 
 final class PostFileProcessor
 {
@@ -28,9 +29,11 @@ final class PostFileProcessor
      * @param Node[] $nodes
      * @return Node[]
      */
-    public function traverse(array $nodes): array
+    public function traverse(array $nodes, ?SmartFileInfo $smartFileInfo = null): array
     {
         foreach ($this->postRectors as $postRector) {
+            $postRector->addCurrentSmartFileInfo($smartFileInfo);
+
             $nodeTraverser = new NodeTraverser();
             $nodeTraverser->addVisitor($postRector);
             $nodes = $nodeTraverser->traverse($nodes);
--- a/packages/post-rector/src/Application/PostFileProcessor.php
+++ b/packages/post-rector/src/Application/PostFileProcessor.php
@@ -8,7 +8,6 @@ use PhpParser\Node;
 use PhpParser\NodeTraverser;
 use Rector\Core\Exception\ShouldNotHappenException;
 use Rector\PostRector\Contract\Rector\PostRectorInterface;
-use Symplify\SmartFileSystem\SmartFileInfo;
 
 final class PostFileProcessor
 {
@@ -29,11 +28,9 @@ final class PostFileProcessor
      * @param Node[] $nodes
      * @return Node[]
      */
-    public function traverse(array $nodes, ?SmartFileInfo $smartFileInfo = null): array
+    public function traverse(array $nodes): array
     {
         foreach ($this->postRectors as $postRector) {
-            $postRector->addCurrentSmartFileInfo($smartFileInfo);
-
             $nodeTraverser = new NodeTraverser();
             $nodeTraverser->addVisitor($postRector);
             $nodes = $nodeTraverser->traverse($nodes);
--- a/packages/post-rector/src/Contract/Rector/PostRectorInterface.php
+++ b/packages/post-rector/src/Contract/Rector/PostRectorInterface.php
@@ -6,7 +6,6 @@ namespace Rector\PostRector\Contract\Rector;
 
 use PhpParser\NodeVisitor;
 use Rector\Core\Contract\Rector\RectorInterface;
-use Symplify\SmartFileSystem\SmartFileInfo;
 
 interface PostRectorInterface extends NodeVisitor, RectorInterface
 {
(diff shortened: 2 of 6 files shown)


packages/post-rector/src/Application/PostFileProcessor.php
```diff
@@ -28,9 +29,11 @@ public function __construct(array $postRectors)
      * @param Node[] $nodes
      * @return Node[]
      */
-    public function traverse(array $nodes): array
+    public function traverse(array $nodes, ?SmartFileInfo $smartFileInfo = null): array
     {
         foreach ($this->postRectors as $postRector) {
+            $postRector->addCurrentSmartFileInfo($smartFileInfo);
```
TomasVotruba, line 35: `traverse()` contract should be 1:1 to original `NodeTraverser`, to prevent cofussion

`CurrentFileInfoProvider` services should be used instead.

packages/post-rector/src/Contract/Rector/PostRectorInterface.php
```diff
@@ -6,11 +6,14 @@
...
+use Symplify\SmartFileSystem\SmartFileInfo;
 
 interface PostRectorInterface extends NodeVisitor, RectorInterface
 {
     /**
      * Higher values are executed first
      */
     public function getPriority(): int;
+
+    public function addCurrentSmartFileInfo(?SmartFileInfo $smartFileInfo = null): void;
```
TomasVotruba, line 18: This can be dropped as `CurrentFileInfoProvider` will be used

TomasVotruba:
The idea makes sense to me :+1: Let's add it :) 

Please update your fork with current `master`, so CI can pass. There is old bug on your branch fork now that makes most jobs fail.

packages/post-rector/src/Rector/AbstractPostRector.php
```diff
@@ -6,9 +6,42 @@
...
+     */
+    private $skipper;
+
+    /**
+     * @required
+     */
+    public function autowireAbstractPostRector(Skipper $skipper): void
+    {
+        $this->skipper = $skipper;
+    }
```
TomasVotruba, line 33: :+1: 

This is the place to add `CurrentFileInfoProvider`

packages/post-rector/src/Rector/NameImportingPostRector.php
```diff
@@ -55,6 +55,10 @@ public function enterNode(Node $node): ?Node
             return null;
         }
 
+        if ($this->shouldSkip($this)) {
+            return null;
+        }
```
TomasVotruba, line 60: What about ignoring other PostRectors? I guess there should be different place to add the skipper to.

src/Skip/Skipper.php
```diff
@@ -34,7 +34,7 @@ public function __construct(ParameterProvider $parameterProvider)
         $this->parameterProvider = $parameterProvider;
     }
 
-    public function shouldSkipFileInfoAndRule(SmartFileInfo $smartFileInfo, AbstractRector $rector): bool
+    public function shouldSkipFileInfoAndRule(SmartFileInfo $smartFileInfo, RectorInterface $rector): bool
```
TomasVotruba, line 37: :+1:

packages/post-rector/src/Rector/NameImportingPostRector.php
sabbelasichon: What would you suggest? Add this check to every PostRector?

TomasVotruba: It should be in their parent, as the `AbstractRector` handle this in one place too

sabbelasichon: And then call parent::enterNode on every PostRector?

TomasVotruba: Not sure about exact implementation, I'd have to look closer when I have more time.

Thr implementation should work for any post Rector rule, and be at 1 place. That's the goal here.

sabbelasichon: Will try to figure something out, but without the inheritance hell.

TomasVotruba: Thank you

TomasVotruba:
I see PHPStan is unrelated. Is this ready to merge?

sabbelasichon:
Maybe the easiest solution is to just add the skip guard to every PostRector. Done. Drawback is that every other PostRector has to take care of it.

TomasVotruba:
You're right. It would probably deserve some refactoring to have `Skipper` outside rules themselves. Not sure where now.

As a quick solution, I'd mimic the same behavior normal Rectors have:
https://github.com/rectorphp/rector/blob/054baf96d12de58f56e183e257f98c69f7a269c4/src/Rector/AbstractRector.php#L221-L224

sabbelasichon:
I am going to add this to all the other PostRectors

sabbelasichon:
@TomasVotruba I had a quick look to the other post rectors. For me it makes no sense to add the skippable behaviour there. So despite no testing for this new behaviour for the NameImportingPostRector i would this is ready to merge.

TomasVotruba:
If that's only on *some* rules, how could you recognize which of those lines works?

```yaml
   FirstRector::class => ['somePath'],
   SecondRector::class => ['somePath'],
]);
```

That could be pretty WTF. The previous solution looked good to me.

sabbelasichon:
Hi Tomas, for the moment i am gonna close this issue. I solved the problem for us by decorating the current NamingPostRector. Works like a charm. I will test it under real conditions and see if we encounter some flaws. If not, i will get back to you.

TomasVotruba:
Fine with me :+1:

PR #4669, opened 2020-11-22
[Solid] Using Variable node type for for MoveVariableDeclarationNearReferenceRector

samsonasik:


samsonasik:
@TomasVotruba it seems really more usable with `Variable::class` for type node, as it can detect precisely of its usage. I will enable it in config set and let see if it works.

tests/Configuration/ComposerJsonParserTest.php
```diff
@@ -12,7 +12,7 @@
 final class ComposerJsonParserTest extends AbstractKernelTestCase
 {
     /**
-     * @dataProvider dataProvider
+     * @dataProviderdataProviderdataProvider
```
samsonasik, line 15: oh no

samsonasik: I will check this, meanwhile #4659 can be merged first.

PR #4670, opened 2020-11-22
[Solid] use Variable node type for for MoveVariableDeclarationNearReferenceRector

samsonasik:


samsonasik:
Using Variable node type really make it works :tada: :tada: :tada: @TomasVotruba it is ready for review.

TomasVotruba:
That's amazing and hard work. Thank you for pushing this :100: :+1: 


cc @afilina I think you might like this. Now Rector codebase is using this rule and CI passes :+1:

PR #4678, opened 2020-11-23
[Solid] Fixes #4677 Skip MoveVariableDeclarationNearReferenceRector on multiple usage in Switch -> cases

### LINKED ISSUE #4677: MoveVariableDeclarationNearReferenceRector false positive
# Bug Report

| Rector version | last master              |

I tried to run MoveVariableDeclarationNearReferenceRector on my code but had a false positive when a variable is used multiple time in a switch where each case can return. The issue is that the rector moves the variable on the first case so it's undefined later

```php
     public static function get_price($do_trt): array {
-        $do_sfe = $do_trt->get_sfe_do();
         switch($do_trt->get_mar_do()->get_cod_s()) {
             case 'MAR1':
+              $do_sfe = $do_trt->get_sfe_do();
                 return class1::get_price_all($do_sfe);
             case 'MAR2':
                 return class1::get_price_all($do_sfe);
```

## Expected Behaviour

The variable shouldn't move in this case

samsonasik:


samsonasik:

TomasVotruba:
Thank you :+1:

PR #4679, opened 2020-11-23
[DX] Fixes #4205 Change phpversion from string to php constant int type

### LINKED ISSUE #4205: [DX] Change phpversion from string to php constant int type
## Tasks

- Change features types we use in 
https://github.com/rectorphp/rector/blob/master/src/ValueObject/PhpVersionFeature.php
into PHP integers

- update related behaviors in 
    - https://github.com/rectorphp/rector/blob/88d57114364c1f86b7a46cbb3e0dcf06838e8389/src/Rector/AbstractRector.php#L266-L269
    - https://github.com/rectorphp/rector/blob/88d57114364c1f86b7a46cbb3e0dcf06838e8389/src/Php/PhpVersionProvider.php

---

E.g. `70400`

<br>

This way here could re-used another numeric value.

What do you think?

...

`70400` <= I like it, it simplifies the comparison, and you can directly check against `PHP_VERSION_ID`.

Better doing `-1` instead of `-0.1` then it's always an `int`:

```php
public const BEFORE_NULLABLE_TYPE = NULLABLE_TYPE - 1;
```

_Originally posted by @TomasVotruba amd @leoloso in https://github.com/rectorphp/rector/pull/4192#discussion_r487393083_

samsonasik:


samsonasik:

src/Php/PhpVersionProvider.php
```diff
@@ -51,9 +52,20 @@ public function provide(): string
         return PHP_VERSION;
     }
 
-    public function isAtLeastPhpVersion(string $phpVersion): bool
+    /**
+     * @param string|int $phpVersion
+     */
+    public function isAtLeastPhpVersion($phpVersion): bool
```
samsonasik, line 58: make allow both string and int to avoid bc break in custom rector class that uses it.

TomasVotruba: I think we can go for `int` here, as next version will be 0.9 and BC break is expected.

samsonasik: implemented

samsonasik:

packages/testing/src/PHPUnit/AbstractRectorTestCase.php
```diff
@@ -378,7 +378,7 @@ private function configureEnabledRectors(EnabledRectorsProvider $enabledRectorsP
 
     private function configurePhpVersionFeatures(): void
     {
-        if ($this->getPhpVersion() === '') {
+        if ($this->getPhpVersion() === 0) {
```
TomasVotruba, line 381: This might be a bit confusing.
This zero and one above should be implemented in a named local constant, e.g

```php
private const PHP_VERSION_UNDEFINED = 0;
```

rules/code-quality/tests/Rector/Class_/CompleteDynamicPropertiesRector/CompleteDynamicPropertiesRectorTest.php
```diff
@@ -30,7 +30,7 @@ protected function getRectorClass(): string
         return CompleteDynamicPropertiesRector::class;
     }
 
-    protected function getPhpVersion(): string
+    protected function getPhpVersion(): int
     {
         return PhpVersionFeature::BEFORE_UNION_TYPES;
```
TomasVotruba, line 35: The `BEFORE_` constants now can be removed.

Their calls should be replaced with:

```suggestion
        return PhpVersionFeature::UNION_TYPES - 1;
```

packages/testing/src/PHPUnit/AbstractRectorTestCase.php
samsonasik: implemented.

rules/code-quality/tests/Rector/Class_/CompleteDynamicPropertiesRector/CompleteDynamicPropertiesRectorTest.php
samsonasik: implemented.

samsonasik:

phpstan.neon
```diff
@@ -741,5 +741,31 @@ parameters:
                 - packages/changes-reporting/src/Output/JsonOutputFormatter.php # 14
                 - src/Console/Output/RectorConfigurationFormatter.php # 13
 
+        -
+            message: '#Do not use scalar or array as constructor parameter\. Use "Symplify\\PackageBuilder\\Parameter\\ParameterProvider" service instead#'
```
TomasVotruba, line 745: Here I only changed the message, so it should be updatd in the other place in `phpstan.neon`.

```diff
-Do not use scalar or array as constructor parameter\. Use "ParameterProvider" service instead
+Do not use scalar or array as constructor parameter\. Use "Symplify\\PackageBuilder\\Parameter\\ParameterProvider" service instead
```

TomasVotruba (commented):
Thank you :+1: 

One last detail

phpstan.neon
samsonasik: implemented.

samsonasik:

TomasVotruba:
Thank you :clap:

PR #4684, opened 2020-11-24
update: drop migrify/php-config-printer for symplify/php-config-printer 

HDVinnie:
- Package migrify/php-config-printer is abandoned. Using symplify/php-config-printer instead.


composer.json
```diff
@@ -36,6 +35,7 @@
         "symplify/composer-json-manipulator": "dev-master",
         "symplify/console-color-diff": "dev-master",
         "symplify/package-builder": "dev-master",
+        "symplify/php-config-printer": "^0.4.2",
```
TomasVotruba, line 38: `dev-master` should be used instead, as other `symplify/*` packages. It's only temporary, till Symplify 9 gets release in a week or two

```suggestion
        "symplify/php-config-printer": "dev-master",
```

TomasVotruba:
Thannk you

Issue #4693, opened 2020-11-26
ChangeAndIfToEarlyReturnRector  should use `continue` within loops, instead of `return`


staabm:
# Bug Report

early returns introduced in foreach-loops alter logic

## Minimal PHP Code Causing Issue

https://getrector.org/demo/8ed4081f-01c5-4bab-9fec-6d001c73fbcb#result

## Expected Behaviour

since rector adds `return` the loop does not iterate all elements, as it did before the change.

input
```php
<?php

declare(strict_types=1);

final class DemoFile
{
	public function abc() {
		foreach($this->columns as $column) {
			// guess validation patterns based on fieldnames
			if (!empty($field_validators[$column->Field]) && empty($this->validates_format_of[$column->Field])) {
					$this->validates_format_of[$column->Field] = [
						'with' => $field_validators[$column->Field],
					];
			}
		}
	}
}
```

current result
```php
<?php

declare(strict_types=1);

final class DemoFile
{
	public function abc() {
        foreach ($columns as $column) {
            // guess validation patterns based on fieldnames
            if (empty($field_validators[$column->Field])) {
                return;
            }
            if (!empty($this->validates_format_of[$column->Field])) {
                return;
            }
            $this->validates_format_of[$column->Field] = [
                'with' => $field_validators[$column->Field],
            ];
        }
}
```

expected result
```php
<?php

declare(strict_types=1);

final class DemoFile
{
	public function abc() {
        foreach ($columns as $column) {
            // guess validation patterns based on fieldnames
            if (empty($field_validators[$column->Field])) {
                continue;
            }
            if (!empty($this->validates_format_of[$column->Field])) {
                continue;
            }
            $this->validates_format_of[$column->Field] = [
                'with' => $field_validators[$column->Field],
            ];
        }
	}
}
```

TomasVotruba:
Thanks for reporting :+1: 

We'll need a failing test for `ChangeAndIfToEarlyReturnRector`

It's matter of combining input + exptected to 1 file: https://github.com/rectorphp/rector/blob/master/docs/how_to_add_test_for_rector_rule.md

samsonasik:
It seems already fixed. I am closing this.

PR #4696, opened 2020-11-26
ChangeAndIfToEarlyReturnRector: added a failing testcase for continue-in-foreach

clxmstaab:
refs https://github.com/rectorphp/rector/issues/4693

this failing test just represents a example. same is true for `for` and `while` loops


TomasVotruba:
Thanks :+1: 

Just curious, why do you have 2 ids on Github? Is that on purpose or a different person? :)

clxmstaab:
> Just curious, why do you have 2 ids on Github? Is that on purpose or a different person? :)

one personal account and one corporate account :)

TomasVotruba:
Ah :). Why you use both of them in Rector? 
It very confusing for me and hard to track what I wrote to whom :D

clxmstaab:
> Ah :). Why you use both of them in Rector?
> It very confusing for me and hard to track what I wrote to whom :D

hehe sorry. depends on which browser I surf.

from the company I only have commit access with clxmstaab.

TomasVotruba:
Allright :) 

Would you like to fix this one? It's a good first issue

clxmstaab:
> Would you like to fix this one? It's a good first issue

I will give it a try.

having a look at the code, I found out:
- I need to adjust this line https://github.com/rectorphp/rector/blob/080e5cbe00b30073395091d162e8934ca29ce261/rules/solid/src/Rector/If_/ChangeAndIfToEarlyReturnRector.php#L207
- I need to check whether the current if has a foreach/while/for loop as one of its parents. If so, use "continue" otherwise use a "return" as we have right now

now I need to figure out, how to detect whether the current if is within a loop

clxmstaab:
I think I got it right for my new testcase. the change regresses a already existing test though.

I am pretty sure though that the test-expectations of `if_in_foreach.php.inc` is wrong.
the expected result changes the meaning of the original code, because it no longer guarantees that the loop will only process the first element

clxmstaab:
IMO `if_in_foreach` should be adjusted like

```
Index: rules/solid/tests/Rector/If_/ChangeAndIfToEarlyReturnRector/Fixture/if_in_foreach.php.inc
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/rules/solid/tests/Rector/If_/ChangeAndIfToEarlyReturnRector/Fixture/if_in_foreach.php.inc b/rules/solid/tests/Rector/If_/ChangeAndIfToEarlyReturnRector/Fixture/if_in_foreach.php.inc
--- a/rules/solid/tests/Rector/If_/ChangeAndIfToEarlyReturnRector/Fixture/if_in_foreach.php.inc	(revision ddbe858d5eb66c17410222c62194cfd6f9757a20)
+++ b/rules/solid/tests/Rector/If_/ChangeAndIfToEarlyReturnRector/Fixture/if_in_foreach.php.inc	(date 1606466114298)
@@ -42,6 +42,8 @@
                 return;
             }
             $this->processNameOrIdentifier($node);
+
+            return;
         }
 
         return;
```

but then I still don't know how to identify this case properly to choose correctly between `return` and `continue`

TomasVotruba:
> IMO if_in_foreach should be adjusted like

I've checked and you're right :+1:

TomasVotruba:
> but then I still don't know how to identify this case properly to choose correctly between return and continue.

That could be done. But we'd need a new rule to handle returns and their combination. I think it would be easier to maintain one rule that decides what to use in one place.

rules/solid/src/Rector/If_/ChangeAndIfToEarlyReturnRector.php
```diff
@@ -279,4 +286,18 @@ private function isLastIfOrBeforeLastReturn(If_ $if): bool
         }
         return $nextNode instanceof Return_;
     }
+
+    private function isIfInLoop(If_ $if): bool
+    {
+        $parent = $if->getAttribute(AttributeKey::PARENT_NODE);
```
TomasVotruba, line 292: Here might be used: 

```php
$this->betterNodeFinder->findFirstParentInstanceOf(...)
```

clxmstaab:
> That could be done. But we'd need a new rule to handle returns and their combination. I think it would be easier to maintain one rule that decides what to use in one place.

I agree. but I don't know how/where to place this logic in the rule, without breaking the existing stuff  :-)

TomasVotruba:
Personally, I'd start with 1 fixture only, fix and then merge it. 
It's easier to get the work done, because there is less problems at once.

Which of those 2 fixtures do you find easier to fix?

rules/solid/src/Rector/If_/ChangeAndIfToEarlyReturnRector.php
clxmstaab: done

clxmstaab:
> Which of those 2 fixtures do you find easier to fix?

the new one is already fixed. it regresses `if_in_foreach` atm.

rules/solid/src/Rector/If_/ChangeAndIfToEarlyReturnRector.php
```diff
@@ -198,13 +201,17 @@ private function getBooleanAndConditions(BooleanAnd $booleanAnd): array
...
      */
-    private function createInvertedIfNodesFromConditions(array $conditions): array
+    private function createInvertedIfNodesFromConditions(bool $isIfInLoop, array $conditions): array
     {
         $ifs = [];
         foreach ($conditions as $condition) {
             $invertedCondition = $this->conditionInverter->createInvertedCondition($condition);
             $if = new If_($invertedCondition);
-            $if->stmts = [new Return_()];
+            if ($isIfInLoop) {
```
TomasVotruba, line 210: I'm looking int to this and the only difference is extra `return` after the `if`.

<br>

So basically here must be check for the `return;` in $foreach->stmts after the `if`.

Something like:

```suggestion
            if ($isIfInLoop || $this->isFollowedByReturn((array) $foreach->stmts) {
```

TomasVotruba:
> the new one I introduced in the first step of this PR is already fixed. but the PR regresses if_in_foreach atm.

I see. Personaly I'd try to fix the regressoin first without adding a new test fixture. It's much faster to work with lower cognitive load. I'm looking into this PR and I'm already exhausted from switching between 2 cases :D.

clxmstaab:
> I'm looking into this PR and I'm already exhausted from switching between 2 cases :D.

ok, let me put the fix of the existing test in a separate PR

clxmstaab:
I feel this should now be fine. 

After https://github.com/rectorphp/rector/pull/4711 was merged, we just need to rebase.

just make sure we merge https://github.com/rectorphp/rector/pull/4711 before, so the diff of this PR gets smaller (it contains #4711 right now)

clxmstaab:
should be good to go now

TomasVotruba (approved):

TomasVotruba:
Thank you, very good job :+1:

PR #4708, opened 2020-11-27
use more specific phpdoc

clxmstaab:
allows static analysis to report errors for `findFirstParentInstanceOf()` calls

### DIFF
--- a/packages/node-nesting-scope/src/ValueObject/ControlStructure.php
+++ b/packages/node-nesting-scope/src/ValueObject/ControlStructure.php
@@ -20,7 +20,7 @@ use PhpParser\Node\Stmt\While_;
 final class ControlStructure
 {
     /**
-     * @var class-string[]
+     * @var class-string<Node>[]
      */
     public const BREAKING_SCOPE_NODE_TYPES = [
         For_::class,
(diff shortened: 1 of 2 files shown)


clxmstaab:
hmmm tried to reproduce the error we see in the phpstan build, but no chance atm

my last try is 
https://phpstan.org/r/fa20a8ec-8a16-4017-a780-33739393b50e

TomasVotruba:
I'd try running it with empty `phpstan.neon`, just on the single file and without symplify error formatter.

TomasVotruba:
PHPStan is not giving more info in this matter and actually breaks PHPStorm support. 
I want to avoid this mess (been there with similar tried in the past). Closing to keep PHPStorm support on.

clxmstaab:
> PHPStan is not giving more info in this matter and actually breaks PHPStorm support.
> I want to avoid this mess (been there with similar tried in the past). Closing to keep PHPStorm support on.

this should no longer be the case with the recent phpstorm release which adds support for phpstan/psalm based annotations

https://blog.jetbrains.com/phpstorm/2020/12/phpstorm-2020-3-release/

TomasVotruba:
I'm running on 2020.3 and still got troubles. 
If CI is passing in next PR, I'll give it try.

PR #4709, opened 2020-11-27
drop monorepo, packages moved to symplify

TomasVotruba:


TomasVotruba:
Merged in https://github.com/rectorphp/rector/pull/4709

PR #4559, opened 2020-11-05
[Prefixed Rector] Add package-scoper approach

TomasVotruba:
Pre-requisite for https://github.com/rectorphp/rector/issues/4913

Closes https://github.com/rectorphp/rector/issues/4406

<br>

## Reasoning

Why the switch from phar to scoped package? See https://github.com/symplify/symplify/issues/2465#issuecomment-723644671

tl;dr; This allows release of downgraded PHP 7.1, 7.0, 5.6 version. So Rector will not need Docker to run on the most legacy systems.	

## Todo

- push code here: https://github.com/rectorphp/rector-prefixed
- the CI there must pass

PR #4715, opened 2020-11-27
drop MinimalVersionChecker, is checked by composer itself

TomasVotruba:
Partical split from https://github.com/rectorphp/rector/pull/4559


src/PhpParser/Node/BetterNodeFinder.php
```diff
@@ -49,7 +49,7 @@ public function __construct(
     }
 
     /**
-     * @param class-string|class-string[] $type
+     * @param string|string[] $type
      */
```
TomasVotruba, line 53: To make PHPStorm prevent reporting it's not a string

Issue #4705, opened 2020-11-27
Use more labels for issues


simivar:
Personally, I find it difficult right now to navigate through Issues here as their topics could both mean an issue with an existing rule and/or a feature request for a new rule. I would propose to add new labels that would clearly state that issue is:
- a request for a new rule
- a bug in the existing rule/in general
- reproducible bug 

What do you think?

TomasVotruba:
It could help to navigate through issues, but there is currently no one who would maintain these labels.

simivar:
If you want that and need any help I can go through current issues and fill them with proper labels. If not I think this issue can be closed thanks to #4719.

TomasVotruba:
> If you want that and need any help I can go through current issues and fill them with proper labels

That would be very helpful. Thanks!
What access level do you need to be able to do that?

simivar:
From what I read on [this page](https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/managing-labels):
> To create, edit, **apply**, or delete a label, you must have **write access** to the repository.

TomasVotruba:
I've just send you the invitation

simivar:
I went through all open issues that currently are here in the repository and added (what I thought) are proper labels.

I created as well some new ones to make it better:
- [dx](https://github.com/rectorphp/rector/issues?q=is%3Aopen+is%3Aissue+label%3Adx) (7 issues)
- [docs](https://github.com/rectorphp/rector/issues?q=is%3Aopen+is%3Aissue+label%3Adocs) (1 issue)
- [feature](https://github.com/rectorphp/rector/issues?q=is%3Aopen+is%3Aissue+label%3Afeature) (70 issues)
- [on-hold](https://github.com/rectorphp/rector/issues?q=is%3Aopen+is%3Aissue+label%3Aon-hold) (1 issue)
- [performance](https://github.com/rectorphp/rector/labels/performance) (2 issues)
- [support](https://github.com/rectorphp/rector/issues?q=is%3Aopen+is%3Aissue+label%3Asupport) (7 issues)

TomasVotruba:
I'll go trought it :) 

Could you narrow it down to 5 top most used and drop the rest? That way we can maintain in the future and work with them effectively without adding too much maintenance

PR #4719, opened 2020-11-27
Add issue labels to issue templates

staabm:
With this PR issue templates are adjusted to automatically preselect issue-labels.

That way issues get some basic categorization and should be easier to navigate.
Its not exactly what was requested in #4705 but its a start which does not need big maintenance efforts

Refs https://github.com/rectorphp/rector/issues/4705


TomasVotruba:
Cool, I didn't know the label can be set-up there.

Love the 0-work for issue creators and maintainers :+1: Thank you

Issue #4606, opened 2020-11-14
Rector-Prefixed 0.8.50 not compatible with PHPStan PHAR


michaljusiega:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.8.50             |
| Installed as   | prefixed Rector PHAR                      |

```
Fatal error: Cannot redeclare PHPStan\dumpType() (previously declared in phar:///var/www/intranet/vendor/rector/rector-prefixed/rector/vendor/phpstan/phpstan-src/src/dumpType.php:9) in phar:///var/www/intranet/vendor/phpstan/phpstan/phpstan.phar/src/dumpType.php on line 9

Call Stack:
    0.0194     355632   1. {main}() /var/www/intranet/vendor/rector/rector-prefixed/rector:0
    0.1129    7362112   2. require('phar:///var/www/intranet/vendor/rector/rector-prefixed/rector/bin/rector') /var/www/intranet/vendor/rector/rector-prefixed/rector:6
    0.2181    7545440   3. _HumbugBox56c424d11509\AutoloadIncluder->includeDependencyOrRepositoryVendorAutoloadIfExists() phar:///var/www/intranet/vendor/rector/rector-prefixed/rector/bin/rector:33
    0.2182    7545400   4. _HumbugBox56c424d11509\AutoloadIncluder->loadIfNotLoadedYet() phar:///var/www/intranet/vendor/rector/rector-prefixed/rector/bin/rector:94
    0.2212    7545920   5. require_once('phar:///var/www/intranet/vendor/rector/rector-prefixed/rector/vendor/autoload.php') phar:///var/www/intranet/vendor/rector/rector-prefixed/rector/bin/rector:143
    0.2573    7620192   6. class_exists() phar:///var/www/intranet/vendor/rector/rector-prefixed/rector/vendor/autoload.php:17
    0.2573    7620256   7. spl_autoload_call() phar:///var/www/intranet/vendor/rector/rector-prefixed/rector/vendor/autoload.php:17
    0.2576    7620320   8. {closure:/var/www/intranet/vendor/phpstan/phpstan/bootstrap.php:3-46}() phar:///var/www/intranet/vendor/rector/rector-prefixed/rector/vendor/autoload.php:17
[...]

PR #4726, opened 2020-11-28
fix rector autoload when used as a global dependency

### LINKED ISSUE #4725: Class 'Rector\Core\Console\Style\SymfonyStyleFactory' not found
# Bug Report

| Rector version | d962e3f              |
| Installed as   | composer dependency                   |

just installed rector as a `composer global` dependency and starting it leads to the following exception

```
$ composer rector
> rector process --config=.tools/rector/rector.php
PHP Fatal error:  Uncaught Error: Class 'Rector\Core\Console\Style\SymfonyStyleFactory' not found in C:\dev\composer\vendor\rector\rector\bin\rector.php:35
Stack trace:
#0 C:\dev\composer\vendor\rector\rector\bin\rector(4): require_once()
#1 {main}
  thrown in C:\dev\composer\vendor\rector\rector\bin\rector.php on line 35
Fatal error: Uncaught Error: Class 'Rector\Core\Console\Style\SymfonyStyleFactory' not found in C:\dev\composer\vendor\rector\rector\bin\rector.php:35
Stack trace:
#0 C:\dev\composer\vendor\rector\rector\bin\rector(4): require_once()
#1 {main}
  thrown in C:\dev\composer\vendor\rector\rector\bin\rector.php on line 35
Script rector process --config=.tools/rector/rector.php handling the rector event returned with error code 255
```

```
$ php -v
PHP 7.3.0 (cli) (built: Dec  6 2018 02:17:00) ( ZTS MSVC15 (Visual C++ 2017) x86 )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.0-dev, Copyright (c) 1998-2018 Zend Technologies
    with blackfire v1.24.1~win-x32-zts73, https://blackfire.io, by Blackfire
```

on windows10x64

staabm:
like in other tools which work with the autoloader in global space, we need to use a proper path:

example https://github.com/sabre-io/dav/blob/d596e2b49cfc9cee27628fc52adb194502607af4/bin/sabredav.php#L28-L38

closes https://github.com/rectorphp/rector/issues/4725

### DIFF
--- a/bin/rector.php
+++ b/bin/rector.php
@@ -30,6 +30,7 @@ $autoloadIncluder = new AutoloadIncluder();
 $autoloadIncluder->includeCwdVendorAutoloadIfExists();
 $autoloadIncluder->autoloadProjectAutoloaderFile();
 $autoloadIncluder->includeDependencyOrRepositoryVendorAutoloadIfExists();
+$autoloadIncluder->autoloadComposerGlobalAutoloaerIfExists();
 $autoloadIncluder->autoloadFromCommandLine();
 
 $symfonyStyleFactory = new SymfonyStyleFactory(new PrivatesCaller());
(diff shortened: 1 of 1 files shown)


TomasVotruba:
As the method name states, this should handle project's autoload in case of install in `vendor`.
That would be broken now.

Also, Rector does not support global install, as there are many edge cases where it fails.

staabm:
> Also, Rector does not support global install, as there are many edge cases where it fails.

hmm I see. I have adjusted the PR to add 'experimental' composer global autoloading and restored the previous behaviour. it works for me as a workaround because [rector-prefixed is not usable atm](https://github.com/rectorphp/rector/issues/4606).

TomasVotruba:
I see. It would be better have scopes Rector fixed, rather then adding another standard that can break. 

Here is some work in progress for the scope fix:
https://github.com/rectorphp/rector/pulls

Help is welcomed

PR #4728, opened 2020-11-28
ChangeAndIfToEarlyReturnRector: added failing test

staabm:


rules/solid/tests/Rector/If_/ChangeAndIfToEarlyReturnRector/Fixture/nested_if_in_foreach.php.inc
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Rector\SOLID\Tests\Rector\If_\ChangeAndIfToEarlyReturnRector\Fixture;
+
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Name;
+
+class NestedIfInForeachClass
```
staabm, line 8: this class gets transformed into the following which is no longer equivalent to the previous

```
<?php

namespace Rector\SOLID\Tests\Rector\If_\ChangeAndIfToEarlyReturnRector\Fixture;

use PhpParser\Node\Identifier;
use PhpParser\Node\Name;

class NestedIfInForeachClass
{
    private function refactorStmts(array $nodes): void
    {
        $g = [];

        foreach ($nodes as $node) {
            $x = ['A', 'B'];
            if (rand(1,2)) {
                // something
            } else {
                if (!$node instanceof Name) {
                    continue;
                }
                if (!$node instanceof Identifier) {
                    continue;
                }
                $x = [];
            }

            $g = $x;
        }
    }
}

?>
```

TomasVotruba: What would be the correct transformation of `if`?

staabm: I guess no tranformation shoud happen

TomasVotruba: Agreed. The test fixture name should start with `skip_` and the class with `Skip`

Care to fix it at the same PR?

staabm: Will try later

staabm: Do you have suggestions on how we should fix this case / detect this case?

TomasVotruba: Probably check the if + while combination

staabm:
hmm I guess I am pretty close, but don't know where I have a bug here.

staabm:
Just realized I was missing some use-statments.

Wondering whether the phpstan errors within the GithubActions do intentionally not emit proper errors which get annotated in the PR tab?

clxmstaab:
is there a online tool like the rector demo or the phpstan online sandbox which takes a php-source code and visualizes its AST?

it could help to get an idea how the AST finally looks like of certain snippets and what needs to be done.

TomasVotruba:
> is there a online tool like the rector demo or the phpstan online sandbox which takes a php-source code and visualizes its AST?

Nothing I know of.

There is nodes overview, with other way around: https://github.com/rectorphp/rector/blob/master/docs/nodes_overview.md
But both examples are there

TomasVotruba:
> Wondering whether the phpstan errors within the GithubActions do intentionally not emit proper errors which get annotated in the PR tab?

It was rather confusing, as the other actions are in GitHub action list only and it was missing the context/fatal errors.

staabm:
> > is there a online tool like the rector demo or the phpstan online sandbox which takes a php-source code and visualizes its AST?
> 
> Nothing I know of.
> 
> There is nodes overview, with other way around: https://github.com/rectorphp/rector/blob/master/docs/nodes_overview.md
> But both examples nodes + code are there

Just found https://astexplorer.net/ And https://github.com/ircmaxell/php-ast-visualizer/blob/master/demo.php

TomasVotruba:
Just a tip, I use the `dump()` and it's working good so far.

samsonasik:
@staabm I cherry-picked your commits at https://github.com/rectorphp/rector/pull/4757

staabm:
@samsonasik could you gibe me a hint where you think my code is buggy?

samsonasik:
probably because of find loop types as well in the very first, then re-try check its parent is in loop type again.

PR #4730, opened 2020-11-28
Fix permissions on /tmp in Docker

### LINKED ISSUE #4568: Docker Image runs as root
# Bug Report

Rector runs as root. The permissions of all files are broken.

Overriding the user does not help:

```
❯ docker run --rm -v (pwd):/project -u1000 rector/rector:latest process src

 [ERROR] Unable to write in the "cache" directory (/tmp/_rector).     
```

JanMikes:
Fixes #4568

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,8 @@ COPY . .
 # To warmup opcache a little
 RUN bin/rector list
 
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thanks :)

Issue #4723, opened 2020-11-28
`rector --version` is super slow


staabm:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.8.56           |
| Installed as   | composer dependency via global command                    |

invoking rector even without doing transformations is super slow.
actually all rector commands are slow on my machine, but I guess for a minimal repro it would be great to just get `rector --version` fast.

I am wondering why the symfony container get dumped every time and is not persisted. is this the normal thing how it is expected to work?

## Minimal PHP Code Causing Issue

```
$ time php /c/Users/staabm/AppData/Roaming/Composer/vendor/rector/rector/bin/rector --version
Rector 0.8.56

real    0m46.023s
user    0m0.015s
sys     0m0.077s
```

### blackfire profile

https://blackfire.io/profiles/2fed9c66-6fd4-4a37-a3b5-92e5bb2f4c3b/graph (might break after 30 days)

![grafik](https://user-images.githubusercontent.com/120441/100497550-7c792b00-315c-11eb-8895-dc480c49cdff.png)

it seems the container gets dumped over and over on every invocation

## Expected Behaviour

running `rector --version` should be nearly instant

```
$ php -v
PHP 7.3.0 (cli) (built: Dec  6 2018 02:17:00) ( ZTS MSVC15 (Visual C++ 2017) x86 )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.0-dev, Copyright (c) 1998-2018 Zend Technologies
    with blackfire v1.24.1~win-x32-zts73, https://blackfire.io, by Blackfire
```

staabm:
@nicolas-grekas do you have a hint what could be the reason that the symfony container need to do so much IO?

nicolas-grekas:
I don't sorry. Looking at the graph, there is some globbing happening, but I can't tell what this is for...

TomasVotruba:
Global install is not supported. Does this happen in `composer require`?

staabm:
I get similar profiles and other VMs, in which the dependency is added on a per project basis:

```bash
$ blackfire run vendor/bin/rector                                                                                                                                                                             
Rector 0.8.49

Usage:
  command [options] [arguments]

Options:
  -h, --help                     Display this help message
  -V, --version                  Display this application version
      --ansi                     Force ANSI output
      --no-ansi                  Disable ANSI output
[...]

PR #4743, opened 2020-12-01
remove double `**` glob to speedup bootstrapping

staabm:
with this change a `php bin/rector` on my machine finishes in ~13 seconds instead of ~50 seconds.
blackfire shows a 70% performance improvement

grafik

refs https://github.com/rectorphp/rector/issues/4723

I did very rough testing, by e.g. compare the result of `php bin/rector dump-nodes` before/after the PR.
this did not show any issues. lets see what CI thinks about such a change.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -7,8 +7,8 @@ use Rector\Core\ValueObject\ProjectType;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 
 return static function (ContainerConfigurator $containerConfigurator): void {
-    $containerConfigurator->import(__DIR__ . '/../packages/*/config/config.*');
-    $containerConfigurator->import(__DIR__ . '/../rules/*/config/config.*');
+    $containerConfigurator->import(__DIR__ . '/../packages/*/config/config.php');
+    $containerConfigurator->import(__DIR__ . '/../rules/*/config/config.php');
     $containerConfigurator->import(__DIR__ . '/services.php');
     $containerConfigurator->import(__DIR__ . '/../utils/*/config/config.php', null, true);
 
--- a/config/config.php
+++ b/config/config.php
@@ -7,10 +7,10 @@ use Rector\Core\ValueObject\ProjectType;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 
 return static function (ContainerConfigurator $containerConfigurator): void {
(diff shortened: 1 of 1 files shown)


config/config.php
```diff
@@ -7,10 +7,10 @@
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 
 return static function (ContainerConfigurator $containerConfigurator): void {
-    $containerConfigurator->import(__DIR__ . '/../packages/**/config/config.*');
-    $containerConfigurator->import(__DIR__ . '/../rules/**/config/config.*');
+    $containerConfigurator->import(__DIR__ . '/../packages/*/config/config.*');
+    $containerConfigurator->import(__DIR__ . '/../rules/*/config/config.*');
```
staabm, line 11: does rector still support `.yaml` based rules, or can this be restricted to `config.php` ?

TomasVotruba: The latter, we use only `*.php` now

TomasVotruba:
Wow, this is amazing improvement :+1: 
How did you find this out?

Are there any other globs/imports that can be improved?

staabm:
> Wow, this is amazing improvement 👍
> How did you find this out?

Its part of the investigation in https://github.com/rectorphp/rector/issues/4723
The blackfire profile was pointing me into the direction and some echo-debug statements later things got clearer.

> 
> Are there any other globs/imports that can be improved?

At least not within the bootstrapping. I will keep an eye open while working on other rector stuff.

config/config.php
staabm: Ok, adjusted the patter. i guess this will fuether speedup the bootstrapping a bit

TomasVotruba:
I see. That sounds like very detailed work :+1: 

> Hopefully the build is returning green.

GitHub Actions are down last 3 hours :/ https://www.githubstatus.com/

staabm:
I think this is good to go. Did not test the latest small adjustments, though.

staabm:
It seems GH actions recover slowly. I guess you could restart the builds

TomasVotruba:
Restarded :+1: 

It's still lagging though. My other PRs wait for 7 minutes now.

staabm:
Greeeen ;-)

TomasVotruba:
Looks as green as it could be :heavy_check_mark: 

Let's shipt it

PR #4753, opened 2020-12-02
Fix CallTypeAnalyzer when methodName is null

lulco:
Argument 1 passed to PHPStan\Reflection\ClassReflection::hasMethod() must be of the type string, null given

Happens when I have this code:

```
protected function createFormControl(Container $form): IControl
{
    $method = 'add' . ucfirst($this->type);
    $control = $form->$method($this->getKey(), $this->getTitle(), $this->multiple);
    return $control;
}
```

Probably all $foo->$bar calls fails on this

### DIFF
--- a/rules/generic/src/NodeTypeAnalyzer/CallTypeAnalyzer.php
+++ b/rules/generic/src/NodeTypeAnalyzer/CallTypeAnalyzer.php
@@ -83,6 +83,10 @@ final class CallTypeAnalyzer
         /** @var string $methodName */
         $methodName = $this->nodeNameResolver->getName($node->name);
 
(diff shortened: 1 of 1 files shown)


rules/generic/src/NodeTypeAnalyzer/CallTypeAnalyzer.php
```diff
@@ -83,6 +83,10 @@ private function getMethodParameterTypes(string $className, Node $node): array
         /** @var string $methodName */
```
TomasVotruba, line 83:
```suggestion
```

TomasVotruba:
Thank you :+1:

PR #4769, opened 2020-12-04
Added tests for samples - proof of concept

lulco:
Hi,
I was thinking if it would be possible to automatically test code samples which are required in rule definition, so I prepared this proof of concept and add it to over 100 rules.
It works now for all "simple" replacing rules like  strpos -> contains etc.
Not working for rules where the type of class / interface is needed. Probably because of some missing visitor(s) or something like that. Maybe you will give me some hint how to use it in this test.

### DIFF
--- a/rules/code-quality/src/Rector/Assign/CombinedAssignRector.php
+++ b/rules/code-quality/src/Rector/Assign/CombinedAssignRector.php
@@ -17,6 +17,8 @@ use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
  */
 final class CombinedAssignRector extends AbstractRector
 {
+    public $testSamples = true;
+
     /**
      * @var AssignAndBinaryMap
      */
--- a/rules/code-quality/src/Rector/Catch_/ThrowWithPreviousExceptionRector.php
+++ b/rules/code-quality/src/Rector/Catch_/ThrowWithPreviousExceptionRector.php
@@ -29,6 +29,8 @@ use Throwable;
  */
 final class ThrowWithPreviousExceptionRector extends AbstractRector
 {
+    public $testSamples = true;
+
     /**
      * @var int
      */
--- a/rules/code-quality/src/Rector/Concat/JoinStringConcatRector.php
+++ b/rules/code-quality/src/Rector/Concat/JoinStringConcatRector.php
@@ -18,6 +18,8 @@ use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
  */
 final class JoinStringConcatRector extends AbstractRector
 {
+    public $testSamples = true;
+
     /**
      * @var int
      */
--- a/rules/code-quality/src/Rector/Foreach_/UnusedForeachValueToArrayKeysRector.php
+++ b/rules/code-quality/src/Rector/Foreach_/UnusedForeachValueToArrayKeysRector.php
@@ -18,6 +18,8 @@ use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
  */
 final class UnusedForeachValueToArrayKeysRector extends AbstractRector
 {
+    public $testSamples = true;
+
     public function getRuleDefinition(): RuleDefinition
     {
         return new RuleDefinition(
--- a/rules/code-quality/src/Rector/FuncCall/InArrayAndArrayKeysToArrayKeyExistsRector.php
+++ b/rules/code-quality/src/Rector/FuncCall/InArrayAndArrayKeysToArrayKeyExistsRector.php
@@ -16,6 +16,8 @@ use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
  */
 final class InArrayAndArrayKeysToArrayKeyExistsRector extends AbstractRector
 {
+    public $testSamples = true;
+
     public function getRuleDefinition(): RuleDefinition
     {
         return new RuleDefinition(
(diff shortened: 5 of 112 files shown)


TomasVotruba:
Hi,

tl;dr;

Where is the main changed file?

What is the goal of this?

lulco:
Yeah :) sorry, I didn't count there will be so many changes O:-)

Main file is [tests/Samples/SamplesTest.php](https://github.com/rectorphp/rector/pull/4769/files#diff-cbfa1946005a01a7d45c9571f72660d767e65d545b21fdc43be0064dbb5eee5b) and it tests all rectors against their rule definitions (all its code samples)

As I wrote in description, the goal of this PR is to test Code Samples from Rector Rules:
- we have to write them, so we can use them as test cases (I noticed that fixtures are very often the same as code samples so we can skip creating these types of fixtures)
- test will check if Code Samples are correct
- we will increase code coverage


- final goal is to test all Code Samples, but while it is not possible for now, I added public property `$testSamples` to those which are testable in this moment (I would like to remove it and test all rectors)

TomasVotruba:
Ah, I see :) good observation. We're actually doing this, just the other way round: https://github.com/rectorphp/rector/blob/master/docs/rector_recipe.md

Writing rules manually is waste of time. We generate them instead, so first fixture and code sample is the same snippet.

That way code is both valid and tested, right from the start.

lulco:
OK, I see :)

But code sample itself isn't tested, is it? If it is changed or added manually, developer will not know there is some problem. 

You can see some rectors with errors in samples in diff of this PR.

For example missing semicolon:
https://github.com/rectorphp/rector/pull/4769/files#diff-40193dea3781382ce687f28e26d0d28b2382a23db7434b967e166e511447748c

Or missing `function` keyword:
https://github.com/rectorphp/rector/pull/4769/files#diff-bf49a492d9cf13a703cbec4b70ff3f5950b4cbf4bcfe56e50244bac1b89707ba

And several minor indentation issues.

With the test created in this PR, you can skip step for generating the first fixtures from code samples. The CODE will be tested, not only FIXTURES. And I believe that CodeSamples are used for generating documentation so also examples in documentation will be valid.

TomasVotruba:
The code snippets could be linted to find bugs above, but they should not be part of test battery. That woudl mix concerts. Rules snippets are for documentaiton, fixture are for testing.

> With the test created in this PR, you can skip step for generating the first fixtures from code samples.

That would actually create 2 soucres of testing. Where to put tests then?

<br>

Could you remove the `$testSamples` property and make it work without it?

lulco:
> The code snippets could be linted to find bugs above, but they should not be part of test battery. That would mix concerns. Rules snippets are for documentaiton, fixture are for testing.

Yeah I understand that. I just try to remove duplicities.

> That would actually create 2 soucres of testing. Where to put tests then?
Well yes. One test is global for all rectors and second source will stay as is near rector. As I said, I was just trying to remove duplicities and increase code coverage, because Code Samples are part of code whether you want to or not and they should be tested.
 
> The `$testSamples` property is now extra work, that adds maintenance cost.
Yes it is. For now, you just add it to Rectors you want to test. Default value in AbstractRector is false.

> Could you remove the `$testSamples` property and make it work without it?
Not now, while all Code samples will pass tests :( 

As I wrote in PR description:
> Not working for rules where the type of class / interface is needed. Probably because of some missing visitor(s) or something like that. Maybe you will give me some hint how to use it in this test.

But if you are not interested to test Code Samples, why should I do that?

TomasVotruba:
The property is only blocker here. It should be doable without it as well.

TomasVotruba:
This could be solve in more generic way, e.g. as part of ECS: https://github.com/symplify/symplify/issues/2589

lulco:
> This could be solve in more generic way, e.g. as part of ECS: [symplify/symplify#2589](https://github.com/symplify/symplify/issues/2589)

I don’t think this is just linter of samples. Primary target was to test Rector itself with code sample(s) as test case(s).

lulco:
> The property is only blocker here. It should be doable without it as well.

How it can be doable without some property or interface or whatever which marks Rectors to test and those which should not be tested?

My problem here is that I didn’t find way how to run all rectors with samples the same way as process command do that. I think the reason is that I can’t use some node visitors which set scope of class somehow based on file path - test samples have no file path, it is just string, so in rectors with conditions isObjectType() I always get false, because resolved type is always MixedType (because node scope is null)

TomasVotruba:
> How it can be doable without some property or interface or whatever which marks Rectors to test and those which should not be tested?

The same way all 100 % test fixtures are tested. If just *some* test would be run, it would kill the design of PHPUnit. We need the same ideology here to make it work in the longterm.

lulco:
I agree we need 100%. Can you please read also the second part of my answer and maybe give je some hints?

TomasVotruba:
I've read that but didn't get what is the problem. There is no need to run Rectors, code samples are isolated pieces of code.

TomasVotruba:
There might be some inspiration: https://github.com/symplify/symplify/blob/b485035efd12da69c8e2177d70f4762ffc2eab7d/packages/rule-doc-generator/src/RuleDefinitionsResolver.php#L29

The symplify/rule-doc-generator package uses there definitions and works with them, but does not work with the rest of rule class at all.

TomasVotruba:
php-parser is quite heavy tool just to lint. `token_get_all()` might do better

lulco:
OK, it is completely different approach, so I'm closing this PR.

PR #4771, opened 2020-12-04
[DX] Fixes #3675 Report running rule without configuration

### LINKED ISSUE #3675: [DX] Report running rule without configuration
To prevent empty run like these https://twitter.com/realFlowControl/status/1281145781165207553

E.g. throw exception:

```bash
The rule X is running without any configraution, is that on purpose?
```

samsonasik:


samsonasik:
There are empty configuration usage in test that if no check on running phpunit test, it will got: 

```php
59) Rector\Php74\Tests\Rector\Property\TypedPropertyRector\ImportedTest::test with data set #0 (Symplify\SmartFileSystem\SmartFileInfo Object (...))
Rector\Core\Exception\ShouldNotHappenException: Rule "Rector\Php74\Rector\Property\TypedPropertyRector" is running without any configuration, is that on purpose?
```

I disable check on empty configuration on running tests.

samsonasik:
@TomasVotruba For `TypedPropertyRector`, empty configuration means fallback to false value:

```php
    public function configure(array $configuration): void
    {
        $this->classLikeTypeOnly = $configuration[self::CLASS_LIKE_TYPE_ONLY] ?? false;
    }
```

which make absolutely fine. That means, if this report applied, this can break existing usage?

TomasVotruba:
> which make absolutely fine. 

Agreed. What od you think would be the best solution?

samsonasik:
probably add a flag check, eg: `$isAllowEmptyConfig` in the rector class, default to true ?

TomasVotruba:
That would reuqire extra maintenance of these rules.

What about something without changing rules, that would respect default behavior and warn about missing configuration?

I think only empty arrays are the problem, everything else is ok.

samsonasik:
How to warn and make keep working ? with `ShouldNotHappenException` seems just stop the execution, eg in phpunit, it shows above error https://github.com/rectorphp/rector/pull/4771#issuecomment-738595603

TomasVotruba:
Empty arrays should really throw an exception, as the rule will run with no effect. 

That's either missconfiguration, or broken Rector configuration part.

samsonasik:
Ok, so, this change is now correct? I disabled check on empty configuration on running tests.

TomasVotruba:
Yes, that looks good :+1:

PR #4772, opened 2020-12-04
[Downgrade PHP 8.0] Add DowngradeClassOnObjectToGetClassRector

samsonasik:


samsonasik:
I don't have PHP 8.0 in my local dev, and it seems there is no PHP 8.0 in github actions? How to test it?

rules/downgrade-php80/src/Rector/ClassConstFetch/DowngradeClassOnObjectToGetClassRector.php
```diff
@@ -0,0 +1,88 @@
...
+        $nameIdentifier = $node->name;
+        if (! $nameIdentifier instanceof Identifier) {
+            return null;
+        }
+
+        /** @var string $name */
+        $name = $nameIdentifier->toString();
+        if ($name !== 'class') {
+            return null;
+        }
```
TomasVotruba, line 80:
```suggestion
	    if ($this->isName($node->name, 'class')) {
	    	return null;
	    }
```

This could work too

samsonasik: updated.

TomasVotruba:
Thank you :+1:

TomasVotruba:
> I don't have PHP 8.0 in my local dev, and it seems there is no PHP 8.0 in github actions? How to test it?

Someone has to add PHP 8.é GitHub Actions first.

samsonasik:
I will try.

PR #4755, opened 2020-12-02
Downgrade PHP 7.4 covariant return type

leoloso:
Fixes #4163

PR #4774, opened 2020-12-04
Downgrade PHP 7.4 Covariant return types - Handle also interfaces

leoloso:
#4755 added support for covariant return types from class inheritance, but interfaces were missing. Fixed now


leoloso:
I had a typo in the class name: `DowngradeCovarianReturnTypeRector` instead of `DowngradeCovariantReturnTypeRector` (a "t" was missing!!!). For convenience, I'm also fixing it in this PR

TomasVotruba:
Thank you :+1:

PR #4778, opened 2020-12-04
[DX] Fixes #4777 Update ConfiguredCodeSample to CodeSample for no configuration in Rector class

### LINKED ISSUE #4777: Report running rule without configuration will cause issue in AbstractDowngradeRector with empty config
# Bug Report

| Rector version | dev-master (invoke `vendor/bin/rector --version`)              |

@TomasVotruba as DX PR https://github.com/rectorphp/rector/pull/4771 , it seems will cause issue in class extends `AbstractDowngradeRector` which using empty configuration, eg: on `DowngradeClassConstantVisibilityRector`:

```php
final class DowngradeClassConstantVisibilityRector extends AbstractDowngradeRector
{
    public function getRuleDefinition(): RuleDefinition
    {
        return new RuleDefinition(
            'Downgrade class constant visibility',
            [
                new ConfiguredCodeSample(
                    <<<'CODE_SAMPLE'
<?php

class SomeClass
{
   public const PUBLIC_CONST_B = 2;
   protected const PROTECTED_CONST = 3;
   private const PRIVATE_CONST = 4;
}
CODE_SAMPLE
                    ,
                    <<<'CODE_SAMPLE'
<?php

class SomeClass
{
   const PUBLIC_CONST_B = 2;
   const PROTECTED_CONST = 3;
   const PRIVATE_CONST = 4;
}
CODE_SAMPLE
                    ,
                    []
                ),
            ]
        );
    }
```

The `ConfiguredCodeSample` needs to be changed to `CodeSample`, and its class itself need to extends directly to `AbstractRector`

samsonasik:


samsonasik:

TomasVotruba:
Thank you :partying_face:

PR #4791, opened 2020-12-05
[CodeQuality] Add DateTimeToDateTimeInterfaceRector

### LINKED ISSUE #4478: Fixer to migrate from `DateTime` to `DateTimeInterface`
# Feature Request
Is there a fixer to allow migration from `DateTime` to `DateTimeInterface`? If not, would you think such a fixer would be useful?

`DateTime` is mutable, whereas `DateTimeImmutable` is immutable. Call to methods `add`, `modify`, `setDate`, `setISODate`, `setTime`, `setTimestamp`, `setTimezone` and `sub` modify the original object in `DateTime`, whereas a new object is returned in case of `DateTimeImmutable`. Other than that, the behaviour is exactly the same between `DateTime` and `DateTimeImmutable`. And all instances of `DateTime` can be replaced with `DateTimeInterface`.

Switching from `DateTime` to `DateTimeInterface` this way will cause static analyzers to spew warnings, since the above methods are not part of `DateTimeInterface`. If it is possible to add a `@param \DateTimeImmutable|\DateTime $var` phpdoc annotation on changing the function signature, that would solve this.

`DateTimeInterface` [cannot be implemented](https://github.com/php/php-src/blob/41b096b392e551df51c80d29f6f30c163a0630ea/ext/date/php_date.c#L1565) by other classes, so it is safe to assume that `DateTime` and `DateTimeImmutable` are the only classes implementing `DateTimeInterface`.
 
## Diff 

```diff
- function nextDay(\DateTime $datetime) {
-     $datetime->modify('+1 day');
      return $datetime;
+ /**
+  * @param \DateTime|\DateTimeImmutable $datetime
+  */
+ function nextDay(\DateTimeInterface $datetime) {
+     $datetime = $datetime->modify('+1 day');
      return $datetime;
}

- function getDay(\DateTime $datetime) {
+ /**
+  * @param \DateTime|\DateTimeImmutable $datetime
+  */
+ function getDay(\DateTimeInterface $datetime) {
     return $datetime->format('n');
}

- /**
-  * @param \DateTime $datetime
-  */
- function foo(\DateTime $datetime) {
-     return getDay($datetime->modify('+1 day'));
+ /**
+  * @param \DateTime|\DateTimeImmutable $datetime
+  */
+ function foo(\DateTimeInterface $datetime) {
+     return getDay($datetime = $datetime->modify('+1 day'));
}
```

simivar:
Closes #4478

### DIFF
--- a/rules/code-quality/src/Rector/ClassMethod/DateTimeToDateTimeInterfaceRector.php
+++ b/rules/code-quality/src/Rector/ClassMethod/DateTimeToDateTimeInterfaceRector.php
@@ -177,7 +177,7 @@ CODE_SAMPLE
             return true;
         }
 
-        if (! in_array($this->getName($methodCall->name), self::METHODS_MAP)) {
+        if (! $this->isNames($methodCall->name, self::METHODS_MAP)) {
             return true;
         }
 
--- a/rules/code-quality/src/Rector/ClassMethod/DateTimeToDateTimeInterfaceRector.php
+++ b/rules/code-quality/src/Rector/ClassMethod/DateTimeToDateTimeInterfaceRector.php
@@ -1,0 +1,196 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\CodeQuality\Rector\ClassMethod;
+
+use DateTime;
+use DateTimeImmutable;
+use DateTimeInterface;
+use PhpParser\Node;
+use PhpParser\Node\Arg;
+use PhpParser\Node\Expr\Assign;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Name\FullyQualified;
+use PhpParser\Node\NullableType;
+use PhpParser\Node\Param;
+use PhpParser\Node\Stmt\ClassMethod;
+use PHPStan\Type\NullType as PHPStanNullType;
+use PHPStan\Type\ObjectType as PHPStanObjectType;
+use PHPStan\Type\UnionType as PHPStanUnionType;
+use Rector\BetterPhpDocParser\PhpDocInfo\PhpDocInfo;
+use Rector\Core\Exception\ShouldNotHappenException;
+use Rector\Core\Rector\AbstractRector;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+use Rector\NodeTypeResolver\Node\AttributeKey;
+use Rector\NodeTypeResolver\NodeTypeResolver;
+
+/**
+ * @see \Rector\CodeQuality\Tests\Rector\ClassMethod\DateTimeToDateTimeInterfaceRector\DateTimeToDateTimeInterfaceRectorTest
+ */
+final class DateTimeToDateTimeInterfaceRector extends AbstractRector
+{
+    private const METHODS_MAP = [
+        'add', 'modify', '__set_state', 'setDate', 'setISODate', 'setTime', 'setTimestamp', 'setTimezone', 'sub',
+    ];
+
+    /**
+     * @var NodeTypeResolver
+     */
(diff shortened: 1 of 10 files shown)


rules/code-quality/src/Rector/ClassMethod/DateTimeToDateTimeInterfaceRector.php
```diff
@@ -0,0 +1,196 @@
...
+        $parentNode->expr = $newAssignNode;
+    }
+
+    private function shouldSkipMethodCallRefactor(string $paramName, MethodCall $methodCall): bool
+    {
+        if (! $this->isName($methodCall->var, $paramName)) {
+            return true;
+        }
+
+        if (! in_array($this->getName($methodCall->name), self::METHODS_MAP)) {
```
TomasVotruba, line 180:
```suggestion
        if (! $this->isNames($methodCall->name, self::METHODS_MAP)) {
```

rules/code-quality/src/Rector/ClassMethod/DateTimeToDateTimeInterfaceRector.php
```diff
@@ -0,0 +1,196 @@
...
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+use Rector\NodeTypeResolver\Node\AttributeKey;
+use Rector\NodeTypeResolver\NodeTypeResolver;
+
+/**
+ * @see \Rector\CodeQuality\Tests\Rector\ClassMethod\DateTimeToDateTimeInterfaceRector\DateTimeToDateTimeInterfaceRectorTest
+ */
+final class DateTimeToDateTimeInterfaceRector extends AbstractRector
+{
+    private const METHODS_MAP = [
```
TomasVotruba, line 35: What kind of map is this? Excluded or included?
More specific naming could help

TomasVotruba (commented):
Very clean work :+1:

rules/code-quality/src/Rector/ClassMethod/DateTimeToDateTimeInterfaceRector.php
```diff
@@ -0,0 +1,196 @@
...
+    {
+        return [ClassMethod::class];
+    }
+
+    /**
+     * @param ClassMethod $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        $isModifiedNode = false;
```
TomasVotruba, line 90: I saw that `DateTimeInterface` was added only in PHP 5.5: https://www.php.net/manual/en/class.datetimeinterface.php

Here should be check for PHP version, as minimal is 5.3. Just so we don't break code during downgrades.

```suggestion
        		if (! $this->isAtLeastPhpVersion(PhpVersionFeature::DATE_TIME_INTERFACE)) {
                	return null;
   		        }
        
                $isModifiedNode = false;
        

```


The `DATE_TIME_INTERFACE` constant must be added to `PhpVersionFeature`

TomasVotruba (commented):
I've left few comments. After they get resolved, it's ready to merge :+1:

rules/code-quality/src/Rector/ClassMethod/DateTimeToDateTimeInterfaceRector.php
simivar: I've renamed it to `METHODS_RETURNING_CLASS_INSTANCE_MAP`.

src/ValueObject/PhpVersionFeature.php
```diff
@@ -16,6 +16,11 @@ final class PhpVersionFeature
      */
     public const ELVIS_OPERATOR = 50300;
 
+    /**
+     * @var int
+     */
+    public const DATE_TIME_INTERFACE = 50300;
```
TomasVotruba, line 22:
```suggestion
    public const DATE_TIME_INTERFACE = 50500;
```

The docs shows `>= 5.5.0`

image

TomasVotruba:
Thank you :+1:

Issue #4765, opened 2020-12-03
[PHP 8.0] Undocumented BC breaks


TomasVotruba:
These should be part of PHP 8 set

See https://github.com/rectorphp/rector/pull/4762/commits/82e92c0c0dca26287b4fc43ec2ce9ce98265f5e7

- `private` and `final` cannot be used together on class method
	→ drop `final`, as `private` is stating it clearly for both

- parent cannot be called, if no parent class
	→ remove parent (handled by code quality rule)

- `__setState()` must be `static`
	→ add visibility to it

simivar:
Closing as it was done in #4792, #4794 and #4795.

TomasVotruba:
Thank you :+1:

PR #4792, opened 2020-12-05
[PHP 8.0] Add SetStateToStaticRector

simivar:
Part of #4765. I didn't use `ChangeMethodVisibilityRector` as it required a class name to be specified and this one should be a generic one.

A way to use generic `ChangeMethodVisibilityRector` could be to refactor it to not require class name but I don't know if that's a visible solution for you.

### DIFF
--- a/rules/php80/src/Rector/ClassMethod/SetStateToStaticRector.php
+++ b/rules/php80/src/Rector/ClassMethod/SetStateToStaticRector.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Php80\Rector\ClassMethod;
+
+use PhpParser\Node;
+use PhpParser\Node\Stmt\ClassMethod;
+use Rector\Core\Rector\AbstractRector;
+use Rector\Core\ValueObject\MethodName;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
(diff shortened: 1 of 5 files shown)


simivar:
@TomasVotruba I might need some guidance on how to make it work with PHP 8 tests.

TomasVotruba:
I'd skip the test on PHP 8, as the code is not PHP 8 compatible.

Something like:

```php
@requires PHP < 8
```

rules/php80/src/Rector/ClassMethod/SetStateToStaticRector.php
```diff
@@ -0,0 +1,80 @@
...
+            return true;
+        }
+
+        if ($classMethod->isStatic()) {
+            return true;
+        }
+
+        return false;
+    }
+}
```
TomasVotruba, line 80: :clap:

TomasVotruba:
> I didn't use ChangeMethodVisibilityRector as it required a class name to be specified and this one should be a generic one.

That's probably a good choice... I wonder, would this doo?

```php
[stdClass::class, '__set_state', 'static']
```

TomasVotruba:
Ah, it doesn't seem to work: https://3v4l.org/7RJ0X

TomasVotruba:
Thank you :+1:

rules/php80/src/Rector/ClassMethod/SetStateToStaticRector.php
simivar: Glad you liked it 👍

PR #4806, opened 2020-12-06
[Downgrade][PHP 7.3] Add SetCookieOptionsArrayToArgumentsRector

### LINKED ISSUE #4328: [Downgrade PHP 7.3] Set(raw)cookie accepts $option Argument
# Feature Request

Downgrade from PHP 7.3 to 7.2.

From [here](https://www.php.net/manual/en/migration73.other-changes.php#migration73.other-changes.core.setcookie):

<a href="https://www.php.net/manual/en/function.setcookie.php" class="function">setcookie()</a> and <a href="https://www.php.net/manual/en/function.setrawcookie.php" class="function">setrawcookie()</a> now also support the following signature:

```php
 setcookie ( string $name [, string $value = "" [, array $options = [] ]] ) : bool
```

where `$options` is an associative array which may have any of the keys <code class="literal">"expires"</code>, <code class="literal">"path"</code>, <code class="literal">"domain"</code>, <code class="literal">"secure"</code>, <code class="literal">"httponly"</code> and <code class="literal">"samesite"</code>.

simivar:
Closes #4328

There is one question that I didn't handle here: what to do when `samesite`  key is specified in `options` array? Such parameter is not available for PHP <= 7.2. Should we throw some exception? 

One of the solutions could be as well translating `setcookie` to `header` call with defined `SameSite` option, e.g.:
```diff
- setcookie('name', 'value', ['samesite' => 'Lax', 'secure' => true]);
+ header('Set-Cookie: name="value"; SameSite=Lax; Secure'); 
```
but I feel like with this approach we could have some security & escaping issues.


rules/downgrade-php73/tests/Rector/FuncCall/SetCookieOptionsArrayToArgumentsRector/Fixture/options_scalars.php.inc
```diff
@@ -0,0 +1,23 @@
+<?php
+if (false) {
+    setcookie('name', 'value', ['expires' => 3600]);
```
TomasVotruba, line 3: These snippets of code can be executed.

Better wrap it in `class SomeClass {}` and `run()` method as other fixture have.
That also prevent missing classes and keeps test fixture format united.

config/set/downgrade-php73.php
```diff
@@ -3,7 +3,7 @@
 declare(strict_types=1);
 
 use Rector\Downgrade\Rector\LNumber\ChangePhpVersionInPlatformCheckRector;
-use Rector\DowngradePhp73\Rector\FuncCall\SetCookieOptionsArrayToArguments;
+use Rector\DowngradePhp73\Rector\FuncCall\SetCookieOptionsArrayToArgumentsRector;
```
TomasVotruba, line 6: I wonder... this name typo should not happen with [`generate` command](https://github.com/rectorphp/rector/blob/master/docs/rector_recipe.md). I recall there was a bug in naming, but I think we fixed it.

How did you create a rule and tests?

TomasVotruba (commented):
I've added 1 tip for test fixture, the rest is good

config/set/downgrade-php73.php
simivar: I created the files by hand.

TomasVotruba: Oh. I use only the generate command for last 2 years. It handles the boring parts and gives more time for coding the `refactor()` logic :)

TomasVotruba:
Thank you :+1:

PR #4808, opened 2020-12-06
dont run weekly jobs on forks, because gh-tokens are not available

staabm:
Try to fix a error which I get in GitHub Action every week

7F430463-989D-4DC9-B5FC-AD250530B8F8

### DIFF
--- a/.github/workflows/weekly_pull_requests.yaml
+++ b/.github/workflows/weekly_pull_requests.yaml
@@ -7,6 +7,10 @@ on:
 
 jobs:
     weekly_pull_requests:
+
+        # dont run jobs on forks, because gh-tokens are not available 
+        if: github.event.pull_request.head.repo.full_name == github.repository
+
         strategy:
             fail-fast: false
             matrix:


.github/workflows/weekly_pull_requests.yaml
```diff
@@ -7,6 +7,10 @@ on:
 
 jobs:
     weekly_pull_requests:
+
+        # dont run jobs on forks, because gh-tokens are not available 
+        if: github.event.pull_request.head.repo.full_name == github.repository
+
```
TomasVotruba, line 13: This whole workflows should only run on main repository.

Can you move the `if` to first level to make sure no other jobs will be run? As `on` is?

staabm: not sure what you mean. I am not aware of the place where a `if:` can be placed above this per-job level.

TomasVotruba: This one:
https://github.community/t/conditional-jobs-and-workflows/16296/10

It seems it's not possible :/

TomasVotruba:
Thank you

TomasVotruba:
I think the daily cron should be covered too:
https://github.com/rectorphp/rector/blob/master/.github/workflows/daily_pull_requests.yaml

staabm:
> I think the daily cron should be covered too:

will create a followup

staabm:
here we go: https://github.com/rectorphp/rector/pull/4813

TomasVotruba:
Thank you

PR #4813, opened 2020-12-06
dont run jobs on forks, because gh-tokens are not available

staabm:
similar to https://github.com/rectorphp/rector/pull/4808


samsonasik (approved):

samsonasik:
@staabm thank you.

PR #4815, opened 2020-12-07
Added test cases for RenameParamToMatchTypeRector

leoloso:
Test case to reproduce bug #4394 

The test case works, it evidences no bug. However, in getrector.org the problem still happens! Or is it that the printed results are not complete?

This one is using the current version from master:

https://getrector.org/demo/13a8099f-4b05-486f-889b-6501107d48e0#result

Screenshot_2020-12-07 Try Rector Online Rector - Automated Way to Instantly Upgrade and Refactor any PHP code


TomasVotruba:
Thank you

Issue #4754, opened 2020-12-02
[Downgrade PHP 7.2] Parameter type widening


leoloso:
# Feature Request

Downgrade new feature from PHP 7.2 to its equivalent PHP 7.1 code. [Description](https://www.php.net/manual/en/migration72.new-features.php#migration72.new-features.param-type-widening).

Because the child class parameter type can be more generic than the parent class (check [argument contravariance](https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters)), we can't simply copy the same parameter type from parent to child, since it could generate an error from invoking the method in the child with a more generic type.

A possible solution is to remove the type on the parent:

```diff
interface A
{
-    public function test(array $input);
+    /**
+     * @param array $input
+     */
+    public function test($input);
}

class B implements A
{
    public function test($input){} // type omitted for $input
}
```

**Watch out:** then the type must also be removed everywhere, and maybe this is difficult to achieve. Because the change to be done does not happen on the child class, but on the parent class, the Rector rule should check if a method on the parent class has been overloaded removing the parameter type on any child class and, if so, remove the parameter in ALL methods from all extending classes!

For instance, this class must also be modified, even though it didn't introduce any change in the method signature:

```diff
class C implements A
{
-    public function test(array $input){}
+    public function test($input){}
}
```

Is it too messy? Is there a better approach?

leoloso:
The same strategy has been implemented for `ParamTypeDeclarationRector`. I'll use this rule as source

leoloso:
The strategy is: for every method from a class and interface, check if the same method in child/implementing classes has some parameter with a different typehint. If so, remove the parameter typehint must be removed everywhere: both in the parent class/interface, and in ALL the child/implementing classes (not just the one with a different signature).

leoloso:
There is a big issue: the downgrade script for Rector is currently downgrading package by package, limiting the input files to those of a certain package. Hence, we won't be able to modify signatures from other packages than the current one being downgraded.

2 solutions:

1. Do not downgrade package by package, but the whole project together
2. Directly remove all parameter types, ALWAYS, even if there's no need to modify the code (ugly, but it will work)

leoloso:
Argument types were introduced to PHP 7.0: https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.scalar-type-declarations

Hence, we could implement and execute that rule for the Rector downgrade script

Issue #4195, opened 2020-09-11
[Downgrade PHP 7.1] iterable pseudo-type


leoloso:
# Feature Request

Downgrade new feature from PHP 7.1 to its equivalent PHP 7.0 code. [Feature description](https://www.php.net/manual/en/language.types.iterable.php)

Docblock tags should be converted to `array|\Traversable`, not to `iterable`. From the description:

> It accepts any array or object implementing the Traversable interface
 
## Diff 

```diff
class SomeClass
{
-    public function run(iterable $iterator): iterable
+    /**
+     * @param array|\Traversable $iterator
+     * @return array|\Traversable
+     */
+    public function run($iterator)
     {
     }
}
```

norberttech:
Hey, just want to confirm one thing. 
Currently the behavior of translating param/return type `iterable` into the docblock looks like this:

```
* @param mixed[] $iterator
```

and it should be changed into:

```
@param array|\Traversable $iterator
```

?

leoloso:
In [the description](https://www.php.net/manual/en/language.types.iterable.php#language.types.iterable), it says that an iterable is either:

- an array
- an object implementing the `Traversable` interface

Hence, when downgrading, we do not know which of these 2 options the downgraded code is, and to be safe we add both of them: `array|\Traversable`

norberttech:
ok so the current behavior should stay as it is and only when downgrading we should change `iterable` into `array|\Traversable` correct? 

btw. `array` or maybe `mixed[]`? Not sure which one is preferred

leoloso:
> ok so the current behavior should stay as it is and only when downgrading we should change `iterable` into `array|\Traversable` correct?

Just the signature of the method must change: remove any `iterable` return or param type declaration, and add the docblock with `array|\Traversable` in its place.

Btw, there are plenty several rules doing some version of this, such as [`DowngradeParamObjectTypeDeclarationRector`](https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#downgradeparamobjecttypedeclarationrector) and [`DowngradeReturnObjectTypeDeclarationRector`](https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#downgradereturnobjecttypedeclarationrector).

> btw. array or maybe mixed[]? Not sure which one is preferred

Maybe `mixed[]` is better, so that PHPStan doesn't give a warning on it (even though, even if it does, we shouldn't care so much for downgraded code, which we already knows works well)

[...]

PR #4819, opened 2020-12-07
[Downgrade PHP 7.0] Remove param and return types

leoloso:
Remove [param type](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.scalar-type-declarations) and [return type](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.return-type-declarations) declarations, added to PHP 7.0.

This downgrade will be needed for the downgrade script (#4447) if we want to downgrade package by package (the alternative is to downgrade the whole project in a single run).

Explanation in https://github.com/rectorphp/rector/issues/4754#issuecomment-738031609 and https://github.com/rectorphp/rector/issues/4754#issuecomment-738043907:

> the downgrade script for Rector is currently downgrading package by package, limiting the input files to those of a certain package. Hence, we won't be able to modify signatures from other packages than the current one being downgraded.
>
> 2 solutions:
>
>    - Do not downgrade package by package, but the whole project together
>    - Directly remove all parameter types, ALWAYS, even if there's no need to modify the code (ugly, but it will work)

> Argument types were introduced to PHP 7.0: https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.scalar-type-declarations
>
> Hence, we could implement and execute that rule for the Rector downgrade script


rules/downgrade-php70/tests/Rector/FunctionLike/DowngradeTypeReturnDeclarationRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * @return string
+     */
+    public function run()
+    {
+        return 'Hello world';
+    }
+}
+
+?>
```
TomasVotruba, line 30: Could you add test case for array type?

leoloso: Ok, added

rules/downgrade-php70/tests/Rector/FunctionLike/DowngradeTypeReturnDeclarationRector/Fixture/array.php.inc
```diff
@@ -0,0 +1,30 @@
...
+<?php
+
+namespace Rector\DowngradePhp70\Tests\Rector\FunctionLike\DowngradeTypeReturnDeclarationRector\Fixture;
+
+class ReturnArrayTypeOnClassMethod
+{
+    /**
+     * @return mixed[]
+     */
+    public function run()
```
TomasVotruba, line 24: :+1:

rules/downgrade-php70/tests/Rector/FunctionLike/DowngradeTypeParamDeclarationRector/Fixture/array.php.inc
```diff
@@ -0,0 +1,28 @@
...
+<?php
+
+namespace Rector\DowngradePhp70\Tests\Rector\FunctionLike\DowngradeTypeParamDeclarationRector\Fixture;
+
+class ArrayFunctionParam
+{
+    /**
+     * @param mixed[] $value
+     */
+    public function run($value)
```
TomasVotruba, line 23: The `array` type should remain here:

It works on `@param` since PHP 5.3:
https://3v4l.org/haRck

image

leoloso: In some other issue, we decided to use `mixed[]` instead of `array` to avoid complaints by PHPStan: https://github.com/rectorphp/rector/issues/4195#issuecomment-730049526. It seems this logic is being applied everywhere. Do you want to change that?

TomasVotruba: I don't talk about docblock. I mean the exact line:

```diff
-public function run($value)
+public function run(array $value)
```

leoloso: ahhhh ok got it. I'll fix it. I can see can receive `array` and also `callable`

leoloso: Fixed

TomasVotruba:
Looks good, thank you :+1:

PR #4828, opened 2020-12-09
[Downgrade PHP 8.0]  Add DowngradeTrailingCommasInParamUseRector

### LINKED ISSUE #4160: [Downgrade PHP 8.0] Trailing comma in parameter lists
# Feature Request

Downgrade new feature from PHP 8.0 to its equivalent PHP 7.4 code
 
## Diff 

```diff
public function(
    string $parameterA,
    int $parameterB,
-    Foo $objectfoo,
+    Foo $objectfoo
) {
    // …
}
```

samsonasik:
Fixes #4160

### DIFF
--- a/rules/downgrade-php80/src/Rector/ClassMethod/DowngradeTrailingCommasInParamUseRector.php
+++ b/rules/downgrade-php80/src/Rector/ClassMethod/DowngradeTrailingCommasInParamUseRector.php
@@ -78,32 +78,32 @@ CODE_SAMPLE
         if ($node instanceof Closure) {
             $node = $this->processUses($node);
         }
-        
+
         return $this->processParams($node);
     }
 
     private function processParams(Node $node): ?Node
     {
-        if ($node->params !== []) {
-            $node->setAttribute(AttributeKey::ORIGINAL_NODE, null);
-            $last = $node->params[array_key_last($node->params)];
-            $last->setAttribute(AttributeKey::FUNC_ARGS_TRAILING_COMMA, false);
-
-            return $node;
+        if ($node->params === []) {
+            return null;
         }
 
-        return null;
+        $node->setAttribute(AttributeKey::ORIGINAL_NODE, null);
+        $last = $node->params[array_key_last($node->params)];
+        $last->setAttribute(AttributeKey::FUNC_ARGS_TRAILING_COMMA, false);
+
+        return $node;
     }
 
     private function processUses(Closure $node): Closure
     {
-        if ($node->uses !== []) {
-            $node->setAttribute(AttributeKey::ORIGINAL_NODE, null);
-            $last = $node->uses[array_key_last($node->uses)];
-            $last->setAttribute(AttributeKey::FUNC_ARGS_TRAILING_COMMA, false);
-
+        if ($node->uses === []) {
             return $node;
         }
+
+        $node->setAttribute(AttributeKey::ORIGINAL_NODE, null);
+        $last = $node->uses[array_key_last($node->uses)];
+        $last->setAttribute(AttributeKey::FUNC_ARGS_TRAILING_COMMA, false);
 
         return $node;
     }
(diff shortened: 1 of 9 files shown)


samsonasik:

rules/downgrade-php80/src/Rector/ClassMethod/DowngradeTrailingCommasInParamUseRector.php
```diff
@@ -0,0 +1,111 @@
...
+    {
+        return [ClassMethod::class, Function_::class, Closure::class];
+    }
+
+    /**
+     * @param ClassMethod|Function_|Closure $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $node instanceof Closure) {
```
simivar, line 78: We could skip the negation here and run `$this->processUses` only when the instance is `Closure`. Thanks to that we will only have one `return` statement and a simpler condition (without negation).

rules/downgrade-php80/src/Rector/ClassMethod/DowngradeTrailingCommasInParamUseRector.php
```diff
@@ -0,0 +1,111 @@
...
+            return $this->processParams($node);
+        }
+
+        $node = $this->processUses($node);
+        return $this->processParams($node);
+    }
+
+    private function processParams(Node $node): ?Node
+    {
+        if ($node->params !== []) {
```
simivar, line 88: Here as well we could flip the condition and return immediately `null` when `$node->params === []` and do the rest outside of if statement.

rules/downgrade-php80/src/Rector/ClassMethod/DowngradeTrailingCommasInParamUseRector.php
TomasVotruba: I'm not sure what exactly you mean, the code in words is very hard to read.
Could you maybe diff a suggestion?

Like:

```suggestion
-        if (! $node instanceof Closure) {
+        if (! $node instanceof Closure....) {
```

rules/downgrade-php80/src/Rector/ClassMethod/DowngradeTrailingCommasInParamUseRector.php
```diff
@@ -0,0 +1,111 @@
...
+     * @param ClassMethod|Function_|Closure $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $node instanceof Closure) {
+            return $this->processParams($node);
+        }
+
+        $node = $this->processUses($node);
+        return $this->processParams($node);
```
simivar, line 83:
```suggestion
        if ($node instanceof Closure) {
            $node = $this->processUses($node);
        }
        
        return $this->processParams($node);
```

rules/downgrade-php80/src/Rector/ClassMethod/DowngradeTrailingCommasInParamUseRector.php
simivar: Done in a separate discussion.

rules/downgrade-php80/src/Rector/ClassMethod/DowngradeTrailingCommasInParamUseRector.php
samsonasik: Closure can have params, that will make closure with params not updated

samsonasik: oh, ok, I understand

rules/downgrade-php80/src/Rector/ClassMethod/DowngradeTrailingCommasInParamUseRector.php
samsonasik: implemented.

samsonasik:
@simivar all incorporated 👍

samsonasik:
I moved cleaning trailing comma process to separate method.

samsonasik:

TomasVotruba:
Thank you :+1:

simivar:
Very nice work, thank you @samsonasik 👍

PR #4836, opened 2020-12-09
Add function call inside string test case

budziam:
This MR provides a test case which fails with the current implementation of `DowngradeTrailingCommasInFunctionCallsRector`

```
--- Original
+++ New
@@ -1,3 +1,3 @@
 <?php

-$value = "{$this->method("test")}";
+$value = "{{$this->method("test")}}";
```


TomasVotruba:
Test case looks good, why closing?

budziam:
@TomasVotruba I closed it because there was the following issue with tests
```
Rector\Core\Exception\Testing\SuperfluousAfterContentFixtureException: The part after "-----" can be removed in "rules/downgrade-php73/tests/Rector/FuncCall/DowngradeTrailingCommasInFunctionCallsRector/Fixture/function_call_inside_string.php.inc" file. It is the same as top half, so no change is required.
```
https://github.com/rectorphp/rector/runs/1525995354

TomasVotruba:
That's correct. The message is telling what to do:
> The part after "-----" can be removed

Issue #4824, opened 2020-12-08
[Symfony 5.2] Cover upgrade path and new features


TomasVotruba:
Cover changes and new features from https://github.com/symfony/symfony/blob/5.x/UPGRADE-5.2.md

simivar:
Should those be done in a new set, for example `symfony-52`?

TomasVotruba:
Yes! We have one config per Symfony version: https://github.com/rectorphp/rector/blob/c68fd43b65247ee0b79298038f1985dc55b7fa13/packages/set/src/ValueObject/SetList.php#L642

simivar:
List of things to do so it's easier to track:

## DependencyInjection
- [ ] Deprecated Definition::setPrivate() and Alias::setPrivate(), use setPublic() instead

## FrameworkBundle
- [ ] Deprecated the public form.factory, form.type.file, translator, security.csrf.token_manager, serializer, cache_clearer, filesystem and validator services to private.
- [ ] If you configured the framework.cache.prefix_seed option, you might want to add the %kernel.environment% to its value to keep cache namespaces separated by environment of the app
- [ ] Deprecated the lock.RESOURCE_NAME and lock.RESOURCE_NAME.store services and the lock, LockInterface, lock.store and PersistingStoreInterface aliases, use lock.RESOURCE_NAME.factory, lock.factory or LockFactory instead.

## Form
- [ ] Deprecated PropertyPathMapper in favor of DataMapper and PropertyPathAccessor.

## HttpFoundation
- [ ] Deprecated not passing a Closure together with FILTER_CALLBACK to ParameterBag::filter(); wrap your filter in a closure instead.
- [ ] Deprecated the Request::HEADER_X_FORWARDED_ALL constant, use either Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_HOST | Request::HEADER_X_FORWARDED_PORT | Request::HEADER_X_FORWARDED_PROTO or Request::HEADER_X_FORWARDED_AWS_ELB or Request::HEADER_X_FORWARDED_TRAEFIKconstants instead.
- [ ] Deprecated BinaryFileResponse::create(), use __construct() instead

## Lock
- [ ] MongoDbStore does not implement BlockingStoreInterface anymore, typehint against PersistingStoreInterface instead.
- [ ] deprecated NotSupportedException, it shouldn't be thrown anymore.
[...]

PR #4842, opened 2020-12-10
[Symfony 5.2][Security] Migrate PUBLIC_ACCESS to the new class

simivar:
Part of #4824

### DIFF
--- a/config/set/symfony52.php
+++ b/config/set/symfony52.php
@@ -2,8 +2,10 @@
 
 declare(strict_types=1);
 
+use Rector\Renaming\Rector\ClassConstFetch\RenameClassConstantRector;
 use Rector\Renaming\Rector\MethodCall\RenameMethodRector;
 use Rector\Renaming\ValueObject\MethodCallRename;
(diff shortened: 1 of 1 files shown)


config/set/symfony52.php
```diff
@@ -21,4 +23,17 @@
                 new MethodCallRename('Symfony\Component\Mime\Address', 'fromString', 'create'),
             ]),
         ]]);
+
+    # https://github.com/symfony/symfony/blob/5.x/UPGRADE-5.2.md#security
```
samsonasik, line 27:
```suggestion
    // @see https://github.com/symfony/symfony/blob/5.x/UPGRADE-5.2.md#security
```

simivar: We use `#` comment instead of `@see` in other Symfony-related sets like `symfony-43` so I went with it to keep the style the same.

samsonasik: ok

TomasVotruba (approved):

TomasVotruba:
Thank you

PR #4845, opened 2020-12-10
Do not enforce interface-typing for classes with wider public API than defined by interface

### LINKED ISSUE #4844: Class implementing interface but providing more methods than those implemented should not be swapped with interface
# Bug Report

| Rector version | 0.8.56 |
| Installed as   | composer dependency |

After upgrading Symfony from [5.1.8](https://github.com/symfony/security-core/blob/v5.1.8/Security.php) to [5.2.0](https://github.com/symfony/security-core/blob/v5.2.0/Security.php) Rector tries to enforce `AuthorizationCheckerInterface` of a used-on-purpose `Security` class. However, this class has more public methods than those implemented from interface, so such change would lead to other problems with IDE autocompletion and SA tools.

I have `SentryListener` based on @B-Galati's [work](https://github.com/B-Galati/monolog-sentry-handler/blob/HEAD/doc/guide-symfony.md) and it uses `Symfony\Component\Security\Core\Security` in order to get user (`Security::getUser()`). `Security` class implements `AuthorizationCheckerInterface` which defines only `isGranted()`, so `getUser()` and `getToken()` are methods defined on `Security` class' level, hence class and interface are not interchangeable.

```
1) src/Infrastructure/Sentry/SentryListener.php

--- Original
+++ New
@@ -2,6 +2,7 @@

 namespace App\Infrastructure\Sentry;

+use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 use Psr\Log\LoggerInterface;
 use ReflectionClass;
 use Sentry\State\HubInterface;
@@ -26,7 +27,7 @@
     private Security $security;
     private LoggerInterface $logger;

-    public function __construct(HubInterface $hub, Security $security, LoggerInterface $logger)
+    public function __construct(HubInterface $hub, AuthorizationCheckerInterface $security, LoggerInterface $logger)
     {
         $this->hub = $hub;
         $this->security = $security;


Applied rules:

 * Rector\SOLID\Rector\ClassMethod\UseInterfaceOverImplementationInConstructorRector
 * Rector\TypeDeclaration\Rector\ClassMethod\AddParamTypeDeclarationRector
```

and `$this->security->getUser()` is used inside `SentryListener::onKernelRequest()`

## Expected Behaviour

Rector should not enforce interface if code uses object's methods defined outside of that interface (on class level or from traits). I don't know if it's possible to detect usage of an argument which type is about to be changed from concrete class to interface, but honestly such enforcement in my opinion looks too optimistic without such check. Class can have 1 interface implemented (one of the `UseInterfaceOverImplementationInConstructorRector::refactor()`'s condition) but still can have wider public API than defined in the interface.

Wirone:
Closes #4844

EDIT: update closing keyword to "Closes" by @TomasVotruba

### DIFF
--- a/rules/solid/src/Rector/ClassMethod/UseInterfaceOverImplementationInConstructorRector.php
+++ b/rules/solid/src/Rector/ClassMethod/UseInterfaceOverImplementationInConstructorRector.php
@@ -9,6 +9,8 @@ use PhpParser\Node\Name\FullyQualified;
 use PhpParser\Node\Stmt\ClassMethod;
 use Rector\Core\Rector\AbstractRector;
 use Rector\Core\ValueObject\MethodName;
+use ReflectionClass;
+use ReflectionMethod;
 use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
 use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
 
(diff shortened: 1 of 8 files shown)


TomasVotruba:
Looks good, thank you :+1:

PR #4846, opened 2020-12-10
[Symfony 5.2] Add BinaryFileResponseCreateToNewInstanceRector

simivar:
Part of #4824


TomasVotruba:
Better move it to "Symfony" namespace

simivar:
Don't you think it will be a little crowded there as we want to provide Rectors for 5.0 -> 5.1, 5.1 -> 5.2, and, possibly in the future, more of such migrations? Thanks to a separate namespace it will be as well easy to say which Rector do the changes for which version and what is still lacking in the migration.

TomasVotruba:
It might. But we should not have 2 standards here, now it looks pretty unclear where to put next Symfony rule.

It might be worth to have major as a split here, rather then 20 directories per minor.

- Symfony2
- Symfony3
- Symfony4
- Symfony5

simivar:
Should I change the namespace to `Symfony5` now and in separate MR split the current `Symfony` to their respective ones?

TomasVotruba:
The current namespace is only `Symfony`. Please use it, then handle renaming/splitting in next PR

TomasVotruba:
Thank you :+1:

PR #4848, opened 2020-12-11
Renamed NoParticularNodeRule to ForbiddenNodeRule

leoloso:
The Rector CI was broken after Symplify's rule was renamed


leoloso:
I have a question regarding how this problem could be avoided: Could the Monorepo Builder handle 2 different projects, each with its own set of packages, within the same repo?

i.e. instead of having a single root, could the repo have many, as to handle different configurations?

For instance:

- projects/Rector/packages/...
- projects/Symplify/packages/...

Then, both projects/Rector and projects/Symplify would each be a root, providing the config for their own packages

leoloso:
CI is still failing for tests on PHP 7.3: it's still using `symplify/phpstan-rules` version `9.0.1`, not `9.0.3`. It works well everywhere else. Why could that be?

leoloso:
I guess version `9.0.2` or `9.0.3` introduced a dependency that requires PHP 7.4

leoloso:
> I guess version `9.0.2` or `9.0.3` introduced a dependency that requires PHP 7.4

No, can't be that, since now it's working OK after explicitly bumping to `9.0.3`

TomasVotruba:
Thank you :+1:

TomasVotruba:
This could be avoided by locking symplify version to minor. But this was rather last-call change of wrong class naming. Stable was released 2 days ago, so this is quite rare situation that does not need a special behavior.

TomasVotruba:
@leoloso Because it's using lowest deps: https://github.com/rectorphp/rector/blob/ff2ea8617693d3f39f4520df01329af356da0d02/.github/workflows/tests.yaml#L16

leoloso:
But so are the other tests. Why did PHP 7.4 not use `9.0.1`? I checked that

leoloso:
Oh sorry, now I see. I checked the prefer lowest option from `composer update`

leoloso:
@TomasVotruba Btw, please check my question from before https://github.com/rectorphp/rector/pull/4848#issuecomment-743015282: is this avoidable with Monorepo Builder? I'll start using it soon for my own project

TomasVotruba:
This is purely semver, not related to any tooling.

PR #4850, opened 2020-12-11
[Downgrade PHP 7.4] Fixed bug for CovariantReturnType downgrading

leoloso:
There was a bug: returning`?string` was transformed as `?\string` and then everything exploded


simivar (approved):

leoloso:
I discovered another bug: In the code below, return type `SelfB` in the 2nd function must be transformed to `self`:

```php
class SelfA
{
    public function covariantReturnTypes(): self
    { /* … */ }
}

class SelfB extends SelfA
{
    public function covariantReturnTypes(): SelfB
    { /* … */ }
}
```

But it is printing `\self` instead.

Will fix it in this PR.

TomasVotruba:
Thank you

PR #4851, opened 2020-12-11
[Downgrade PHP 7.4] Fix bug on ContravariantArgumentType: reflection on __construct may not return type

leoloso:
Inspecting the type from params in function `__construct` may return `null` instead of the expected `ReflectionNamedType` (may this be a bug on PHP?), and was throwing an error.

Fixed the bug: if `null` then consider we can't decide if to do the downgrade, then skip it.


leoloso:
CI failing because of #4848

TomasVotruba:
Thank you

PR #4853, opened 2020-12-11
Typo in the helper text, uses undefined constant

ordago:
Here is the constant:

https://github.com/rectorphp/rector/blob/668753b63826dce3f59951e3320482d50295234b/src/Configuration/Option.php#L104


simivar (approved):

TomasVotruba:
Thank you

PR #4861, opened 2020-12-12
[CodingStyle] Add PostIncDecToPreIncDecRector

### LINKED ISSUE #4859: [CodingStyle] Add post inc/dec to pre inc/dec rule
Mirror to PHPStan rule, just applying the change:
https://github.com/symplify/phpstan-rules/blob/master/docs/rules_overview.md#nopostincpostdecrule

Beware change in dim fetches:

```diff
-$values[$key++];
+$values[$key];
+++$key;
```

samsonasik:
Fixes #4859


samsonasik:
@TomasVotruba I think it can cause a bug when used inside `if` or inside an operation, eg: https://3v4l.org/G9fDO vs https://3v4l.org/u8t2C have different results.

It can only usefull if it not inside other operation, eg: https://3v4l.org/pY8FS vs https://3v4l.org/Ym1Gc have same result.

samsonasik:
I think the check inside other operation ( if, while, array) should only happen in `PHPStan`, not in rector rule.

samsonasik:

TomasVotruba:
> I think the check inside other operation or expr ( if, while, array) should only happen in PHPStan, not in rector rule.

You're right, that would be safer.

rules/coding-style/tests/Rector/PostInc/PostIncDecToPreIncDecRector/Fixture/skip_as_array_key.php.inc
```diff
@@ -0,0 +1,14 @@
...
+namespace Rector\CodingStyle\Tests\Rector\PostInc\PostIncDecToPreIncDecRector\Fixture;
+
+class SkipAsArrayKey
+{
+    public function run($value = 1)
+    {
+        $x[$value++];
+        $x[$value--];
+    }
+}
```
TomasVotruba, line 12: This can refactorable like described in the issue:

```diff
-$values[$key++];
+$values[$key];
+++$key;
```

samsonasik: implemented.

samsonasik:
I've added handling as for loop's expression

samsonasik:

TomasVotruba:
Looks good, thank you :+1:

PR #4877, opened 2020-12-14
[Downgrade] Fix parameter type widening issue, when method lives on the ancestor's interface

leoloso:
Fix issue: When running `DowngradeParameterTypeWideningRector` on this code:

```php
interface SomeInterface
{
    public function test(string $input);
}

abstract class AbstractSomeAncestorClass implements SomeInterface
{
}

class SomeChildClass extends AbstractSomeAncestorClass
{
    public function test($input) // type omitted for $input
    {
        /* ... */
    }
}
```

it is also attempting to refactor method `test` in `AbstractSomeAncestorClass`, because `RefactorMethod` indicates that this class contains this method. But it does not: this method lives only on `SomeInterface`.

Now, if the method is not found on the class, then skip it. This works because the interface is also refactored, so the method will be converted there.


TomasVotruba:
Thank you :+1:

PR #4878, opened 2020-12-14
[Downgrade] Apply trailing commas rule also on func/method/static call

leoloso:
`DowngradeTrailingCommasInParamUseRector` was downgrading code on the definition of the function only, but not when calling the function:

```php
var_dump('foo',);
```

This PR fixes it:

```diff
-var_dump('foo',);
+var_dump('foo');
```

### DIFF
--- a/rules/downgrade-php80/src/Rector/ClassMethod/DowngradeTrailingCommasInParamUseRector.php
+++ b/rules/downgrade-php80/src/Rector/ClassMethod/DowngradeTrailingCommasInParamUseRector.php
@@ -9,6 +9,7 @@ use PhpParser\Node\Expr\Closure;
 use PhpParser\Node\Expr\ClosureUse;
 use PhpParser\Node\Expr\FuncCall;
 use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\New_;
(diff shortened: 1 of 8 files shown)


leoloso:
Must be fixed also for constructor calls:

```diff
$object = new ConfiguredCodeSample(
    $codeSampleFrom,
-    $codeSampleTo,
+    $codeSampleTo
);
```

Working on it now

simivar:
Are your PRs based on each other? I see the same changes here as in #4877 and #4879

leoloso:
Nops, they are all based on master. All those same changes are being added by the Rector Rectify bot

TomasVotruba:
Thank you :+1: 

Could you make a failing test fixture for the class const fetch rename?

leoloso:
> Could you make a failing test fixture for the class const fetch rename?

Sorry for asking, but I'm a bit confused. You mean to test downgrading `new RenameClassConstant('SomeClass', 'OLD_CONSTANT', 'NEW_CONSTANT', )`? (it won't fail). Or to test the `ClassConstFetch` node? (It doesn't receive params). Or if not, what?

TomasVotruba:
For these https://github.com/rectorphp/rector/pull/4878#issuecomment-744307228

PR #4879, opened 2020-12-14
[Downgrade] Composer's platform check must use the PHP version below

leoloso:
When using set `downgrading-php80`, we're downgrading from PHP 8.0 to its previous version, i.e. 7.4. Hence, Composer's platform check (on `vendor/composer/platform_check.php`) must do `if (!(PHP_VERSION_ID >= 70400)) {` (not `80000`).

Fixed for all downgrade sets.

### DIFF
--- a/rules/symfony5/src/Rector/New_/PropertyAccessorCreationBooleanToFlagsRector.php
+++ b/rules/symfony5/src/Rector/New_/PropertyAccessorCreationBooleanToFlagsRector.php
@@ -7,6 +7,7 @@ namespace Rector\Symfony5\Rector\New_;
 use PhpParser\Node;
 use PhpParser\Node\Expr\BinaryOp\BitwiseOr;
 use PhpParser\Node\Expr\New_;
+use PhpParser\Node\Name;
 use Rector\Core\Rector\AbstractRector;
 use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
 use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
@@ -21,7 +22,7 @@ final class PropertyAccessorCreationBooleanToFlagsRector extends AbstractRector
     {
         return new RuleDefinition('Changes first argument of PropertyAccessor::__construct() to flags from boolean', [
             new CodeSample(
-                <<<'PHP'
+                <<<'CODE_SAMPLE'
 class SomeClass
 {
     public function run()
@@ -29,9 +30,9 @@ class SomeClass
         $propertyAccessor = new PropertyAccessor(true);
     }
 }
-PHP
+CODE_SAMPLE
                 ,
-                <<<'PHP'
+                <<<'CODE_SAMPLE'
 class SomeClass
 {
     public function run()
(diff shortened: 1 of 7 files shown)


rules/symfony5/src/Rector/New_/PropertyAccessorCreationBooleanToFlagsRector.php
```diff
@@ -62,43 +63,41 @@ public function refactor(Node $node): ?Node
...
         $isTrue = $this->isTrue($node->args[0]->value);
-        $flags = $this->prepareFlags($isTrue);
-        $node->args[0] = $this->createArg($flags);
+        $bitwiseOr = $this->prepareFlags($isTrue);
+        $node->args[0] = $this->createArg($bitwiseOr);
 
         return $node;
     }
 
+    private function shouldSkip(New_ $new): bool
```
simivar, line 72: Changes for this file seem out of the scope of this PR.

simivar (changes requested):

rules/symfony5/src/Rector/New_/PropertyAccessorCreationBooleanToFlagsRector.php
```diff
@@ -62,43 +63,41 @@ public function refactor(Node $node): ?Node
...
+        if (! $this->isName($new->class, 'Symfony\Component\PropertyAccess\PropertyAccessor')) {
+            return true;
+        }
+        return ! $this->isBool($new->args[0]->value);
+    }
+
     private function prepareFlags(bool $currentValue): BitwiseOr
     {
-        $magicGet = $this->createClassConstFetch('Symfony\Component\PropertyAccess\PropertyAccessor', 'MAGIC_GET');
+        $classConstFetch = $this->createClassConstFetch(
```
simivar, line 86: Name `classConstFetch` can be misleading as we have multiple of them created here (one line below we have `magicSet` for example).

simivar (changes requested):

leoloso:
@simivar all these unsuitable changes were applied automatically by Rector Rectify: https://github.com/rectorphp/rector/pull/4879/commits/182aef1bbc73966ce3ee74ec951988c52d42fdde

That's why this is happening in all my PRs!

If this is bug, I can create a new issue about it...

TomasVotruba:
> If this is bug, I can create a new issue about it...


It is indeed unrleated to PR. Please create failing test case for the broken rule

TomasVotruba:
Thank you

leoloso:
> Please create failing test case for the broken rule

There's no broken rule. It's just Rector Rectify working properly, on 1 file (for rule `PropertyAccessorCreationBooleanToFlagsRector`) that, somehow, had not been previously rectified...

I guess when this happens again, we can switch to master, and do some simple change just to trigger Rector Rectify, and commit the rectified code so it doesn't spill over the other PRs

TomasVotruba:
The rule can do better. It should be fixed now, because it changes our code for worse

leoloso:
Ok, hopefully I get it now. The problem is that if there are 2 or more variables (`$magicGet` and `$magicSet`) that can be renamed to the same name (as `$classConstFetch`), then the rule should be skipped...

Then, `$flags` renamed as `$bitwiseOr`, that's OK.

Is this right?

TomasVotruba:
Yes, exactly.

leoloso:
Added now in #4898. But I wonder if I got it right? The tests are failing in a different way to Rector Rectify's behavior...

PR #4888, opened 2020-12-15
[Symfony 5.2] Rename setProviderKey()/getProviderKey() to setFirewallName()/getFirewallName()

simivar:
Part of #4824

### DIFF
--- a/config/set/symfony52.php
+++ b/config/set/symfony52.php
@@ -51,4 +51,18 @@ return static function (ContainerConfigurator $containerConfigurator): void {
                 ),
             ]),
         ]]);
+
+    $services->set(RenameMethodRector::class)
+        ->call('configure', [[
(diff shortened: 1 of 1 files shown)


TomasVotruba (approved):

TomasVotruba:
Thank you

Feel free to send multiple PRs. It's almost always ready to merge so it would be easier to merge them at once :)

simivar:
I mostly add them as soon as I have them ready to not get any conflicts. I will try to group them a little to open a few at once :)

TomasVotruba:
Adding new rule is usually 100 % isolated from other each other.

In case of buch of rules for single set, I create 5 PRs with each one rule. Then 1 PR that edits the sets. 

We merge very quickly, so conflicts usually apear only on PRs that are being worked on for weeks.

PR #4889, opened 2020-12-15
[Downgrade] Add ".0" after downgrading the numeric literal separator on floats

### LINKED ISSUE #4881: [Downgrade Script] The numeric literal separator downgrade replaces `0.0` with `0`
# Feature Request

Rector downgraded to PHP 7.1 (#4447) works well, except on 1 instance: The tests for `AssertEqualsParameterToSpecificMethodsTypeRectorTest` [are failing](https://github.com/rectorphp/rector/pull/4447/checks?check_run_id=1549449844#step:11:96), because the [Downgrade PHP 7.4] Numeric Literal Separator Rector (#4361) is applied always on the number (even if it doesn't have a `_`), and it replaces `0.0` with `0`.

Then, [this comparison](https://github.com/rectorphp/rector/blob/a7705ed0a59c7c0ff8619c07ec48f489fb90787f/rules%2Fphpunit%2Fsrc%2FRector%2FMethodCall%2FAssertEqualsParameterToSpecificMethodsTypeRector.php#L145) to `0.0` is replaced to `0` on `AssertEqualsParameterToSpecificMethodsTypeRector`, and it then becomes always true:

```php
private function processAssertEqualsWithDelta(Node $node): void
{
    if (isset($node->args[3])) {
        // add new node only in case of non-default value
        if ($this->getValue($node->args[3]->value) !== 0.0) {
            //...
        }

        unset($node->args[3]);
    }
}
```

Unfortunately `DowngradeNumericLiteralSeparatorRector` is [applied always](https://github.com/rectorphp/rector/blob/a7705ed0a59c7c0ff8619c07ec48f489fb90787f/rules%2Fdowngrade-php74%2Fsrc%2FRector%2FLNumber%2FDowngradeNumericLiteralSeparatorRector.php#L70), even if the number doesn't contain `_`, with this code:

```php
$node->value = (string) $node->value;
```

That's because the number available in the node is the final number. The `_` is lost along the way.

Proposal: When parsing the node, if it's an `LNumber` or `DNumber`, then keep the data if it contains `_` or not. Then the rule won't be applied on`0.0`:

```php
$hasLiteralSeparator = $node->getAttribute(AttributeKey::HAS_LITERAL_SEPARATOR);
if ($hasLiteralSeparator) {
  $node->value = (string) $node->value;
}
```

@TomasVotruba I need your feedback on this one

leoloso:
Fixes #4881.

As explained in https://github.com/rectorphp/rector/issues/4881#issuecomment-745127760:

> if the downgraded float doesn't have `.`, then quite likely it was forced to be a float by doing `0.0` instead of `0` (for instance, Symfony does this approach). Then add `.0` again on the downgraded number (which is a string).

### DIFF
--- a/rules/downgrade-php74/src/Rector/LNumber/DowngradeNumericLiteralSeparatorRector.php
+++ b/rules/downgrade-php74/src/Rector/LNumber/DowngradeNumericLiteralSeparatorRector.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Rector\DowngradePhp74\Rector\LNumber;
 
+use Nette\Utils\Strings;
 use PhpParser\Node;
 use PhpParser\Node\Scalar\DNumber;
 use PhpParser\Node\Scalar\LNumber;
@@ -68,6 +69,17 @@ CODE_SAMPLE
         }
 
         $node->value = (string) $node->value;
+
+        /**
+         * This code follows a guess, to avoid modifying floats needlessly.
+         * If the node is a float, but it doesn't contain ".",
+         * then it's likely that the number was forced to be a float
+         * by adding ".0" at the end (eg: 0.0).
+         * Then, add it again.
+         */
+        if ($node instanceof DNumber && ! Strings::contains($node->value, '.')) {
+            $node->value .= '.0';
+        }
         return $node;
     }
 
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thank you, nice fix

PR #4897, opened 2020-12-15
Access classes installed on DEV through their strings

leoloso:
Packages `"phpunit/phpunit"` and `"nette/application"` are installed for DEV, hence the Rector rules, which run on PROD, must not access their classes directly, as these may not exist.

This was making the downgrade script fail when running PHPStan on the downgraded source, which includes only code for PROD, not DEV.

So I replaced `Class::class` with the string `"\Path\To\Class"`, so now it always works.

(The alternative solution would be to require `"nette/application"` for PROD. For `"phpunit/phpunit"` no need, since the only fix is on a `getRuleDefinition`, which outputs just an example of code)


rules/nette/src/Rector/ClassMethod/RemoveParentAndNameFromComponentConstructorRector.php
```diff
@@ -40,6 +39,14 @@ final class RemoveParentAndNameFromComponentConstructorRector extends AbstractRe
      */
     private const NAME = 'name';
 
+    /**
+     * Package "nette/application" is required for DEV, might not exist for PROD.
+     * So access the class throgh the string
+     *
+     * @var string
+     */
+    private const CONTROL_CLASS = '\Nette\Application\UI\Control';
```
TomasVotruba, line 48: The first slash is not needed, as it's not valid class name. Other cases too

leoloso: Ok, fixing now

TomasVotruba:
Thanks for clear description.


The class name reference should exists, even if class missing. What exactly failed where? 

I'm afraid about maintenance, how do we make sure this is covered in the future?

leoloso:
For instance, rule `OnPropertyMagicCallProvider` references `Nette\Application\UI\Control`. But package `"nette/application"` is under `require-dev` in `composer.json`, not under `require`, then that library won't be installed when doing `composer install --no-dev`. You may never run into this issue because everyone runs Rector in DEV, but the Rector release can't make that assumption.

In particular, the downgrade script (#4447) is only downgrading code from PROD. Then, it doesn't include `"nette/application"`, and running PHPStan on it to test the downgraded code was failing, with error:

```
 ------ -------------------------------------------------------------------- 
  Line   rules/nette-kdyby/src/DataProvider/OnPropertyMagicCallProvider.php  
 ------ -------------------------------------------------------------------- 
  85     Class Nette\Application\UI\Control not found.                       
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
 ------ -------------------------------------------------------------------- 
```

leoloso:
> I'm afraid about maintenance, how do we make sure this is covered in the future?

Maybe my proposal from above:

The alternative solution would be to require `"nette/application"` for PROD.

Otherwise, can do a combination of the 2:

```php
if (class_exists('Nette\Application\UI\Control')) {
  // Now access the class
  dump(Control::class);
}
```

and maybe an `assertEquals` of these in the tests.

TomasVotruba:
Typical solution for this use case is to create a class in stubs. That fakes the class event without dependency.

leoloso:
You decide. I haven't never used stubs, but if you want it that way, I'll check  it out and do it

TomasVotruba:
Go for it

leoloso:
@TomasVotruba So I did the replacement: instead of using strings for classnames, now the code uses stubs. But it's so terribly not worth it! The code has gotten worse (as I'll explain below), and it took me a few hours to pull off, compared to seconds coding the string.

First, there's an incredible unexplainable bug happening. Check the difference for the last commit (https://github.com/rectorphp/rector/pull/4897/commits/1c458025bb281f340b8463ed21761b4a8c8b330a):

Before:

```php
if (! interface_exists('Nette\ComponentModel\IComponent')) {
    return;
}

interface IComponent
{
    /**
     * Returns component specified by name or path.
     * @throws \Nette\InvalidArgumentException  if component doesn't exist
     */
    function getComponent(string $name): ?IComponent;
}
```

After:

```php
if (! interface_exists('Nette\ComponentModel\IComponent')) {
    interface IComponent
    {
        /**
         * Returns component specified by name or path.
         * @throws \Nette\InvalidArgumentException  if component doesn't exist
         */
        function getComponent(string $name): ?IComponent;
    }
}
```

They are the same, right? Well, with the first one it doesn't work, it keeps giving [this error message](https://github.com/rectorphp/rector/pull/4897/checks?check_run_id=1575958360):

```
PHP Fatal error:  Class Nette\Forms\Controls\TextInput contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Nette\ComponentModel\IComponent::getComponent) in /home/runner/work/rector/rector/vendor/nette/forms/src/Forms/Controls/TextInput.php on line 19
```

It's as if PHP is still loading this interface, even though it never reaches its declaration. I made a test adding a `die` before it, and it never reaches that line of code, yet on runtime it still loads the interface from the stub!

The other issue is that you want to gain maintainability. But it's getting less maintainable: the classname will seldom change (if ever), but the code within the classes or interfaces in the stub does change! That's the situation right now, even on `master`. Look at [interface `IComponent` from the stub](https://github.com/rectorphp/rector/blob/c50aedca5c2c8542573787c8ee5e2e5d27cdb4f4/stubs%2FNette%2FComponentModel%2FIComponent.php#L17-L18):

```php
interface IComponent
{
    /**
     * Returns component specified by name or path.
     * @throws \Nette\InvalidArgumentException  if component doesn't exist
     */
    function getComponent(string $name): ?IComponent;
}
```

And compare it against the [original in Nette](https://github.com/nette/component-model/blob/master/src/ComponentModel/IComponent.php):

```php
/**
 * Provides functionality required by all components.
 */
interface IComponent
{
	/** Separator for component names in path concatenation. */
	public const NAME_SEPARATOR = '-';

	function getName(): ?string;

	/**
	 * Returns the parent container if any.
	 */
	function getParent(): ?IContainer;

	/**
	 * Sets the parent of this component.
	 * @return static
	 */
	function setParent(?IContainer $parent, string $name = null);
}
```

They have changed completely! And this matters: the method `getComponent` is causing the errors from that unexplainable bug.

Another issue is that the stubs may need to load their own other stubs for parent classes interfaces. Hence needing 1 stub might require adding 10 stubs, as it happened in this case.

And finally, the stub files in `composer.json` must be loaded in order, manually (unless they can be loaded using `psr-4`?). It took me a few minutes to get it right, and that's just with less than 10 files.

Conclusion: I'd suggest using the strings for classnames. It's easy and, most likely, they will never change. To make it more maintainable, we could add a test with an `assertEquals`, then it's safe.

leoloso:
Made it draft to be sure you don't merge it 🤣

composer.json
```diff
@@ -63,7 +63,16 @@
...
+            "rules/restoration/tests/Rector/Use_/RestoreFullyQualifiedNameRector/Source/ShortClassOnly.php",
+            "stubs/Nette/ComponentModel/IComponent.php",
+            "stubs/Nette/ComponentModel/IContainer.php",
+            "stubs/Nette/ComponentModel/Container.php",
+            "stubs/Nette/Application/Routers/RouteList.php",
+            "stubs/Nette/Application/UI/IRenderable.php",
+            "stubs/Nette/Application/UI/ISignalReceiver.php",
+            "stubs/Nette/Application/UI/IStatePersistent.php",
+            "stubs/Nette/Application/UI/Component.php",
+            "stubs/Nette/Application/UI/Control.php"
```
TomasVotruba, line 75: This is not needed, look at other files in `/stubs` that are not here.

stubs/Nette/ComponentModel/Component.php
```diff
@@ -4,6 +4,10 @@
 
 namespace Nette\ComponentModel;
 
+if (class_exists('Nette\ComponentModel\Component')) {
+    return;
+}
```
TomasVotruba, line 9: :+1:

TomasVotruba:
> They are the same, right? Well, with the first one it doesn't work, it keeps giving this error message: PHP Fatal error:  Class Nette\Forms\Controls\TextInput contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Nette\ComponentModel\IComponent::getComponent) in /home/runner/work/rector/rector/vendor/nette/forms/src/Forms/Controls/TextInput.php on line 19

The error is clear. `TextInput` class must implement `getComponent()` method.


Not sure where you got lost, but this usually takes 10-15 minutes to setup.

leoloso:
No. `TextInput` must not implement `getComponent()`. It doesn't on Nette's original package (I already compared their code side by side on my previous comment), then why should you replicate something that has already been removed? You shouldn't recreate the component, if possible, but, if you have to, certainly not with code that doesn't belong there anymore.

> Not sure where you got lost, but this usually takes 10-15 minutes to set up.

It took 10 min to set up. Check commit "Added stubs to autoload for PROD", everything should've worked already then. I was surprised to see it did not. From then on, it took almost 10 more commits to fix the errors.

> This is not needed, look at other files in `/stubs` that are not here.

All those stubs? Are needed because a class either extending another class or implementing an interface, which also need to have stubs or otherwise the test run fails.

Once again, check all the commits after "Added stubs to autoload for PROD", the tests were failing without the stubs. 

So it took 2 hs because the tests were always failing, because the test expected code, provided by some stub, that was always incomplete. And fixing them was a lengthy manual process of recreating the stub hierarchy. And then the `getComponent` error happened which shows the stub contains outdated code.

In my honest opinion, this is not worth it.

stubs/Nette/ComponentModel/Component.php
leoloso: It should be 👍 but it doesn't work! So it's 👎 then 😛

leoloso:
But also, just implementing `getComponent` would've failed anyway, because that error demonstrates a bug in PHP that its loading the interface from the stub, even though I had the Nette package so the stub must not be loaded. Then you're in a rabbit hole in which the real package is replaced with the stub and errors happen somewhere else.

leoloso:
To be clear: it didn't take me 2 hs. It took 1 hour more, used to try to understand why it was not working and fix it. It was 1 hour that was a waste of time and effort, and there was nothing wrong in the code (no, I didn't get lost)

leoloso:
I removed the stubs from PROD, using strings for classnames again

TomasVotruba:
Looks good :+1:

TomasVotruba (approved):

PR #4898, opened 2020-12-15
Added tests for RenameVariableToMatchNewTypeRector 

leoloso:
As requested in https://github.com/rectorphp/rector/pull/4879#issuecomment-745132901

However, the tests are not failing in the expected manner:

- Test `with_method_call.php.inc` should not fail, but it does
- Test `with_double_method_call.php.inc` replicates the issue from Rector Rectify. It fails, but it a different way than expected: `$flag` was not renamed as `$bitwiseOr`, and `$magicGet` was left intact

Am I missing something?


leoloso:
Changes to fix, applied by Rector Rectify, can be seen here: https://github.com/rectorphp/rector/commit/182aef1bbc73966ce3ee74ec951988c52d42fdde

rules/naming/tests/Rector/ClassMethod/RenameVariableToMatchNewTypeRector/Fixture/with_double_method_call.php.inc
```diff
@@ -0,0 +1,77 @@
...
+                $magicGet,
+            ),
+            $magicSet,
+        );
+    }
+
+    protected function createClassConstFetch(string $class, string $constant): ClassConstFetch
+    {
+        return new ClassConstFetch($class, $constant);
+    }
```
TomasVotruba, line 36: There is too much going on. Please spilt it to standlone files per case

TomasVotruba:
The test has probably different scope then real file.

Is `RenameVariableToMatchNewTypeRector` the source of origin?

leoloso:
You were right, I got the wrong rule. The actual one is `RenameVariableToMatchMethodCallReturnTypeRectorTest`.

> There is too much going on. Please spilt it to standlone files per case

I had kept the added test so long, to find out what was different. Now that I found out the problem, I made the test just minimal.

TomasVotruba:
@leoloso Thank you :+1:

TomasVotruba:
@samsonasik Could you check this one?

It should be either skipped or better renamed with type suffix, like this:

```diff
-$magicGet = $this->createClassConstFetch('SomeClass', 'MAGIC_GET');
+$magicGetClassConstFetch = $this->createClassConstFetch('SomeClass', 'MAGIC_GET');

-$magicSet = $this->createClassConstFetch('SomeClass', 'MAGIC_SET');
+$magicSetClassConstFetch = $this->createClassConstFetch('SomeClass', 'MAGIC_SET');
```

samsonasik:
I will try.

simivar (approved):

samsonasik:
fixed.

samsonasik:

TomasVotruba (approved):

TomasVotruba:
Thank you

PR #4901, opened 2020-12-15
[Utils] Run validate-fixture-namespace and validate-fixture-classname  for /tests/ directory

samsonasik:


rules/privatization/src/Rector/Property/PrivatizeFinalClassPropertyRector.php
```diff
@@ -106,7 +106,7 @@ private function isPropertyVisibilityGuardedByParent(Property $property, Class_
     }
 
     /**
-     * @return class-string[]
+     * @return class-string[]|bool[]
      */
     private function getParentClasses(Class_ $class): array
     {
```
TomasVotruba, line 112: It would be better to modify code bellow to return only strings.

samsonasik: fixed.

samsonasik:

TomasVotruba:
Thank you :+1:

Issue #4905, opened 2020-12-16
Unable to install ECS 9.0 along with Rector 0.8.56


Wirone:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | e.g. v0.8.56 |
| Installed as   | composer dependency |

I want to upgrade ECS to `^9.0` but dependencies cannot be resolved:

```
$ composer update symplify/easy-coding-standard --with-dependencies
Loading composer repositories with package information
Restricting packages listed in "symfony/symfony" to "^5.1"
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - rector/rector is locked to version 0.8.56 and an update of this package was not requested.
    - rector/rector 0.8.56 requires symplify/autowire-array-parameter ^8.3.48 -> found symplify/autowire-array-parameter[8.3.48] but it conflicts with another require.

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.

$ composer why-not symplify/easy-coding-standard ^9.0        
symplify/easy-coding-standard  9.0.11      requires          symplify/coding-standard (^9.0.11)                           
__root__                       dev-master  does not require  symplify/coding-standard (but 8.3.48 is installed)           
symplify/easy-coding-standard  9.0.11      requires          symplify/package-builder (^9.0.11)                           
__root__                       dev-master  does not require  symplify/package-builder (but 8.3.48 is installed)           
symplify/easy-coding-standard  9.0.11      requires          symplify/set-config-resolver (^9.0.11)                       
__root__                       dev-master  does not require  symplify/set-config-resolver (but 8.3.48 is installed)       
symplify/easy-coding-standard  9.0.11      requires          symplify/symplify-kernel (^9.0.11)                           
__root__                       dev-master  does not require  symplify/symplify-kernel (but 8.3.48 is installed)           
symplify/easy-coding-standard  9.0.11      requires          symplify/smart-file-system (^9.0.11)                         
__root__                       dev-master  does not require  symplify/smart-file-system (but 8.3.48 is installed)         
symplify/easy-coding-standard  9.0.11      requires          symplify/autowire-array-parameter (^9.0.11)                  
__root__                       dev-master  does not require  symplify/autowire-array-parameter (but 8.3.48 is installed)  
symplify/easy-coding-standard  9.0.11      requires          symplify/console-color-diff (^9.0.11)                        
[...]

PR #4907, opened 2020-12-16
[UPGRADE] Add Rector 0.9 upgrade docs

TomasVotruba:
Ref https://github.com/rectorphp/rector/issues/4905#issuecomment-746399661

### DIFF
--- a/UPGRADE_09.md
+++ b/UPGRADE_09.md
@@ -1,0 +1,63 @@
+# How to Upgrade From Rector 0.8 to 0.9
+
+## In Symfony project, clear `config/bundles.php`
+
+- drop `ComposerJsonManipulatorBundle` class
+- drop `ConsoleColorDiffBundle` class
+
+## `rector.php`
+
+Since Rector 0.9 we switched from internal skipping to [`symplify/skipper` package](https://tomasvotruba.com/blog/2020/12/10/new-in-symplify-9-skipper-skipping-files-and-rules-made-simple/). Now there is only one `Option::SKIP` parameter to handle both paths and classes.
+
+Replace deprecated `Option::EXCLUDE_RECTORS` parameters with `Option::SKIP`:
+
+```diff
+ use Rector\Core\Configuration\Option;
+ use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+ return static function (ContainerConfigurator $containerConfigurator): void {
+     $parameters = $containerConfigurator->parameters();
+
+-    $parameters->set(Option::EXCLUDE_RECTORS, [
++    $parameters->set(Option::SKIP, [
+         SomeRector::class,
+         SomeRector::class,
+     ]);
+ };
+```
+
+Replace deprecated `Option::EXCLUDE_PATHS` parameters with `Option::SKIP`:
+
+```diff
+ use Rector\Core\Configuration\Option;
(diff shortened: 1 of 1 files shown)


TomasVotruba:
@Wirone Thanks for the input to make this +1

Could you check what else should be mentioned?

Wirone:
Is it correct to have one `SKIP` option for rectors and paths? Rectors should be passed as keys (where value can be `null` or array of paths) where paths rather as values 🤔 I did not look at code, would it automatically check if key is integer and resolve it correctly? Even so, this is less readable than previous options...

TomasVotruba:
> Is it correct to have one SKIP option for rectors and paths?

Yes, that's the goal.

It might be confusing at first, but in long term it saves 2 extra parameters.

```php
    // path or rule?
    // be explicit!
]);

$parameters->set(Option::SKIP_PATH, [
    // path
]);

$parameters->set(Option::SKIP_RULE, [
    RuleRector::class
]);

// ← or maybe just a path? →

$parameters->set(Option::SKIP_RULE_AND_PATH, [
    RuleRector::class => [
       //
	],
]);
```

I explain the reasoning in a post:
https://tomasvotruba.com/blog/2020/12/10/new-in-symplify-9-skipper-skipping-files-and-rules-made-simple/

---

> would it automatically check if key is integer and resolve it correctly?

Yes!

Wirone:
I've read blog post and must say that reasoning doesn't convince me. Tools have different parameters' naming and it's not a problem if there is proper documentation and/or IDE support. It can be schema for XML or options defined as constants like in Rector/ECS. You can't make a typo in such scenarios, just like in Symfony bundle with proper configuration definition - compiler will throw exception for unsupported option. Using the same parameter name for different purposes (excluding files from processing vs. limiting processing rules/scope) is IMHO violation of SRP in terms of configuration. As a user I prefer more small-scope options to tweak everything rather than multi-purpose, "magic" options. But this is only my subjective opinion, you can treat it as a friendly feedback :wink:

TomasVotruba:
Thank you for feedback. I imagine change from quite ok options is hard and I wish I figured this earlier, before we introduced those options.

> You can't make a typo in such scenarios, just like in Symfony bundle with proper configuration definition - compiler will throw exception for unsupported option.

We had dozens of reported bugs just because of typos. In YAML times and even in PHP times. We live in cognitive overload world. Users don't have time to read docs carefuly step by step. I don't read docs untill I have a problem. 

The 3 options above were really confusing. Even for us. The goal of this feature is to "skip", whatever it is. Not to read docs about how this or that or that and this should be skipped. That's not Rector's domain. If we're a e.g. symfony\validator, that would be different :)

It also opens for future format like ECS or PHPStan has (`"#this is a message \d"`), without forcing users to memorize new keyword.

<br>

Despite different opinions. This is highly experimental feature, as there is no such component in PHP. If it goes wrong, we're ready to split it.

nelson6e65:
Drop support for PHP 7.2 should be documented:

```
Problem 1
    - Root composer.json requires rector/rector ^0.9.15 -> satisfiable by rector/rector[0.9.15].
    - rector/rector 0.9.15 requires php ^7.3|^8.0 -> your php version (7.2.34) does not satisfy that requirement.
```

TomasVotruba:
Be sure follow up our blog :) 
https://getrector.org/blog/2020/12/28/rector-09-released

nelson6e65:
Hi. Sorry, I did not know about the blog. I use to check changes directly in repos 😅

Thanks for the link!

PR #4909, opened 2020-12-17
Bundle bin/rector.php in dist archive

Wirone:
See: https://github.com/rectorphp/rector/issues/4905#issuecomment-747320544


TomasVotruba:
Why not reuse the existing .gitignore config?

TomasVotruba:
Could you also refer the issue in comment there? This is weird edge cases, so future generations know why is that :)

TomasVotruba (approved):

TomasVotruba:
Thank you. Perfect line to put it on!

PR #4915, opened 2020-12-19
[EarlyReturn] Enable ReturnBinaryAndToEarlyReturnRector in early-return config set

samsonasik:


samsonasik:

TomasVotruba:
Thanks!

Oh, we don't have it enabled in `rector-ci.php`:
https://github.com/rectorphp/rector/blob/11d36a4bf99062cb5159eba29a013ef145b40a24/rector-ci.php#L59

Should be added there too. I forgot to handle that in the past.

There might be many false positives in this set, that will need to be fixed.

samsonasik:
I've enable EARLY_RETURN in rector-ci.php

samsonasik:
added handling to keep comment if exists.

rules/dead-code/src/Comparator/Parameter/ParameterDefaultsComparator.php
```diff
@@ -53,8 +53,10 @@ private function isMutuallyExclusiveNull(ParameterReflection $parameterReflectio
         if ($parameterReflection->getDefaultValue() === null && $param->default !== null) {
             return true;
         }
-
-        return $parameterReflection->getDefaultValue() !== null && $param->default === null;
+        if (! ($parameterReflection->getDefaultValue() !== null)) {
```
samsonasik, line 56: not identical need to be changed to identical and vice versa

samsonasik: handled.

samsonasik:

packages/better-php-doc-parser/src/ValueObject/PhpDocNode/AbstractTagValueNode.php
```diff
@@ -189,8 +189,10 @@ private function shouldPrintEmptyBrackets(): bool
         )) {
             return true;
         }
-
-        return $this->tagValueNodeConfiguration->hasOpeningBracket() && $this->tagValueNodeConfiguration->hasClosingBracket();
+        if (! $this->tagValueNodeConfiguration->hasOpeningBracket()) {
+            return false;
+        }
+        return (bool) $this->tagValueNodeConfiguration->hasClosingBracket();
```
TomasVotruba, line 195: `hasClosingBracket()` returns bool, so type-casting is not needed here

packages/node-type-resolver/src/NodeTypeCorrector/PregMatchTypeCorrector.php
```diff
@@ -108,7 +108,10 @@ private function getVariableUsages(Variable $variable): array
         }
 
         return $this->betterNodeFinder->find((array) $scope->stmts, function (Node $node) use ($variable): bool {
-            return $node instanceof Variable && $node->name === $variable->name;
+            if (! $node instanceof Variable) {
+                return false;
+            }
+            return (bool) ($node->name === $variable->name);
```
TomasVotruba, line 114: This also returns bool, so no type cast is needed

```suggestion
            return $node->name === $variable->name;
```

rules/coding-style/src/Rector/Class_/AddArrayDefaultToArrayPropertyRector.php
```diff
@@ -176,8 +176,6 @@ private function clearNotNullBeforeCount(Class_ $class, array $propertyNames): v
             if (! $node instanceof BooleanAnd) {
                 return null;
             }
-
-            // $this->value !== null
```
staabm, line 180: Were this comments deleted manually or should the rector preserve them?

samsonasik: it seems by rectify command that delete commented code

packages/better-php-doc-parser/src/ValueObject/PhpDocNode/AbstractTagValueNode.php
samsonasik: implemented.

packages/node-type-resolver/src/NodeTypeCorrector/PregMatchTypeCorrector.php
samsonasik: implemented.

samsonasik:

rules/coding-style/src/Rector/Class_/AddArrayDefaultToArrayPropertyRector.php
TomasVotruba: @samsonasik Could you return manually those comments? 

It's out of scope this pull-request, so just to preserve them here

samsonasik: reverted manually.

TomasVotruba:
Thank you :clap:

samsonasik:

Issue #3225, opened 2020-04-23
Couldn't find constant expose:true


ThomasTr:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | v0.7.17              |
| Installed as   | composer dependency                       |

Rector changes options={"expose"=true} in @ Route annotation, causes error:

[Semantical Error] Couldn't find constant expose:true, method App\Controller\API\AccountingController::contact() in .... 
Make sure annotations are installed and enabled.

## Minimal PHP Code Causing Issue

https://getrector.org/demo/e0fc28be-0156-41a9-bc98-5ccabf9b76d3

## Expected Behaviour

Rector should not touch options={"expose"=true} in @Route annotation.

Expose=true comes from FOSJsRoutingBundle.

ThomasTr:
Also in phpdoc but here in @Assert\Choice:
```
      * @Assert\Choice(
-     *         choices = {"Open", "Close", "Deleted"},
-     *         message = "Bitte gültigen Status wählen.",
-     *         strict = true
-     * )
+     *         {"Offen", "In Bearbeitung", "Geschlossen", "Rechnung", "Veröffentlichung", "Feedback"}, strict=true )
```
and here:
```
-     * @ORM\Column(name="customer_id", type="integer", nullable=false, options={"default" = 0})
+     * @ORM\Column(name="customer_id", type="integer", nullable=false, options={default:0})
```

TomasVotruba:
You need to remove spaces around ` = `, see https://github.com/rectorphp/rector/issues/3212#issuecomment-617792010

ThomasTr:
changed the spacing, but problem still exists here for me, here it suggests expose without quotes, which breaks:

```diff
     /**
-     * @Route("/{contactId}/edit/{page}", name="accounting_contact_edit", options={"expose"=true})
+     * @Route("/{contactId}/edit/{page}", name="accounting_contact_edit", options={expose:true})
      */
     public function edit(Request $request, Contact $contact, int $page = 0): Response

**Comment by ThomasTr** [comment]
Here it whipes the message out (event without spaces)
```diff
      * @Assert\Choice(
      *         choices={"Offen", "In Bearbeitung", "Geschlossen", "Rechnung", "Veröffentlichung", "Feedback"},
-     *         message="Bitte gültigen Status wählen.",
-     *         strict=true
-     * )
+     *         strict=true )
```

ThomasTr:
[...]

PR #4924, opened 2020-12-20
Keep colon when it's used in Doctrine tag's original content.

Charl13:
Rector changes use of colon array separator to equal by default.

Example tag `(options={"key":"value"})` is changed to `(options={"key"="value"})`.

This change enables Rector to first check if a colon is used before changing it to equal.

See related issue at https://github.com/rectorphp/rector/issues/3225


TomasVotruba:
Thank you

PR #4926, opened 2020-12-20
Upgrade to php-parser 4.10.4 and PHPStan 0.12.63

TomasVotruba:


samsonasik:
cleanupDuplicateContent() method should no longer required?

https://github.com/rectorphp/rector/blob/2c8a76ef12e58b2de2eba1f2a558b26b60bb4e96/src/PhpParser/Printer/BetterStandardPrinter.php#L133

TomasVotruba:
@samsonasik Yes, that's right.

Could you revert the patching PR? https://github.com/rectorphp/rector/pull/4553
Send it as a new pull-request to current `master`

samsonasik:
I will try.

PR #4934, opened 2020-12-21
[Doc] Using transparent image for space.png

samsonasik:
To support github dark mode.

Before:

Screen Shot 2020-12-21 at 6 00 04 PM

After:
Screen Shot 2020-12-21 at 6 03 01 PM


TomasVotruba:
Splendid :+1:

PR #4960, opened 2020-12-23
[CodingStyle] Skip CatchExceptionNameMatchingTypeRector on variable previously defined

### LINKED ISSUE #4946: CatchExceptionNameMatchingTypeRector overrides variable
# Bug Report

| Rector version | dev-master            |

CatchExceptionNameMatchingTypeRector renames exception variable event when it already exists. Which effectively changes code behavior and overrides variable.

This happens usually when there are nested try..catch blocks.

https://getrector.org/demo/410f8a1c-9113-4f4f-895a-55e4305f9dc5#demo_result

## Expected Behaviour

To NOT renames variables like this OR generate some unique name. I would prefer no do renaming only when name is short.

samsonasik:
Fixes #4946

### DIFF
--- a/src/Rector/AbstractRector/BetterStandardPrinterTrait.php
+++ b/src/Rector/AbstractRector/BetterStandardPrinterTrait.php
@@ -79,30 +79,4 @@ trait BetterStandardPrinterTrait
     {
         return $this->betterStandardPrinter->areNodesEqual($firstNode, $secondNode);
     }
-
-    /**
-     * Check if node found in previous node
-     */
-    protected function isNodeFoundInPrevious(Node $node, Node $find): bool
-    {
-        $previous = $node->getAttribute(AttributeKey::PREVIOUS_NODE);
-        while ($previous) {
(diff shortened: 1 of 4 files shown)


samsonasik:

src/Rector/AbstractRector/BetterStandardPrinterTrait.php
```diff
@@ -78,4 +79,30 @@ protected function areNodesEqual($firstNode, $secondNode): bool
...
+            $previous = $previous->getAttribute(AttributeKey::PREVIOUS_NODE);
+        }
+
+        $parent = $node->getAttribute(AttributeKey::PARENT_NODE);
+        if (! $parent instanceof Node) {
+            return false;
+        }
+
+        return $this->isNodeFoundInPrevious($parent, $find);
+    }
```
TomasVotruba, line 107: Thai should be part of not finde. Printer should only přitom.

samsonasik: implemented.

samsonasik:

TomasVotruba:
Thank you

PR #4975, opened 2020-12-24
[EarlyReturn] Register ChangeOrIfReturnToEarlyReturnRector to early-return set

samsonasik:


packages/node-name-resolver/src/NodeNameResolver.php
```diff
@@ -115,15 +115,10 @@ public function isName(Node $node, string $name): bool
...
     {
         if ($node instanceof MethodCall || $node instanceof StaticCall) {
-            if ($node->name instanceof MethodCall) {
-                return null;
-            }
-            if ($node->name instanceof StaticCall) {
-                return null;
-            }
-            if ($node->name instanceof Identifier) {
+            if ($this->isCallOrIdentifier($node->name)) {
```
TomasVotruba, line 118: :+1:

rules/code-quality/src/Rector/Return_/SimplifyUselessVariableRector.php
```diff
@@ -79,11 +79,8 @@ public function refactor(Node $node): ?Node
         /** @var AssignOp|Assign $previousNode */
         $previousNode = $previousNode->expr;
         $previousVariableNode = $previousNode->var;
-        // has some comment
-        if ($previousVariableNode->getComments() !== []) {
-            return null;
-        }
-        if ($previousVariableNode->getDocComment() !== null) {
+
+        if ($this->hasSomeComment($previousVariableNode)) {
```
TomasVotruba, line 83: :+1:

rules/defluent/src/NodeAnalyzer/FluentChainMethodCallNodeAnalyzer.php
```diff
@@ -88,6 +85,15 @@ public function isFluentClassMethodOfMethodCall(MethodCall $methodCall): bool
         return true;
     }
 
+    private function isCall(Expr $expr): bool
+    {
+        if ($expr instanceof FuncCall) {
```
TomasVotruba, line 90: There is actually `MethodCall` above, not `FuncCall`

rules/order/src/PropertyRanker.php
```diff
@@ -72,4 +66,30 @@ public function rank(Property $property): int
...
+            return true;
+        }
+
+        if ($type instanceof IntegerType) {
+            return true;
+        }
+
+        if ($type instanceof BooleanType) {
+            return true;
+        }
```
TomasVotruba, line 82: This should be refactored into array map constant, e.g.:

```php
private const TYPE_TO_RANK = [
    StringType::class => 5,
    // ...
    ArrayType::class => 10,
];
```

Then foreached.

rules/defluent/src/NodeAnalyzer/FluentChainMethodCallNodeAnalyzer.php
samsonasik: fixed.

TomasVotruba:
It seems lots of cases are `instanceof` on multiple types.

It might be word having helper static class, e.g.:

```php
StaticInstanceOf::isOneOf($node, [MethodCall::class, StaticCall::class]
```

What do you think?

rules/order/src/PropertyRanker.php
samsonasik: implemented.

samsonasik:
Yes, probably create `Rector\Core\Util\StaticInstanceOf` ?

samsonasik:

TomasVotruba:
Thank you :+1:

PR #4979, opened 2020-12-24
[Util] Add StaticInstanceOf

samsonasik:


samsonasik:

TomasVotruba:
Thanks!

Could you apply it on places with multiple instanceof in https://github.com/rectorphp/rector/pull/4975 ?

samsonasik:
I will try.

samsonasik:
implemented.

src/Util/StaticInstanceOf.php
```diff
@@ -0,0 +1,32 @@
...
+     * @param class-string[] $array
+     */
+    public static function isOneOf(?object $object, array $array): bool
+    {
+        if ($object === null) {
+            return false;
+        }
+
+        try {
+            Assert::isAnyOf($object, $array);
```
samsonasik, line 25: updated with Webmozart's Assert isAnyOf(), wrapped with try catch to return false.

composer.json
```diff
@@ -49,6 +49,7 @@
         "ocramius/package-versions": "^1.9",
         "php-parallel-lint/php-parallel-lint": "^1.2",
         "phpstan/phpstan-nette": "^0.12.12",
+        "phpstan/phpstan-webmozart-assert": "^0.12.8",
```
samsonasik, line 52: added `phpstan/phpstan-webmozart-assert` to require-dev to handle Assert::isA/AnyOf

src/Util/StaticInstanceOf.php
samsonasik: it seems the check is not works for object with property call:

```bash

  src/PhpParser/Node/Manipulator/IdentifierManipulator.php:93
  - '#Access to an undefined property PhpParser\\Node\:\:\$name#'
 ```
 
 while actually it already works. I will retry with `instanceof` check instead of previous `is_a`

samsonasik:

TomasVotruba:
Thank you :+1:

PR #4988, opened 2020-12-25
[DeadDocBlock] Skip RemoveNonExistingVarAnnotationRector on has other comment before var

### LINKED ISSUE #4955: RemoveNonExistingVarAnnotationRector: duplicates comments
# Bug Report

| Rector version | dev-master              |

Code example is self-explaining.

https://getrector.org/demo/19fe13bc-65df-4f78-be71-58170e358711#demo_result

samsonasik:
Fixes #4955


samsonasik:

rules/dead-doc-block/tests/Rector/Node/RemoveNonExistingVarAnnotationRector/Fixture/skip_other_comment_before_var.php.inc
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Rector\DeadDocBlock\Tests\Rector\Node\RemoveNonExistingVarAnnotationRector\Fixture;
+
+class SkipOtherCommentBeforeVar
+{
+    public function get()
+    {
+        /** @var \stdClass $nonExisting */
+        // Load data also with projekt...
+        $return[] = $this->getReturnData();
```
TomasVotruba, line 11: The `@var` annotation should be removed here, as it relates to non-existing variable

```suggestion
        // Load data also with projekt...
        $return[] = $this->getReturnData();
```

samsonasik: I tried that, it somehow remove all comments together:

```php
        $comments = $node->getComments();
        if (isset($comments[1]) && $comments[1] instanceof Comment) {
            $node->setAttribute(AttributeKey::PHP_DOC_INFO, $phpDocInfo);
            $node->setAttribute(AttributeKey::COMMENTS, [$comments[1]]);

            return $node;
        }
```

TomasVotruba: I see.

It might be related to adding docblocks back to nodes in BetterStandardPrinter:
https://github.com/rectorphp/rector/blob/a4f8a23ecf5019f9344159707890c90722a0ae52/src/PhpParser/Printer/BetterStandardPrinter.php#L585

samsonasik: I tried add: 

```php
$node->setAttribute(AttributeKey::COMMENTS, null);
```
in 

https://github.com/rectorphp/rector/blob/d4883d44d925b80ce4293f8f71dada290baf06ef/packages/node-type-resolver/src/PhpDoc/NodeAnalyzer/DocBlockManipulator.php#L140

That remove all comments, but I can't re-add the comment back in the rule

samsonasik: finally fixed.

TomasVotruba:
Thank you :+1:

rules/dead-doc-block/src/Rector/Node/RemoveNonExistingVarAnnotationRector.php
```diff
@@ -118,8 +120,16 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
-        $phpDocInfo->removeByType(VarTagValueNode::class);
+        $comments = $node->getComments();
+        if (isset($comments[1]) && $comments[1] instanceof Comment) {
+            $node->setAttribute(AttributeKey::COMMENTS, null);
+            $node->setDocComment(new Doc($comments[1]->getText()));
+            $node->setAttribute(AttributeKey::PHP_DOC_INFO, null);
```
TomasVotruba, line 127: This will be probably needed in more generic way in one place, where comments are removed.

samsonasik: need a Util class to rollback comment?

TomasVotruba: Something like that. We'll see what else is reported

PR #4990, opened 2020-12-25
failing test: ClassPropertyAssignToConstructorPromotionRector: should ignore early returns

jkuchar:
Currently it promotes variable which is assigned AFTER early return. [Which changes behaviour.](https://getrector.org/demo/2676fbab-e519-4f3a-883d-daf1c92619d6)

(Maybe another) question is what should it do with [excetions](https://getrector.org/demo/fad9ecfe-4307-4f7c-aa98-d39c349cef98). Probably nothing, as class is NOT created when there is an exception thrown in constructor?

### DIFF
--- a/rules/php80/tests/Rector/Class_/ClassPropertyAssignToConstructorPromotionRector/earlyReturn.php.inc
+++ b/rules/php80/tests/Rector/Class_/ClassPropertyAssignToConstructorPromotionRector/earlyReturn.php.inc
@@ -1,0 +1,17 @@
+<?php
+
+namespace Rector\Php80\Tests\Rector\Class_\ClassPropertyAssignToConstructorPromotionRector\Fixture;
+
+final class DemoFile
+{
+  public int $x = 60;
+    
+	public function __construct(int $x)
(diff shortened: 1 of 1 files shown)


TomasVotruba:
The red cross makes me confused if CI is passing or not. Could we keep emojis out of titles? :)

jkuchar:
emoji: sure. I wasn't able to make CI to execute my test.

[Here is the code](https://getrector.org/demo/2676fbab-e519-4f3a-883d-daf1c92619d6), which is broken after the rector processing. It looks to me that test is not executed at all.

TomasVotruba:
I see it now. The file must be located in `/Fixture` directory, not next to test class.

I'm trying to run demo with PHP 7.4, but nothing is reported:
https://getrector.org/demo/9b70b5c5-42bb-4f70-824a-480990e0bf76

TomasVotruba:
Ah, got it :) it's PHP 8.
https://getrector.org/demo/b998dd6f-4952-4f01-aab2-56b3eb1a3163

PR #4991, opened 2020-12-25
failing test: ClassPropertyAssignToConstructorPromotionRector: Create parameter is not passed as a reference

jkuchar:
# The original code

I have tried to migrated [this code](https://3v4l.org/JoUf8):
```php
<?php

class X {
    
    public string $x;
    
    function __construct(string $x) {
        $x = 'preprended-' . $x;
        $this->x = $x;
    }
    
}

$x = new X('input');
echo $x->x; // === "preprended-input"
```

# [Option 1](https://3v4l.org/lW0Hd): Used `$this->x` instead of local variable

As argument passed to constructor is NOT passed as reference, it is not possible to change value of property by assigning into local variable.

```php
<?php

class X {
    
    function __construct(public string $x) {
        $this->x = 'preprended-' . $this->x;
    }
    
}

$x = new X('input');
echo $x->x;
```

# [Option 2](https://3v4l.org/OQOpR): Overriding argument with reference:
```php
<?php

class X {
    
    function __construct(public string $x) {
        $x = & $this->x;
        $x = 'preprended-' . $x;
    }
    
}

$x = new X('input');
echo $x->x;
```

# [this PR] [Option 3](https://3v4l.org/eEeK1): Reassign changed property at the end

```php
<?php

class X {
    
    function __construct(public string $x) {
        $x = 'preprended-' . $x;
        $this->x = $x;
    }
    
}

$x = new X('input');
echo $x->x;
```

# Summary

<del>To me, option 1 looks as the most readable and simplest to implement.</del> I have updated PR to option 3.

### DIFF
--- a/rules/php80/tests/Rector/Class_/ClassPropertyAssignToConstructorPromotionRector/parameter-is-not-passed-as-reference.php.inc
+++ b/rules/php80/tests/Rector/Class_/ClassPropertyAssignToConstructorPromotionRector/parameter-is-not-passed-as-reference.php.inc
@@ -21,7 +21,8 @@ namespace Rector\Php80\Tests\Rector\Class_\ClassPropertyAssignToConstructorPromo
 class X {
         
     function __construct(public string $x) {
-        $this->x = 'preprended-' . $this->x;
+        $this->x = 'preprended-' . $x;
+        $this->x = $x;
     }
     
 }
--- a/rules/php80/tests/Rector/Class_/ClassPropertyAssignToConstructorPromotionRector/parameter-is-not-passed-as-reference.php.inc
+++ b/rules/php80/tests/Rector/Class_/ClassPropertyAssignToConstructorPromotionRector/parameter-is-not-passed-as-reference.php.inc
@@ -1,0 +1,28 @@
+<?php
+
+namespace Rector\Php80\Tests\Rector\Class_\ClassPropertyAssignToConstructorPromotionRector\Fixture;
+
+class X {
+    
+    public $x;
+    
+    function __construct(string $x) {
+        $x = 'preprended-' . $x;
+        $this->x = $x;
+    }
+    
+}
+?>
+-----
+<?php
+
+namespace Rector\Php80\Tests\Rector\Class_\ClassPropertyAssignToConstructorPromotionRector\Fixture;
+
+class X {
+        
+    function __construct(public string $x) {
+        $this->x = 'preprended-' . $this->x;
+    }
+    
+}
+?>


jkuchar:
I have added explanation directly into this PR comment above.

rules/php80/tests/Rector/Class_/ClassPropertyAssignToConstructorPromotionRector/parameter-is-not-passed-as-reference.php.inc
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Rector\Php80\Tests\Rector\Class_\ClassPropertyAssignToConstructorPromotionRector\Fixture;
+
+class X {
```
TomasVotruba, line 5: @samsonasik Hi, I guess these miss-names in fixture should be reported in CI, right? 
Is it working correctly?

TomasVotruba:
The CI is passing here too :thinking:  How does it perform in your local environment?

TomasVotruba:
Thanks for the options. Option 3 would be a way to go here, as purpose of this rule is only property promotion.

The inline change would be role of another rule e.g. in code quality set.

rules/php80/tests/Rector/Class_/ClassPropertyAssignToConstructorPromotionRector/parameter-is-not-passed-as-reference.php.inc
samsonasik: it seems because it uses dash instead of underscore, so it not detected

TomasVotruba: I see. How can we deal with those too?

samsonasik: I think the validate-fixture-* need dash check too

jkuchar: I have used a new feature from https://getrector.org/demo, using create PR. Code is unmodified from what has been generated by the tool.

jkuchar:
It looks to me, that test is not executed. As rector in demo environment breaks the code. (I'm sorry I have forgotten to pass the link here). Maybe it would be great if create PR adds link to DEMO also to code that produced the error.

https://getrector.org/demo/bfeb41bb-f2f3-4f1d-b519-946b857744ad

rules/php80/tests/Rector/Class_/ClassPropertyAssignToConstructorPromotionRector/parameter-is-not-passed-as-reference.php.inc
TomasVotruba: The filename is the problem

TomasVotruba:
I'l l check it locally

TomasVotruba:
Ah, I see the location bug now :)

jkuchar:
Thank you! 🙂

PR #4999, opened 2020-12-26
[phpstan] enable file lenght rule

TomasVotruba:
@staabm Hah, I just realized PHPStan ignores `*.php.inc` by default files :facepalm: 
With ECS we could use "only" parameter, but PHPStan does not have it


.github/workflows/code_analysis.yaml
```diff
@@ -46,6 +46,10 @@ jobs:
                         name: 'Validate PHPStan Mappers Compatibility'
                         run: bin/rector check-static-type-mappers --ansi
 
+                    -
+                        name: 'Validate Max File Length'
+                        run: bin/rector validate-file-length
```
TomasVotruba, line 51: @staabm @clxmstaab Custom command will do :)  
I've also shortened all long files reported by CI, so your local development is a bit smoother from now on

clxmstaab: ahh thx for the headsup. didn't realize rector also has custom commands :-).

initially wasn't sure whether to check this with rector or phpstan,.. didn't thought about that there could be a 3rd option ;)

TomasVotruba: Yea, same here. Sometimes it I have to write 3 rules for 3 different tools, before I get the right one :D that's how we learn.

At the moment I'm writing custom printer packages over PHPStan phpdoc parser https://github.com/rectorphp/phpdoc-parser-printer
It might be the same trip :D

PR #5006, opened 2020-12-27
[PHP 8.0] Skip promotion property if worked with before assign

TomasVotruba:
Closes https://github.com/rectorphp/rector/pull/4990


jkuchar:
Thank you! 🙂

PR #5008, opened 2020-12-27
[PHP 8.0] Skip accessed variable before property assign

TomasVotruba:
Closes https://github.com/rectorphp/rector/pull/4991


TomasVotruba:
@jkuchar I looked into this, fixed the website for PR link path and added demo link to PR description.

This case won't be refactored, as variable is worked with before the property access.
It should be handled by a code-quality set rule, e.g. this one: https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#simplifyuselessvariablerector

PR #5013, opened 2020-12-27
Add failing test fixture for AddArrayReturnDocTypeRector - array shape is replaced by structural type

jkuchar:
# Failing Test for AddArrayReturnDocTypeRector

Based on https://getrector.org/demo/4afc86f8-de78-4f13-bed4-3644f1b16d6d

`AddArrayReturnDocTypeRector` replaces array shape with less specific structural type.

BTW: current assigned type `int[][]|float[][]|string[][]` is wrong, as it does NOT allow to pass int|float|string for each value. It allows only int-only array, or float-only array, etc. Correct (still less specific type) would be `array<string, int|float|string>[]` or `(int|float|string)[][]`

### DIFF
--- a/rules/type-declaration/tests/Rector/ClassMethod/AddArrayReturnDocTypeRector/Fixture/b.php.inc
+++ b/rules/type-declaration/tests/Rector/ClassMethod/AddArrayReturnDocTypeRector/Fixture/b.php.inc
@@ -1,0 +1,19 @@
+<?php
+
+namespace Rector\TypeDeclaration\Tests\Rector\ClassMethod\AddArrayReturnDocTypeRector\Fixture;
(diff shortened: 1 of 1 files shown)


jkuchar:
Probably same problem with overrides method. https://getrector.org/demo/27963b19-4d20-4f99-b376-267d21aa828c

TomasVotruba:
Thank you! These array shapes should be skipped, as they're more advanced than simple arrays.

I think people who array shapes do it on purpose and want to keep it that way :)

jkuchar:
Thank you! I have tried this fix, it is much better. It does not fix this problem with inherited methods. Follow up 👉 #5017

PR #5016, opened 2020-12-28
[TypeDeclaration] Skip array shrapnels

TomasVotruba:
Fixes #5013

PR #5017, opened 2020-12-28
Add failing test fixture for AddArrayReturnDocTypeRector – shape arrays inherited from parent

jkuchar:
Follow up from #5016

# Failing Test for AddArrayReturnDocTypeRector

Based on https://getrector.org/demo/27963b19-4d20-4f99-b376-267d21aa828c

Rector should also ignore array shapes which are inherited from parent parent class method.

### DIFF
--- a/rules/type-declaration/tests/Rector/ClassMethod/AddArrayReturnDocTypeRector/Fixture/a.php.inc
+++ b/rules/type-declaration/tests/Rector/ClassMethod/AddArrayReturnDocTypeRector/Fixture/a.php.inc
@@ -1,0 +1,17 @@
+<?php
+
+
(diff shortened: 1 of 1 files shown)


rules/type-declaration/tests/Rector/ClassMethod/AddArrayReturnDocTypeRector/Fixture/a.php.inc
```diff
@@ -0,0 +1,17 @@
+<?php
+
+
+namespace Rector\TypeDeclaration\Tests\Rector\ClassMethod\AddArrayReturnDocTypeRector\Fixture;
+
+abstract class A {
```
TomasVotruba, line 6: Could you use more descriptive name, like `SkipShapeTypeWithParentDeclaratoin`?

The fixture is named afte this and it makes debugging months later much easier.

PR #5020, opened 2020-12-28
Add support for inheritdoc in parent type use

TomasVotruba:
Fixes https://github.com/rectorphp/rector/pull/5017

Based on https://getrector.org/demo/27963b19-4d20-4f99-b376-267d21aa828c


rules/type-declaration/tests/Rector/ClassMethod/AddArrayReturnDocTypeRector/Fixture/child_has_priority.php.inc
```diff
@@ -0,0 +1,66 @@
+<?php
+
+namespace Rector\TypeDeclaration\Tests\Rector\ClassMethod\AddArrayReturnDocTypeRector\Fixture;
+
+class ChildHasPriority extends ParentClassWithDefinedReturn
+{
+    public function getData()
```
TomasVotruba, line 7: Here the problem is not related to array shape of parent, but to missing information of any array type here.

Why type must be defined regardless parent definition?

- If parent methods can be source of truth, it would prevent more specific children. 
- Static tools and PHPStorm work better with explicit definition, then guessing what user ment by leaving the doc block. E.g what type is this property?

   ```php
   private $application;
   ```

- If the inheritance is on purpose, it must be explicitly stated. This feature was missing here, so I've added it.

```
    /**
     * @inheritdoc
     */
    public function getData()
```

This case would be skipped. Not sure how PHPStorm or static analysis tool deals with it though. Personally I put type everyhere, where tools or me might use it.

TomasVotruba: cc @jkuchar

jkuchar: I see two acceptable ways:
- I would be totally OK if rector copy the array shape type to child property. This would be the same behaviour as what you get when you use native type hints.
- It would also make sense to me to have `@inheritdoc`, but from what I know it is the same thing as leaving the docblock empty.

jkuchar: I have tried current behaviour and I strongly lean to the first variant with automatically adding array shape type from parent.

TomasVotruba: The 1st way assumes a lot and parent types can be invalid. Copying invalid docblock does not make sense.

2nd way explicitly states you've decided to trust the parent. `@inheritdoc` is standard way to state that.

<br>

Some developers prefer respecting the parent, as it might happen in `/vendor`. 


Personally, I'd go with value object in this particular case, that has exactly 1 way to use :)

jkuchar: Sure, that is THE way to go. However I wanted to type hint old code before refactoring. This helps me to better understand what is going on there. :-)

jkuchar: I'm still not 100% sure if `@inheritdoc` is the way to go. Even ECS considers this as spam, and removes this in default config. source: https://github.com/symplify/easy-coding-standard/blob/30d5721b16029e8f307da9e028af78c9f55f2499/config/set/common/docblock.php#L50

TomasVotruba: Personally, I never use it and just duplicate the docs. It's the easiest step to PHP migration then.

jkuchar: I have tried current PHPStan behaviour and it looks like there is no needs for inherit doc. So the best default would be to NOT add php doc to child methods, when parent already has one. What do you think?

[phpstan demo](https://phpstan.org/r/cf95b91b-9d23-46aa-8df1-d455de78ba1c)

jkuchar: This comment has been probably lost. What about this [phpstan demo](https://phpstan.org/r/cf95b91b-9d23-46aa-8df1-d455de78ba1c)? :-)

TomasVotruba: That direct use, easy :), the method call is the devil.
Some tools take only first docblock they find, without traversing to parents.

Issue #5024, opened 2020-12-28
Fatal error redeclare class: class_exists check triggers re-including project file via composer based autoloading


ktomk:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | e.g. v0.9.1 (invoke `vendor/bin/rector --version`)              |
| Installed as   | composer rector/rector-prefixed  (rector/rector fatals, confirm of #4559 fixed to prefix (?!))               |

## Minimal PHP Code Causing Issue

I could not isolate the problem. What is causing the issue is the following line in a method docblock:

~~~
 * @param \DateTime $time1
~~~

## Expected Behaviour + More Information

`class_exists()` check on the parameter type should reveal that the PHP global DateTime class is loaded.

However the leading backslash `\` is taken into the class_exists() check and the whole suffixed to the files `namespace` declaration
making the class_exists() check report it does not exists *and* triggers the auto-loader which then tries to load the class (but does backslash namespace normalization) and therefore re-includes a file leading to a fatal error as that class in the namespace already exists.

The error message:

> Fatal error:  Cannot declare class Name\Space\DateTime, because the name is already in use in ...

Excerpt from stack-trace w/ Xdebug:
~~~
    6.5453   80157776  41. Rector\NodeTypeResolver\ClassExistenceStaticHelper::doesClassLikeExist($classLike = 'Name\\Space\\\\DateTime') /home
    6.5453   80157776  42. class_exists($classname = 'Name\\Space\\\\DateTime') /home/user/.config/composer/vendor/rector/rector-prefixed/packag
    6.5453   80157864  43. spl_autoload_call($class_name = 'Name\\Space\\\\DateTime') /home/user/.config/composer/vendor/rector/rector-prefixed/
    6.5453   80157952  44. Composer\Autoload\ClassLoader->loadClass($class = 'Name\\Space\\\\DateTime') ...
~~~

`class_exists()` call that triggers `spl_autoload()`: https://github.com/rectorphp/rector/blob/master/packages/node-type-resolver/src/ClassExistenceStaticHelper.php#L11

([full output](https://gist.github.com/ktomk/bf1b0fbd9b4c430efdf88deb19562aa1))

## Initial Guesswork

[...]

PR #5026, opened 2020-12-28
Fix fatal error auto-loading file for already existing class by invalid class-name

ktomk:
... filing fix for #5024

- [x] Regression test for invalid class-names
- [x] Give Phpstan a helping hand on anonymous class in test-suite
- [x] Add the fix (ltrim)

Using invalid class-names, specifically with two consecutive backslashes, while performing `class_exist()` checks, triggers auto-loading for the invalid class-name which then may result in the re-inclusion of existing files (due to the mapping of the class-name to a file-system-path) which then may trigger a fatal error if such a file contains the class-definition of an already defined
class.

          load class: Name\Space\DoubleTrouble
    file-system path: ../Name/Space/DoubleTrouble.php

    class_exists(Name\Space\\DoubleTrouble)
                           ^^
       file-system path: ../Name/Space//DoubleTrouble.php
                                      ^^
    already loaded file: ../Name/Space/DoubleTrouble.php

This fatal error brings rector to halt.

To stabilize class_exist() while auto-loading is active (it is always active w/ rector), prevent invalid class-names, specifically containing two or more consecutive backslashes.

Fix is to prevent consecutive backslashes when putting the name-space and the class-name together by dropping all leading backslashes from the class-name.

### DIFF
--- a/rules/renaming/tests/Rector/FileWithoutNamespace/PseudoNamespaceToNamespaceRector/PseudoNamespaceToNamespaceRectorTest.php
+++ b/rules/renaming/tests/Rector/FileWithoutNamespace/PseudoNamespaceToNamespaceRector/PseudoNamespaceToNamespaceRectorTest.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace Rector\Renaming\Tests\Rector\FileWithoutNamespace\PseudoNamespaceToNamespaceRector;
 
 use Iterator;
+use PHPUnit\Framework\ExpectationFailedException;
+use PHPUnit\Framework\TestCase;
 use Rector\Generic\ValueObject\PseudoNamespaceToNamespace;
 use Rector\Renaming\Rector\FileWithoutNamespace\PseudoNamespaceToNamespaceRector;
 use Rector\Testing\PHPUnit\AbstractRectorTestCase;
@@ -17,7 +19,11 @@ final class PseudoNamespaceToNamespaceRectorTest extends AbstractRectorTestCase
      */
     public function test(SmartFileInfo $fileInfo): void
     {
+        $autoloadClassNameAssertion = $this->getAutoloadClassNameAssertion();
+
         $this->doTestFileInfo($fileInfo);
+
+        $autoloadClassNameAssertion->assert($fileInfo);
     }
 
     public function provideData(): Iterator
@@ -41,5 +47,66 @@ final class PseudoNamespaceToNamespaceRectorTest extends AbstractRectorTestCase
                 ],
             ],
         ];
+    }
+
+    /** @noinspection PhpMissingReturnTypeInspection */
+    // @phpstan-ignore-next-line
+    private function getAutoloadClassNameAssertion()
+    {
+        return new class ($this) { // @phpstan-ignore-line
+            // ref: https://www.php.net/manual/en/language.oop5.basic.php
+            private const CLASS_NAME = '/^
+                # valid PHP class-name (incl. optional leading \\ name-space separator)
+                \\\\?
+                [a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*
+                (?: \\\\ [a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]* )*
+                $/Dx';
+
+            /**
+             * @var TestCase
+             */
+            private $testCase;
+
+            /**
+             * @var string[]
+             */
+            private $sequence = [];
+
+            /**
+             * @param TestCase $testCase
+             * @noinspection InterfacesAsConstructorDependenciesInspection
+             */
+            public function __construct(TestCase $testCase)
+            {
+                $this->testCase = $testCase;
+                \spl_autoload_register($this, true ,true);
+            }
+
+            public function __invoke(string $className): void
+            {
+                $this->sequence[] = $className;
+            }
+
+            /**
+             * @param SmartFileInfo $fileInfo
+             * @throws ExpectationFailedException
+             */
+            public function assert(SmartFileInfo $fileInfo): void
+            {
+                foreach ($this->sequence as $className) {
+                    $this->testCase::assertMatchesRegularExpression(
+                        self::CLASS_NAME,
+                        $className,
+                        $fileInfo->getBasename() . ' class-name validity'
+                    );
+                }
+                $this->sequence = [];
+            }
+
+            public function __destruct()
+            {
+                \spl_autoload_unregister($this);
+            }
+        };
     }
 }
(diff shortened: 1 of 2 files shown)


rules/renaming/tests/Rector/FileWithoutNamespace/PseudoNamespaceToNamespaceRector/Fixture/fixture6.php.inc
```diff
@@ -12,6 +12,13 @@ class Fixture6
     public function test(\PHPUnit_Framework_TestCase $test): \PHPUnit_Framework_TestCase
     {
     }
+
+    /**
+     * @param \DateTime $time
+     */
+    public function testMore(\DateTime $time)
+    {
+    }
```
TomasVotruba, line 21: Could you decoupe this to a new test fixture file? Just so the case is isolated

rules/type-declaration/src/PHPStan/Type/ObjectTypeSpecifier.php
```diff
@@ -128,7 +128,9 @@ private function matchSameNamespacedObjectType(Node $node, ObjectType $objectTyp
             return null;
         }
 
-        $namespacedObject = $namespaceName . '\\' . $objectType->getClassName();
+        // ltrim: prevents double-backslash that gives fatal error with existing class for class_exists() while
+        // auto-loading #5024
+        $namespacedObject = $namespaceName . '\\' . ltrim($objectType->getClassName(), '\\');
```
TomasVotruba, line 133: Does this fix the bug you had?

Could you run first CI without this fix, so we can see the reported bug? Thanks :+1:

TomasVotruba:
Not sure if that would help, but there is also `doTestFileInfoWithoutAutoload()` method, that will test the file without autoloading it.

rules/type-declaration/src/PHPStan/Type/ObjectTypeSpecifier.php
ktomk: Yes, this is identified as effective against the fatal error. I'll do the regression test first, this includes auto-loading.

rules/renaming/tests/Rector/FileWithoutNamespace/PseudoNamespaceToNamespaceRector/Fixture/fixture6.php.inc
ktomk: Sure, I put it into a new fixture file, this one was just a quick shot to get this running on CI here.

rules/type-declaration/src/PHPStan/Type/ObjectTypeSpecifier.php
ktomk: CI without this fix is in: https://github.com/rectorphp/rector/pull/5026/commits/88163d6c5f6b315d9649392f2aeb34fa7f3424d7

Phpstan is exceptionally low on anonymous classes and over-estimates itself to not infer the anonymous type (Phpstorm does this). If you know by chance a docblock notation to handle this please let me know, also about suppressing these, I'm not as fluent with Phpstan as you are.

ktomk: Noise reduced by suppression. The one is a known issue with Phpstan (https://github.com/phpstan/phpstan/issues/3144) and the other is an issue with Simplify (just reported: https://github.com/symplify/symplify/issues/2717).

rules/renaming/tests/Rector/FileWithoutNamespace/PseudoNamespaceToNamespaceRector/PseudoNamespaceToNamespaceRectorTest.php
```diff
@@ -42,4 +48,65 @@ protected function getRectorsWithConfiguration(): array
...
+                        $className,
+                        $fileInfo->getBasename() . ' class-name validity'
+                    );
+                }
+                $this->sequence = [];
+            }
+
+            public function __destruct()
+            {
+                \spl_autoload_unregister($this);
```
TomasVotruba, line 108: This must be handled without hacking autoload, so it's easy to maintain in years to come

ktomk: Yes sure this must be stable as the foundations of mountains. Therefore it would help me if you could elaborate a bit where this spl_autoload_register / spl_autoload_unregister scheme poses a maintenance burden? Actually I can not imagine that it will go away nor that composer drops it. @TomasVotruba

TomasVotruba: Autoload is handled by composer, so we can autoload it by putting in psr-4 scope or not autoloadit by putting anywhere else. No need to create 2nd composer :)

TomasVotruba:
I'm looking into this and I can't find the test fixture in `/Fixtures` directory.
Where is it located?

rules/renaming/tests/Rector/FileWithoutNamespace/PseudoNamespaceToNamespaceRector/PseudoNamespaceToNamespaceRectorTest.php
ktomk: Yes, sure. But this is all not happening with the change-set :) It just keeps a list of class-names to assert it later ...

TomasVotruba:
I'm running https://github.com/ktomk/rector-report now, I see the bug I look for, so all good for now :+1:

ktomk:
@TomasVotruba: Yes, I also put the bugfix-branch as a job on rector-report so this extra test is visible.

rules/renaming/tests/Rector/FileWithoutNamespace/PseudoNamespaceToNamespaceRector/PseudoNamespaceToNamespaceRectorTest.php
TomasVotruba: See https://github.com/rectorphp/rector/pull/5040/files

TomasVotruba:
Manaully merged in https://github.com/rectorphp/rector/pull/5040

Thank you :+1:

ktomk:
Huh, why did you merge changes that weren't accepted? Are you that much in a rush?

And thanks for having me.

TomasVotruba:
I merged only the accepted changes :) 

This was an edge case that required local fixture of `DateTime.php`, which is not allowed. It needed to excluded from typical fixtures, no `.php.inc`

PR #5027, opened 2020-12-28
Add failing test fixture for AddArrayDefaultToArrayPropertyRector – must NOT initialize property as type after intialization is incompatible

jkuchar:
must NOT initialize property as type after intialization is incompatible

# Failing Test for AddArrayDefaultToArrayPropertyRector

Based on https://getrector.org/demo/0d6a92ef-1369-4fc9-a08a-e394937298ae

### DIFF
--- a/rules/coding-style/tests/Rector/Class_/AddArrayDefaultToArrayPropertyRector/Fixture/array_shape_uninitialied_property.php.inc
+++ b/rules/coding-style/tests/Rector/Class_/AddArrayDefaultToArrayPropertyRector/Fixture/array_shape_uninitialied_property.php.inc
@@ -1,0 +1,14 @@
+<?php
+
+namespace Rector\CodingStyle\Tests\Rector\Class_\AddArrayDefaultToArrayPropertyRector\Fixture;
+
+final class ArrayShapeUninitialiedProperty
+{
+    /**
+	 * @var array{
+	 *   price: float,
+	 * }
+	 */
+	public array $array;
+}
+?>


TomasVotruba:
That's actually what `AddArrayDefaultToArrayPropertyRector` does :) 

Acceptation criteria for this rule is that property is foreachable.
I recommend disabling this one, until you switch to objects.

jkuchar:
I understand, but adding empty array produces type mismatch, therefore invalid code. This one should be automatically skipped when specified type does not accept `[]` empty array. That is basically all array shapes.

TomasVotruba:
That's what this rule does. Also such property is invalid - https://3v4l.org/5UEKa 
Excluding some arrays from being an array would be WTF.

I'd open to having an opt-in parameter, which would disable this for array shapes. If you add such feature, we can merge it.

jkuchar:
I assume that:
- if some add array shape php doc that he has done that intentionally
- if there is an array shape php doc, there almost no use case for iterating over an array. Array shape says that this array is and implicit value object.
- if you want to cover migration to php7.4, if there is already `array` in property type, migration should be skipped, as it has already been done (or this should be coordinated with rector adding array type to a property)

jkuchar:
What is your opinion? I still think that it does not make sense to add incompatible default value to property. So disabling this for array shapes makes sense in all cases (should not be optional).

TomasVotruba:
I'm still standing behind my answer it's what rule should do, but if you need it, I'm open to PR for optional drop off.

PR #5040, opened 2020-12-29
[Renaming] Add test fixture for same-namespaced short class fail

### LINKED ISSUE #5024: Fatal error redeclare class: class_exists check triggers re-including project file via composer based autoloading

TomasVotruba:
- Closes https://github.com/rectorphp/rector/pull/5026
- Closes https://github.com/rectorphp/rector/issues/5024

### DIFF
--- a/rules/renaming/tests/Rector/FileWithoutNamespace/PseudoNamespaceToNamespaceRector/Fixture/DateTime.php
+++ b/rules/renaming/tests/Rector/FileWithoutNamespace/PseudoNamespaceToNamespaceRector/Fixture/DateTime.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace Rector\Renaming\Tests\Rector\FileWithoutNamespace\PseudoNamespaceToNamespaceRector\Fixture;
+
+class DateTime
+{
+}
--- a/rules/renaming/tests/Rector/FileWithoutNamespace/PseudoNamespaceToNamespaceRector/Fixture/preslashed_bug.php.inc
+++ b/rules/renaming/tests/Rector/FileWithoutNamespace/PseudoNamespaceToNamespaceRector/Fixture/preslashed_bug.php.inc
@@ -1,0 +1,15 @@
+<?php
+
+namespace Rector\Renaming\Tests\Rector\FileWithoutNamespace\PseudoNamespaceToNamespaceRector\Fixture;
+
+class PreslashedBug
+{
+    /**
+     * @param \DateTime $time1
+     * @return int
+     */
+    public function test(\DateTime $time1)
+    {
+        return $time1->getOffset();
+    }
+}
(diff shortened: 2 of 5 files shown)


TomasVotruba:
@ktomk Here is the bug we needed to replicated: https://github.com/rectorphp/rector/pull/5040/checks?check_run_id=1622614004#step:6:59

image

ktomk:
> @ktomk Here is the bug we needed to replicate: https://github.com/rectorphp/rector/pull/5040/checks?check_run_id=1622614004#step:6:59
> 
> image

Yes, beware that the order of the fixtures is undefined. This may not apply right now as testing with collecting the class-names has shown that these bogus classes are triggered in auto-loading multiple times each even if there is only a single reference (three times per one reference b/c of the class/interface/trait_exists line-up that must be), but this is maybe not that well maintainable. Just saying, can be done this way, but it does not test for the regression just for the unspecified behaviour (which can be done with a single fixture-file btw.).

TomasVotruba:
Sure, any further improvents are welcomed :+1:

PR #5041, opened 2020-12-29
[Test] Move Re-generate code coverage to weekly pull request

samsonasik:


.github/workflows/weekly_pull_requests.yaml
```diff
@@ -32,7 +50,7 @@ jobs:
                 uses: shivammathur/setup-php@v2
                 with:
                     php-version: 7.4
```
samsonasik, line 52: @TomasVotruba should it be 7.3 same as code coverage action?

TomasVotruba: I think 7.4 is ok here

TomasVotruba (approved):

TomasVotruba:
Thank you

samsonasik:
Just realized that regenerate code coverage doesn't need a pull request, it may just need create branch - run coverage - delete branch

PR #5046, opened 2020-12-30
Skip variadic arguments from property promotion

### LINKED ISSUE #5045: `ClassPropertyAssignToConstructorPromotionRector` promotes variadic argument, which is not supported
# Bug Report

| Rector version | v0.9.3 |
| Installed as   | composer dependency |

`ClassPropertyAssignToConstructorPromotionRector` introduces not supported code syntax by using constructor property promotion on variadic arguments.

From [RFC](https://wiki.php.net/rfc/constructor_promotion):

> Variadic parameters cannot be promoted

```php
class Bar
{
    private array $ints;

    public function __construct(int ...$ints)
    {
        $this->ints = $ints;
    }
}
```

is changed like this:

```diff
--- Original
+++ New
@@ -6,10 +6,7 @@

 class Bar
 {
-    private array $ints;
-
-    public function __construct(int ...$ints)
+    public function __construct(private int ...$ints)
     {
-        $this->ints = $ints;
     }
 }

```

Which causes `Fatal error: Cannot declare variadic promoted property in ...`

## Expected Behaviour

Constructor's variadic arguments shouldn't be promoted.

Wirone:
Fixes #5045

### DIFF
--- a/rules/php80/tests/Rector/Class_/ClassPropertyAssignToConstructorPromotionRector/Fixture/skip_variadic_arguments.php.inc
+++ b/rules/php80/tests/Rector/Class_/ClassPropertyAssignToConstructorPromotionRector/Fixture/skip_variadic_arguments.php.inc
@@ -1,0 +1,37 @@
+<?php
+
+namespace Rector\Php80\Tests\Rector\Class_\ClassPropertyAssignToConstructorPromotionRector\Fixture;
+
+class SkipVariadicArguments
+{
+    private string $foo;
+    /** @var string[] */
+    private array $bars;
+
+    public function __construct(
+        string $foo,
+        string ...$bars
+    ) {
+        $this->foo = $foo;
+        $this->bars = $bars;
+    }
+}
+?>
+-----
+<?php
+
+namespace Rector\Php80\Tests\Rector\Class_\ClassPropertyAssignToConstructorPromotionRector\Fixture;
+
+class SkipVariadicArguments
+{
+    /** @var string[] */
+    private array $bars;
+
+    public function __construct(
+        private string $foo,
+        string ...$bars
+    ) {
+        $this->bars = $bars;
+    }
+}
+?>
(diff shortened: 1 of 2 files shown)


rules/php80/tests/Rector/Class_/ClassPropertyAssignToConstructorPromotionRector/Fixture/skip_variadic_arguments.php.inc
```diff
@@ -0,0 +1,37 @@
...
+
+class SkipVariadicArguments
+{
+    private string $foo;
+    /** @var string[] */
+    private array $bars;
+
+    public function __construct(
+        string $foo,
+        string ...$bars
```
TomasVotruba, line 13: Skipping the variadic would be better here, so it's clear why this fixture was added.
The other property can be dropped

Wirone: I wanted to make sure that it will still handle properly supported promotion while skipping not supported one. But if you think that single, skipped argument will be good enough, I will remove `$foo`. Maybe it should be tested in other fixture (ensure promoting and skipping can be done in the same constructor).

TomasVotruba: I understood that and used to write tests the same way. Recently I'm more into minimal-scope appraoch, with the minimal code possible that shows the bug. It makes reading fixtures from years hindsigh much easier than longer snippets.

Here in this particular rule each property is resolved standalone, so they don't influence each other. It's already covered in many other fixtures.

Wirone: OK, `$foo` removed :+1:

TomasVotruba:
Thank you for test and the fix :+1:

TomasVotruba (approved):

PR #5052, opened 2020-12-30
Add failing test fixture for RemoveParentAndNameFromComponentConstructorRector

lulco:
# Failing Test for RemoveParentAndNameFromComponentConstructorRector

Based on https://getrector.org/demo/12f10dc7-7ffb-4fe1-8004-7bf82e7aa9ce

There are multiple problems in this rule:
1) only $name is removed ($parent should be also removed) **EDIT** there was missing use for IContainer
2) $name (and if it would work also $parent) is removed also when there is some future use of it (e.g. assigning to private property $this->name = $name)


rules/nette/tests/Rector/ClassMethod/RemoveParentAndNameFromComponentConstructorRector/Fixture/control.php.inc
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+?>
+-----
+<?php
+
+namespace Rector\Nette\Tests\Rector\ClassMethod\RemoveParentAndNameFromComponentConstructorRector\Fixture;
+
+// what is expected code?
+// should remain the same? delete part bellow ----- (included)
```
TomasVotruba, line 54: This section is missing. The expected output must be here

lulco: Yeah, I have to check how to do it O:-)
But there is another problem:
```
PHP Fatal error:  Cannot declare interface Nette\ComponentModel\IContainer, because the name is already in use in /tmp/_temp_fixture_easy_testing/input_bb1a98f5d08ae38bde39_control.php on line 5
```

TomasVotruba: The class is duplicated :) It's already in the stubs: https://github.com/rectorphp/rector/blob/master/stubs/Nette/ComponentModel/IContainer.php

TomasVotruba: > Yeah, I have to check how to do it O:-) 

You can modify [it here](https://github.com/lulco/rector/edit/patch-1/rules/nette/tests/Rector/ClassMethod/RemoveParentAndNameFromComponentConstructorRector/Fixture/control.php.inc?pr=%2Frectorphp%2Frector%2Fpull%2F5052), or better on creating pull-request next time.

lulco: @TomasVotruba Failing tests added, now build "failed correctly"

samsonasik:
@lulco I cherry-picked your commit at PR #5098

PR #5074, opened 2021-01-02
Full upgrade of codebase including changes in composer.json

### LINKED ISSUE #5054: Full upgrade of codebase including changes in composer.json
# Feature Request
If this is already possible, stop reading here and please tell me how to do it :)

What if Rector can update codebase of application and also change composer.json (ideally with executing composer update)?

I am upgrading Nette applications from 2.4 to 3.0 (but this would be usable for all upgrades and also downgrades), I run Rector and then I have to change my composer.json manually and run composer update manually.

I was thinking about some non-php rule which will change composer.json, configuration should be like this:

```
$services->set(ComposerRector::class)
    ->call('configure', [[
        'change_package_version' => [  // change version of package which is already in composer.json
            'nette/application' => '^3.0',
            'nette/forms' => '^3.0',
            ...
        ],
        'add_package' => [    // add new package or change version of existing package
            'nette/mail' => '^3.1',
            ...
        ],

         // maybe some other changes like remove packages and change packages
    ]]);
```

All sets could include this rule.

For example. Update php to 8.0 should change composer.json this way:
```diff
-"php": "^7.1",
+"php": "^8.0",
```

If there would be some interest for this feature, I can try to implement it. The only issue I found for now is how to find composer.json file while executing rector process command

lulco:
First version of implementation, please review

Resolves #5054


lulco:
@TomasVotruba build is finally green, can you please check this?

lulco:
@TomasVotruba @samsonasik Hi, I don’t want to be rude. Can I get some feedback here? I know you have a lot of work, just please look at this PRs, please :)

TomasVotruba:
Thank you for your work! I'm sorry, the first week of the year is more busy ever :)

From rough first check, it would help to have data in tests as a data providers. It would make it more clean and easy to spot what is input and what expected.

Maybe even test fixtures like this:

```diff
{
   "json": "before"
}
-----
{
   "json": "after"
}
```


I'll check in more detailed way right after

rules/composer/tests/Modifier/ComposerModifierTest.php
```diff
@@ -0,0 +1,223 @@
...
+use Rector\Composer\ValueObject\ChangePackageVersion;
+use Rector\Composer\ValueObject\MovePackage;
+use Rector\Composer\ValueObject\RemovePackage;
+use Rector\Composer\Modifier\ComposerModifier;
+
+final class ComposerModifierTest extends TestCase
+{
+    public function testRefactorWithOneAddedPackage(): void
+    {
+        $composerModifier = new ComposerModifier();
```
TomasVotruba, line 18: This one should be service and used with `AbstractKernelTestCase`. See other tests how to get a service from container

rules/composer/src/ValueObject/AddOrChangePackage.php
```diff
@@ -0,0 +1,58 @@
...
+     * @param string $packageName name of package (vendor/package)
+     * @param string $version target package version (1.2.3, ^1.2, ~1.2.3 etc.)
+     * @param string $section require or require-dev
+     */
+    public function __construct(string $packageName, string $version, string $section = ComposerModifier::SECTION_REQUIRE)
+    {
+        Assert::oneOf($section, [ComposerModifier::SECTION_REQUIRE, ComposerModifier::SECTION_REQUIRE_DEV]);
+
+        $this->packageName = $packageName;
+        $this->version = $version;
```
TomasVotruba, line 36: It would be great if version is validated here and standardized as an object. We have a `VersionFactory` to check this, or `Schema` service from composer can be used

rules/composer/src/ValueObject/AddPackage.php
```diff
@@ -0,0 +1,51 @@
...
+
+        $this->packageName = $packageName;
+        $this->version = $version;
+        $this->section = $section;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function modify(array $composerData): array
```
TomasVotruba, line 43: Working with array is kind-of unreliable. There is no autocomplete, static analysis nor automated refactoring.

It would be better to work with object `ComposerJson` wich exact API, see
https://tomasvotruba.com/blog/2020/10/31/new-in-symplify-9-composer-json-manipulator-in-object-api/

rules/composer/src/ValueObject/ChangePackage.php
```diff
@@ -0,0 +1,50 @@
...
+namespace Rector\Composer\ValueObject;
+
+use Rector\Composer\Modifier\ComposerModifier;
+use Rector\Composer\Contract\ComposerModifier\ComposerModifierInterface;
+
+/**
+ * Changes one package for another
+ * @see \Rector\Composer\Tests\ValueObject\ChangePackageTest
+ */
+final class ChangePackage implements ComposerModifierInterface
```
TomasVotruba, line 14: Is this about rename package or change package version? Class name should tell the story

rules/composer/src/ValueObject/MovePackage.php
```diff
@@ -0,0 +1,53 @@
...
+
+use Rector\Composer\Modifier\ComposerModifier;
+use Rector\Composer\Contract\ComposerModifier\ComposerModifierInterface;
+use Webmozart\Assert\Assert;
+
+/**
+ * Moves package to selected section, if package is not in composer data, nothing happen, also if package is already in selected section
+ * @see \Rector\Composer\Tests\ValueObject\MovePackageTest
+ */
+final class MovePackage implements ComposerModifierInterface
```
TomasVotruba, line 15: Move package where? 

I see from code bellow the require/require-dev is the target. It would be better to have 2 objects, that could explain this in a class name. E.g. MovePackageToRequire, MovePackageToRequireDev

TomasVotruba:
I've added few comments. The code looks very clean and is easy to read, good job :+1: 
I also like how tests are separated to own test case files :+1:

rules/composer/src/ValueObject/ChangePackage.php
lulco: Well, it is both. When you are changing kdyby/translation to contributte/translation you need to check package and its version too

rules/composer/src/ValueObject/MovePackage.php
lulco: Ok I will split it to two classes

rules/composer/tests/Modifier/ComposerModifierTest.php
lulco: Ok I will try to rewrite it

rules/composer/src/ValueObject/AddPackage.php
lulco: True, didn’t know the package. I can try how it will work here and also in tests

rules/composer/src/ValueObject/ChangePackage.php
TomasVotruba: I see. The `ChangePackage` name is like `ChangeClass`. It could be anything :) That's why we have `RenameClassRector`, `RemoveDeadClassRector`, `PrivatizeClassRector` etc.

IMO these should be 2 different classes, one for package rename and other for version change

lulco: I agree that we need another name of this class but imho it should does what it does now - change the single line of composer.json - package name and also set the version of new package. I think it is more user friendly like run two modifiers

rules/composer/src/ValueObject/AddOrChangePackage.php
lulco: I am trying to find one of these classes in Rector (or vendor), but no success. I just found `PhpVersionFactory` and `Composer\Semver\VersionParser` with method `normalize()`, but it will be just some validation and it will not create version as an object. Can you navigate me to classes you have in mind?

rules/composer/src/ValueObject/AddPackage.php
lulco: I've check this ComposerJson class and I have some reservations:
- I can't add or remove package by calling some method, I have to get packages, then unset key from array and set packages back. Same with replace one package to another. So in my code, I will still work with arrays :(
- it sorts packages always when `setRequire` / `setRequireDev` are called, even if there is no sort-packages in composer.json

rules/composer/src/ValueObject/MovePackage.php
lulco: AddPackage has also this $section flag in __construct. Should I split it too?

rules/composer/src/ValueObject/ChangePackage.php
lulco: What about `ReplacePackage`?

rules/composer/src/ValueObject/AddPackage.php
TomasVotruba: New methods should be added to Symplify to allow this

rules/composer/src/ValueObject/MovePackage.php
TomasVotruba: Yes, makes sense

rules/composer/src/ValueObject/AddOrChangePackage.php
TomasVotruba: Something like this: https://github.com/symplify/symplify/blob/master/packages/monorepo-builder/src/ValueObjectFactory/VersionFactory.php

It might require combination or new object. The goal is object API, not a string, so it's reliable

rules/composer/src/ValueObject/ChangePackage.php
TomasVotruba: Sounds good!

rules/composer/src/ValueObject/AddPackage.php
lulco: Ok, so I will prepare PR

lulco: PR ready for review https://github.com/symplify/composer-json-manipulator/pull/1

rules/composer/src/ValueObject/AddOrChangePackage.php
lulco: done

rules/composer/src/ValueObject/ComposerModifier/AddPackageToRequireDev.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->version = new Version($version);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function modify(array $composerData): array
+    {
+        if (!isset($composerData[ComposerModifier::SECTION_REQUIRE][$this->packageName]) && !isset($composerData[ComposerModifier::SECTION_REQUIRE_DEV][$this->packageName])) {
+            $composerData[ComposerModifier::SECTION_REQUIRE_DEV][$this->packageName] = $this->version->getVersion();
```
TomasVotruba: `ComposerJsonSection` should be re-used: https://github.com/symplify/symplify/blob/614d76d68a92a7382d7e87d0b9ce6be4e23d1884/packages/composer-json-manipulator/src/ValueObject/ComposerJsonSection.php#L19

rules/composer/src/ValueObject/ComposerModifier/AddPackageToRequireDev.php
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        return $composerData;
+    }
```
TomasVotruba: I'm not sure if this is value object or a service.
It modified outer data `$composerData` without keeping a state, so it's a service.

It should be separated to value object and modifier service

lulco: this is some older version, actual is:

```
public function modify(ComposerJson $composerJson): ComposerJson
    {
        $composerJson->addRequiredDevPackage($this->packageName, $this->version->getVersion());
        return $composerJson;
    }
```

rules/composer/src/ValueObject/ComposerModifier/AddPackageToRequireDev.php
lulco: this method is completely rewritten using ComposerJson object from symplify/composer-json-manipulator package

rules/composer/src/ValueObject/ComposerModifier/ChangePackageVersion.php
```diff
@@ -2,21 +2,22 @@
...
+ * @see \Rector\Composer\Tests\ValueObject\ComposerModifier\ChangePackageVersionTest
  */
 final class ChangePackageVersion implements ComposerModifierInterface
 {
     /** @var string */
     private $packageName;
 
-    /** @var string */
+    /** @var Version */
     private $targetVersion;
```
TomasVotruba, line 21: :+1:

rules/composer/src/ValueObject/ComposerModifier/AddPackageToRequireDev.php
TomasVotruba: I see. That's the same, just with different argument.

lulco: So basically, you want me, to split all these classes to two classes:
1) ValueObject with only constructor (and probably some getters)
and
2) service with modify method which will have 2 arguments ComposerJson $composerJson and ValueObject for getting configuration (package name, version etc.).

Right?

lulco: @TomasVotruba Hello, I didn't get an answer here :) I just want to say that first implementation get this comment from you: `The code looks very clean and is easy to read` :) With 2x more classes it imho will not be clean and easy to read. Also first implementation was really easy to add another modifiers like ChangeName, ChangeDescription etc. because you just manipulate an array in modify() method of ComposerModifierInterface implementation. It is also possible now with ComposerJson object (if it has all needed methods), but adding another layer just unnecessarily add some complexity which is imho not needed here... But it is your package, so if you want, I will rewrite it as you wish, just answer to my question. As I am thinking now, maybe I will have to add some Factory to select correct service depending on which value object will be processing

TomasVotruba: Sorry, I've missed it. Yes you got it right 👍

TomasVotruba:
An idea to keep your cognitive load smaller... Let's merge it as it is and you can make additional PR to process unresolved comments.

What do you think? :)

lulco:
I agree :) it is working as it is. I will just change internal implementation

TomasVotruba:
Done :)

PR #5084, opened 2021-01-04
[DeadCode] Skip RemoveUnusedPrivatePropertyRector on property fetch in unset

### LINKED ISSUE #5076: Incorrect behavior of Rector\DeadCode\Rector\Property\RemoveUnusedPrivatePropertyRector
# Bug Report

| Rector version | dev-master@900b60c             |
| Installed as   | git clone                            |

Parameter `$packageName` is removed from constructor as unused, but it is used in the next method. Maybe problem is it is in unset function?
```php
unset($composerData[$section][$this->packageName]);
```

https://getrector.org/demo/2ae1ce38-7504-4fae-84f5-6f78a8988d69

## Expected Behaviour

Do not remove private property from construct if it is in use

samsonasik:
Fixes #5076


samsonasik:

lulco:
@samsonasik I check another use case: https://getrector.org/demo/a8dfb6cc-936e-4f0b-8adf-5c50b81cf9c6

there is isset instead of unset and demo failed again, so probably unset isn't only problematic construction

TomasVotruba:
@samsonasik Thank you :+1:

Issue #5083, opened 2021-01-04
Call to undefined function Symfony\Component\DependencyInjection\Loader\Configurator\service()


tomasfejfar:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | rector/rector:latest             |
| Installed as   | dockerimage                          |

```
λ docker run -it --rm --entrypoint bash -v %CD%:/project rector/rector:latest
root@440c45a4a3fb:/project# cd legacy-app
root@440c45a4a3fb:/project/legacy-app# which rector
/rector/bin/rector
root@440c45a4a3fb:/project/legacy-app# rector process . --dry-run -vvvv


 [ERROR] Call to undefined function Symfony\Component\DependencyInjection\Loader\Configurator\service()


root@440c45a4a3fb:/project/legacy-app# rector --version


 [ERROR] Call to undefined function Symfony\Component\DependencyInjection\Loader\Configurator\service()

```

```
<?php

declare(strict_types=1);

use Rector\Core\Configuration\Option;
use Rector\Php74\Rector\Property\TypedPropertyRector;
use Rector\Set\ValueObject\SetList;
use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;

return static function (ContainerConfigurator $containerConfigurator): void {
    // get parameters
    $parameters = $containerConfigurator->parameters();

    // Define what rule sets will be applied
    $parameters->set(Option::SETS, [
        'php72',
    ]);

    // get services (needed for register a single rule)
    // $services = $containerConfigurator->services();

    // register a single rule
    // $services->set(TypedPropertyRector::class);
};

```

## Minimal PHP Code Causing Issue

No idea where to start the debug. It seems that rector does not output stacktrace for its exceptions even with `-vvv`

## Expected Behaviour

This does not happen :)

TomasVotruba:
Thanks for reporting. Try `--debug` option.
That looks like Docker autoloading issue.

If you include the function manually in composer.json files, does it help?

It's here:
https://github.com/symfony/dependency-injection/blob/5.x/Loader/Configurator/ContainerConfigurator.php#L109

Docker might be crappy here. How does [prefixed version do](https://github.com/rectorphp/rector-prefixed)? If PHP 7.3 is not a blocker, give it a try. Might be easier to use.

tomasfejfar:
```
root@cfd7f4f3afdb:/project/legacy-app# rector --debug


[...]

PR #5086, opened 2021-01-04
Publish all tags as docker images

tomasfejfar:
No tags except latest were published because tags no longer contain leading `v`

### DIFF
--- a/.github/workflows/publish_docker_images.yaml
+++ b/.github/workflows/publish_docker_images.yaml
@@ -7,7 +7,7 @@ on:
 
         # Publish `v1.2.3` tags as releases.
         tags:
-            - v*
+            - *
 
 jobs:
     publish_images:


tomasfejfar:
https://hub.docker.com/r/rector/rector/tags?page=1&ordering=last_updated

tomasfejfar:
Noticed when trying to debug https://github.com/rectorphp/rector/issues/5083

TomasVotruba:
Thanks for spotting this :+1:

tomasfejfar:
Can you start the action manually for some previous tag? So that I have something beyond "latest" to test?

TomasVotruba:
Sure.

It seems the syntax is invalid though:
https://github.com/rectorphp/rector/actions/runs/461203210/workflow

tomasfejfar:
Yes, I just noticed https://github.com/rectorphp/rector/pull/5087

tomasfejfar:
I do not have any GH actions myself, so I didn't know how to test it... :/

TomasVotruba:
> Yes, I just noticed #5087

Nice :+1: 

> I do not have any GH actions myself, so I didn't know how to test it... :/

One way to test GitHub Action on pull request is to invoke it on pull-request.

Temporarily, by adding these 2 lines:

https://github.com/rectorphp/rector/blob/1ddaf53bee8a628c8bd55399fac187d38cb4e49e/.github/workflows/code_analysis.yaml#L3-L4

That way at least syntax is checked.

tomasfejfar:
Wouldn't that push the image to dockerhub?

TomasVotruba:
Isn't that point of all this?

(It probably would not, as only repo write-users have access to security keys)

PR #5096, opened 2021-01-05
Fixed typo "lenght"

florisluiten:
Hi,

I was reading the docs, and found a typo that I wanted to fix. I did a grep on the entire codebase, and found a few other instances where "lenght" instead of "length" was used. One instance even did a check on `class_exists` on "lenght" where the class was called "length", and thus the `class_exists` call did not work as expected.

So, here is the mr for fixing docs and one small bug :)

Floris


samsonasik:
Thank you @florisluiten

samsonasik (approved):

PR #5100, opened 2021-01-06
Add failing test fixture for ExplicitBoolCompareRector

jkuchar:
# Failing Test for ExplicitBoolCompareRector

Based on https://getrector.org/demo/3cb57eee-954d-4f12-9d0f-7ce703672f60

As in this case rector has no idea what is inside the variable, this case should be skipped.


samsonasik:
I cherry-picked your commit at PR #5101

PR #5108, opened 2021-01-08
 [Downgrade PHP 7.4] Skip DowngradeNumericLiteralSeparatorRector on value contains + char

### LINKED ISSUE #4401: DowngradeNumericLiteralSeparatorRector must not convert 1.0 to 1
# Bug Report

| Rector version | v0.8.21       |

Running rector `DowngradeNumericLiteralSeparatorRector` (to replace `1_000` with `1000`) on an `int` such as `1.0` currently replaces it with `1`:

```diff
-if (0 > $beta = $beta ?? 1.0) {
+if (0 > $beta = $beta ?? 1) {
```

It must not.

https://getrector.org/demo/715a5acd-2f03-4db3-a7f2-d9e0a5d02ba5#result

## Expected Behaviour

Do not replace `1.0` with `1`, keep it as it is.

samsonasik:
Fixes #4401

### DIFF
--- a/rules/downgrade-php74/src/Rector/LNumber/DowngradeNumericLiteralSeparatorRector.php
+++ b/rules/downgrade-php74/src/Rector/LNumber/DowngradeNumericLiteralSeparatorRector.php
@@ -68,7 +68,9 @@ CODE_SAMPLE
             return null;
         }
 
-        if ($node->value === PHP_FLOAT_MAX) {
+        $numberNode      = clone $node;
+        $numberNodeValue = (string) $numberNode->value;
+        if (strpos($numberNodeValue, '+') !== false) {
             return null;
         }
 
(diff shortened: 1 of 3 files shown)


leoloso:
This solves the specific case of the max float, but it doesn't fix the underlying problem: that the rule should not be applied on numbers using the scientific notation. However, that cannot be fixed currently.

leoloso (approved):

samsonasik:
@leoloso do you think it needs check the string version, and verify if the value contains like `+` character?

leoloso:
Exactly. But that can't be done since the number you're dealing with is already a float, and PHP-Parser doesn't preserve the parsed number, where you can find the `+`

samsonasik:
Fixed now 🎉 , the solution is clone the node for tested against `+` so it doesn't use original behaviour.

samsonasik:

TomasVotruba:
Thank you :+1:

PR #5126, opened 2021-01-10
Fix weekly code coverage github action

samsonasik:
it got error at https://github.com/rectorphp/rector/actions/runs/474871156

```

The workflow is not valid. .github/workflows/weekly_code_coverage.yaml (Line: 34, Col: 51): Unrecognized named-value: 'secrets'. Located at position 1 within expression: secrets.GITHUB_TOKEN
```

The `secrets` seems only needed for automatic PR.


samsonasik:
phpstan notice is unrelated.

TomasVotruba:
:+1:

PR #5128, opened 2021-01-10
Another fix weekly code coverage action

samsonasik:
`secrets.GITHUB_TOKEN` should be `secrets.ACCESS_TOKEN` like in other action.


samsonasik:
@TomasVotruba since weekly code coverage action  this doesn't create automatic PR, can we make it daily? we can check if any error early before another next week.

TomasVotruba:
That could be doable, just think about environment. This is not really something crutial for our project, it's just nice to have number.

PR #5129, opened 2021-01-10
[PhpDocParser] re-use nodes from external package

TomasVotruba:


TomasVotruba:
This is not possible untill full package switch

PR #5000, opened 2020-12-26
[mysql-to-mysqli] Added fixture for issue#4977

genesiscz:
I hope I didn't do it completely wrong, if yes, please correct me so my next contribution can be better :-)

https://github.com/rectorphp/rector/issues/4977

PR #5130, opened 2021-01-10
[mysql-to-mysqli] Handle second parameter on mysqli_query swap for sql - conn variable 

### LINKED ISSUE #4977: [mysql-to-mysqli] Not renaming mysql_query to mysqli_query on legacy PHP5 code
https://github.com/rectorphp/rector/blob/9433616c634e3e2513112dd441b8163dc12f6767/config/set/mysql-to-mysqli.php#L39

Here it says it should swap the parameters and then rename the function. mysql_query, mysql_real_escape_string are two that are not being renamed.

On another note, second parameter in most of these functions are optional, which is connection handler. Rector doesn't correctly work in this situation. I understand that this is because of rector not being able to find the $connection, but shouldn't this be stated somewhere in the execution/documentation of the set?

samsonasik:
Close #5000 Fixes #4977


samsonasik:
seems cause another error, I will recheck.

samsonasik:
The issue is because the check is rely on variable named `connection`:

https://github.com/rectorphp/rector/blob/20588b78a08b241b1830867dfbdc24febc1cec4c/rules/mysql-to-mysqli/src/Rector/FuncCall/MysqlQueryMysqlErrorWithLinkRector.php#L176

which it will be unreliable if variable changed to `conn` or `con` or `koneksi`

samsonasik:
fixed 🎉

samsonasik:

TomasVotruba (approved):

TomasVotruba:
Thank you :+1:

PR #5141, opened 2021-01-11
InferParamFromClassMethodReturnRector: Don't fail with empty array

matthiasnoback:
When inferring param from class method return, an empty array should be fine. In the current situation it throws a `ShouldNotHappenException`. This PR returns `null` instead, which solves the issue.

### DIFF
--- a/rules/restoration/src/Type/ConstantReturnToParamTypeConverter.php
+++ b/rules/restoration/src/Type/ConstantReturnToParamTypeConverter.php
@@ -33,7 +33,7 @@ final class ConstantReturnToParamTypeConverter
         return $this->unwrapConstantTypeToObjectType($type);
     }
 
-    private function unwrapConstantTypeToObjectType(Type $type): Type
+    private function unwrapConstantTypeToObjectType(Type $type): ?Type
     {
         if ($type instanceof ConstantArrayType) {
             return $this->unwrapConstantTypeToObjectType($type->getItemType());
@@ -44,11 +44,14 @@ final class ConstantReturnToParamTypeConverter
         if ($type instanceof UnionType) {
             $types = [];
             foreach ($type->getTypes() as $unionedType) {
-                $types[] = $this->unwrapConstantTypeToObjectType($unionedType);
+                $type = $this->unwrapConstantTypeToObjectType($unionedType);
+                if ($type !== null) {
+                    $types[] = $type;
+                }
             }
             return $this->typeFactory->createMixedPassedOrUnionType($types);
         }
 
-        throw new NotImplementedYetException();
+        return null;
     }
 }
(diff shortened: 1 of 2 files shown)


rules/restoration/src/Type/ConstantReturnToParamTypeConverter.php
```diff
@@ -44,11 +44,14 @@ private function unwrapConstantTypeToObjectType(Type $type): Type
...
+                $type = $this->unwrapConstantTypeToObjectType($unionedType);
+                if ($type !== null) {
+                    $types[] = $type;
+                }
             }
             return $this->typeFactory->createMixedPassedOrUnionType($types);
         }
 
-        throw new NotImplementedYetException();
+        return null;
```
TomasVotruba, line 55: For unknown type we use PHPStan convetions of `MixedType`.
Can we apply it here?

rules/restoration/tests/Rector/ClassMethod/InferParamFromClassMethodReturnRector/Fixture/skip_empty_array.php.inc
```diff
@@ -0,0 +1,21 @@
...
+        return [];
+    }
+
+    /**
+     * @param String_ $node
+     */
+    public function process(\PhpParser\Node $node)
+    {
+    }
+}
```
TomasVotruba, line 21: Nice minimalism :) :+1:

rules/restoration/src/Type/ConstantReturnToParamTypeConverter.php
matthiasnoback: I haven't tried that, but my guess is that that will modify the `@param` annotation to contain `mixed`.

rules/restoration/tests/Rector/ClassMethod/InferParamFromClassMethodReturnRector/Fixture/skip_empty_array.php.inc
matthiasnoback: :)

rules/restoration/src/Type/ConstantReturnToParamTypeConverter.php
matthiasnoback: Okay, I just tried that and my suspicion was correct. So I think `null` is better, meaning, there is actually no type to be derived.

rules/restoration/tests/Rector/ClassMethod/InferParamFromClassMethodReturnRector/Fixture/skip_empty_array.php.inc
matthiasnoback: I've just made it more minimal.

PR #5146, opened 2021-01-11
[Downgrade PHP 7.4] Remove self return type

### LINKED ISSUE #5145: [Downgrade PHP 7.4] Remove self return type
# Feature Request

Thanks to [Covariant Returns and Contravariant Parameters](https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters) introduced in PHP 7.4, returning `self` works across different class inheritance and implementing interfaces, while it didn't work before. Check: https://3v4l.org/ttop8.

Hence, for PHP 7.3 and below, returning `self` must be removed.
 
## Diff 

```diff
class A
{
-    public function foo(): self
+    public function foo()
    {
        return $this;
    }
}
```

leoloso:
Fixes #5145


TomasVotruba:
Looks good :+1:

PR #5155, opened 2021-01-12
[DeadCode] Skip when $this->isOpenSourceProjectType() on RemoveUnusedPublicMethodRector

samsonasik:


TomasVotruba:
Could you also add test case for this one? Into new standalone test class.

Just a public method that is not used, but will not be removed

samsonasik:
implemented.

TomasVotruba:
Thanks :+1:

PR #5160, opened 2021-01-12
Split composer modifiers to ValueObjects and services

lulco:


lulco:
@TomasVotruba I rewrote composer modifiers as you suggested. I am not sure if the implementation is what you will like, because I don't like it much O:-) so please comment if I did something wrong or another way you like

TomasVotruba:
Hi, 

I'm not sure this is the goal I wanted to see. I don't like it either :)
I'll have to look into the code to understand it better.

PR #5161, opened 2021-01-12
Added usage of composer modifier for nette upgrade to 3.0

lulco:


lulco:

TomasVotruba:
Thank you :+1:

lulco:
No problem :) I probably just find some bug... But I don't understand it for now... 
Why it is in `/tmp/_rector/Container...` generated file called `getComposerProcessorService` which has this inside:
```
$a = new \Rector\Composer\Modifier\ComposerModifier();
$a->configure([
    0 => new \Rector\Composer\ValueObject\ComposerModifier\ChangePackageVersion('nette/nette', new \Rector\Composer\ValueObject\Version\Version('^3.0')),
]);
```

It throws error `Uncaught TypeError: Argument 2 passed to Rector\Composer\ValueObject\ComposerModifier\ChangePackageVersion::__construct() must be of the type string, object given`

How is the string ^3.0 changed to Version object? Is it some magic?

TomasVotruba:
Intersting :) we should have a test for this too. Could you add it in next PR?

I'd try removing Inliner if it's in it or somewhere else.

lulco:
Removing Inliner throws this error:
```
Cannot use values of type "Rector\Composer\ValueObject\ComposerModifier\ChangePackageVersion" in service configuration files.
```

Changing version back to string helps. Now validation is doing before assign to private property:
```
    /** @var string */
    private $packageName;

    /** @var string */
    private $targetVersion;

    public function __construct(string $packageName, string $targetVersion)
    {
        $this->packageName = $packageName;
        new Version($targetVersion);
        $this->targetVersion = $targetVersion;
    }
```

It works but it is not what you wanted before :)


Also I found out that I cannot use 
```
$services->set(ComposerModifier::class)
        ->call('configure', [
             ...
]);
```
twice (or more). Only one is used.

I tried to use `get` and it works, but I am not sure if it is correct
```
$services->get(ComposerModifier::class)
        ->call('configure', [
            ...
]);
```

Issue #5152, opened 2021-01-12
Bug with dead code: leaveNode() returned invalid value of type integer


brucealdridge:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.9.6              |
| Installed as   |  prefixed Rector                           |

## Minimal PHP Code Causing Issue

https://getrector.org/demo/b55f427c-3b82-4f7f-9246-41c8bdc4d612

~~~
    public function testFind(): void
    {
        assert($user = User::find('foo bar'));
        assert(User::find('foo bar'));
    }
~~~

## Expected Behaviour

~~~
-       assert($user = User::find('foo bar'));
+       assert(User::find('foo bar'));
~~~

samsonasik:
@brucealdridge  could you run with `--debug` to get the exact rule causing this, thank you.

brucealdridge:
not sure it helps

```
⇒  bin/rector process tests/unit/models/UserTest.php --debug
Rector 0.9.6
Config file: rector.php

[parsing] tests/unit/models/UserTest.php
[refactoring] tests/unit/models/UserTest.php
    [applying] Rector\PSR4\Rector\FileWithoutNamespace\NormalizeNamespaceByPSR4ComposerAutoloadRector
    [applying] Rector\CodingStyle\Rector\Use_\SplitGroupedUseImportsRector
    [applying] Rector\DeadCode\Rector\Class_\RemoveUnusedDoctrineEntityMethodAndPropertyRector
    [applying] Rector\PHPUnit\Rector\Class_\AddSeeTestAnnotationRector
    [applying] Rector\DeadCode\Rector\Property\RemoveUnusedPrivatePropertyRector
    [applying] Rector\DeadCode\Rector\Property\RemoveSetterOnlyPropertyAndMethodCallRector
    [applying] Rector\Php74\Rector\Property\RestoreDefaultNullToNullableTypePropertyRector
    [applying] Rector\Php74\Rector\Property\TypedPropertyRector
    [applying] Rector\DeadCode\Rector\PropertyProperty\RemoveNullPropertyInitializationRector
    [applying] Rector\DeadCode\Rector\ClassMethod\RemoveEmptyClassMethodRector
    [applying] Rector\DeadCode\Rector\ClassMethod\RemoveUnusedParameterRector
    [applying] Rector\DeadCode\Rector\FunctionLike\RemoveOverriddenValuesRector
[...]

PR #5165, opened 2021-01-13
[DeadCode] Failing fixture remove unused variable assign in expr

samsonasik:
#5152


samsonasik:
fixed 🎉

brucealdridge:
> fixed 🎉

Thank you! If I understood more how it worked and could get the dev env setup I would start fixing my own bugs!

samsonasik:
After some thinking, I think assignment inside Expr should be skipped, as it may cause side effect.

samsonasik:
I've updated to skip change when parent of assign is not dedicated `Expression` statement

samsonasik:

samsonasik:
rebased.

TomasVotruba:
Thanks :+1:

PR #5167, opened 2021-01-13
fix cs

TomasVotruba:


TomasVotruba:
Replaced by https://github.com/rectorphp/rector/pull/5170

PR #5169, opened 2021-01-13
Fix: do not process composer.json if file doesn’t exist

### LINKED ISSUE #4173: getrector.org/demo broken
# Bug Report

Plain source code with `function_exists()` seems not to work in getrector.org/demo. 

```
<?php
function d() {}

if (!function_exists('d')) {
	function d() {}
}

d();
```
See (not working) getrector.org/demo here:

https://getrector.org/demo/acf76090-c0ab-4e95-83f1-0c7ed7897568#result

## Expected Behaviour

Some output

## Actual Behaviour

500 Internal Server Error

lulco:
resolves #4173


TomasVotruba:
Thanks!

I wasn't sure when you get to this, so I prepared PR few minutes ago :)
  https://github.com/rectorphp/rector/pull/5168

lulco:
yeah, I noticed after creating PR. Sorry for troubles

TomasVotruba:
No problem, the assumption was correct. 
We just use non-standard setup :)

PR #5186, opened 2021-01-14
[Docker] Install libicu-dev to allow install intl

samsonasik:


samsonasik:
@TomasVotruba I verified the build is working locally with install libicu-dev 🎉 

Screen Shot 2021-01-14 at 20 42 24

TomasVotruba:
Thank you

PR #5191, opened 2021-01-15
(fix) RemoveDoubleAssignRector: properly skip if the assigned expression is a call

j2L4e:
RemoveDoubleAssignRector was checking if the _current_ assignment includes a call, rather than the preceding one. This was resulting in wrong removals:

```
$a = echo(''); // was removed before, is skipped now.
$a = 7;
```

### DIFF
--- a/rules/dead-code/src/Rector/Assign/RemoveDoubleAssignRector.php
+++ b/rules/dead-code/src/Rector/Assign/RemoveDoubleAssignRector.php
@@ -80,7 +80,7 @@ CODE_SAMPLE
             return null;
         }
 
-        if ($this->isCall($node->expr)) {
+        if ($this->isCall($previousStatement->expr->expr)) {
             return null;
         }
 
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thank you :+1:

TomasVotruba:
Thank you :+1:

TomasVotruba (approved):

j2L4e:
@TomasVotruba still doesn't work correctly. `$a = '' . foo();` would still be removed

TomasVotruba:
What fix is neede to support it?

j2L4e:
`SideEffectNodeDetector::detect` could generally do it, but the way it detects impure functions seems to be a little off. it detects `array_pop($foo)` as pure for example.

TomasVotruba:
I see. It's not perfect yet. Are you able to extend functionality to fix it?

j2L4e:
#5284

PR #5193, opened 2021-01-15
[phpstan] Remove unreported errors and duplicated methods

TomasVotruba:


samsonasik:
@TomasVotruba the duplicated phpstan error message seems only happen in this PR ? The behaviour autoload seems changed somewhere here

TomasVotruba:
@samsonasik Yes, it happens here.

PR #5194, opened 2021-01-15
[DeadCode] Add RemoveDeadConditionAboveReturnRector

### LINKED ISSUE #4925: [DeadCode] Remove dead condition above return anyway
```diff
-        if ($node instanceof Expr) {
-            return null;
-        }

        return null;
```

samsonasik:
Fixes #4925


rules/dead-code/tests/Rector/Return_/RemoveDeadConditionAboveReturnRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,31 @@
...
+
+class Fixture
+{
+    public function run()
+    {
+        if (1 === 1) {
+            return 'yes';
+        }
+
+        return 'yes';
```
TomasVotruba, line 13: I see. This case should be covered too, right?

```diff
-if (1 === 'hey') {
-    return 'yes';
-}

 return 'yes';
```

samsonasik: the check is not in the cond of if, but on its stmt, so that's already covered.

samsonasik: I added new fixture https://github.com/rectorphp/rector/pull/5194/commits/08a197fef44aa782690191ef7414edc1a5ecb6e0 for different cond

samsonasik:

TomasVotruba:
Thank you :+1:

PR #5195, opened 2021-01-15
Upgrade to Nette 3.1

### LINKED ISSUE #5142: [Nette 3.1] Add upgrade set
See https://forum.nette.org/cs/34080-novinky-v-nette-3-1-je-venku

lulco:
Resolves #5142


config/set/nette-31.php
```diff
@@ -0,0 +1,71 @@
...
+    ]]);
+
+    // TODO not working now because of https://github.com/rectorphp/rector/issues/5171
+    $services->set(ComposerModifier::class)
+        ->call('configure', [
+            ValueObjectInliner::inline([
+                // meta package
+                new ChangePackageVersion('nette/nette', '^3.1'),
+                // https://github.com/nette/nette/blob/v3.0.0/composer.json vs https://github.com/nette/nette/blob/v3.1.0/composer.json
+                new ChangePackageVersion('php', '^7.2'),
```
lulco, line 49: What if I have already some higher php version in composer?

lulco:
@TomasVotruba Hi, I just finished with renaming things. I didn't see any other issues except those I marked as TODOs. Are there any rector rules to cover those or we have to write new rules?

config/set/nette-31.php
```diff
@@ -0,0 +1,130 @@
...
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+use Symplify\SymfonyPhpConfig\ValueObjectInliner;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
+    $services = $containerConfigurator->services();
+
+    $services->set(RenameClassRector::class)->call('configure', [[
+        RenameClassRector::OLD_TO_NEW_CLASSES => [
+            // https://github.com/nette/application/compare/v3.0.7...v3.1.0
+            'Nette\Application\IRouter' => 'Nette\Routing\Router',  // TODO not sure about this, it is not simple rename, Nette\Routing\Router already exists in nette/routing
```
TomasVotruba, line 22: Rector should handle it. Did you came across any bug? 
If so, we'll need a test fixture to fix it.

config/set/nette-31.php
```diff
@@ -0,0 +1,130 @@
...
+
+    $services->set(RenameStaticMethodRector::class)
+        ->call('configure', [[
+            RenameStaticMethodRector::OLD_TO_NEW_METHODS_BY_CLASSES => ValueObjectInliner::inline([
+                // https://github.com/nette/utils/commit/8a4b795acd00f3f6754c28a73a7e776b60350c34
+                new RenameStaticMethod('Nette\Utils\Callback', 'closure', 'Closure', 'fromCallable'),
+            ]),
+        ]]);
+
+    // TODO change $router[] = new Router() to $router->addRoute() because of deprecated flags
```
TomasVotruba, line 93: Something like `DimFetchAssignToMethodCallRector`? :) 

We don't have such a rule yet, but should be easy to add.
This might be a good start: https://github.com/rectorphp/rector/blob/master/rules/magic-disclosure/src/Rector/Assign/GetAndSetToMethodCallRector.php

TomasVotruba, line 93: Something like `DimFetchAssignToMethodCallRector`? :) 

We don't have such a rule yet, but should be easy to add.
This might be a good start: https://github.com/rectorphp/rector/blob/master/rules/magic-disclosure/src/Rector/Assign/GetAndSetToMethodCallRector.php

config/set/nette-31.php
```diff
@@ -0,0 +1,130 @@
...
+
+    // TODO Presenter->getContext() and Presenter->context is deprecated
```
TomasVotruba, line 95: We have a similar rule to handle Symfony `container`/`getContainer()`

It's here: https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#containergettoconstructorinjectionrector

With slight modification, it should handle context too.

TomasVotruba:
Good job with use of generic rules :+1:  lot of case are covered now

config/set/nette-31.php
lulco: I thought it will change also usage of Nette\Application\IRouter in use and there could be two same uses:
```php
<?php

use Nette\Application\IRouter;
use Nette\Routing\Router;

class RouterTest
{
    public static function createRouter()
    {
        $router = new Router();
        $router->addRoute('a', 'b', IRouter::ONE_WAY);
        return $router;
    }
}
```
but it only replace usage in code, so no problem :)

config/set/nette-31.php
lulco: Tomorrow I will try to create something :)

config/set/nette-31.php
lulco: I found this :) https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#contextgetbytypetoconstructorinjectionrector

TomasVotruba: Even better :)

config/set/nette-31.php
lulco: I didn't write generic rule for now. Just one for this use case inspired by ArrayAccessSetControlToAddComponentMethodCallRector

lulco:
@TomasVotruba I think it is ready for review. Code Analysis PHP 8.0 / PHPStan is failing but I believe it is not connected to this PR

rules/nette/src/Rector/Assign/ArrayAccessAddRouteToAddRouteMethodCallRector.php
```diff
@@ -0,0 +1,103 @@
...
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\Variable;
+use Rector\Core\Rector\AbstractRector;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @see \Rector\Nette\Tests\Rector\Assign\ArrayAccessAddRouteToAddRouteMethodCallRector\ArrayAccessAddRouteToAddRouteMethodCallRectorTest
+ */
+final class ArrayAccessAddRouteToAddRouteMethodCallRector extends AbstractRector
```
TomasVotruba, line 19:
```suggestion
final class ArrayAccessAddRouteToAddRouteMethodCallRector extends AbstractRector implements \Rector\Core\Contract\Rector\ConfigurableRectorInterface
```

rules/nette/src/Rector/Assign/ArrayAccessAddRouteToAddRouteMethodCallRector.php
```diff
@@ -0,0 +1,103 @@
...
+{
+    public function getRuleDefinition(): RuleDefinition
+    {
+        return new RuleDefinition(
+            'Change magic array access add Route, to explicit $routeList->addRoute(...) method',
+            [
+                new CodeSample(
```
TomasVotruba, line 26:
```suggestion
                new \Symplify\RuleDocGenerator\ValueObject\CodeSample\ConfiguredCodeSample(
```

rules/nette/src/Rector/Assign/ArrayAccessAddRouteToAddRouteMethodCallRector.php
```diff
@@ -0,0 +1,103 @@
...
+        if (! $this->isObjectType($assign->expr, 'Nette\Application\Routers\Route')) {
+            return false;
+        }
+
+        $arrayDimFetch = $assign->var;
+        if (! $arrayDimFetch->var instanceof Variable) {
+            return false;
+        }
+
+        return $this->isObjectType($arrayDimFetch->var, 'Nette\Application\Routers\RouteList');
```
TomasVotruba, line 101: The class and method should be passed via `configure()` method, so this rule can be used as generic one.

```php
$services->set(ArrayAccessAddRouteToAddRouteMethodCallRector::class)
    ->call('configure', [[
        ArrayAccessAddRouteToAddRouteMethodCallRector::TYPE_TO_METHOD_CALL => [
           'Nette\Application\Routers\Route' => 'addRoute'
        ]
    ]]);
```

Is the second class 'Nette\Application\Routers\RouteList' really neccessary or can we remove it?

TomasVotruba (commented):
Very nice work with the array dim fetch rule :+1:

TomasVotruba:
The PHPStan CI fail is unrelated, I'll look into it

rules/nette/src/Rector/Assign/ArrayAccessAddRouteToAddRouteMethodCallRector.php
lulco: I'm not sure now. I have to go deeper :) there are at least two classes with method addRoute() but I don't know if both can be used in createRouter method...

But let assume that you have router which just adds Routes to some array and then iterates over this array and adds it to RouteList

Something like this:
```
public function createRouter()
{
    $routeArray = [];
    $routeList = new RouteList();

    $routeArray[] = new Route('<module>/<presenter>/<action>[/<id>]', 'Homepage:default');

    foreach ($routeArray as $route) {
        $routeList[] = $route;
    }
    return $routeList;
}
```
With ignoring RouteList, our code will call ->addRoute on $routeArray

lulco:
> Very nice work with the array dim fetch rule 👍

As I wrote, it is inspired by existing rule ArrayAccessSetControlToAddComponentMethodCallRector

rules/nette/src/Rector/Assign/ArrayAccessAddRouteToAddRouteMethodCallRector.php
TomasVotruba: I see. Thanks for the code.
The right part must be checked for `New_` node instance.

lulco: check added

lulco:
So, are we done? :)

rules/nette/src/Rector/Assign/ArrayAccessAddRouteToAddRouteMethodCallRector.php
TomasVotruba: This needs to be resolved. Rule must be configurable

TomasVotruba: This needs to be resolved. Rule must be configurable

TomasVotruba:
This is not resolved yet
https://github.com/rectorphp/rector/pull/5195/files#r559254124

TomasVotruba:
This is not resolved yet
https://github.com/rectorphp/rector/pull/5195/files#r559254124

rules/nette/src/Rector/Assign/ArrayAccessAddRouteToAddRouteMethodCallRector.php
TomasVotruba: Only time will tell. This abstraction is obvious (class and method name), so better do it now when we have focus, then in 6 month create duplicated rule. Yes that happens :)

lulco: OK I will try something. If I have 3+ config strings (listClass, itemClass, listMethod) I will need some ValueObject right?

TomasVotruba: Thank you :+1: 

> I will need some ValueObject right?

Yes, that would be way to go for 3+ values to check

TomasVotruba: Thank you :+1: 

> I will need some ValueObject right?

Yes, that would be way to go for 3+ values to check

lulco: Generic rule is created, but probably some renamings should be done...

rules/generic/src/Rector/Assign/DimFetchAssignToMethodCallRector.php
```diff
@@ -0,0 +1,131 @@
...
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\ConfiguredCodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+use Webmozart\Assert\Assert;
+
+/**
+ * @see \Rector\Generic\Tests\Rector\Assign\DimFetchAssignToMethodCallRector\DimFetchAssignToMethodCallRectorTest
+ */
+final class DimFetchAssignToMethodCallRector extends AbstractRector implements ConfigurableRectorInterface
+{
+    public const DIM_FETCH_ASSIGN_TO_METHOD_CALL = 'dim_fetch_assign_to_method_call';
```
lulco, line 25: what about just ASSIGN_TO_METHOD_CALL?

lulco:
Code Analysis PHP 8.0 / PHPStan failed again

rules/generic/src/Rector/Assign/DimFetchAssignToMethodCallRector.php
TomasVotruba: Assign to method call is this:

```php
$this->someMethod() = 1000;
```

Assign to dim fetch is:

```php
$values[] = 1000;
```

New assing to dim fetch (case for this rule) is:

```php
$values[] = new SomeObject()
```

TomasVotruba: I re-checked the rule now. Maybe `DimFetchAssignToMethodCall` would be better

lulco: that's its name now :)

lulco:
@TomasVotruba I merged master into branch and build is green again. Generic rule is done. Is there anything you want to add / change in this PR?

rules/generic/src/Rector/Assign/DimFetchAssignToMethodCallRector.php
TomasVotruba: Then it's verified :D

TomasVotruba:
Looks very good, thanks for your detailed work :+1:

TomasVotruba (approved):

PR #5175, opened 2021-01-13
ValueObjectInliner failing test case

lulco:

PR #5196, opened 2021-01-15
[Composer] Fix value object inliner with Version

TomasVotruba:
Mirror of https://github.com/rectorphp/rector/pull/5175, with fix


rules/composer/src/ValueObject/ComposerModifier/AddPackageToRequire.php
```diff
@@ -20,23 +19,18 @@ final class AddPackageToRequire implements ComposerModifierInterface
...
 
     /**
-     * @var Version
+     * @var string
      */
     private $version;
 
-    /**
-     * @param string $packageName name of package (vendor/package)
-     * @param string $version target package version (1.2.3, ^1.2, ~1.2.3 etc.)
```
lulco, line 29: Do you really want to remove this?

rules/composer/src/ValueObject/ComposerModifier/AddPackageToRequireDev.php
```diff
@@ -20,23 +19,18 @@ final class AddPackageToRequireDev implements ComposerModifierInterface
...
     private $version;
 
-    /**
-     * @param string $packageName name of package (vendor/package)
-     * @param string $version target package version (1.2.3, ^1.2, ~1.2.3 etc.)
-     */
     public function __construct(string $packageName, string $version)
     {
         $this->packageName = $packageName;
-        $this->version = new Version($version);
```
lulco, line 34: now there is no validation, what about at least call `new Version($version);` at first line to validate version string?

lulco (commented):
The final version will look like one of my version 7 days ago :P

rules/composer/tests/Processor/Fixture/composer_before.json
```diff
@@ -0,0 +1,5 @@
+{
+    "require": {
+        "nette/nette": "^3.0"
```
lulco, line 3: this should be 2.4 here

rules/composer/src/ValueObject/ComposerModifier/AddPackageToRequire.php
TomasVotruba: Yea, these are obvious comments that do not any value.

rules/composer/src/ValueObject/ComposerModifier/AddPackageToRequireDev.php
TomasVotruba: WIP :) validation should be in 1 place.

rules/composer/tests/Processor/Fixture/composer_before.json
TomasVotruba: Good catch! Already changed in next PR.

TomasVotruba:
> The final version will look like one of my version 7 days ago :P

It's possible :D give me an hour

PR #5198, opened 2021-01-16
Refactor composer modifier to ComposerRector

TomasVotruba:
Closes https://github.com/rectorphp/rector/pull/5160


TomasVotruba:
@lulco This is the idea I failed to share :) 

To be continue, good night

rules/composer/src/ComposerModifier.php
```diff
@@ -0,0 +1,51 @@
...
+
+use Rector\Composer\Contract\Rector\ComposerRectorInterface;
+use Symplify\ComposerJsonManipulator\ValueObject\ComposerJson;
+
+final class ComposerModifier
+{
+    /**
+     * @var string
+     */
+    private const COMPOSER_UPDATE = 'composer update';
```
lulco, line 15: There is no need to have it here with getter if it is constant. In my code it was possible to change the executed command via setter - if somebody need to run it with parameters or use composer.phar etc

config/set/nette-30-composer.php
```diff
@@ -0,0 +1,65 @@
...
+use Rector\Composer\Rector\ReplacePackageAndVersionRector;
+use Rector\Composer\ValueObject\PackageAndVersion;
+use Rector\Composer\ValueObject\ReplacePackageAndVersion;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+use Symplify\SymfonyPhpConfig\ValueObjectInliner;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
+    $services = $containerConfigurator->services();
+
+    $services->set(ChangePackageVersionRector::class)
```
lulco, line 16: This was something I didn’t want to do because somewhere it my head was warning which said I need to know exact order of modifications (AddPackage -> RemovePackage has different output like RemovePackage ->AddPackage if we are changing same package). But maybe there is no real world use case for it

config/set/nette-30-composer.php
```diff
@@ -0,0 +1,65 @@
...
+        ->call('configure', [[
+            ChangePackageVersionRector::PACKAGES_AND_VERSIONS => ValueObjectInliner::inline([
```
lulco, line 18: Did you test it with multiple sets? For example nette30 + php7.1 (which change php to ^7.1)? Did not work for me in old code.

rules/composer/src/ComposerModifier.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public function getCommand(): string
+    {
+        return self::COMPOSER_UPDATE;
+    }
+
+    public function modify(ComposerJson $composerJson): void
+    {
+        foreach ($this->composerRectors as $composerChanger) {
```
lulco, line 47: composerChanger -> composerRector

config/set/nette-30-composer.php
```diff
@@ -0,0 +1,65 @@
...
+                new PackageAndVersion('contributte/event-dispatcher-extra', '^0.8'),
+                new PackageAndVersion('contributte/forms-multiplier', '3.1.x-dev'),
+                // other packages
+                new PackageAndVersion('radekdostal/nette-datetimepicker', '^3.0'),
+            ]),
+        ]]);
+
+    $services->set(RemovePackageRector::class)
+        ->call('configure', [[
+            RemovePackageRector::PACKAGE_NAMES => ['nette/deprecated', 'nette/reflection'],
```
lulco, line 52: This is little bit incosistent with other types - here is only array of strings of package names, other types has objects and value object inline

rules/composer/src/ComposerModifier.php
TomasVotruba: It was used only in tests. This should be configured on parameter level and documented, not on service setter level.

config/set/nette-30-composer.php
TomasVotruba: If there is a conflict, it must be resolved on logical level. As PHP rules are.

config/set/nette-30-composer.php
TomasVotruba: Not sure what you mean. Test case would help

rules/composer/src/ComposerModifier.php
TomasVotruba: Ref https://github.com/rectorphp/rector/issues/5203

config/set/nette-30-composer.php
TomasVotruba: Indeed. We use a convention:

- arrays for for simple types, like string array or list
- value objects for 2+ items, like class name and method name and potision 

See e.g. https://github.com/rectorphp/rector/blob/4b887f863f44dca93889e96b5d59c4bf993dc228/config/set/nette-30.php#L64-L73

config/set/nette-30-composer.php
TomasVotruba: The best practise is to use one set at a time during big upgrade.
Only then add multiple sets to CI.

TomasVotruba: The best practise is to use one set at a time during big upgrade.
Only then add multiple sets to CI.

Issue #5205, opened 2021-01-16
Incorrect behavior of ReturnTypeDeclarationRector


genesiscz:
# Bug Report

| Subject        | Details              |
| :------------- | :--------------------|
| Rector version | 0.8.56, demo ver ?     |
| Installed as   | composer dependency  |

## Minimal PHP Code Causing Issue

See https://getrector.org/demo/9cba5a13-fa55-4f06-880a-ccdd8259db78

```php
<?php

class DemoFile
{
    public function run()
    {
        throw new Exception("not implemented");
    }
}

class DemoFile2 extends DemoFile {
    
    public function run() : Baba {
       return new Baba();        
    }
}

class Baba {
}
```

Rector diff:
```patch
 
 class DemoFile2 extends DemoFile {
     
-    public function run() : Baba {
+    public function run() : void {
        return new Baba();        
     }
 }
 
 class Baba {
}
```
### Responsible rules

* `ReturnTypeDeclarationRector`

## Expected Behavior

Should be skipped.

In my own code it has a little bit different behaviour for some reason: it does add `: void` to the throwing function as a return type, which makes the code screaming about compatibility between the overriden function.

genesiscz:
Ok I got the first part of the issue narrowed down. I had app/model/XXAuthenticator.php files while having App\Authenticators namespace not properly mapped in PSR-4. Sorry for that. But after I made it right, I am getting the following

```
Rector 0.8.56
Config file: rector.php

 28/28 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

7 files with changes
====================

1) app/Authenticators/BaseAuthenticator.php

    ---------- begin diff ----------
--- Original
+++ New
@@ -44,7 +44,7 @@
        /**
         * @param string ...$credentials
         */
-       public function login(...$credentials)
+       public function login(...$credentials): void
        {
                throw new Security\AuthenticationException();
        }
    ----------- end diff -----------


Applied rules:

 * Rector\TypeDeclaration\Rector\FunctionLike\ReturnTypeDeclarationRector


2) app/Authenticators/FacebookAuthenticator.php

    ---------- begin diff ----------
--- Original
+++ New
@@ -7,7 +7,7 @@
 class FacebookAuthenticator extends BaseAuthenticator
 {

-       public function login(...$credentials): Identity
+       public function login(...$credentials): void
        {
[...]

PR #5209, opened 2021-01-16
Issue #5205 fixture

genesiscz:


rules/type-declaration/tests/Rector/FunctionLike/ReturnTypeDeclarationRector/Fixture/respect_children_return_type.php.inc
```diff
@@ -0,0 +1,21 @@
...
+
+class RespectChildrenReturnType2 extends RespectChildrenReturnType {    
+    public function run() : Baba {
+       return new Baba();        
+    }
+}
+
+class Baba {
+}
+?>
```
TomasVotruba, line 21: Same here, here only the tested class should be present.
The rest should be moved to `/Source` directory.


If you look at other tests, the structure is following:

```bash
SomeTest.php // the test
/Fixture // directory with tested files, that should or should not be refactored
/Source // all other files, that might be needed, simulates /vendor in normal project
/config // if config is needed
```

genesiscz: This one is fine just like this I think. I have BaseAuthenticator throwing an exception and children which returns some class. Rector makes the parent return void and then changes the return type of the other one to void too.

samsonasik:
@genesiscz I cherry-picked your commit at PR https://github.com/rectorphp/rector/pull/5404

PR #5213, opened 2021-01-17
[CodeCoverage] Fixes weekly code coverage by move COVERALLS_REPO_TOKEN on run

samsonasik:
It previously got error:

```bash
The workflow is not valid. .github/workflows/weekly_code_coverage.yaml (Line: 34, Col: 51): Unrecognized named-value: 'secrets'. Located at position 1 within expression: secrets.ACCESS_TOKEN
```

at https://github.com/rectorphp/rector/actions/runs/490896171 . This change moved COVERALLS_REPO_TOKEN env definition on `run` part.


samsonasik:

PR #5214, opened 2021-01-17
[TypeDeclaration] Add FlipTypeControlToUseExclusiveTypeRector

### LINKED ISSUE #5137: [TypeDeclaration] Flip type control to use exclusive type
```diff
-        /** @var PhpDocInfo|null $phpDocInfo */
        $phpDocInfo = $functionLike->getAttribute(AttributeKey::PHP_DOC_INFO);
-        if ($phpDocInfo === null) {
+        if (! $phpDocInfo instanceof PhpDocInfo) {
            return;
        }
```

samsonasik:
Fixes #5137

### DIFF
--- a/rules/type-declaration/src/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector.php
+++ b/rules/type-declaration/src/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector.php
@@ -16,9 +16,11 @@ use Rector\Core\Rector\AbstractRector;
 use Rector\NodeTypeResolver\Node\AttributeKey;
 use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
 use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+use PHPStan\PhpDocParser\Ast\PhpDoc\VarTagValueNode;
+use Rector\AttributeAwarePhpDoc\Ast\Type\AttributeAwareIdentifierTypeNode;
 
 /**
- * @see \Rector\TypeDeclaration\Tests\Rector\Assign\FlipTypeControlToUseExclusiveTypeRector\FlipTypeControlToUseExclusiveTypeRectorTest
+ * @see \Rector\TypeDeclaration\Tests\Rector\Identical\FlipTypeControlToUseExclusiveTypeRector\FlipTypeControlToUseExclusiveTypeRectorTest
  */
 final class FlipTypeControlToUseExclusiveTypeRector extends AbstractRector
 {
@@ -81,6 +83,7 @@ CODE_SAMPLE
             ? $node->right
             : $node->left;
 
+        /** @var Assign|null $assign */
         $assign = $this->betterNodeFinder->findFirstPrevious($node, function (Node $node) use ($variable): bool {
             if (! $node instanceof Assign) {
                 return false;
@@ -88,6 +91,10 @@ CODE_SAMPLE
 
             return $this->areNodesEqual($node->var, $variable);
         });
+
+        if (! $assign instanceof Assign) {
+            return null;
+        }
 
         $expression = $assign->getAttribute(Attributekey::PARENT_NODE);
         if (! $expression instanceof Expression) {
@@ -99,7 +106,12 @@ CODE_SAMPLE
             return null;
         }
 
+        /** @var VarTagValueNode|null $tagValueNode */
         $tagValueNode = $phpDocInfo->getVarTagValueNode();
+        if (! $tagValueNode instanceof VarTagValueNode) {
+            return null;
+        }
+
         if (! $tagValueNode->type instanceof AttributeAwareUnionTypeNode) {
             return null;
         }
@@ -108,17 +120,25 @@ CODE_SAMPLE
             return null;
         }
 
-        if ($tagValueNode->type->types[0]->name === $tagValueNode->type->types[1]->name) {
+        /** @var AttributeAwareIdentifierTypeNode[] $types */
+        $types = $tagValueNode->type->types;
+        foreach ($tagValueNode->type->types as $type) {
+            if (! $type instanceof AttributeAwareIdentifierTypeNode) {
+                return null;
+            }
+        }
+
+        if ($types[0]->name === $types[1]->name) {
             return null;
         }
 
-        if ($tagValueNode->type->types[0]->name !== 'null' && $tagValueNode->type->types[1]->name !== 'null') {
+        if ($types[0]->name !== 'null' && $types[1]->name !== 'null') {
             return null;
         }
 
-        $type = $tagValueNode->type->types[0]->name === 'null'
-            ? $tagValueNode->type->types[1]->name
-            : $tagValueNode->type->types[0]->name;
+        $type = $types[0]->name === 'null'
+            ? $types[1]->name
+            : $types[0]->name;
 
         if (class_exists($type) || interface_exists($type)) {
             return new Instanceof_($variable, new FullyQualified($type));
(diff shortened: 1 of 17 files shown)


samsonasik:

rules/type-declaration/src/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector.php
```diff
@@ -0,0 +1,186 @@
...
+CODE_SAMPLE
+                    ,
+                    <<<'CODE_SAMPLE'
+class SomeClass
+{
+    public function __construct(array $values)
+    {
+        /** @var PhpDocInfo|null $phpDocInfo */
+        $phpDocInfo = $functionLike->getAttribute(AttributeKey::PHP_DOC_INFO);
+        if ($phpDocInfo instanceof PhpDocInfo) {
```
TomasVotruba, line 57:
```suggestion
        if (! $phpDocInfo instanceof PhpDocInfo) {
```

TomasVotruba, line 57:
```suggestion
        if (! $phpDocInfo instanceof PhpDocInfo) {
```

rules/type-declaration/tests/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,51 @@
...
+        $stdClass = $this->getStdClass();
+        if ($stdClass === null) {
+            return;
+        }
+
+        /** @var null|stdClass $stdClass */
+        $stdClass = $this->getStdClass();
+        if ($stdClass === null) {
+            return;
+        }
```
TomasVotruba, line 21: Oh, I see :). I probably described incorrectly.

It's not about flipping the type in docs. It's rather about flipping the condition of comparison.

Why? In this case, if the code continues, the `$stdClass` is anything but `null`:

```php
 if ($stdClass === null) {
    return;
 }

 // what is stdClass exactly?
```

But in this case, we know it's exactly one type - `SomeType`:

```php
 if (! $stdClass instanceof SomeType) {
    return;
 }

 // we know $stdClass is SomeType
```

So the idea is to converted `=== null` comparison to `instanceof` checks.

```diff
-if ($stdClass === null) {
+if (! $stdClass instanceof SomeType) {
    return;
 }
```

That way code is more robus and we don't have to relly on doblocks above. So they can be droped too.


```diff
-/** @var SomeType|null $stdClass */
-if ($stdClass === null) {
+if (! $stdClass instanceof SomeType) {
    return;
 }
```

Let me know if you need more examples.

rules/type-declaration/src/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector.php
samsonasik: fixed.

rules/type-declaration/tests/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector/Fixture/fixture.php.inc
samsonasik: implemented.

samsonasik: if it not rely on types in the doc, the following should seems still need to be skipped as in the fixture:

```php
        /** @var stdClass|null|Foo|Bar $stdClass */
        $stdClass = $this->getStdClass();
        if ($stdClass === null) {
            return;
        }
 ```

**Comment by samsonasik** [comment]

File: rules/type-declaration/tests/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector/Fixture/fixture.php.inc (hunk 2 above)
Comment (TomasVotruba) [reply rules/type-declaration/tests/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector/Fixture/fixture.php.inc:21]: > the following should seems still need to be skipped as in the fixture:

Agreed :+1: 

It should keep only 1 type + `null`

Comment (TomasVotruba) [reply rules/type-declaration/tests/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector/Fixture/fixture.php.inc:21]: > the following should seems still need to be skipped as in the fixture:

Agreed :+1: 

It should keep only 1 type + `null`, other should skip

File: rules/type-declaration/src/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector.php [hunk 3]
```diff
@@ -0,0 +1,186 @@
...
+     */
+    private function processConvertToExclusiveType(array $types, Expr $expr, PhpDocInfo $phpDocInfo): ?BooleanNot
+    {
+        $type = $types[0]->name === 'null'
+            ? $types[1]->name
+            : $types[0]->name;
+
+        if (! class_exists($type) && ! interface_exists($type)) {
+            return null;
+        }
```
TomasVotruba, line 123: Instead of these checks, the direct object API should be used.

E.g . `$type instanceof IdentifierTypeNode` and similar.

What is the goal here?

samsonasik: to ensure that the type is a FQCN type that exists, I just know about `IdentifierTypeNode`, I updated it.

samsonasik:

rules/type-declaration/src/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector.php
samsonasik: @TomasVotruba it cause failure on string type, eg:

```php
        /** @var string|null $string */
        $string = $this->getString();
        if ($string === null) {
            return;
        }
```

see https://github.com/rectorphp/rector/pull/5214/commits/735ea9e6796955b6c6af50e4ed2f7cd674487a46

samsonasik: I reverted https://github.com/rectorphp/rector/pull/5214/commits/aa046dff740b621e881d27e3a51e3a3a3c73a59c as it cause issue on https://github.com/rectorphp/rector/pull/5214/commits/735ea9e6796955b6c6af50e4ed2f7cd674487a46

samsonasik:

rules/type-declaration/tests/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector/Fixture/skip_no_doc.php.inc
```diff
@@ -0,0 +1,18 @@
...
+namespace Rector\TypeDeclaration\Tests\Rector\Identical\FlipTypeControlToUseExclusiveTypeRector\Fixture;
+
+use stdClass;
+
+class SkipNoDoc
+{
+    public function run()
+    {
+        $stdClass = $this->getStdClass();
+        if ($stdClass === null) {
```
TomasVotruba, line 12: This case should be covered too.
If `getStdClass()` returns `null|SomeObject` it should be processed the same way.

I'll merge it as it is, so you can add in next PR with less code to worry about

TomasVotruba:
Thank you :+1:

rules/type-declaration/tests/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector/Fixture/skip_no_doc.php.inc
samsonasik: I will try.

johanadivare:
Something worth to note is that the performance of `null` checking is a bit faster see: https://gist.github.com/KVytyagov/57509333daaac69eb980777c7d28ea18

PR #5223, opened 2021-01-18
[TypeDeclaration] Register FlipTypeControlToUseExclusiveTypeRector to type-declaration set

samsonasik:

### DIFF
--- a/rules/naming/src/VariableRenamer.php
+++ b/rules/naming/src/VariableRenamer.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Rector\Naming;
 
-use Closure;
 use PhpParser\Node;
 use PhpParser\Node\Expr\Assign;
 use PhpParser\Node\Expr\Closure;
--- a/rules/naming/src/VariableRenamer.php
+++ b/rules/naming/src/VariableRenamer.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Rector\Naming;
 
+use Closure;
 use PhpParser\Node;
 use PhpParser\Node\Expr\Assign;
 use PhpParser\Node\Expr\Closure;
@@ -94,9 +95,8 @@ final class VariableRenamer
 
     private function isParamInParentFunction(Variable $variable): bool
     {
-        /** @var Closure|null $closure */
         $closure = $variable->getAttribute(AttributeKey::CLOSURE_NODE);
-        if ($closure === null) {
+        if (! $closure instanceof Closure) {
             return false;
         }
 
--- a/rules/nette-code-quality/src/NodeAdding/FunctionLikeFirstLevelStatementResolver.php
+++ b/rules/nette-code-quality/src/NodeAdding/FunctionLikeFirstLevelStatementResolver.php
@@ -57,9 +57,8 @@ final class FunctionLikeFirstLevelStatementResolver
      */
     private function matchMultiplierClosure(Node $node): ?Closure
     {
-        /** @var Closure|null $closure */
         $closure = $node->getAttribute(AttributeKey::CLOSURE_NODE);
-        if ($closure === null) {
+        if (! $closure instanceof \Closure) {
             return null;
         }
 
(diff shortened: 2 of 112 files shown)


rules/naming/src/VariableRenamer.php
```diff
@@ -4,6 +4,7 @@
 
 namespace Rector\Naming;
 
+use Closure;
```
samsonasik, line 7: this is false positive as there is already `use PhpParser\Node\Expr\Closure;` means `\Closure` is invalid.

rules/nette-code-quality/src/NodeAdding/FunctionLikeFirstLevelStatementResolver.php
```diff
@@ -57,9 +57,8 @@ public function resolveFirstLevelStatement(Node $node): Node
      */
     private function matchMultiplierClosure(Node $node): ?Closure
     {
-        /** @var Closure|null $closure */
         $closure = $node->getAttribute(AttributeKey::CLOSURE_NODE);
-        if ($closure === null) {
+        if (! $closure instanceof \Closure) {
```
samsonasik, line 61: this is false positive as there is already use PhpParser\Node\Expr\Closure; means \Closure is invalid.

rules/naming/src/VariableRenamer.php
samsonasik: fixed.

rules/nette-code-quality/src/NodeAdding/FunctionLikeFirstLevelStatementResolver.php
samsonasik: fixed.

samsonasik:

rules/type-declaration/src/Rector/Identical/FlipTypeControlToUseExclusiveTypeRector.php
```diff
@@ -125,7 +125,7 @@ private function processConvertToExclusiveType(array $types, Expr $expr, PhpDocI
         /** @var VarTagValueNode $tagValueNode */
         $tagValueNode = $phpDocInfo->getVarTagValueNode();
         $phpDocInfo->removeTagValueNodeFromNode($tagValueNode);
-        return new BooleanNot(new Instanceof_($expr, new FullyQualified($type)));
+        return new BooleanNot(new Instanceof_($expr, new Name($type)));
```
TomasVotruba, line 128: This should work with FQN, as we use it everywhere in Rector by convention.

The name resolution of `Closure` must be fixed above, probably in `$type`.

TomasVotruba, line 128: This should work with FQN, as we use it everywhere in Rector by convention.

The name resolution of `Closure` must be fixed above, probably in `$type`.

TomasVotruba, line 128: This should work with FQN, as we use it everywhere in Rector by convention.

The name resolution of `Closure` must be fixed above, probably in `$type`.

samsonasik: The type name returns actual value like in the doc, eg:

```php
@var stdClass|null
``` 

The types[0] will be 'stdClass'

and for :


```php
@var \stdClass|null
``` 

The types[0] will be '\stdClass'

any specific type converter for it?

TomasVotruba: I think `StaticTypeMapper` would do here.


Proably `mapPHPStanPhpDocTypeToPHPStanType()` method

samsonasik: implemented.

samsonasik:

TomasVotruba:
Thank you, this is a huge improvement in our coding :+1:

TomasVotruba (approved):

PR #5225, opened 2021-01-18
[PHPStan] Fix duplicate report PreventDuplicateClassMethodRule

samsonasik:


samsonasik:
@TomasVotruba the `PreventDuplicateClassMethodRule` already in symplify static rules config, so re-register it cause a duplicate report. 

This patch should is fix the issue finally 🎉

samsonasik:

TomasVotruba:
Ah, there it is :clap:  Thanks!

I wish @phpstan would report this by default.

PR #5236, opened 2021-01-19
Fix parsing JoinTable without table name explicitly set

JarJak:
See https://github.com/rectorphp/rector-prefixed/pull/12


TomasVotruba:
Thanks :+1:

PR #5237, opened 2021-01-19
Fix match without return

JarJak:
I am getting left with something like:

```php
function x($value)
{
	match $value { // notice missing braces
		1 => y($value);
		default => z($value);
	}
}
```


TomasVotruba:
The class name must be the same in both halfs

TomasVotruba:
Thanks!

The CI is passing, so the fixture is correct. Where seems to be an issue?

TomasVotruba (approved):

PR #5255, opened 2021-01-20
Support string tags for Symfony ServiceMapProvider

pierredup:
For some services (E.G Workflow definitions), the tag has a string value instead of an array with the attributes. This causes the error `Cannot unset string offsets` when running Rector.

An example of such a tag is

```xml
<tag name="application" type="state_machine">workflow.definition</tag>
```

When the xml element gets converted, the `$tag` variable just contains the string `workflow.definition`. So this PR helps to cater for these scenarios and not cause an error

### DIFF
--- a/rules/symfony/src/ServiceMapProvider.php
+++ b/rules/symfony/src/ServiceMapProvider.php
@@ -234,6 +234,12 @@ final class ServiceMapProvider
     {
         $tagValueObjects = [];
         foreach ($tagsData as $key => $tag) {
+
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Hi,

thanks for the feature. As this is very sensitive part of Rector that uses 3rd party config, we'll need a test case to cover it.
Symfony can change anytime, so be better ready for it.

TomasVotruba:
Let's give it a try then

PR #5261, opened 2021-01-20
extend Nette 3.1 set

TomasVotruba:


config/set/nette-31.php
```diff
@@ -82,6 +89,7 @@
             new MethodCallRename('Nette\Forms\Container', 'addImage', 'addImageButton'),
             // https://github.com/nette/utils/commit/d0427c1811462dbb6c503143eabe5478b26685f7
             new MethodCallRename('Nette\Utils\Arrays', 'searchKey', 'getKeyOffset'),
+            new MethodCallRename('Nette\Configurator', 'addParameters', 'addStaticParameters'),
```
lulco, line 92: hm this is just an alias, addParameters is not marked as deprecated... I would like to keep using addParameters :)

TomasVotruba: It's more explicit way to go:

```php
	/**
	 * Alias for addStaticParameters()
	 * @return static
	 */
	public function addParameters(array $params)
	{
		return $this->addStaticParameters($params);
	}
```

The idea is to separate it from dynamic parameters. 

`addStaticParameters()` name describes its job better.

Also promoted in tests: https://github.com/nette/bootstrap/commit/019ae17d405d9c1425562592810289c812a3827d#diff-b67a6103a0abb927e6e8b973037c5747cbd81e5445ac637301f8cf03c501b487

TomasVotruba: Upgrades are not only about fixing deprecated features, but rather using the topmost tier that new PHP or packages allows. E.g. typed properties are not the only way that works in PHP 7.4, the old is still supported, but the set upgrades, so devs can enjoy the versoin there is :)

PR #5262, opened 2021-01-20
Add failing test for importing function names

pierredup:
When setting `auto_import_names` to true and using the `performance` set, function names are imported as class names, instead of with `use function`

Demo:

https://getrector.org/demo/c22d225a-4828-4f47-9b4b-b21f251ade7a

**Example:**

```php
namespace Test;

class SomeClass
{
    public function shorten($value)
    {
        return trim($value);
    }
}
```

**Expected:**

```diff
namespace Test;

+use function trim;
+
 class SomeClass
 {
     public function shorten($value)
     {
        return trim($value);
     }
 }
```

**Actual:**

```diff
namespace Test;

+ use trim;
+
 class SomeClass
 {
     public function shorten($value)
     {
        return trim($value);
     }
 }
```

### DIFF
--- a/rules/performance/tests/Rector/FuncCall/PreslashSimpleFunctionRector/Fixture/auto_import_names.php.inc
+++ b/rules/performance/tests/Rector/FuncCall/PreslashSimpleFunctionRector/Fixture/auto_import_names.php.inc
@@ -1,0 +1,29 @@
+<?php
+
+namespace Rector\Performance\Tests\Rector\FuncCall\PreslashSimpleFunctionRector\Fixture;
+
+class AutoImportNames
+{
+    public function shorten($value)
+    {
+        return trim($value);
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\Performance\Tests\Rector\FuncCall\PreslashSimpleFunctionRector\Fixture;
+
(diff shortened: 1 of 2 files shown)


rules/performance/tests/Rector/FuncCall/PreslashSimpleFunctionRector/Fixture/auto_import_names.php.inc
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Rector\Performance\Tests\Rector\FuncCall\PreslashSimpleFunctionRector\Fixture;
+
+class SomeClass
```
samsonasik, line 5: class name should be `AutoImportNames`

samsonasik:
@TomasVotruba is there a `use function` adder functionality for it?

samsonasik:
@TomasVotruba the `PreslashSimpleFunctionRector` seems should skip import function even `Option::AUTO_IMPORT_NAMES` set to true as it seems contradiction, what do you think?

samsonasik:
@pierredup I cherry-picked your commit at https://github.com/rectorphp/rector/pull/5268

PR #5263, opened 2021-01-20
NodeFactory: Allow passing a Cast node to createArrayItems

j2L4e:
Allows passing a Cast node to NodeFactory::createArrayItems. Not much to say here.


TomasVotruba:
Well said, thanks :+1:

PR #5264, opened 2021-01-20
Added missing symfony/process to composer.json

lulco:
It was used only as dev dependency so it doesn't appear in build and prefixed-rector. Throws an error:
```
Class 'RectorPrefix20210117\Symfony\Component\Process\Process' not found
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,6 +41,7 @@
         "symfony/dependency-injection": "^5.1",
         "symfony/finder": "^4.4.8|^5.1",
         "symfony/http-kernel": "^4.4.8|^5.1",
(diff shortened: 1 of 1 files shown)


lulco:
I would like to use this in prefixed rector, but I saw you released a version few minutes ago :( When will be released next version?

lulco:
ideally with package sorting fix in composer json manipulator :)

TomasVotruba:
No plan yet. For now you can use `dev-master` version :)

lulco:
:) ok I will try, maybe deliver some other fixes ;)

PR #5268, opened 2021-01-21
[Performance] Handle PreslashSimpleFunctionRector on Option::AUTO_IMPORT_NAMES => true

samsonasik:
Close #5262


samsonasik:

rules/performance/tests/Rector/FuncCall/PreslashSimpleFunctionRector/Fixture/skip_auto_import_names.php.inc
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Rector\Performance\Tests\Rector\FuncCall\PreslashSimpleFunctionRector\Fixture;
+
+class SkipAutoImportNames
+{
+    public function shorten($value)
+    {
+        return trim($value);
```
pierredup, line 9: If functions aren't imported, then this should be changed to `\trim` as the `PreslashSimpleFunctionRector` rule should still run normally

samsonasik: fixed.

samsonasik:

rules/performance/src/Rector/FuncCall/PreslashSimpleFunctionRector.php
```diff
@@ -75,6 +76,11 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
+        $autoImportNames = $this->parameterProvider->provideParameter(Option::AUTO_IMPORT_NAMES);
+        if ($autoImportNames) {
+            $this->parameterProvider->changeParameter(Option::AUTO_IMPORT_NAMES, false);
```
pierredup, line 81: I'm not sure if this approach should be taken here.
This disables an option that a user explicitly enabled, and would cause no class names to be imported.

samsonasik: No, that will still works for import class, I created new proof fixture for it https://github.com/rectorphp/rector/pull/5268/commits/7f10e32e16ac969204643787df60108c3b7bbac2

samsonasik:

rules/performance/src/Rector/FuncCall/PreslashSimpleFunctionRector.php
pierredup: @samsonasik That fixture only tests importing class names, there is no function call, so the `PreslashSimpleFunctionRector` rector never runs. The issue comes in when the both are combined.

Can you please try the following test case (or something similar that uses a function call and a class import):

```php
<?php

namespace Rector\Performance\Tests\Rector\FuncCall\PreslashSimpleFunctionRector\Fixture;

class AutoImportNames
{
    public function run()
    {
        $count = count([]);
        $val = new \stdClass();
    }
}

?>
-----
<?php

namespace Rector\Performance\Tests\Rector\FuncCall\PreslashSimpleFunctionRector\Fixture;

use stdClass;
class AutoImportNames
{
    public function run()
    {
        $count = \count([]);
        $val = new stdClass();
    }
}

?>
```

samsonasik: Thank you, I understood, I updated the fixture https://github.com/rectorphp/rector/pull/5268/commits/80e2025c9702f9192e9924c84fe3e349ef6ca4ca

samsonasik: The possible safe solution I found right now is:

- keep use classname working
- skip add `\` prefix for function when auto import name enabled.

see https://github.com/rectorphp/rector/pull/5268/commits/8402fb7b3bb035cb7dd2bb7cd0513840297fe7aa

samsonasik:

rules/performance/tests/Rector/FuncCall/PreslashSimpleFunctionRector/PreslashSimpleFunctionRectorTest.php
```diff
@@ -16,6 +17,12 @@ final class PreslashSimpleFunctionRectorTest extends AbstractRectorTestCase
      */
     public function test(SmartFileInfo $fileInfo): void
     {
+        $filename = sprintf('%s.php.inc', Option::AUTO_IMPORT_NAMES);
+
+        if ($fileInfo->endsWith($filename)) {
+            $this->setParameter(Option::AUTO_IMPORT_NAMES, true);
+        }
+
         $this->doTestFileInfo($fileInfo);
```
TomasVotruba, line 26: This should be handled in a stanalone new test case.

samsonasik: implemented.

TomasVotruba:
Thank you :clap:

TomasVotruba (approved):

PR #5272, opened 2021-01-21
Add failing test fixture for ConsoleExecuteReturnIntRector

vladyslavstartsev:
# Failing Test for ConsoleExecuteReturnIntRector

Based on https://getrector.org/demo/2ce24767-1f3c-4f0c-b74f-bbea4b441678

it should change nothing because the value before return is casted to `int`

### DIFF
--- a/rules/symfony4/tests/Rector/ClassMethod/ConsoleExecuteReturnIntRector/Fixture/demo_command.php.inc
+++ b/rules/symfony4/tests/Rector/ClassMethod/ConsoleExecuteReturnIntRector/Fixture/demo_command.php.inc
@@ -1,0 +1,31 @@
+<?php
+
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+namespace Rector\Symfony4\Tests\Rector\ClassMethod\ConsoleExecuteReturnIntRector\Fixture;
+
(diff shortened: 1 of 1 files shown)


rules/symfony4/tests/Rector/ClassMethod/ConsoleExecuteReturnIntRector/Fixture/demo_command.php.inc
```diff
@@ -0,0 +1,31 @@
+<?php
+
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+namespace Rector\Symfony4\Tests\Rector\ClassMethod\ConsoleExecuteReturnIntRector\Fixture;
+
+final class DemoCommand extends \Symfony\Component\Console\Command\Command
```
samsonasik, line 8: fixture with no different between `-----` should only written once above only, and use `skip_` prefix, for example:

https://github.com/rectorphp/rector/blob/bf94f4c420ee577403edf1da3ee445830462695d/rules/symfony4/tests/Rector/ClassMethod/ConsoleExecuteReturnIntRector/Fixture/skip_already_int.php.inc

samsonasik:
@vladyslavstartsev I cherry-picked your commits at PR https://github.com/rectorphp/rector/pull/5275

PR #5275, opened 2021-01-21
[Symfony4] Skip ConsoleExecuteReturnIntRector on cast int return 

samsonasik:
Close #5272


samsonasik:

TomasVotruba:
Thank you :+1:

PR #5280, opened 2021-01-22
Don't skip add `\` prefix for function when auto import name enabled

pierredup:
This is a follow-up to #5268, which enabled the `\` prefix to be added again to functions when using auto-import names

### DIFF
--- a/rules/coding-style/src/Node/NameImporter.php
+++ b/rules/coding-style/src/Node/NameImporter.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Rector\CodingStyle\Node;
 
 use Nette\Utils\Strings;
+use PhpParser\Node;
 use PhpParser\Node\Expr\ConstFetch;
 use PhpParser\Node\Expr\FuncCall;
 use PhpParser\Node\Name;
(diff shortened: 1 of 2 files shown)


rules/coding-style/src/Node/NameImporter.php
```diff
@@ -191,6 +191,12 @@ private function isNamespaceOrUseImportName(Name $name): bool
     private function isFunctionOrConstantImportWithSingleName(Name $name): bool
     {
         $parentNode = $name->getAttribute(AttributeKey::PARENT_NODE);
+
+        $fullName = $name->toCodeString();
+        if (null === $parentNode && Strings::contains($fullName, '\\') && function_exists($fullName)) {
```
samsonasik, line 196: auto import option need to be checked as well to avoid side effect.

samsonasik, line 196: use parentNode instanceof Node instead of null comparison

rules/coding-style/src/Node/NameImporter.php
```diff
@@ -191,6 +192,13 @@ private function isNamespaceOrUseImportName(Name $name): bool
     private function isFunctionOrConstantImportWithSingleName(Name $name): bool
     {
         $parentNode = $name->getAttribute(AttributeKey::PARENT_NODE);
+
+        $fullName = $name->toCodeString();
```
samsonasik, line 196: use `->toString()` instead to get cleaned up FQN, eg, when Name is a `FullyQualified`, after that, use `! Strings::contains($fullName, '\\')` to ensure this patch only applied to simple function.

```suggestion
        $fullName = $name->toString();
```

rules/coding-style/src/Node/NameImporter.php
```diff
@@ -191,6 +192,13 @@ private function isNamespaceOrUseImportName(Name $name): bool
...
+        $autoImportNames = $this->parameterProvider->provideParameter(Option::AUTO_IMPORT_NAMES);
+        if ($autoImportNames && ! $parentNode instanceof Node && Strings::contains($fullName, '\\') && function_exists($fullName)) {
```
samsonasik, line 198: Use `! Strings::contains($fullName, '\\')` followed previous suggestion:

```suggestion
        if ($autoImportNames && ! $parentNode instanceof Node && ! Strings::contains($fullName, '\\') && function_exists($fullName)) {
```

samsonasik (approved):

samsonasik:
Thank you @pierredup

PR #5260, opened 2021-01-20
Add failing test fixture for AddDefaultValueForUndefinedVariableRector (relates to #5253)

vladyslavstartsev:
relates to https://github.com/rectorphp/rector/issues/5253#issuecomment-763537452

# Failing Test for AddDefaultValueForUndefinedVariableRector

Based on https://getrector.org/demo/62ab1e88-6035-4fa6-87bb-c600dac418b2

PR #5293, opened 2021-01-23
[PHP 5.6] Handle empty array initialization for Undefined array in AddDefaultValueForUndefinedVariableRector

### LINKED ISSUE #5253: Incorrect behavior of AddDefaultValueForUndefinedVariableRector
# Bug Report

| Rector version | 0.9.10      |
| Installed as   | rector-prefixed  |

See https://getrector.org/demo/62ab1e88-6035-4fa6-87bb-c600dac418b2

```php
<?php

final class DemoFile
{
    public function run()
    {
        $a['a'] = 'b';
        $a['a'] = 'b';
        $a['a'] = 'b';
    }
}
```

### Responsible rules

* `AddDefaultValueForUndefinedVariableRector`

## Expected Behavior

if you go to the link, rector always sets the default value to `null`. But it can guess, based on the type that is used, which default value to set. In example that I gave, it is reasonable to have empty array as default value. I haven't discovered other cases yet.

samsonasik:
Closes #5260 Fixes #5253

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -584,3 +584,7 @@ parameters:
 
                 # weird case
                 - packages/better-php-doc-parser/src/PhpDocParser/AnnotationContentResolver.php
+
+        -
+            message: '#Class cognitive complexity is 32, keep it under 30#'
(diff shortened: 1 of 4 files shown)


samsonasik:

PR #5299, opened 2021-01-24
[BetterPhpDocParser] Handle parseString() got ShouldNotHappenException 

samsonasik:
Fixes #5267

PR #5307, opened 2021-01-25
Add test for `PropertyFetchTypeResolver`

vhenzl:
Adds missing test for `PropertyFetchTypeResolver`. See https://github.com/rectorphp/rector/pull/5299#issuecomment-766452307 for the motivation.

Currently based on v0.9.10 which is the last version where property fetch type resolution worked (although with some errors as the test reveals). Will be rebased later.


vhenzl:
@TomasVotruba Is it possible to have tests and fixtures targeting only PHP 7.4+ (or even PHP 8.0+) to test typed properties (and native union types)? What's the recommended way to exclude files like `ClassWithNativeProps.php` from linting, analysing and testing on PHP 7.3?

samsonasik:
You can create a test like Php74Test in:

https://github.com/rectorphp/rector/blob/dba5b8d7146ac098e9775e3995877f9e9740ecf2/rules/dead-code/tests/Rector/FunctionLike/RemoveDeadReturnRector/Php74Test.php

that uses `@requires PHP 7.4` and use FixturePhp74 directory for it:

https://github.com/rectorphp/rector/tree/dba5b8d7146ac098e9775e3995877f9e9740ecf2/rules/dead-code/tests/Rector/FunctionLike/RemoveDeadReturnRector/FixturePhp74

vhenzl:
This is ready now. Tests are expected to fail because `PropertyFetchTypeResolver` is broken: instead of the type of a fetched property, it returns the type of the object the property is fetched on (`ClassWithNativeProps`, `ClassWithPhpDocProps`).

samsonasik:
I cherry-picked your commit at PR at https://github.com/rectorphp/rector/pull/5334

samsonasik:
Handled at https://github.com/rectorphp/rector/pull/5334

PR #5210, opened 2021-01-16
Issue #5205 fixture

genesiscz:

PR #5308, opened 2021-01-25
[TypeDeclaration] Skip ReturnTypeDeclarationRector when class parent extends is in 'vendor'

### LINKED ISSUE #5205: Incorrect behavior of ReturnTypeDeclarationRector

samsonasik:
Closes #5210 Fixes #5205


samsonasik:

rules/type-declaration/src/Rector/FunctionLike/ReturnTypeDeclarationRector.php
```diff
@@ -248,6 +256,27 @@ private function addReturnType(FunctionLike $functionLike, Node $inferredReturnN
...
+
+        $parentName = $parent->extends->toString();
+        $reflectionClass = new ReflectionClass($parentName);
+        if ($reflectionClass->isInternal()) {
+            return true;
+        }
+
+        $fileName = $reflectionClass->getFileName();
+        return Strings::contains((string) $fileName, 'vendor');
+    }
```
TomasVotruba, line 278: This logic should be extracted to `packages/vendor-locker`

There might be some useful class like `Rector\VendorLocker\NodeVendorLocker\ClassMethodVendorLockResolver`

phpstan.neon
```diff
@@ -577,3 +577,7 @@ parameters:
 #            paths:
                 # weird case
 #                - packages/better-php-doc-parser/src/PhpDocParser/AnnotationContentResolver.php
+
+        -
+            message: '#Class cognitive complexity is 32, keep it under 30#'
+            path: rules/type-declaration/src/Rector/FunctionLike/ReturnTypeDeclarationRector.php
```
TomasVotruba, line 583: Crossing cognitive complexity treshold should not be ignored.
Instead, should be solved by a new service extraction.

It works well with duplicated method PHPStan rule. That way we keep small consistent services with specific work and without duplication.

rules/type-declaration/src/Rector/FunctionLike/ReturnTypeDeclarationRector.php
samsonasik: implemented with use `VendorLockResolver->isReturnChangeVendorLockedIn()`

phpstan.neon
samsonasik: fixed.

samsonasik:

TomasVotruba:
Thank you :+1:

TomasVotruba:
@samsonasik There is one pull-request that is related to this issue.

Could you verify it was fixed or handle it?
https://github.com/rectorphp/rector/pull/5209

samsonasik:
I will try.

PR #5313, opened 2021-01-25
[Phalcon] Drop custom rules, the framework does not have enough traction

TomasVotruba:


glensc:
Drops `config/set/phalcon40.php` and related files

PR #5325, opened 2021-01-26
[Symfony 5.2] Upgrade Definition/Alias::setPrivate to setPublic

samsonasik:
part of #4824

### DIFF
--- a/config/set/symfony52.php
+++ b/config/set/symfony52.php
@@ -107,4 +107,13 @@ return static function (ContainerConfigurator $containerConfigurator): void {
                 ),
             ]),
         ]]);
+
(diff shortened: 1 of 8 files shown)


samsonasik:
@TomasVotruba change `Definition/Alias`::`setPrivate()` to `Definition/Alias`::`setPublic()` need to include parameter change to pass boolean not the passed value, any param value changer that can be used for it?

TomasVotruba:
There is couple of rules that change default value, look for "change arg value".

It might require a new rule. What do you think?

samsonasik:
New rule seems ok. I will try create new rule for it.

samsonasik:
I've added `DefinitionSetPrivateToSetPublicRector` for it.

samsonasik:

TomasVotruba:
Thank you :+1:

PR #5344, opened 2021-01-28
[PHPStan] Typo fix: Refator -> Refactor

samsonasik:
New symplify release 9.0.46 with this update message https://github.com/symplify/symplify/commit/0fd31c08d3a77abf78c69169463c80202f5d503b


samsonasik:
Also exclude phpstan non-dynamic exclude notice in AnnotationReaderFactory

PR #5345, opened 2021-01-28
Fixed case sensitivity for SetList

lulco:


samsonasik (approved):

samsonasik:
PHPStan notice handled at https://github.com/rectorphp/rector/pull/5344

TomasVotruba:
Nice catch :+1:

PR #5350, opened 2021-01-28
Demonstrate infinite recursion [DO NOT MERGE]

matthiasnoback:
This PR demonstrates an infinite recursion issue I ran into, which is caused by the lack of relation attributes in newly created nodes. The solution I think, and one that might be useful in other situations too, would be to traverse a by `refactor()` returned node using the `NodeConnectingVisitor`, so it sets the correct attributes.


TomasVotruba:
Resolved by https://github.com/rectorphp/rector/pull/6057

Thank you for clear failing example :+1:

PR #5356, opened 2021-01-29
[ActionInjectionToConstructorInjectionRector] Deprecate for nishes market

TomasVotruba:
This feature is nowhere to be seen, luckily, so we can deprecate it as rather showcase then Rector set to be used


TomasVotruba:
It is needed actually: https://getrector.org/blog/2021/02/01/how-to-instantly-refactor-symfony-action-injects-to-constructor-injection#comment-5259251859

PR #5363, opened 2021-01-29
[Testing] Add non .php.inc suffix check on AbstractRectorTestCase

### LINKED ISSUE #5349: Show an error if the user creates files in Fixtures/ with extensions other than `.php.inc`
# Feature Request

I often make the mistake of calling my fixture files `.inc.php` instead of `.php.inc`. I think in general there should be no `.php` files in `Fixtures/` anyway, so it would be helpful if Rector would recognize those files and throw an exception. It could even suggest putting supporting code and classes in `Source/`, next to `Fixtures/`. If the filename ends with `.inc.php` it could suggest to rename it to `.php.inc`.

samsonasik:
Fixes #5349

### DIFF
--- a/packages/testing/src/PHPUnit/AbstractRectorTestCase.php
+++ b/packages/testing/src/PHPUnit/AbstractRectorTestCase.php
@@ -287,6 +287,14 @@ abstract class AbstractRectorTestCase extends AbstractKernelTestCase
      */
     protected function doTestFileInfo(SmartFileInfo $fixtureFileInfo, array $extraFiles = []): void
     {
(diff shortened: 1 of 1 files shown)


samsonasik:
It seems the patch need to be in Symplify\EasyTesting https://github.com/symplify/symplify/blob/master/packages/easy-testing/src/DataProvider/StaticFixtureFinder.php .

Closing.

TomasVotruba:
Could you resolve it in @symplify then?

samsonasik:
I will try.

TomasVotruba:
It could be a new method, that would require stricly files with the suffix to be present. Nothing else

PR #5375, opened 2021-01-30
[Nette 3.0] [Code Quality] Add template converting from Presenter

TomasVotruba:


rules/nette/tests/Rector/Class_/ExtractTemplateClassForPresenterRector/Source/extra_file.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Rector\Nette\Tests\Rector\Class_\ExtractTemplateClassForPresenterRector\Source;
+
+final class SomeTemplate
```
lulco, line 5: each template should extend Nette\Bridges\ApplicationLatte\Template so we can call methods like setFile(), render() etc. on it

rules/nette/tests/Rector/Class_/ExtractTemplateClassForPresenterRector/Fixture/some_class.php.inc
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Rector\Nette\Tests\Rector\Class_\ExtractTemplateClassForPresenterRector\Fixture;
+
+use Nette\Application\UI\Presenter;
+
+final class SomePresenter extends Presenter
```
lulco, line 7: what about presenter with more actions?

rules/nette/tests/Rector/Class_/ExtractTemplateClassForPresenterRector/Source/extra_file.php
lulco: I think the best way would be to extend [DefaultTemplate ](https://github.com/nette/application/blob/master/src/Bridges/ApplicationLatte/DefaultTemplate.php) but it is final

rules/nette/tests/Rector/Class_/ExtractTemplateClassForPresenterRector/Fixture/some_class.php.inc
TomasVotruba: That was my first question when I saw this feature. I haven't used it yet in our projects.

How is that handled in practise?

lulco: I don't know what is the best practise or recommended way but I use it this way:

```
use Nette\Application\UI\Presenter;

final class SomePresenter extends Presenter
{
    public function renderDefault()
    {
        /** @var SomeDefaultTemplate $template */
        $template = $this->template;
    }

    public function actionCreate()
    {
        /** @var SomeCreateTemplate $template */
        $template = $this->template;
    }

    public function actionEdit(string $id)
    {
        /** @var SomeEditTemplate $template */
        $template = $this->template;
    }
}
```

TomasVotruba: I see. Btw, how does Nette instantiates e.g. `SomeEditTemplate` object for last method? I could not find it

lulco: I've read somewhere that it should work automatically based on the name of presenter and template (probably only one-action-presenters). But now it seems you have to pass template class name to createTemplate() method and the template with the same type will be created.

So my previous example should be corrected to:

```
use Nette\Application\UI\Presenter;

final class SomePresenter extends Presenter
{
    public function renderDefault()
    {
        /** @var SomeDefaultTemplate $template */
        $template = $this->createTemplate(SomeDefaultTemplate::class);
    }

    public function actionCreate()
    {
        /** @var SomeCreateTemplate $template */
        $template = $this->createTemplate(SomeCreateTemplate::class);
    }

    public function actionEdit(string $id)
    {
        /** @var SomeEditTemplate $template */
        $template = $this->createTemplate(SomeEditTemplate::class);
    }
}
```

lulco: > I've read somewhere that it should work automatically based on the name of presenter and template (probably only one-action-presenters). 

This works thanks to method Control::formatTemplateClass() but those per-action templates are not implemented. Maybe it is not possible

lulco: > I see. Btw, how does Nette instantiates e.g. `SomeEditTemplate` object for last method? I could not find it

I've added PR to solve this feature https://github.com/nette/application/pull/277

rules/nette/src/Rector/Class_/ExtractTemplateClassForPresenterRector.php
```diff
@@ -0,0 +1,73 @@
...
+
+// some_template.latte
+CODE_SAMPLE
+
+                ,
+                <<<'CODE_SAMPLE'
+use Nette\Application\UI\Presenter;
+
+/**
+ * @property SomeTemplate $template
```
lulco, line 42: would it be possible also to create this class with all properties?

rules/nette/src/Rector/Class_/ExtractTemplateClassForPresenterRector.php
```diff
@@ -0,0 +1,73 @@
...
+ */
+final class SomePresenter extends Presenter
+{
+    public function render()
+    {
```
lulco, line 47: I am thinking if I need custom Template class if there are not parameters passed to it.

rules/nette/src/Rector/Class_/ExtractTemplateClassForPresenterRector.php
TomasVotruba: Yes. I think the properties should be collected from the related template.

Is there a some convention algorithm to find template based on presenter + method? That would be useful

rules/nette/src/Rector/Class_/ExtractTemplateClassForPresenterRector.php
TomasVotruba: I don't think so

rules/nette/tests/Rector/Class_/ExtractTemplateClassForPresenterRector/Fixture/some_class.php.inc
TomasVotruba: Thank you for looking into this. If you find any code, please share it. Nette documentation does not explain this much

rules/nette/src/Rector/Class_/ExtractTemplateClassForPresenterRector.php
lulco: Convention (from nette/application) is to find class which extends `Nette\Bridges\ApplicationLatte\Template` in same namespace as Presenter with the same name as Presenter but word "Presenter" is replaced to "Template". You implemented it well.

For presenter + method there is no convention. I send PR, we will see :)

TomasVotruba: I saw this but it seemed rather weird. Now we have to add `Template` to magic place + mention the annotation in the class. Otherwise one or the other won't work.

TomasVotruba: I like this approach of yours:

```php
        /** @var SomeDefaultTemplate $template */
        $template = $this->createTemplate(SomeDefaultTemplate::class);
```

Explicit, and clear.

lulco: Well, yes it is explicit and clear. And also it is the same for all presenters one-action and multi-actions. And I like one other thing with this approach - you can have your Templates in another namespace like Presenters. It is not possible for now with the magic template finding.

lulco:
It will be super awesome if rector will find corresponding template file and add
```
{templateType SomeTemplate}
```
on top of it :)

rules/nette/src/Rector/Class_/ExtractTemplateClassForPresenterRector.php
lulco: Feature was merged  https://github.com/nette/application/pull/277, but it changes nothing for Rector :)

lulco: Maybe one thing. You can use $this->template now instead of $this->createTemplate()

lulco: createTemplate() will not work when you will have actionSomething and renderSomething and in both you will need the template - https://github.com/nette/application/issues/282

TomasVotruba:
Project that required this rule is now paused, so we don't have dedicated time to finish it.

If anyone wants to take over, go for it :+1:

PR #5377, opened 2021-01-30
VarTagRemover should not remove arrays of interfaces

ruudk:
Found out that the VarTagRemover only checks for `class_exists`. When you have an array of interfaces it will remove them. 

Based on https://getrector.org/demo/8aae9749-6dad-4fda-a281-27b25c5afe68


ruudk:
@TomasVotruba Ready 🎉

rules/php74/tests/Rector/Property/TypedPropertyFromStrictConstructorRector/Fixture/iterable_interface_var_tag.php.inc
```diff
@@ -0,0 +1,45 @@
...
+<?php
+
+namespace Rector\Php74\Tests\Rector\Property\TypedPropertyFromStrictConstructorRector\Fixture;
+
+use DateTimeInterface;
+
+final class IterableInterfaceVarTag
+{
+    /**
+     * @var iterable|DateTimeInterface[]
```
ruudk, line 33: @TomasVotruba Question: Is there also a Rector that can transform `iterable|DateTimeInterface[]` to `iterable<DateTimeInterface>`? (same for `array`).

TomasVotruba: Hi, it's possible, but that woud require a new rule.

rules/dead-doc-block/src/TagRemover/VarTagRemover.php
```diff
@@ -104,6 +104,19 @@ private function isNonBasicArrayType(Node $node, VarTagValueNode $varTagValueNod
...
+            return true;
+        }
+
+        if (interface_exists((string) $type->type)) {
+            return true;
+        }
+
+        return false;
+    }
+
```
TomasVotruba, line 119: Good catch!

There is `Symplify\PackageBuilder\Reflection\ClassLikeExistenceChecker` service that can be re-used. It will help us to catch all existance edge cases without maintenance.

ruudk: Great, added in https://github.com/rectorphp/rector/pull/5377/commits/32abc0b37dd69eb46050952e7c3d9f11b6c9d653

ruudk:
@TomasVotruba PR updated and ✅

TomasVotruba:
Thank you :+1:

Issue #5259, opened 2021-01-20
Aliased class not found


lulco:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | e.g. v0.9.10 |
| Installed as   | prefixed Rector                           |

`
Could not process "..." file, due to: "Analyze error: "Class GuzzleClient was not found while trying to analyse it - discovering symbols is probably not configured properly.". Include your files in "$parameters->set(Option::AUTOLOAD_PATHS, [...]);" in "rector.php" config. 
`

## Minimal PHP Code Causing Issue

I can't reproduce this issue in rector demo. It works there for some reason. So here is the simplified class.
```
<?php

use Exception;
use GuzzleHttp\Client as GuzzleClient;

class ArchivationHandler
{
    private function callApi($apiUrl, $data, $token, $retry = 3)
    {
        $client = new GuzzleClient();
        try {
            $response = $client->post($apiUrl, [
                'json' => $data,
                'headers' => [
                    'Authorization' => 'Bearer ' . $token
                ]
            ]);
            return (string) $response->getBody();
        } catch (Exception $e) {
            if ($retry) {
                return $this->callApi($data, $token, $retry - 1);
            }
            return $e->getMessage();
        }
    }
}
```

If I remove alias, it works. But I can't because in original class I have two classes with the name `Client` so I need it.

TomasVotruba:
This might require failing test case one of responsible rules.

lulco:
well but in that case it would be the same problem on demo site too or not?

TomasVotruba:
In ideal world, yes :). But Github CI is different to Docker on website. Also Docker on website and Docker in CI can be different, as they're build in different time on different machines.

lulco:
I am not able to create failing test case. Everything works in Rector itself. There is some problem on my app maybe.

TomasVotruba:
Failing repository is good enough.

lulco:
Here it is https://github.com/lulco/rector-aliased-class

Probably there is some problem with NETTE_30 set, but I don't know why :) all other (tested) sets works fine. Error occured only with this set.

lulco:
@TomasVotruba I've dig little bit deeper and I found one responsible rule `Rector\TypeDeclaration\Rector\MethodCall\FormerNullableArgumentToScalarTypedRector`. Failing test case added https://github.com/rectorphp/rector/pull/5397

PR #5397, opened 2021-01-31
Failing test case for Aliased class not found #5259

lulco:


TomasVotruba:
The class must really exists. Otherwise it's just a string, as any missing class is in PHP.

lulco:
I don’t think it is true. Without alias it works

TomasVotruba:
That's how type resolution works here. It's not related to alias

lulco:
I rewrote the test case. I used existing class `NowString` and reuse whole fixture.php.inc with only one change - I added alias to NowString class. Result is the same: without alias it works, with alias it doesn't. For me it is the proof that it is related to alias.

samsonasik:
@lulco I cherry-picked your commit at PR https://github.com/rectorphp/rector/pull/5408

PR #5398, opened 2021-01-31
Use ChangePackageVersionComposerRector for PHP sets

lulco:


config/set/php52.php
```diff
@@ -22,4 +24,11 @@
                 new RemoveFuncCallArg('ldap_first_attribute', 2),
             ]),
         ]]);
+
+    $services->set(ChangePackageVersionComposerRector::class)
+        ->call('configure', [[
+            ChangePackageVersionComposerRector::PACKAGES_AND_VERSIONS => ValueObjectInliner::inline([
+                new PackageAndVersion('php', '^5.2'),
+            ]),
```
TomasVotruba, line 32: Btw, do we have a test to avoid lowering PHP version? 

So we can avoid downgrading bugs like these:

```diff
 {
-   "php": "^7.3"
+   "php": "^5.2"
 }
```


If not, this PR should contain this test too

lulco: You mean if it compare version strings and change only if target version is higher than original one? No those test don’t exist and it is good because you should use this rules also for downgrading

TomasVotruba: > You mean if it compare version strings and change only if target version is higher than original one?

Yes :+1:

TomasVotruba:
@berezuev Hi, I'm curious, why :-1: ?

berezuev:
> @berezuev Hi, I'm curious, why :-1: ?

My opinion that it wouldn't be convenient to use this in real work.

I use Rector almost everyday in my company to upgade dozens of internal legacy services and libs. Most of them require more then 20 dependencies and would be broken if you just increment minor PHP version in composer file. So, we need to resolve a lot of dependency conflicts manually (hello, php7.3 bc breaks!). 

Rector is a useful tool for small unary updates, eg.:
1) update dependencies to compatible with new php or framework version and release
2) fix all deprecations and release
3) update php/framework version and release 
4) start using new features
5) repeat

config/set/php52.php
berezuev: > it is good because you should use this rules also for downgrading

We have downgrading sets, haven't we?

lulco: Yes, that's why we shouldn't check if new version is higher as previous

lulco:
@berezuev If you use language features from let's say php7.3, you should change you php version in composer. That's all what is this PR about.

I saw many php libraries where there were typehints used, but composer says the library supports "php": ">=5.3". And that's not true.

lulco:
Also you can skip these rules in your rector.php config if you don't want to use them. Or set your own target php version as it is done for rector itself here https://github.com/rectorphp/rector/pull/5398/files#diff-d933a2ad57daa43419b483d3f1ddedaec7aa44933007fead9f9437390f596f4aR97

berezuev:
@lulco @TomasVotruba I think I have misunderstood this PR at first reading.
You're right, this is useful feature.

lulco:
@TomasVotruba Build is green. If you agree with me, that comparing old version to new version is not necessary (even it will be contra-productive in downgrade set lists), we can merge it.

TomasVotruba:
@lulco I'd merge it already, but test case is still missing: https://github.com/rectorphp/rector/pull/5398#discussion_r567502931

TomasVotruba:
I talk only about increasing the version now, lowering can be handled later

config/set/php52.php
ruudk: Not sure if I follow the discussion correctly: If my composer requires `^7.4` and I run the `PHP 5.6` upgrade set, that this will now change my composer to require `^5.6`?

TomasVotruba: I hope not :)
We need a test case to avoid it.

lulco: Maybe there are some misunderstandings from me how rector is used with php sets in a wild. Here are my assumptions:
If app / library requires php ^7.3, it run:
- only set php7.3 
- or more php sets - let's say php7.1, php7.2, php7.3. Set for actual version is always the last running.

Same steps are used in rector itself in rector.php. 

If your lib / app using different version of php or some range, you can set it in your rector config - like it is in this merge request for rector https://github.com/rectorphp/rector/pull/5398/files#diff-d933a2ad57daa43419b483d3f1ddedaec7aa44933007fead9f9437390f596f4aR94

ruudk: For what it's worth: This is how I used it: Our application runs on PHP 7.4 and requires that in composer. I just discovered Rector and applied sets starting from PHP 5.2, one by one till 7.1 (need to do some more work to do all the upgrades up until 7.4). Turned out that Rector found lots of things that we were doing "the old way". In any case, our composer PHP version should remain 7.4.

lulco: I understand, so you would need just add one setting in your rector.php to keep it unchanged:
```
$services->set(ChangePackageVersionComposerRector::class)
        ->call('configure', [[
            ChangePackageVersionComposerRector::PACKAGES_AND_VERSIONS => ValueObjectInliner::inline([
                new PackageAndVersion('php', '^7.4'),
            ]),
        ]]);
```

TomasVotruba: @lulco Adding work to our users is not a Rector way. This is our responsibility to handle.

ruudk: Can't this automatically be read from `$parameters->set(Option::PHP_VERSION_FEATURES, PhpVersion::PHP_74);`?

PR #5399, opened 2021-01-31
Removed duplicit package sebastian/diff from composer.json

lulco:


TomasVotruba:
It seems used:

```php
use SebastianBergmann\Diff...
```

And just once in `composer.jsonh

lulco:
Where it is used? Ok we can remove it from require-dev section. IDE marks it as diplicated

TomasVotruba:
Here: https://github.com/rectorphp/rector/search?q=use+SebastianBergmann

I see it now. Better remove the dev one.

I thought composer validate is handling this one by default, but it needs a `--strict` flag:

```bash
composer validate --strict
```

Could you update the CI here too to avoid it next time?

lulco:
Is there any test in CI for composer validate? I don't see any. Or should I add another one?

lulco:
Removed package from require-dev and composer validate test added

TomasVotruba:
Thanks :+1:

PR #5408, opened 2021-02-02
[TypeDeclaration] Handle alias usage on FormerNullableArgumentToScalarTypedRector

### LINKED ISSUE #5259: Aliased class not found

samsonasik:
Close #5397 Fixes #5259

### DIFF
--- a/rules/type-declaration/src/PHPStan/Type/ObjectTypeSpecifier.php
+++ b/rules/type-declaration/src/PHPStan/Type/ObjectTypeSpecifier.php
@@ -79,8 +79,13 @@ final class ObjectTypeSpecifier
                     return new AliasedObjectType($alias, $fullyQualifiedName);
                 }
 
-                // B. is aliased classes matching the class name
-                if ($useName === $className) {
+                // B. is aliased classes matching the class name non namespaced
+                if (strpos($useName, '\\') === false && $useName === $className) {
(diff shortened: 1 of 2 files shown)


samsonasik:

TomasVotruba:
Thank you :+1:

PR #5416, opened 2021-02-03
[Renaming] Do not update target new namespace exists on RenameClassRector

### LINKED ISSUE #5181: Incorrect behavior of RenameClassRector
# Bug Report

See https://getrector.org/demo/c16fbf01-a697-4fa7-ba6d-790d4d63e203

```php
<?php

namespace Acme {
   use Acme\Foo\DemoFile;
    
   DemoFile::run();
}

namespace Acme\Foo {

/**
 * @deprecated Use Acme\Bar\DemoFile instead!
 */
final class DemoFile
{
    public function run()
    {
        return 5;
    }
}

}

namespace Acme\Bar {
 
final class DemoFile
{
    public function run()
    {
        return 5;
    }
}   
}
```

### Responsible rules

* `RenameClassRector`

## Expected Behavior

I'm trying to rector a classic class deprecation where one class superseeds an old one. The RenameClassRector works fine, except changing the namespace of the old class itself. I'd expect that it won't touch the namespace.

Kind regards
Matthias

samsonasik:
Fixes #5181

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -549,3 +549,6 @@ parameters:
             paths:
                 # is "require" in PHP
                 - src/Php/ReservedKeywordAnalyzer.php
+
+        - '#Class Acme\\Bar\\DoNotUpdateExistingTargetNamespace not found#'
+        - '#Class Acme\\Foo\\DoNotUpdateExistingTargetNamespace not found#'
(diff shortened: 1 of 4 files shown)


samsonasik:

TomasVotruba:
Thank you :+1:

PR #5418, opened 2021-02-04
[Laminas 4.0] Add laminas-servicemanager-40 set

samsonasik:
@TomasVotruba the upcoming [laminas-servicemanager](https://github.com/laminas/laminas-servicemanager) 4.0 will require `Psr\Container\ContainerInterface` instead of `Interop\Container\ContainerInterface` like the commit https://github.com/laminas/laminas-servicemanager/commit/779d5321b83d0f23fb726ded3dadc2df41ef0145 .

This PR add a config set to allow laminas-servicemanager user to upgrade from v3 to v4 with the `LAMINAS_SERVICEMANGER_40` config set.

### DIFF
--- a/config/set/laminas-servicemanager-40.php
+++ b/config/set/laminas-servicemanager-40.php
@@ -12,6 +12,7 @@ return static function (ContainerConfigurator $containerConfigurator): void {
         ->call('configure', [[
             RenameClassRector::OLD_TO_NEW_CLASSES => [
                 'Interop\Container\ContainerInterface' => 'Psr\Container\ContainerInterface',
+                'Laminas\ServiceManager\FactoryInterface' => 'Laminas\ServiceManager\Factory\FactoryInterface',
             ],
         ]]);
 };
--- a/config/set/laminas-servicemanager-40.php
+++ b/config/set/laminas-servicemanager-40.php
@@ -18,34 +18,4 @@ return static function (ContainerConfigurator $containerConfigurator): void {
                 'Interop\Container\ContainerInterface' => 'Psr\Container\ContainerInterface',
             ],
         ]]);
-
-    $services->set(AddParamTypeDeclarationRector::class)
-        ->call('configure', [[
-            AddParamTypeDeclarationRector::PARAMETER_TYPEHINTS => ValueObjectInliner::inline([
-                new AddParamTypeDeclaration(
-                    'Laminas\ServiceManager\Factory\AbstractFactoryInterface',
-                    '__invoke',
-                    0,
-                    new ObjectType('Psr\Container\ContainerInterface'),
-                ),
-                new AddParamTypeDeclaration(
-                    'Laminas\ServiceManager\Factory\DelegatorFactoryInterface',
-                    '__invoke',
-                    0,
-                    new ObjectType('Psr\Container\ContainerInterface'),
-                ),
-                new AddParamTypeDeclaration(
-                    'Laminas\ServiceManager\Factory\FactoryInterface',
-                    '__invoke',
-                    0,
-                    new ObjectType('Psr\Container\ContainerInterface'),
-                ),
-                new AddParamTypeDeclaration(
-                    'Laminas\ServiceManager\Factory\InvokableFactory',
-                    '__invoke',
-                    0,
-                    new ObjectType('Psr\Container\ContainerInterface'),
-                ),
-            ]),
-        ]]);
 };
--- a/config/set/laminas-servicemanager-40.php
+++ b/config/set/laminas-servicemanager-40.php
@@ -12,8 +12,6 @@ return static function (ContainerConfigurator $containerConfigurator): void {
         ->call('configure', [[
             RenameClassRector::OLD_TO_NEW_CLASSES => [
                 'Interop\Container\ContainerInterface' => 'Psr\Container\ContainerInterface',
-                'Laminas\ServiceManager\ServiceLocatorInterface' => 'Psr\Container\ContainerInterface',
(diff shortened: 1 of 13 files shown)


samsonasik:
//cc @boesing @ocramius here is the PR, the usage can be:

```php
<?php
// rector.php
declare(strict_types=1);

use Rector\Core\Configuration\Option;
use Rector\Set\ValueObject\SetList;

        SetList:: LAMINAS_SERVICEMANGER_40,
    ]);
};
```

samsonasik:
Actually, `RenameClassRector` enough https://getrector.org/demo/f828d2e8-d459-4f2c-8e37-8c77e69e3d09

boesing:
The mentioned commit will be removed. I'll force-push 3.6.x to 4.0.x now as its already messed up.
I'll keep 4.0.x as a backup locally just in case we probably missed some feature we want to re-apply in the new 4.0.x branch.

samsonasik:
@boesing thank you, is there a possibility of user not pass `Interop\Container\ContainerInterface`  to __invoke , eg: `ServiceLocatorInterface` ? If yes, then it need to be reverted https://github.com/rectorphp/rector/pull/5418/commits/f9649f9321d52f939e9fad8f0fa84030416e8874

samsonasik:
@boesing I just re-tried with `ServiceLocatorInterface` and it works for a class like this:

```php
class IndexControllerFactory
{
    public function __invoke(\Laminas\ServiceManager\ServiceLocatorInterface $container)
    {
        return new IndexController();
    }
}
```

For implemenst `implements FactoryInterface`, it seems must be already `Interop` interface. The one that need to be changed is `Laminas\ServiceManager\ServiceLocatorInterface` to `Psr\Container\ContainerInterface`

samsonasik:
I updated to change `Laminas\ServiceManager\ServiceLocatorInterface` to `Psr\Container\ContainerInterface`.

samsonasik:
I verify in my local dev that it works:

1. Interop to Psr

Screen Shot 2021-02-04 at 18 42 15

2. ServiceLocatorInterface to Psr 

Screen Shot 2021-02-04 at 18 43 13

samsonasik:
Added `Zend\ServiceManager\ServiceLocatorInterface` to `Psr\Container\ContainerInterface` if user uses `laminas-zendframework-bridge`

packages/set/src/ValueObject/SetList.php
```diff
@@ -211,6 +211,11 @@ final class SetList
      */
     public const KDYBY_TRANSLATOR_TO_CONTRIBUTTE_TRANSLATION = __DIR__ . '/../../../../config/set/kdyby-translator-to-contributte-translation.php';
 
+    /**
+     * @var string
+     */
+    public const LAMINAS_SERVICEMANGER_40 = __DIR__ . '/../../../../config/set/laminas-servicemanager-40.php';
```
TomasVotruba, line 217: How is laminas project versioned?

If it's a monorepo, we can have just `LAMINAS_40` here

samsonasik: The laminas project is not a monorepo, per-component has its own live so upgrade path may need per-component basis.

TomasVotruba: I see.

Let me know when it's ready to merge

samsonasik:
@boesing I registered more interfaced to be mapped to new interface. I think it is ready for starter. Please review, thank you.

samsonasik:
@boesing I just added new rector rule :tada: RemoveServiceLocatorRector to remove `->getServiceLocator()` method call https://github.com/rectorphp/rector/pull/5418/commits/078080cd39bd72486431a2d6e894f20f99af4680 , the naming and or object required for check against `->getServiceLocator()` can be updated if needed.

config/set/laminas-servicemanager-40.php
```diff
@@ -2,6 +2,7 @@
 
 declare(strict_types=1);
 
+use Rector\LaminasServiceManager4\Rector\MethodCall\RemoveServiceLocatorRector;
```
TomasVotruba, line 5: The namespace should be `Lamaminas\\`, until we support many various version and rules for the framework, like only Symfony/Laravel have now.

config/set/laminas-servicemanager-40.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+use Rector\LaminasServiceManager4\Rector\MethodCall\RemoveServiceLocatorRector;
+use Rector\Renaming\Rector\Name\RenameClassRector;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
```
lulco, line 9: what about to add ChangePackageVersionComposerRector? O:-) see https://github.com/rectorphp/rector/blob/master/config/set/nette-31.php

config/set/laminas-servicemanager-40.php
samsonasik: updated.

samsonasik:
The laminas team is plannning to use own repo for rector plugin development. I am closing it for now.

TomasVotruba:
Could you link it, so people know where to go?

samsonasik:
It is still in planning for next development, here is the repo https://github.com/laminas/laminas-servicemanager-migration

TomasVotruba:
I see, thanks!

It's not very effective to keep migration repo per package, as it would take lot of work to upgrade all the rules in each package, when new Rector is released.

Here are 2 already existing projects maintaining Rector rules on their own, for inspiration:

- https://github.com/sabbelasichon/typo3-rector
- https://github.com/palantirnet/drupal-rector

PR #5419, opened 2021-02-04
[CodeQuality] Add "single" prefix for value var of Foreach on ForToForeachRector when singularize() got same name

### LINKED ISSUE #5338: Incorrect behavior of ForToForeachRector
# Bug Report

See https://getrector.org/demo/679daae2-1fb3-4f01-86a6-a8c44e91a991

```php
<?php

final class DemoFile
{
    public function run()
    {
        $test = ['t', 'e', 's', 't'];

        for($i = 0; $i < count($test); $i++) {
            echo $test[$i];
            echo " ";
        }
    }
}
```

### Responsible rules

* `ForToForeachRector`

## Expected Behavior

Rector modifies the loop like this:

`foreach ($test as $i => $test) {`

which of course does not work. The issue is, that ForToForeachRector tries to guess the variable name by singularizing it using the Inflector. 
Some iterable-names, like $test cannot be singularized so the Inflector returns the same name, which leads to this problem. And in my experience names like this are found a lot in legacy codebases.

At least the singularized name should be checked if it is different from the iterable variable name. If not, the value-variable needs to be constructed in some other way. 
If anyone comes up with an idea how to do this, I'll be happy to submit a PR.

Thank you for your work and this awesome project!

samsonasik:
Fixes #5338

### DIFF
--- a/rules/code-quality/src/Rector/For_/ForToForeachRector.php
+++ b/rules/code-quality/src/Rector/For_/ForToForeachRector.php
@@ -183,6 +183,11 @@ CODE_SAMPLE
             $iteratedVariableSingle = $originalVariableSingle;
         }
 
+        // double check for final $iteratedVariableSingle value
+        if ($this->isValueVarUsedNext($node, $iteratedVariableSingle)) {
+            return null;
+        }
+
         $foreach = $this->createForeach($node, $iteratedVariableSingle);
         $this->mirrorComments($foreach, $node);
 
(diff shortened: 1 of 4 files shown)


mkrauser:
@samsonasik Thank you very much!

TomasVotruba:
Thanks :+1: 

This rule seems huge. I've noticed it's ignored in `phpstan.neon` for complexity.
Could you take it out in next PR and try to decouple it to services

samsonasik:
@TomasVotruba I will try.

PR #5429, opened 2021-02-05
[Utils] Remove Validate Max File Length command

samsonasik:
ref https://github.com/symplify/symplify/pull/2907 for use `MaxFileLengthRule` PHPStan rule.


samsonasik:
@TomasVotruba configurable rules seems disabled: 

https://github.com/rectorphp/rector/blob/2085291ae924d0fe5077a22ba8a098ed92f040d9/phpstan.neon#L9

is the `MaxFileLengthRule` can be registered somewhere?

TomasVotruba:
Oh crap, I forgot PHPStan only analyzers `*.php` files and skips `*.php.inc`.

We'll need to keep this one here. Maybe move the command to `symplify/easy-ci` package so we can re-use it on more places

samsonasik:
Ok, I will try.

PR #5430, opened 2021-02-05
Added Docker image for blackfire profiling

JanMikes:
Hi, im playing with profiling and trying to optimize performance for online demo. I use blackfire and run it in docker, i think it is worth publishing it directly in the repo - either it might serve someone for inspiration and either it might be useful if anyone wants to start profiling as well + i do not have to save it in separe repository 😄


TomasVotruba:
Hi Honza, sure. I look forward for the performance optimizations :)

PR #5431, opened 2021-02-06
[EasyCI] Using symplify/easy-ci for validate-file-length

samsonasik:
ref https://github.com/symplify/symplify/pull/2913 to use `vendor/bin/easy-ci validate-file-length` command so the current `ValidateFileLengthCommand` can be removed.


samsonasik:

TomasVotruba:
Thanks 👍

PR #5432, opened 2021-02-06
[Privatization] Add MakeOnlyUsedByChildrenProtectedRector

### LINKED ISSUE #5078: [Privatization] Make public class method protected, if only used by its children
E.g.

```diff
abstract class AbstractSomeClass
{
-	public function run()
+ 	protected function run()
    {
    }
}

class ChildClass extends AbstractSomeClass
{
	public function go()
    {
        $this->run();
    }
}
```

samsonasik:
Fixes #5078

### DIFF
--- a/rules/privatization/tests/Rector/ClassMethod/MakeOnlyUsedByChildrenProtectedRector/Fixture/skip_used_by_external_in_closure.php.inc
+++ b/rules/privatization/tests/Rector/ClassMethod/MakeOnlyUsedByChildrenProtectedRector/Fixture/skip_used_by_external_in_closure.php.inc
@@ -1,0 +1,17 @@
+<?php
+
+namespace Rector\Privatization\Tests\Rector\ClassMethod\MakeOnlyUsedByChildrenProtectedRector\Fixture;
+
+class SkipUsedByExternalInClosure
+{
+    public function run()
+    {
+    }
+}
+
+function () {
+    $obj = new SkipUsedByExternalInClosure();
+    $obj->run();
+};
+
+?>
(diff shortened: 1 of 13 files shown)


samsonasik:

samsonasik:
I updated to skip extended overridden method

samsonasik:

rules/privatization/tests/Rector/ClassMethod/MakeOnlyUsedByChildrenProtectedRector/Fixture/skip_used_by_external_in_closure.php.inc
```diff
@@ -0,0 +1,17 @@
...
+    {
+    }
+}
+
+function () {
+    $obj = new SkipUsedByExternalInClosure();
+    $obj->run();
+};
+
+?>
```
TomasVotruba, line 17: There should be also use case with parent class method, that has a child, and the child method call is used.

```php
abstract class SomeParent
{
    public function run()
    {
    }
}

final class ChildClass extends SomeParent
{
}

final class SkipSomeUser
{
     public function useIt(ChildClass $childClass)
     {
         // the method is used, so it should be skipped
         $childClass->run();
     }
}
```

TomasVotruba:
Looks very good, just one detail :+1:

rules/privatization/tests/Rector/ClassMethod/MakeOnlyUsedByChildrenProtectedRector/Fixture/skip_used_by_external_in_closure.php.inc
samsonasik: implemented, added fixture for it.

TomasVotruba:
Thanks :+1:

PR #5433, opened 2021-02-06
[Test] Increase test coverage by test Rector object ->getRuleDefinition() instanceof RuleDefinition

samsonasik:


samsonasik:

packages/testing/src/PHPUnit/AbstractRectorTestCase.php
```diff
@@ -307,6 +308,12 @@ protected function doTestFileInfo(SmartFileInfo $fixtureFileInfo, array $extraFi
             return;
         }
 
+        $rectorClass = $this->getRectorClass();
+        if (self::$container->has($rectorClass)) {
+            $rectorObject = self::$container->get($rectorClass);
+            $this->assertInstanceOf(RuleDefinition::class, $rectorObject->getRuleDefinition());
+        }
+
```
TomasVotruba, line 316: This test does not bring much value. This is tested by PHP itself as there is `RuleDefinition` return type

samsonasik: ok 👍

TomasVotruba:
What would be really helpful, is to improve the performance of tests. Now they take 40-50 seconds, because tests with custom `provideConfigFileInfo()` invoke new container rebuild. If there was a way to speed this up, it would save us CI and local times.

PR #5434, opened 2021-02-06
[Privatization] Register MakeOnlyUsedByChildrenProtectedRector to privatization config set

samsonasik:


packages/post-rector/src/Collector/UseNodesToAddCollector.php
```diff
@@ -174,7 +174,7 @@ public function getShortUsesByFileInfo(SmartFileInfo $smartFileInfo): array
     /**
      * @return AliasedObjectType[]|FullyQualifiedObjectType[]
      */
-    public function getUseImportTypesByNode(Node $node): array
+    private function getUseImportTypesByNode(Node $node): array
```
samsonasik, line 177: it is strange that it changed to `private`, as the rule only change to `protected`, it seems the re-use of `getChildrenOfClass()` make it not passed.

samsonasik: fixed by skip final class

samsonasik:

TomasVotruba:
Thank you :+1:

Issue #5234, opened 2021-01-19
Not Implemented Yet exception thrown


olivernybroe:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.9.10             |
| Installed as   | composer dependency                           |

When running rector it fails on a bunch of files, however the errors I was getting looked like this
```
 [ERROR] Could not process "app/Listeners/NotifyInvoiceBooked.php" file, due to:                                        
         "".    
```
So I tried running it again with verbose on and got

```
In AbstractReadNodeAnalyzer.php line 60:
                                                      
  [Rector\Core\Exception\NotImplementedYetException]  
```

## Minimal PHP Code Causing Issue

https://getrector.org/demo/289c2eaf-c214-4f7d-995d-c1f8a9a6dff0

## Expected Behaviour

That rector would be able to run the files.

olivernybroe:
Sorry for the code example 😟 

I took my file and kept deleting stuff and simplifying until I couldnt delete more. Might be some other changes which could have been made, however this was the most basic I could do for the issue.

TomasVotruba:
I'm on it

PR #5445, opened 2021-02-07
[DeadCode] Fix RemoveSetterOnlyPropertyAndMethodCallRector for no-read, no-write expression

TomasVotruba:
Closes https://github.com/rectorphp/rector/issues/5234

@olivernybroe This was the problematic snippet:

```php
public function data()
    {
        $this->data;
    }
```

This method does nothing, no set, no return. Now it's evalueated as no-read property use, so it's skipped.

It looks like dead code. What does it do actually?


olivernybroe:
It's dead code. 
There were a lot of other lines in the method, and I think it was an old property which wasn't used anymore but hadn't been removed completely 👍 

But nice with the fix, will look forward to trying running rector again on our codebase, hopefully it can Analyse all the files now, else I'll just create more issues 😉

TomasVotruba:
The worse code base the better training for Rector :)

There should be a rule for removing property fetches like these. If not in dead-code set, could pay make an issue?

PR #5449, opened 2021-02-07
Add failing test fixture for RenameMethodRector

ruudk:
# Failing Test for RenameMethodRector

Based on https://getrector.org/demo/85509375-fe03-4f79-b861-fe8a4e7d6990

Why doesn't it rename the method definition?

### DIFF
--- a/rules/renaming/tests/Rector/MethodCall/RenameMethodRector/Fixture/demo_file.php.inc
+++ b/rules/renaming/tests/Rector/MethodCall/RenameMethodRector/Fixture/demo_file.php.inc
@@ -1,0 +1,45 @@
+<?php
+
+namespace Rector\Renaming\Tests\Rector\MethodCall\RenameMethodRector\Fixture;
+
+final class DemoFile
+{
+    public function notify()
+    {
+        return 5;
+    }
+
+}
+
+final class Caller
+{
+    public static function execute()
+    {
+        $demo = new DemoFile();
+        $demo->notify();
+    }
+}
+?>
+-----
+<?php
+
+namespace Rector\Renaming\Tests\Rector\MethodCall\RenameMethodRector\Fixture;
+
+final class DemoFile
+{
+    public function __invoke()
+    {
+        return 5;
+    }
+
+}
+
+final class Caller
+{
+    public static function execute()
+    {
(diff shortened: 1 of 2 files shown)


ruudk:
@TomasVotruba Sorry for the ping. Do you maybe have any insights on what this could be? I would love to fix the root cause (like the many other PR's you helped me with 😊)

rules/renaming/tests/Rector/MethodCall/RenameMethodRector/RenameMethodRectorTest.php
```diff
@@ -40,6 +40,7 @@ protected function getRectorsWithConfiguration(): array
                     new MethodCallRename(AbstractType::class, 'setDefaultOptions', 'configureOptions'),
                     new MethodCallRename(Html::class, 'add', 'addHtml'),
                     new MethodCallRename('*Presenter', 'run', '__invoke'),
+                    new MethodCallRename('DemoFile', 'run', 'test'),
```
TomasVotruba, line 43: Here must be fqn class name, with namespace

ruudk: So it's just not working in the Playground because it doesn't have a namespace?

TomasVotruba: Looks like that :+1:

TomasVotruba:
PHPStan reports one more issue

ruudk:
Found out that my problem is caused by this change https://github.com/rectorphp/rector/pull/3707

When I change this:
https://github.com/rectorphp/rector/blob/e822db8625565106fd1b1b7d78662dba4967e145/rules/renaming/src/Rector/MethodCall/RenameMethodRector.php#L125
to return false, my test works, but the https://github.com/rectorphp/rector/blob/master/rules/renaming/tests/Rector/MethodCall/RenameMethodRector/Fixture/skip_self_method_rename.php.inc test fails.... 

Not sure how to fix this.

samsonasik:
@ruudk I cherry-picked your commit at PR https://github.com/rectorphp/rector/pull/5493

Issue #5340, opened 2021-01-28
[BUG] Rector removes class in use


tomasnorre:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.9.16              |
| Installed as   | composer dependency                           |
| PHP| 7.4.14 | 

Regarding version:
My `composer.json` says `0.9.16` and `rector --version` returns 
```shell 
vendor/bin/rector --version                                     
Rector Unknown
``` 

I have build a small Laravel zero application to reproduce this problem.

When running the `vendor/bin/rector process --dry-run --config rector-ci.php` it wants to remove the `Models/Vehicle.php` file, and I don't know why. 

The `Car`-object is used in the `InspiringCommand.php` and calls the inherited method `hasEngine()` therefore I don't see why rector don't recognize this as in use. The `extend Vehicle` itself should in my eyes be enough for that and not want to remove it.  

```shell 
[WARNING] File "app/Models/Vehicle.php" will be removed  
```

If I change the class to be and abstract class it will stay. But not every class that I might want to inherit will be abstract. 

## Minimal PHP Code Causing Issue

I have stripped away all not needed. My file tree looks like this for the files I'll list below
```shell
app
├── Commands
│   └── InspiringCommand.php
├── Models
│   ├── Car.php
│   └── Vehicle.php
``` 
`Commands/InspiringCommand.php`
```php
<?php

namespace App\Commands;
use App\Models\Car;

final class InspiringCommand
{
    protected $signature = 'inspiring {name=Artisan}';
    protected $description = 'Display an inspiring quote';
    public function handle(): void
    {
        $car = new Car();
        $this->info($car->hasEngine());
        $this->info('Simplicity is the ultimate sophistication.');
    }
}
``` 
`Models/Car.php`
```php
<?php

declare(strict_types=1);
namespace App\Models;

class Car extends Vehicle
{
    private string $engineSize;
}
``` 

`Models/Vehicle.php`
```php
<?php

declare(strict_types=1);
namespace App\Models;

class Vehicle
{
    protected string $colour;
    protected bool $engine;

    public function hasEngine(): bool
    {
        return $this->engine;
    }
}
```
`rector-ci.php` The rules are adjusted to this example to only have the errors that troubles me atm. 
```php
<?php

declare(strict_types=1);

use Rector\Core\Configuration\Option;
use Rector\DeadCode\Rector\Cast\RecastingRemovalRector;
[...]

PR #5459, opened 2021-02-08
[WIP] Add tests for issue #5340

tomasnorre:


TomasVotruba:
Test case looks good. Could you fix it as well?

tomasnorre:
I'll try to. Thanks for confirming, that helps me a lot.

tomasnorre:
I haven't found the issue/solution yet. Perhaps you can give me some pointers base on additional information. 

The https://github.com/rectorphp/rector/blob/master/packages/node-collector/src/NodeCollector/ParsedNodeCollector.php#L167 is called and the Car class is seen as `$node instanceof New_` where the `Vehicle`-class is seen as `$node instanceof Class_`. 

The https://github.com/rectorphp/rector/blob/d7353baf2c289b16c763219f46421d774a190f9a/rules/dead-code/src/UnusedNodeResolver/UnusedClassResolver.php#L120 only iterates over the `$this->parsedNodeCollector->getNews() as $newNode` which explains why the class is seen as to be removed.

As said, I don't know the solution to this, so if you have some pointers, I will be happy. I'll see if I can figure out what the problem is :)

rules/dead-code/tests/Rector/Class_/RemoveUnusedClassesRector/Fixture/skip_class_extend.php.inc
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * @var bool
+     */
+    protected $engine;
+
+    public function hasEngine(): bool
+    {
+        return $this->engine;
+    }
+}
```
TomasVotruba, line 34: This file should not be in `/Fixture/`, but in `/Source/` directory, as it's not the tested class

By convention, the test fixture should only contain tested class

TomasVotruba:
Could you submit your WIP? It would be easier to give better feedback on real code

rules/dead-code/tests/Rector/Class_/RemoveUnusedClassesRector/Fixture/skip_class_extend.php.inc
tomasnorre: If I move the Vehicle class to source, then the tests are green.

tomasnorre:
> Could you submit your WIP? It would be easier to give better feedback on real code

The code hasn't changed since the commit. Still trying to figure out what's going on.

tomasnorre:
I'm currently trying to figure out if a function `isParent()` analog to `isAbstract()` would be a solution and how it should look like. 
It's not part of the Class_ implementation. But perhaps it should be. 

The info is there. 
Screenshot from 2021-02-08 20:08:58

rules/dead-code/tests/Rector/Class_/RemoveUnusedClassesRector/Fixture/skip_class_extend.php.inc
TomasVotruba: Ah, so it's not about class that extends, but class that has children, right?
I got confused by the fixture name.

In that case the `Vehicle` should be moved first, and named `SkipClassThatHasUsedChildren`. Then reflection check on:

```
$classReflection = $scope->getClassReflection();

if ($classReflection->getAncestors()) {
    return [];
}
```

tomasnorre: Then reflection check on:

```php
$classReflection = $scope->getClassReflection();

if ($classReflection->getAncestors()) {
    return [];
}
```

Not sure what you mean here? Where should this be done/checked?

TomasVotruba: Probably here https://github.com/rectorphp/rector/blob/9f6834c46ea7599446dce808d21a069a4a6a9189/rules/dead-code/src/Rector/Class_/RemoveUnusedClassesRector.php#L118

samsonasik:
@tomasnorre I cherry-picked your commit at PR https://github.com/rectorphp/rector/pull/5498

PR #5461, opened 2021-02-08
Add failing test fixture for StringifyStrNeedlesRector

ruudk:
# Failing Test for StringifyStrNeedlesRector

Based on https://getrector.org/demo/d2da9632-23c3-4f71-b800-219c8871639c

It should understand that `PHP_EOL` is a constant that is always a `string`. Therefore, no need to cast it to string.

### DIFF
--- a/rules/php73/src/NodeTypeAnalyzer/NodeTypeAnalyzer.php
+++ b/rules/php73/src/NodeTypeAnalyzer/NodeTypeAnalyzer.php
@@ -9,6 +9,7 @@ use PHPStan\Type\Accessory\AccessoryNumericStringType;
 use PHPStan\Type\IntersectionType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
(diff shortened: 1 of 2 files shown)


TomasVotruba:
CI apart tests should be passing. There are 2 jobs that need fix first

ruudk:
Done

TomasVotruba:
Thanks :+1: 

This look like easy pick. Could you fix it here as well?

ruudk:
Does Rector know that `PHP_EOL` is a built-in constant? And what's type it is?

TomasVotruba:
You can try to dump it or use `$this->getValue($node)` on it.

What static type does it provide (`$this->getStaticType($node)`?

ruudk:
@TomasVotruba What do you think?

TomasVotruba:
I think amazing. You could not pick better line to add 💯

TomasVotruba:
Thanks

Issue #5462, opened 2021-02-08
Incorrect behavior of JsonThrowOnErrorRector


ruudk:
# Bug Report

| Subject        | Details              |
| :------------- | :--------------------|
| Rector version | last dev-master      |
| Installed as   | composer dependency  |

## Minimal PHP Code Causing Issue

See https://getrector.org/demo/22986871-208b-4f73-a1f3-21d9d11115a4

```php
<?php

final class DemoFile
{
    public function run() : array
    {
        $campaigns = json_decode($input['campaigns'] ?? '', true) ?? [];
        
        return $campaigns;
    }
}
```

### Responsible rules

* `JsonThrowOnErrorRector`

## Expected Behavior

Rector should skip this because there is a `?? []` fallback that is used when `json_decode` returns false.

TomasVotruba:
Thanks for report!

We'll need a failing test case for this one

ruudk:
My example is wrong anyway.... it uses the Null Coalescing Operator `??` which only falls back when the left side is `null`. 

That means that the fallback is not meant to handle the `json_decode` error.

But when I change it to use the Ternary Operator `?:` the result is the same.

I will create a test for that.

TomasVotruba:
Closing as there is no need anymore: https://github.com/rectorphp/rector/pull/5469#issuecomment-779659279

PR #5469, opened 2021-02-09
Add failing test fixture for JsonThrowOnErrorRector

ruudk:
# Failing Test for JsonThrowOnErrorRector

Based on https://getrector.org/demo/c94533f5-3e58-4ffd-a813-3fc257248a70

See #5462 

When the Ternary Operator `?:` is used, it should not rewrite the code as it prevents the fallback from being used.

## Idea

Maybe it should add a try / catch and set the fallback?

```php
<?php

namespace Rector\Php73\Tests\Rector\FuncCall\JsonThrowOnErrorRector\Fixture;

final class TernaryOperator
{
    public function run() : array
    {
        try {
            $campaigns = json_decode($input['campaigns'] ?? '', true, 512, JSON_THROW_ON_ERROR) ?: [];
        } catch (\JsonException $exception) {
            $campaigns = [];
        }
        
        return $campaigns;
    }
}
?>
```

### DIFF
--- a/rules/php73/tests/Rector/FuncCall/JsonThrowOnErrorRector/Fixture/ternary_perator.php.inc
+++ b/rules/php73/tests/Rector/FuncCall/JsonThrowOnErrorRector/Fixture/ternary_perator.php.inc
@@ -1,0 +1,14 @@
+<?php
+
+namespace Rector\Php73\Tests\Rector\FuncCall\JsonThrowOnErrorRector\Fixture;
+
+final class TernaryOperator
+{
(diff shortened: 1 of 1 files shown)


TomasVotruba:
> Maybe it should add a try / catch and set the fallback?

How would you refactor code manually?

TomasVotruba:
Any updates?

ruudk:
I don't know/want to fix this at the moment. Shall we close?

TomasVotruba:
Sure :+1:

PR #5475, opened 2021-02-09
Add failing test fixture for SimplifyUselessVariableRector

ruudk:
# Failing Test for SimplifyUselessVariableRector

Based on https://getrector.org/demo/692bfec6-24af-4f45-a6d6-ededa13d116a

It should understand that the variable is appended a couple of times.... this doesn't improve the code at all.

### DIFF
--- a/rules/code-quality/tests/Rector/Return_/SimplifyUselessVariableRector/Fixture/demo_file.php.inc
+++ b/rules/code-quality/tests/Rector/Return_/SimplifyUselessVariableRector/Fixture/demo_file.php.inc
@@ -1,0 +1,18 @@
+<?php
+
+namespace Rector\CodeQuality\Tests\Rector\Return_\SimplifyUselessVariableRector\Fixture;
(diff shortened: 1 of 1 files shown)


TomasVotruba:
The fixture name `DemoFile` is not very helpful. Should it be skipped or modified in some way? What is the use case? That should be included in the class/file name already.

It should be either `Skip*`, or have the other half of diff completed.

ruudk:
I understand. But this is what happens when you quickly test this on demo and then click "Create PR". Would be great if it would ask you to describe the problem, and ask you to give it a better classname... 

I will be working on this over the next days when I have time.

TomasVotruba:
> Would be great if it would ask you to describe the problem, and ask you to give it a better classname...

Agreed. The Github link features are limited to passing a string. How would you do it?

ruudk:
Not interested in fixing it at this time. Maybe later.

PR #5480, opened 2021-02-09
[EarlyReturn] Register ChangeAndIfContinueToMultiContinueRector to early-return config set

samsonasik:


src/PhpParser/Printer/BetterStandardPrinter.php
```diff
@@ -519,10 +519,12 @@ private function resolveNewStmts(array $stmts): array
     private function cleanUpStmts(array $stmts): array
     {
         foreach ($stmts as $key => $stmt) {
-            if (! $stmt instanceof ClassLike && ! $stmt instanceof FunctionLike) {
+            if ($stmt instanceof ClassLike) {
+                continue;
+            }
+            if ($stmt instanceof FunctionLike) {
```
samsonasik, line 525: @TomasVotruba the `BooleanNot` check seems not valid, ref https://3v4l.org/Ip0J6 vs https://3v4l.org/MFf4O

samsonasik: The truthy check also not works, ref https://3v4l.org/5IIDX vs https://3v4l.org/gRqe5

samsonasik: @TomasVotruba I think the rule is invalid, it will only works for `BooleanOr`, refs :

https://3v4l.org/UZ2rD vs https://3v4l.org/MFf4O
https://3v4l.org/ZlWch vs https://3v4l.org/L8XB8

While in `BooleanAnd`, that's invalid:

https://3v4l.org/Ip0J6 vs https://3v4l.org/MFf4O
https://3v4l.org/5IIDX vs https://3v4l.org/gRqe5

The check should be for `BooleanOr`, with change name to `ChangeOrIfContinueToMultiContinueRector` instead.

PR #5484, opened 2021-02-09
WIP: Do not upgrade nette to 3.1 in 3.0 set

lulco:
I am upgrading Nette application from 2.4 with NETTE_30 set, but it is directly upgrading to nette 3.1
Probably we will need to change some other packages. I have to find the best combination.

### DIFF
--- a/config/set/nette-30-composer.php
+++ b/config/set/nette-30-composer.php
@@ -16,10 +16,10 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     $services->set(ChangePackageVersionComposerRector::class)
         ->call('configure', [[
             ChangePackageVersionComposerRector::PACKAGES_AND_VERSIONS => ValueObjectInliner::inline([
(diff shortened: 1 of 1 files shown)


TomasVotruba:
This is rather convention-related. Composer promotes `^` for ages, so we should stick with that. Mixing 2 approaches in one `composer.json` open new layer for surprise bugs.

Also Nette 3.0 and 3.1 should be compatible, so this is irelevant.

lulco:
Composer can promote what it wants, but Nette doesn’t follow semver for ages. It is the same now. Versions 3.0 and 3.1 are not compatible, there are several BC breaks. So it is required to use ~ instead of ^. If you don’t want to mix them, we can change all. 

I just wanted to fix it for other developers :) nevermind I can change it locally in my config

TomasVotruba:
Then it's problem of Nette development that must be addressed and worked on there. Project have to be responsible on their own

PR #5493, opened 2021-02-10
[Renaming] Handle RenameMethodRector on new in caller class

samsonasik:
Close #5449

### DIFF
--- a/rules/renaming/src/Rector/MethodCall/RenameMethodRector.php
+++ b/rules/renaming/src/Rector/MethodCall/RenameMethodRector.php
@@ -139,6 +139,14 @@ CODE_SAMPLE
 
     private function shouldSkipForExactClassMethodForClassMethod(ClassMethod $classMethod, string $type): bool
     {
+        $className = $classMethod->getAttribute(AttributeKey::CLASS_NAME);
+        $methodCalls = $this->nodeRepository->findMethodCallsOnClass($className);
+
+        $name = $this->getName($classMethod->name);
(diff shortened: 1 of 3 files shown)


samsonasik:

ruudk (approved):
🙌

TomasVotruba:
Thank you :+1:

ruudk:
Thanks for fixing it. Wanted to give it a try but it doesn't work on the latest version. Turns out it's not tagged yet. 

1) Could it be tagged?
2) Would it be possible to have rector-prefixed always in sync with rector master? Then I can just pull rector-prefix dev-master.

ruudk:
Also, what's the tagging policy? I thought this happened automatically. Or is that a manual process?

TomasVotruba:
The 0.x branch is with BC breaks, so having it tagged on every commit could cause many broken versions.
Thus the tagging is manual.

For now, use the `dev-master` version.

ruudk:
Is it possible then to sync rector-prefixed on every commit with rector master? I cannot install rector master as it conflicts with my installed phpstan version :(

TomasVotruba:
It should be synced on every commit: https://github.com/rectorphp/rector-prefixed, just not tagged

TomasVotruba:
What version do you have?

```bash
vendor/bin/rector --version
```

ruudk:
Amazing, it __does__ work! 

That means that I just verified that this PR fixed my problem! 👏 Thanks @samsonasik  @TomasVotruba

PR #5494, opened 2021-02-10
[Testing] Add autoload option to community test case

TomasVotruba:


packages/testing/src/PHPUnit/AbstractCommunityRectorTestCase.php
```diff
@@ -58,12 +58,13 @@ protected function setUp(): void
         $this->parameterProvider = $this->getService(ParameterProvider::class);
     }
 
-    protected function doTestFileInfo(SmartFileInfo $fixtureFileInfo): void
+    protected function doTestFileInfo(SmartFileInfo $fixturaeFileInfo, bool $shouldAutoload = true): void
```
samsonasik, line 61: @TomasVotruba it seems typo ? `fixturaeFileInfo` -> `fixtureFileInfo`

samsonasik: it seems already fixed in master https://github.com/rectorphp/rector/commit/2f907222a39b0f6d5ac368df7fcd5d7179ff5e7d 👍

PR #5500, opened 2021-02-11
Show diff of changed non-php files instead of simple message

lulco:


TomasVotruba:
Hi, could you provide some description and screenshot? Not sure what this does

lulco:
Sure, if I use RenameClassRector is is also used NonPhpFileClassRenamer in NonPhpFileProcessor. The output of renaming classes in non-php files is just simple message "! [NOTE] File "app/config/config.neon" would be changed ("dry-run" is on now)":
Screenshot 2021-02-11 at 13 15 58

This PR changes it to nice real diff:
Screenshot 2021-02-11 at 13 16 51

TomasVotruba:
Ah, that's purfect :+1:

PR #5502, opened 2021-02-11
Duplicate calls to configure should merge

ruudk:
Failing test for #5496

PR #5503, opened 2021-02-11
[Renaming] Handle duplicate ->call() should merge on RenameClassRector

### LINKED ISSUE #5496: RenameClassRector doesn't work when PHPUNIT_60 and PHPUNIT_70 sets are enabled together
# Bug Report

I ran into a very weird issue. The `RenameClassRector` doesn't work when both `PHPUNIT_60` and `PHPUNIT_70` are enabled. When only enabling one of the two, it works. When removing both, it works. 🤯 

| Rector version | latest              |

:x: Both sets enabled: https://getrector.org/demo/a86c11df-eef0-4ff2-af96-2a4651ad6a51
✅  Only PHPUNIT_60 set enabled: https://getrector.org/demo/bb129f1f-2897-4fdf-9ae8-b4ae1f38c3ee
✅  Only PHPUNIT_70 set enbled: https://getrector.org/demo/10e90400-1473-4f84-a1b6-d42f930fe3a6
✅  No sets enabled: https://getrector.org/demo/687214ff-ae83-4ffb-92bf-d67288b32fe1

## Expected Behaviour

It should change my code

samsonasik:
Closes #5502 Fixes #5496

### DIFF
--- a/src/DependencyInjection/CompilerPass/MergeImportedRectorConfigureCallValuesCompilerPass.php
+++ b/src/DependencyInjection/CompilerPass/MergeImportedRectorConfigureCallValuesCompilerPass.php
@@ -58,7 +58,13 @@ final class MergeImportedRectorConfigureCallValuesCompilerPass implements Compil
         $mergedConfigure = [];
 
         foreach ($configuration as $configure) {
-            $mergedConfigure = array_merge_recursive($mergedConfigure, $configure[1][0] ?? []);
+            if ($methodCall[0] !== 'configure') {
+                continue;
+            }
+
+            foreach ($methodCall[1] as $configureMethodCall) {
+                $mergedConfigure = array_merge_recursive($mergedConfigure, $configureMethodCall);
+            }
         }
 
         return $mergedConfigure;
--- a/src/DependencyInjection/CompilerPass/MergeImportedRectorConfigureCallValuesCompilerPass.php
+++ b/src/DependencyInjection/CompilerPass/MergeImportedRectorConfigureCallValuesCompilerPass.php
@@ -35,12 +35,26 @@ final class MergeImportedRectorConfigureCallValuesCompilerPass implements Compil
 
     private function completeCollectedArguments(string $serviceClass, Definition $definition): void
     {
-        $configureCallValues = $this->configureCallValuesCollector->getConfigureCallValues($serviceClass);
+        $configureCallValues = $definition->hasMethodCall(self::CONFIGURE_METHOD_NAME)
+            ? $this->mergeConfigure($definition->getMethodCalls())
+            : $this->configureCallValuesCollector->getConfigureCallValues($serviceClass);
+
         if ($configureCallValues === []) {
             return;
         }
 
         $definition->removeMethodCall(self::CONFIGURE_METHOD_NAME);
         $definition->addMethodCall(self::CONFIGURE_METHOD_NAME, [$configureCallValues]);
+    }
+
+    private function mergeConfigure(array $configuration): array
+    {
+        $mergedConfigure  = [];
+
+        foreach ($configuration as $configure) {
+            $mergedConfigure = array_merge_recursive($mergedConfigure, $configure[1][0]);
+        }
+
+        return $mergedConfigure;
     }
 }
--- a/src/DependencyInjection/CompilerPass/MergeImportedRectorConfigureCallValuesCompilerPass.php
+++ b/src/DependencyInjection/CompilerPass/MergeImportedRectorConfigureCallValuesCompilerPass.php
@@ -58,11 +58,11 @@ final class MergeImportedRectorConfigureCallValuesCompilerPass implements Compil
         $mergedConfigure = [];
 
         foreach ($configuration as $configure) {
-            if ($methodCall[0] !== 'configure') {
+            if ($configure[0] !== 'configure') {
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


samsonasik:
Other rule may need merge configure as well, this is the demo for `RenameClassRector`

samsonasik:

ruudk:
Shouldn't this be fixed in MergeImportedRectorConfigureCallValuesCompilerPass? Otherwise it will happen a lot more when new Rectors are added.

samsonasik:
Yes, it seems.

OskarStark:
Also this fixes Old to new class, but not old to new method (which is my bug), right?

samsonasik:
Applied the fix in `MergeImportedRectorConfigureCallValuesCompilerPass` https://github.com/rectorphp/rector/pull/5503/commits/c81f6b756ca21d1763176620c567b4fca493538d

samsonasik:
the `ConfigurableRectorImportConfigCallsMergeTest` seems need to be updated, something-something `@runInSeparateProcess`

samsonasik:
fixed 🎉

samsonasik:

TomasVotruba:
@samsonasik Thanks :+1: 

@OskarStark @ruudk Could you confirm this change fixes your use cases?

ruudk:
I cannot test it, but since my failing test passes I assume it's Ok. I will test it as soon as this is in master and rector-prefixed is synced.

samsonasik:
You should can test with `composer update rector/rector:dev-close-5502`

ruudk:
On my project I'm using rector-prefixed as rector/rector refuses to install because it conflicts with phpstan.

TomasVotruba:
@ruudk You can copy paste these couple of lines to your local class. That would verify it.

ruudk:
Sure, will check tonight

ruudk:
It does not work 😞

I added a var_dump in `completeCollectedArguments` and I can see that the `$configureCallValues` value holds the PHPUnit values and not my values.

I think the issue is related to how sets are loaded vs how the project config is loaded.

There should be a test case where rector loads from this config:
```php
<?php

        SetList::PHPUNIT_60,
        SetList::PHPUNIT_70,
    ]);

    $services->set(RenameMethodRector::class)
        ->call('configure', [[
            RenameMethodRector::METHOD_CALL_RENAMES => ValueObjectInliner::inline([
                new MethodCallRename('TicketSwap\**\Handler\*', 'handle', '__invoke'),
                new MethodCallRename('TicketSwap\**\Subscriber\*', 'notify', '__invoke'),
                new MethodCallRename('TicketSwap\**\Query\*', 'handle', '__invoke'),
            ]),
        ]]);
};

```

You'll see that the fix only works for the sets 'configures' and not for the configure that is defined in the main config itself.

src/DependencyInjection/CompilerPass/MergeImportedRectorConfigureCallValuesCompilerPass.php
```diff
@@ -35,12 +35,32 @@ public function process(ContainerBuilder $containerBuilder): void
...
+    /**
+     * @param array<int, array<int, mixed>> $configuration
+     * @return mixed[]
+     */
+    private function mergeConfigure(array $configuration): array
+    {
+        $mergedConfigure = [];
+
+        foreach ($configuration as $configure) {
+            $mergedConfigure = array_merge_recursive($mergedConfigure, $configure[1][0] ?? []);
```
ruudk, line 61: It should be like this:
```suggestion
            if ($methodCall[0] !== 'configure') {
                continue;
            }

            foreach ($methodCall[1] as $configureMethodCall) {
                $mergedConfigure = array_merge_recursive($mergedConfigure, $configureMethodCall);
            }
```

Beause `$configuration` are __all method calls__ we need to filter out anything that is not `configure`.

And then we should loop over `$methodCall[1]` because it can already contain more than 1 call.

samsonasik: applied

samsonasik:
@ruudk I applied your patch, is it working now?

ruudk:
I'm testing it right now.

lulco:
not working for me... now it is like "last registered wins"

samsonasik:
@lulco One more try with re-merge to config from configure call values collector https://github.com/rectorphp/rector/pull/5503/commits/c2d842fe43af55dca76b8904bbeae8645ed196f9 , please try again.

lulco:
One question before test, should it affect all rectors with configure method? I am testing it on Rector\Composer\Rector\ChangePackageVersionRector and it seems like there are some merges done but not the way I expected.

samsonasik:
I am not sure, probably need a real failing fixture to show it.

ruudk:
Oke I found out why things are working weird... https://github.com/symfony/symfony/pull/40167

samsonasik:
Oh, last commit https://github.com/rectorphp/rector/commit/c2d842fe43af55dca76b8904bbeae8645ed196f9 won't work as well since it check its emptyness first.

ruudk:
I was already debugging this for an hour or so. I think I can fix it now.

samsonasik:
@ruudk ok, feel free to continue your PR https://github.com/rectorphp/rector/pull/5502 or create new PR for it

ruudk:
Ok I'll apply it on my PR.

samsonasik:
Ok, I am closing it.

PR #5511, opened 2021-02-11
Add failing test fixture for RenameMethodRector: Check visibility when renaming to magic method __invoke

ruudk:
# Failing Test for RenameMethodRector

Based on https://getrector.org/demo/38775b0b-7f57-4f18-97ac-b39a71fa6d78

It should detect that the old method is not public and that the new magic method `__invoke` can only be applied when public.

### DIFF
--- a/rules/renaming/tests/Rector/MethodCall/RenameMethodRector/Fixture/rename_class_method.php.inc
+++ b/rules/renaming/tests/Rector/MethodCall/RenameMethodRector/Fixture/rename_class_method.php.inc
@@ -2,7 +2,7 @@
 
 namespace Rector\Renaming\Tests\Rector\MethodCall\RenameMethodRector\Fixture;
 
-class NetteToSymfonyPresenter
+class RenameClassMethod
 {
     public function run()
     {
@@ -15,7 +15,7 @@ class NetteToSymfonyPresenter
 
 namespace Rector\Renaming\Tests\Rector\MethodCall\RenameMethodRector\Fixture;
 
-class NetteToSymfonyPresenter
+class RenameClassMethod
 {
     public function __invoke()
     {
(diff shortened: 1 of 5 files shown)


ruudk:
Is there a way to detect if something is a built-in magic method and infer the visibility from there?

Then I could try to fix it.

TomasVotruba:
The class method is unsed `ClassMethod` node.

You can make use of `flags`: https://github.com/rectorphp/php-parser-nodes-docs/blob/master/docs/nodes_overview.md#public-properties-53

It has also some useful methods:

- https://github.com/nikic/PHP-Parser/blob/8165cf69fab95ade34cb73d1dc1c23d08b57cbb2/lib/PhpParser/Node/Stmt/ClassMethod.php#L116-L118


And method that can be re-used for magic method checks:

- https://github.com/nikic/PHP-Parser/blob/8165cf69fab95ade34cb73d1dc1c23d08b57cbb2/lib/PhpParser/Node/Stmt/ClassMethod.php#L153

ruudk:
It seems that this is already fixed. I cannot reproduce it anymore on getrector.org also. 🎉

ruudk:
Turns out I couldn't reproduce because the demo is broken https://github.com/rectorphp/getrector.org/issues/270

samsonasik:
it seems now working https://getrector.org/demo/ee44d8d4-3b94-4f8d-b37a-83e914fffd15 ?

ruudk:
No, __invoke is a magic function that never can be private or protected.

https://3v4l.org/VbJR4

samsonasik:
That's because `VisibilityOnMagicMethod` has `*` prefix, re-execute with remove `*` prefix seems working:

https://getrector.org/demo/47d578fd-cfe8-4f0b-b10b-cd58be4097aa

For `*` prefix, seems related to other PR https://github.com/rectorphp/rector/pull/5564

samsonasik:
This seems conflict with current master, I will try if I can reproduce it, thank you 👍

samsonasik:
@ruudk I created PR https://github.com/rectorphp/rector/pull/5586 for it.

PR #5520, opened 2021-02-12
Failing test for Fully qualified name of function is imported as use instead of use function

lulco:
Description in https://github.com/rectorphp/rector/issues/5521


TomasVotruba:
I see. We'll need a test fixture without creating a new Rector rule.
Please re-use existing one with auto import enabled.

rules/coding-style/tests/Rector/Use_/UseFunction/Source/functions.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Deployer;
+
+function set($name, $value)
+{
+
+}
```
TomasVotruba, line 10: Just a tip for your project: this is not included/autoloaded. 
It must be in `composer.json` > `files`

Without that, functions are invisible :)

lulco: My project uses deployer and it has this in composer.json https://github.com/deployphp/deployer/blob/master/composer.json#L33

I was just trying to create some minimal reprodution of the issue.

TomasVotruba: The same must happen here too

lulco:
@TomasVotruba I used RenameFunctionRector with auto import names, you can check it now.

rules/renaming/tests/Rector/FuncCall/RenameFunctionAutoimportRector/config/configured_rule.php
```diff
@@ -1,15 +1,18 @@
...
 return static function (ContainerConfigurator $containerConfigurator): void {
     $parameters = $containerConfigurator->parameters();
     $parameters->set(Option::AUTO_IMPORT_NAMES, true);
 
     $services = $containerConfigurator->services();
-    $services->set(UseFunctionRector::class);
+    $services->set(RenameFunctionRector::class)
+        ->call('configure', [[
+            RenameFunctionRector::OLD_FUNCTION_TO_NEW_FUNCTION => [
+                'view' => 'Laravel\Templating\render',
```
TomasVotruba: Where is the `Laravel\Templating\render` autoloaded in tests? I can't find it

rules/renaming/tests/Rector/FuncCall/RenameFunctionAutoimportRector/RenameFunctionAutoimportRectorTest.php
```diff
@@ -0,0 +1,30 @@
...
+use Symplify\SmartFileSystem\SmartFileInfo;
+
+final class RenameFunctionAutoimportRectorTest extends AbstractRectorTestCase
+{
+    /**
+     * @dataProvider provideData()
+     */
+    public function test(SmartFileInfo $fileInfo): void
+    {
+        $this->doTestFileInfo($fileInfo);
```
TomasVotruba, line 18: Could you verify the 'Symfony\Component\DependencyInjection\Loader\Configurator\service' is loaded?



```php
$this->assertTrue(function_exists('...'));
```

rules/renaming/tests/Rector/FuncCall/RenameFunctionAutoimportRector/config/configured_rule.php
lulco: This is an old commit. First I was trying it with the same set as in existing renaming tests. But rector just change names and didn't use auto import names, so that's why I used `service` function

rules/renaming/tests/Rector/FuncCall/RenameFunctionAutoimportRector/RenameFunctionAutoimportRectorTest.php
lulco: assert added

TomasVotruba:
I can't find your other long comment about function import.
But there is definitely space for imporovement in name importer. 

Not depend on `function_exists`, but only on context:
https://github.com/rectorphp/rector/blob/59e1063b50b9271e5c7059c210049b8b1f8bafa8/rules/coding-style/src/Node/NameImporter.php#L231

TomasVotruba:
Suprassed by https://github.com/rectorphp/rector/pull/5755

PR #5523, opened 2021-02-12
Fix merging of multiple calls to configure in sets

ruudk:
I spent a lot of time debugging this but when I fixed one issue another one popped up 🤯 

It's all related to the loading of sets that is different from the normal rector.php configuration. 

There were 3 failing scenarios:
- ✅ load one set that configures RenameClassRector + configure your own
- ❌ load two sets that configure both RenameClassRector ( they should merge, they don't )
- ❌ load two sets that configure both RenameClassRector + configure your own

Chained configure calls inside your own config were already working fine. 

@samsonasik @TomasVotruba Any advice? 

Another thing I noticed that drove me completely insane is that the tests are re-using the same container that is stored somewhere in a temp directory on my filesystem. In my case, `/private/var/folders/m4/2_8czq_s1xxxx0000gn/T/_rector`. From time to time I needed to manually purge that directory before new changes would popup. I really think that tests should not rely on those artifacts as that makes debugging issues very hard.

1) can we store the temp stuff relative to the project in `var`?
2) can we by default bust cache on start of the test?

### DIFF
--- a/tests/DependencyInjection/ConfigurableRectorImportConfigCallsMergeTest.php
+++ b/tests/DependencyInjection/ConfigurableRectorImportConfigCallsMergeTest.php
@@ -51,17 +51,17 @@ final class ConfigurableRectorImportConfigCallsMergeTest extends AbstractKernelT
         ];
 
         yield [
-            __DIR__ . '/config/main_config_with_own_value.php', [
+            __DIR__ . '/config/main_config_with_override_value.php', [
                 'old_2' => 'new_2',
                 'old_1' => 'new_1',
-                'old_3' => 'new_3',
             ],
         ];
 
         yield [
-            __DIR__ . '/config/main_config_with_override_value.php', [
+            __DIR__ . '/config/main_config_with_own_value.php', [
                 'old_2' => 'new_2',
                 'old_1' => 'new_1',
+                'old_3' => 'new_3',
             ],
         ];
     }
--- a/tests/DependencyInjection/ConfigurableRectorImportConfigCallsMergeTest.php
+++ b/tests/DependencyInjection/ConfigurableRectorImportConfigCallsMergeTest.php
@@ -98,29 +98,29 @@ final class ConfigurableRectorImportConfigCallsMergeTest extends AbstractRectorT
 
         yield [
             __DIR__ . '/config/two_sets.php', [
+                'Twig_SimpleFilter' => 'Twig_Filter',
+                'Twig_SimpleFunction' => 'Twig_Function',
+                'Twig_SimpleTest' => 'Twig_Test',
                 'PHPUnit_Framework_MockObject_Stub' => 'PHPUnit\Framework\MockObject\Stub',
                 'PHPUnit_Framework_MockObject_Stub_Return' => 'PHPUnit\Framework\MockObject\Stub\ReturnStub',
                 'PHPUnit_Framework_MockObject_Matcher_Parameters' => 'PHPUnit\Framework\MockObject\Matcher\Parameters',
                 'PHPUnit_Framework_MockObject_Matcher_Invocation' => 'PHPUnit\Framework\MockObject\Matcher\Invocation',
                 'PHPUnit_Framework_MockObject_MockObject' => 'PHPUnit\Framework\MockObject\MockObject',
                 'PHPUnit_Framework_MockObject_Invocation_Object' => 'PHPUnit\Framework\MockObject\Invocation\ObjectInvocation',
-                'Twig_SimpleFilter' => 'Twig_Filter',
-                'Twig_SimpleFunction' => 'Twig_Function',
-                'Twig_SimpleTest' => 'Twig_Test',
             ],
         ];
 
         yield [
             __DIR__ . '/config/two_sets_with_own_rename.php', [
+                'Twig_SimpleFilter' => 'Twig_Filter',
+                'Twig_SimpleFunction' => 'Twig_Function',
+                'Twig_SimpleTest' => 'Twig_Test',
                 'PHPUnit_Framework_MockObject_Stub' => 'PHPUnit\Framework\MockObject\Stub',
                 'PHPUnit_Framework_MockObject_Stub_Return' => 'PHPUnit\Framework\MockObject\Stub\ReturnStub',
                 'PHPUnit_Framework_MockObject_Matcher_Parameters' => 'PHPUnit\Framework\MockObject\Matcher\Parameters',
                 'PHPUnit_Framework_MockObject_Matcher_Invocation' => 'PHPUnit\Framework\MockObject\Matcher\Invocation',
                 'PHPUnit_Framework_MockObject_MockObject' => 'PHPUnit\Framework\MockObject\MockObject',
                 'PHPUnit_Framework_MockObject_Invocation_Object' => 'PHPUnit\Framework\MockObject\Invocation\ObjectInvocation',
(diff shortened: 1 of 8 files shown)


TomasVotruba:
My guess is somewhere in import method of `FileLoader` and `ConfigureCallValuesCollector`.

> 1. can we store the temp stuff relative to the project in var?

It might make sense for tests, but not for production use. If you can make it for tests only, then :+1: 

> 2. can we by default bust cache on start of the test?

That would kill the point of container cache. Now bare tests run 1,5 mins, while on 2nd run it's 15 seconds. 

There a way to optimize the cache invalidation based on provided resources:

- https://symfony.com/doc/current/components/config/caching.html
- https://symfony.com/doc/current/performance.html#dump-the-service-container-into-a-single-file

But that would not help, as CompilerPass is special uncacheable part of Symfony DI. When working with CompilerPasses, you always have to clear cache manually.

ruudk:
@TomasVotruba @samsonasik I think I cracked the mystery 🎉 

Tests are green. And I tested this on my own project where I found this issue, and it solves it.

Questions:

1) Is this the right approach?

2) What do you think about https://github.com/rectorphp/rector/pull/5523/commits/7a1b37379bd137e836e0be0a822989178f51b324

tests/DependencyInjection/ConfigurableRectorImportConfigCallsMergeTest.php
```diff
@@ -31,41 +33,95 @@ protected function setUp(): void
...
             ],
         ];
 
         yield [
-            __DIR__ . '/config/main_config_with_override_value.php', 2, [
+            __DIR__ . '/config/main_config_with_own_value.php', [
                 'old_2' => 'new_2',
                 'old_1' => 'new_1',
+                'old_4' => 'new_4',
+                'old_3' => 'new_3',
```
samsonasik, line 72: is 4 and 3 order flipped ?

ruudk: See per commit view > https://github.com/rectorphp/rector/pull/5523/commits/00d3d0834feb7d2a1389f6b59167f33ff018535e

I moved the test so follow the ordering of the files on disk.

samsonasik (approved):

ruudk:
@samsonasik What do you think about 7a1b373

samsonasik:
@ruudk I think as far as there is no replaced config, flip order is ok, unless we found a real use case when flip order make an issue. @TomasVotruba what do you think?

TomasVotruba:
Rector rules should be idempotent, so order does not matter.

Thanks for the fix :+1:

lulco:
If I want to override some value, the order does matter.

For example:
Let's say SomeSet in Rector renames ClassA to ClassB (with other 50 renames), but I want to rename ClassA to ClassC.

So I use:
```
        SetList::SomeSet,
    ]);
```
And then:
```
$services->set(RenameClassRector::class)->call('configure', [[
        RenameClassRector::OLD_TO_NEW_CLASSES => [
            'ClassA' => 'ClassC',
        ]
    ]]);
```
to rewrite just this one setting. Keep other 50 from set as they are.

But I can't do it because it looks like the rules are always run in order:
- imports
- local
- sets

Imho at least local settings should be run as the last.

lulco:
Hm now I realize that `RenameClassRector` maybe should work in order
- local
- imports
- sets
because after renaming the original class name will not exist in code...

But for example `ChangePackageVersionComposerRector` should use another order because we need the last change to win, so here it is
- imports
- sets
- local

Is there any way how to achieve it?

TomasVotruba:
For Rector rules, there is convention of merging parameters, regardless the order.
Only that way nested sets can work together.

lulco:
Maybe we need more debugging here, but my first try would be: it doesn't work with ValueObjects, because they has no keys so the final merged array contains all items and in foreach they are all executed.

```
return static function (ContainerConfigurator $containerConfigurator): void
{
        SetList::NETTE_30,
    ]);

    $services->set(ChangePackageVersionComposerRector::class)
        ->call('configure', [[
            ChangePackageVersionComposerRector::PACKAGES_AND_VERSIONS => ValueObjectInliner::inline([
                new PackageAndVersion('nette/application', '~3.0.6'),
                new PackageAndVersion('latte/latte', '~2.8.3'),
            ])
        ]]);
}
```

Output of ChangePackageVersionComposerRector::refactor foreach
```
string(27) "nette/application -> ~3.0.6"
string(21) "latte/latte -> ~2.8.3"
string(19) "nette/nette -> ^3.0"
string(27) "nette/application -> ^3.0.6"
string(23) "nette/bootstrap -> ^3.0"
string(21) "nette/caching -> ^3.0"
string(29) "nette/component-model -> ^3.0"
string(22) "nette/database -> ^3.0"
string(16) "nette/di -> ^3.0"
string(20) "nette/finder -> ^2.5"
string(19) "nette/forms -> ^3.0"
string(18) "nette/http -> ^3.0"
string(18) "nette/mail -> ^3.0"
string(18) "nette/neon -> ^3.0"
string(27) "nette/php-generator -> ^3.0"
string(26) "nette/robot-loader -> ^3.0"
string(25) "nette/safe-stream -> ^2.4"
string(22) "nette/security -> ^3.0"
string(23) "nette/tokenizer -> ^3.0"
string(19) "nette/utils -> ^3.0"
string(19) "latte/latte -> ^2.5"
string(19) "tracy/tracy -> ^2.6"
string(42) "contributte/event-dispatcher-extra -> ^0.8"
string(41) "contributte/forms-multiplier -> 3.1.x-dev"
string(40) "radekdostal/nette-datetimepicker -> ^3.0"
```

First two are from my "local" rector.php config file. Rest is from nette30 set. Nette set will always rewrite nette/application to ^3.0.6 and latte/latte to ^2.5 because they are in array after my local values

Config:
```
return static function (ContainerConfigurator $containerConfigurator): void
{
    $services->set(ChangePackageVersionComposerRector::class)
        ->call('configure', [[
            ChangePackageVersionComposerRector::PACKAGES_AND_VERSIONS => ValueObjectInliner::inline([
                new PackageAndVersion('nette/application', '~3.0.6'),
                new PackageAndVersion('latte/latte', '~2.8.3'),
            ])
        ]]);

        SetList::NETTE_30,
    ]);
}
```

Has the same output, so how the "last overrides wins" work? For me it is not working at all.

TomasVotruba:
> First two are from my "local" rector.php config file. Rest is from nette30 set. Nette set will always rewrite nette/application to ^3.0.6 and latte/latte to ^2.5 because they are in array after my local values


Apart technical difference, I wonder, what is the use case for using set and override it's work like this?

TomasVotruba:
> Has the same output, so how the "last overrides wins" work? For me it is not working at all.

That works only for parameters, e.g:

```php
$parameters->set('some_key', 'some_value');

// later in another config, that is imported after

$parameters->set('some_key', 'another_value');
```

The `some_key` is now overriden with new value:

```php
$parameters->get('some_key'); // "another_value"
```

TomasVotruba:
@lulco I see what you mean. Feel free to change the order to comly with your use case. It makes sense :+1:

lulco:
Any hint how and where should I change it?

TomasVotruba:
Apart this pull-request, I'm not sure

ruudk:
Maybe start with creating a failing test PR so that others can help and try to fix it?

PR #5529, opened 2021-02-13
[EarlyReturn] Add ReturnAfterToEarlyBreakRector

### LINKED ISSUE #5218: [EarlyReturn] Foreach with continue, then return to early return
```diff
-$pathOK = false;

foreach ($pathConstants as $allowedPath) {
    if ($dirPath == $allowedPath) {
-        $pathOK = true;
+		 return false; 
-        break;
    }
}

-return $pathOK;
+return false;
```

samsonasik:
Fixes #5218

### DIFF
--- a/packages/better-php-doc-parser/src/PhpDocParser/BetterPhpDocParser.php
+++ b/packages/better-php-doc-parser/src/PhpDocParser/BetterPhpDocParser.php
@@ -191,7 +191,7 @@ final class BetterPhpDocParser extends PhpDocParser
         $docContent = $this->annotationContentResolver->resolveFromTokenIterator($originalTokenIterator);
 
         // fallback to original parser
-        if ($tagValueNode === null) {
+        if (! $tagValueNode instanceof PhpDocTagValueNode) {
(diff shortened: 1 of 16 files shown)


samsonasik:
@TomasVotruba the current `BetterNodeFinder::findFirstPrevious()` seems invalid as when it locate parent, the filter also filtered next of current node that located as already in parent. 

I will try to fix the `BetterNodeFinder::findFirstPrevious()` to able to correctly locate findFirstPrevious for the following use case locate `$pathOK` before foreach:

```php
        $pathOK = false;
        foreach ($pathConstants as $allowedPath) {
            if ($dirPath == $allowedPath) {
                $pathOK = true;
                break;
            }
        }

        return $pathOK;
```

samsonasik:
I keep existing `BetterNodeFinder::findFirstPrevious()` as changing it break other tests. 
I added new method `BetterNodeFinder::findFirstPreviousOfNode()` for to get previous node in previous node.

samsonasik:

TomasVotruba:
Thanks :+1:

Issue #5522, opened 2021-02-12
[Docker] Build of docker is failing for unknown reason


TomasVotruba:
![image](https://user-images.githubusercontent.com/924196/107799603-ea95bd00-6d5d-11eb-801b-c65391b40dcb.png)

See https://github.com/rectorphp/rector/runs/1888924456

---

It's happening since https://github.com/rectorphp/rector/commit/47354793691c35d27365b43b0aae33a939ba75d2, but randomly.
Another commit next day passed it green.

Now I have to restart the build (5 times so far) and it still does not get through. No idea what is going on.

@JanMikes Could you help? :)

JanMikes:
Hi, will look into it ;)

JanMikes:
Closing as seems to be fixed via #5531

PR #5531, opened 2021-02-13
[CI][Docker] Push all Docker image tags at once

JanMikes:
Possible fix to #5522


JanMikes:
@TomasVotruba im not sure this will 100% fix it, but i believe this is good start

TomasVotruba:
Let's try it. The :x: is still there even after 15 restarts. That's very frustrating :D

TomasVotruba:
Looks good 
image

:+1:

PR #5542, opened 2021-02-14
[EarlyReturn] Register ReturnAfterToEarlyOnBreakRector to early-return config set

samsonasik:


samsonasik:

TomasVotruba:
Smooth :) thanks :+1:

PR #5543, opened 2021-02-14
[BetterNodeFinder] [WIP] Refactor BetterNodeFinder::findFirstPrevious()

samsonasik:
Ref https://github.com/rectorphp/rector/pull/5529#issuecomment-778628430 the current behaviour of `BetterNodeFinder::findFirstPrevious()` is invalid as when fallback to parent, it re-select current node searched. 

It may need update the code using it.


samsonasik:
still not working, closing for now.

PR #5544, opened 2021-02-14
Prepared xdebug Docker image with instructions

JanMikes:
Similar as #5430 - prepared docker image for quickstart with debugging and guide how to.


TomasVotruba:
Thanks :+1:

PR #5553, opened 2021-02-15
[PHP 7.1 Scoped] Copy composer.lock to build

samsonasik:


samsonasik:
@TomasVotruba is there a way to copy downgraded composer.lock to root `rector-php71` build? I tried this https://github.com/rectorphp/rector/pull/5553/commits/5cc16531451ded0eb370d86d063e80aafec384be but not sure if composer.lock located there, possibly need disable ignore composer.lock in `.gitignore`?

TomasVotruba:
What is it needed for?

The result build should not contain `composer.lock` at all: https://github.com/rectorphp/rector-prefixed

samsonasik:
To install phpunit. Do we need manual download like in https://github.com/rectorphp/rector-php71/pull/1#discussion_r572751142 then ?

samsonasik:
closing to try with `wget` phpunit to get phpunit dep.

PR #5555, opened 2021-02-15
Use `Class_::getProperty()`

ruudk:
As https://github.com/nikic/PHP-Parser/pull/646 is tagged already.

### DIFF
--- a/src/NodeManipulator/ClassInsertManipulator.php
+++ b/src/NodeManipulator/ClassInsertManipulator.php
@@ -154,20 +154,9 @@ final class ClassInsertManipulator
         return false;
     }
 
-    /**
-     * Waits on https://github.com/nikic/PHP-Parser/pull/646
-     */
     private function hasClassProperty(Class_ $class, string $name): bool
     {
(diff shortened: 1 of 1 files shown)


samsonasik:
For reference https://github.com/nikic/PHP-Parser/commit/64f4d5b619e85bb9b2462aa4f6724cf55ee7fcb1

src/NodeManipulator/ClassInsertManipulator.php
```diff
@@ -154,20 +154,9 @@ private function hasClassConstant(Class_ $class, string $constantName): bool
...
-        foreach ($class->getProperties() as $property) {
-            if (! $this->nodeNameResolver->isName($property, $name)) {
-                continue;
-            }
-
-            return true;
-        }
-
-        return false;
+        return $class->getProperty($name) !== null;
```
samsonasik, line 159: probably better use : 

```suggestion
        return $class->getProperty($name) instanceof Property;
```

TomasVotruba:
Thank you

PR #5558, opened 2021-02-15
[DeadCode] Add RemoveEmptyAbstractClassRector

### LINKED ISSUE #3282: [DeadCode] Empty abstract class that does nothing
```diff
-class SomeClass extends SomeAbstractClass
+class SomeClass extends AnotherAbstractClass
{
}

-// nothing, nowhere used as type
-abstract class SomeAbstractClass extends AnotherAbstractClass
-{
-}

abstracst clas AnotherAbstractClass
{
     public function getName()
     { 
          return 'cowo';
     }
}
````

### PR DESCRIPTION
**Pull request by samsonasik**
Fixes #3282

### CODE REVIEW

**Comment by samsonasik** [comment]

File: rules/dead-code/tests/Rector/Class_/RemoveEmptyAbstractClassRector/Fixture/multi_extends.php.inc [hunk 1]
```diff
@@ -0,0 +1,43 @@
...
+     {
+        return 'name';
+     }
+}
+
+class MultiExtends extends AnotherAbstractClassMulti
+{
+}
+
+?>
```
TomasVotruba, line 43: I miss a text fixture like this, where abstact class is actually used and skipped

samsonasik: is the fixture incorrect?

TomasVotruba: This is fixture is correct. But we'll need a new one that uses the abstract class and is skipped

samsonasik: any example?

TomasVotruba: E.g.


```php
final class SomeClass extends AbstractClass
{
}


final class AnyClass
{
	public function process($some)
    {
        if ($some instanceof AbstractClass) {
        }
    } 
}
```

If we revome `AbstractClass` here, the condition would break

samsonasik: I added the fixture https://github.com/rectorphp/rector/pull/5558/commits/1e9fb8bdfde27050e7d14b6a5a6ecf109e1429c1 and it seems working ok 🎉 

Screen Shot 2021-02-16 at 03 59 04

samsonasik: The removal is only for abstract class extends abstract class like in issue description.

samsonasik: Ok, I got the failing fixture https://github.com/rectorphp/rector/pull/5558/commits/b40546f92e42dbb7819e5fdd7f3119730916895d

samsonasik: @TomasVotruba is there a way to class the `FullyQualified` node used not in extends, like vendor locker?

samsonasik: seems need to use BetterNodeFinder->find inside loop children

samsonasik: @TomasVotruba I tried to collect `FullyQualified` node when class is not abstract but seems overlapped as abstract class detected first when abstract class in the top of non-abstract class.

samsonasik: I've tried with `Instanceof_` to collect `FullyQualified` usage, and it overlapped as well, do you have idea to get it?

samsonasik: @TomasVotruba using `FileNode` make it works 🎉

samsonasik:

samsonasik:
I've added handling single empty abstract

samsonasik:

samsonasik:
rebased.

samsonasik:

rules/dead-code/tests/Rector/Class_/RemoveEmptyAbstractClassRector/Fixture/multi_extends.php.inc
TomasVotruba: > The removal is only for abstract class extends abstract class like in issue description.

Ah, I missed that.

> I've tried with Instanceof_ to collect FullyQualified usage, and it overlapped as well, do you have idea to get it?

This would have to be collected before running Rector rule, the same way `Class_` and other nodes are collected, see `RectorRepository`.

samsonasik: I can't found class `RectorRepository`, is it has the service somewhere? or I need to create a new service `RectorRepository` with `FileNode` ?

TomasVotruba: Let's merge it as it is to keep it one step at a time.

Please add a test fixture with 2 files in a next PR.

- first: one with class extending an abstract class + abstract class
- second: with abstract class usage

E.g. this test is using extra files technique: https://github.com/rectorphp/rector/blob/06906a06591c1c635669b269635352cfe4b36a2a/rules/autodiscovery/tests/Rector/FileNode/MoveInterfacesToContractNamespaceDirectoryRector/MoveInterfacesToContractNamespaceDirectoryRectorTest.php#L26

TomasVotruba: > I can't found class RectorRepository, is it has the service somewhere? or I need to create a new service RectorRepository with FileNode ?

I'm sorry, it's `NodeRepository`

samsonasik: Do we need new separate test class ? using `extraFiles` seems need manually added fixture files ?

samsonasik: `NodeRepository`::collect()` seems not collect FullyQualified node.

TomasVotruba: > Do we need new separate test class ? using extraFiles seems need manually added fixture files ?

Yes, that would be the best to keep it narrow :+1: 


> NodeRepository::collect()` seems not collect FullyQualified node.

It will have to be added. It should also collect simple `Name`, to cover cases like these:

```php
use App\Nested\AbstractClass;

public function useClass(-->AbstractClass<-- $abstractClass)
{
}
```

samsonasik: I will try.

PR #5561, opened 2021-02-15
[PHP 7.1 Scoped] install dev dependency as well to allow run phpunit

samsonasik:


samsonasik:
running on root `rector-php71` with `dev` dependency installed got the following error:

```bash
➜  rector-php71 git:(master) vendor/bin/phpunit 
This version of PHPUnit requires PHP >= 7.3.
You are using PHP 7.1.33 (/opt/homebrew/Cellar/php@7.1/7.1.33/bin/php).
```

The php version on build php 7.1 seems need to have php 7.1 as well.

TomasVotruba:
Here the wget PHPUnit for PHP 7.1+ must be used.

The PHPUnit 7 is the only one that works on PHP 7.1: https://phpunit.de/getting-started/phpunit-7.html

PR #5562, opened 2021-02-15
[PHP 7.1 Scoped] Use PHP 7.1 for build php 7.1

samsonasik:


samsonasik:
It got error:

```bash
Run ./.github/workflows/scripts/downgrade_packages.sh 7.1
4
No lock file found. Updating dependencies instead of installing from lock file. Use composer update over composer install if you do not have a lock file.
5
Loading composer repositories with package information
6
Updating dependencies
7
Your requirements could not be resolved to an installable set of packages.
8
Error: Your requirements could not be resolved to an installable set of packages.
9

10
  Problem 1
11
    - Root composer.json requires php ^7.3|^8.0 but your php version (7.1.33) does not satisfy that requirement.
```

TomasVotruba:
The codebase must be build in PHP 7.3+, as Rector needs it.

Only downgraded version can run on PHP 7.1+

samsonasik:
ok, reverting at https://github.com/rectorphp/rector/pull/5563

PR #5554, opened 2021-02-15
[RenameMethodRector] Skip methods that are part of interface

ruudk:
# Failing Test for RenameMethodRector

Based on https://getrector.org/demo/3df11204-06c7-4fec-b909-e3acdffb23b1

I am able to prevent the rename when in the `ClassMethod` context... but I'm not able to do this for the `MethodCall`. 

@samsonasik Any advice?

PR #5564, opened 2021-02-15
Do not rename methods from interface when using wildcard

ruudk:
Failing test for https://github.com/rectorphp/rector/pull/5554

cc @samsonasik

### DIFF
--- a/rules/renaming/tests/Rector/MethodCall/RenameMethodRector/config/configured_rule.php
+++ b/rules/renaming/tests/Rector/MethodCall/RenameMethodRector/config/configured_rule.php
@@ -25,6 +25,11 @@ return static function (ContainerConfigurator $containerConfigurator): void {
                     'old',
                     'new'
                 ),
+                new MethodCallRename(
+                    'Rector\Renaming\Tests\Rector\MethodCall\RenameMethodRector\Fixture\*WildcardSubscriber',
+                    'old',
+                    'new'
+                ),
                 new MethodCallRename('*Presenter', 'run', '__invoke'),
                 new MethodCallRename(
                     \Rector\Renaming\Tests\Rector\MethodCall\RenameMethodRector\Fixture\SkipSelfMethodRename::class,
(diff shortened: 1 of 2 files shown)


rules/renaming/tests/Rector/MethodCall/RenameMethodRector/config/configured_rule.php
```diff
@@ -25,6 +25,11 @@
                     'old',
                     'new'
                 ),
+                new MethodCallRename(
+                    'Rector\Renaming\Tests\Rector\MethodCall\RenameMethodRector\Fixture\*WildcardSubscriber',
```
TomasVotruba, line 29: Could you provide `https://3v4l.org` proof that the `fnmatch()` work with expected class?
That would narrow the scope for the issue

ruudk: Good one. I checked, and it does work: https://3v4l.org/4tCj4

ruudk: The problem is that `getOldClass()` still contains the wildcard and is then passed to `ClassManipulator::hasParentMethodOrInterface`.

Screenshot 2021-02-16 at 09 13 46@2x

TomasVotruba: What is needed to fix it?

ruudk: I don't know. 

@samsonasik added this: https://github.com/rectorphp/rector/blob/8d58a6dd87a0618b70b3cc6be979f4f6f9d843b1/rules/renaming/src/Rector/MethodCall/RenameMethodRector.php#L88-L94

But that doesn't have support for `fnmatch`. 

I think that we should not use `$methodCallRename->getOldClass()` as it can be a FQCN or wildcard matching.

Instead, we should check if the Class `$node` has interfaces that implement `$methodCallRename->getOldMethod()` method. If that's the case, it should not rename. But I tried this already and I couldn't get it to work.

samsonasik:
@ruudk I cherry-picked your commit at PR https://github.com/rectorphp/rector/pull/5622

PR #5569, opened 2021-02-16
[Bin] Add PHPStan Stub ReflectionUnionType and Attribute to bin/rector

samsonasik:
To avoid the following error in PHP < 8.0

```bash
"Class 'ReflectionUnionType' not found". 
```

### DIFF
--- a/bin/rector
+++ b/bin/rector
@@ -1,4 +1,8 @@
 #!/usr/bin/env php
 <?php
 
+// @see https://github.com/phpstan/phpstan/issues/4541#issuecomment-779434916
+require_once 'phar://vendor/phpstan/phpstan/phpstan.phar/stubs/runtime/ReflectionUnionType.php';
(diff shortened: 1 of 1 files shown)


TomasVotruba:
I though I've fixed all cases.
Where does this happen?

samsonasik:
I tried this is `CodeIgniter4` repository, and it require it to make it work in php < 8, ref :

https://github.com/codeigniter4/CodeIgniter4/blob/ae833c1b21cae7ec65ba1b96a6588cd7b6c5b504/rector.php#L21-L23

It uses rector-prefixed 0.9.29 https://github.com/codeigniter4/CodeIgniter4/blob/ae833c1b21cae7ec65ba1b96a6588cd7b6c5b504/composer.json#L25 and latest phpstan 0.12.76 https://github.com/codeigniter4/CodeIgniter4/blob/ae833c1b21cae7ec65ba1b96a6588cd7b6c5b504/composer.json#L22

TomasVotruba:
I see. Let's fix it here then, thank you

Issue #5571, opened 2021-02-16
RemoveUnreachableStatementRector removes function content after exit() or return


addshore:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.9.29              |
| Installed as   | composer dependency                          |

RemoveUnreachableStatementRector removes code that is not actually dead when functions are declared below an `exit()`

## Minimal PHP Code Causing Issue

I couldn't reproduce in the tool, maybe because an `exit()` call is needed?
The code I had reproduced this with was https://gist.github.com/addshore/d46bfacbecc932957d332df1438ed3e4
I imagine a minimal reproducion would be

```php
<?php
foo();
exit(0);
function foo(){
    echo "foo";
}
```

This can be seen to not be dead code https://3v4l.org/C8eDS

The set for reproduction would be the dead code set.

```php
    $parameters->set(Option::SETS, [SetList::DEAD_CODE]);
```

## Expected Behaviour

I would not expect the foo method, or methods declared after the `exit()` to be removed as they will be read by PHP already.

TomasVotruba:
Hi,

thanks for reporting.
We'll need a failing demo link: https://getrector.org/demo

addshore:
> Hi,
> 
> thanks for reporting.
> We'll need a failing demo link to verify the needed code for test: https://getrector.org/demo

The tool doesn't appear to allow me to test this code.

![image](https://user-images.githubusercontent.com/3308769/108050303-58cdce80-7041-11eb-8794-38587c9a0d7b.png)

But a link would be https://getrector.org/demo/90c72ecb-b39e-4fb6-a0d1-80d4717c0e11

I can reproduce it locally though and hope the following output is enough.

This includes:
 - `cat` of the test file
 - `cat` of the config
 - output with --dry-run
 - output of running test.php (which correctly echos "foo")
 - output of rector version

```
(rector)$ $ cat test.php
<?php
foo();
exit(0);
function foo()
{
echo "foo";
}

(rector)$ $ cat rector.php
<?php

use Rector\Core\Configuration\Option;
use Rector\Set\ValueObject\SetList;
use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;

return static function (ContainerConfigurator $containerConfigurator): void {
    $parameters = $containerConfigurator->parameters();

    $parameters->set(Option::SETS, [
        SetList::DEAD_CODE,
        ]);

};

[...]

PR #5575, opened 2021-02-16
[DeadCode] Fix RemoveUnreachableStatementRector issue with global function declarations

addshore:
For #5571


TomasVotruba:
Thank you - good job with the fix :clap:

PR #5583, opened 2021-02-16
[DeadCode] Register RemoveEmptyAbstractClassRector to dead-code config set

samsonasik:


samsonasik:

PR #5591, opened 2021-02-17
[RemoveFuncCallArgRector] Use getName instead of isName to fix closure call

ruudk:
https://getrector.org/demo/cd027c68-3b85-4f98-8b9d-5aa3843bb4bd

### DIFF
--- a/packages/node-name-resolver/src/NodeNameResolver.php
+++ b/packages/node-name-resolver/src/NodeNameResolver.php
@@ -97,11 +97,7 @@ final class NodeNameResolver
     public function isName($node, string $name): bool
     {
         if ($node instanceof MethodCall) {
-            $message = sprintf(
-                'Name called on "%s" is not possible. Use $this->getName($node->name) instead',
-                get_class($node)
-            );
-            throw new ShouldNotHappenException($message);
+            return false;
         }
 
         $nodes = is_array($node) ? $node : [$node];
--- a/packages/node-name-resolver/src/NodeNameResolver.php
+++ b/packages/node-name-resolver/src/NodeNameResolver.php
@@ -97,7 +97,11 @@ final class NodeNameResolver
     public function isName($node, string $name): bool
     {
         if ($node instanceof MethodCall) {
-            return false;
+            $message = sprintf(
+                'Name called on "%s" is not possible. Use $this->getName($node->name) instead',
+                get_class($node)
+            );
+            throw new ShouldNotHappenException($message);
         }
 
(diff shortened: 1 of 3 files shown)


TomasVotruba:
CI is failing. 

The change looks identical at first sight. What is incorrect with `isName()`?

ruudk:
See first commit, it fails: https://github.com/rectorphp/rector/runs/1917435332#step:6:103

```
Rector\Core\Exception\ShouldNotHappenException: Name called on "PhpParser\Node\Expr\MethodCall" is not possible. Use $this->getName($node->name) instead
```

So my second commit uses the suggestion, and it works... but that's forbidden by PHPStan rule.

What to do now?

ruudk:
@TomasVotruba Different approach... what do you think?

TomasVotruba:
Same exception should happen in both `getName()`, `isName()`, as both produce same result. 
It often passed as false negative, becuase name of `$this->getName()` was `false`. It should be either on `$this` or `getName()`.

This should be checked on different layer. Maybe in `RemoveFuncCallArgRector`?

ruudk:
Gotcha! Moved it to the rector instead.

packages/node-name-resolver/src/NodeNameResolver.php
```diff
@@ -97,7 +97,11 @@ public function isNames(Node $node, array $names): bool
     public function isName($node, string $name): bool
     {
         if ($node instanceof MethodCall) {
-            return false;
+            $message = sprintf(
+                'Name called on "%s" is not possible. Use $this->getName($node->name) instead',
```
TomasVotruba:
```suggestion
                'Name called on "%s" is not possible. Use $this->isName($node->name) instead',
```

To avoid same PR in the future

rules/removing/src/Rector/FuncCall/RemoveFuncCallArgRector.php
```diff
@@ -61,6 +62,10 @@ public function getNodeTypes(): array
     public function refactor(Node $node): ?Node
     {
         foreach ($this->removedFunctionArguments as $removedFunctionArgument) {
+            if ($node->name instanceof MethodCall) {
+                continue;
+            }
```
TomasVotruba, line 67:
```suggestion
        if ($node->name instanceof Expr) {
            return null;
        }

        foreach ($this->removedFunctionArguments as $removedFunctionArgument) {
```

This avoid dead looping

TomasVotruba:
Thank you :+1:

PR #5594, opened 2021-02-17
Failing tests for override ChangePackageVersionComposerRector

lulco:
Based on discussion in https://github.com/rectorphp/rector/pull/5523

I've created two test cases for override part of configuration of ChangePackageVersionComposerRector:
1) using NETTE_30 as set
2) using NETTE_30 as import

For both I am trying to override nette/application and latte/latte versions.

With 1) it seems SET is not used at all for some reason. But in real application it is used and behave some as 2): "local" configuration is executed first so it is then overriden by imports

### DIFF
--- a/rules/composer/tests/Rector/OverrideChangePackageVersionComposerRector/Fixture/override_change_package.json
+++ b/rules/composer/tests/Rector/OverrideChangePackageVersionComposerRector/Fixture/override_change_package.json
@@ -1,0 +1,17 @@
+{
+    "require": {
+        "nette/application": "^2.4",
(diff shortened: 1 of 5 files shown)


TomasVotruba:
Finally getting into this :)

This will require some standalone work from you. I don't have time to look around this, as I focus on release of prefixed and downgraded version with static reflection and it's taking my full focus.

I recall you're on 2 months off , so we'll keep this closed to avoid staling. Feel free to restore when ready with a solution that meets your needs :) 

Thank you :+1:

PR #5597, opened 2021-02-17
Pass string to FullyQualified instead of ::class

### LINKED ISSUE #5596: InitializeDefaultEntityCollectionRector suggests \RectorPrefix20210215\Doctrine\Common\Collections\ArrayCollection
# Bug Report

| Rector version | 0.9.29              |

It suggests this code:
```
--- Original
+++ New
@@ @@

     public function __construct(string $journeyId, string $ticketProviderId)
     {
+        $this->validation = new \RectorPrefix20210215\Doctrine\Common\Collections\ArrayCollection();

```

I think this is caused by https://github.com/rectorphp/rector/blob/1561e108c075ca7bc0ff7c87b858a9389625cdf8/rules/doctrine-code-quality/src/Rector/Class_/InitializeDefaultEntityCollectionRector.php#L153

Maybe it shouldn't reference the imported `ArrayCollection::class` but use just a FQCN string?

ruudk:
Fixes #5596

### DIFF
--- a/rules/code-quality/src/Rector/ClassMethod/DateTimeToDateTimeInterfaceRector.php
+++ b/rules/code-quality/src/Rector/ClassMethod/DateTimeToDateTimeInterfaceRector.php
@@ -122,7 +122,7 @@ CODE_SAMPLE
 
     private function refactorParamTypeHint(Param $param): void
     {
-        $fullyQualified = new FullyQualified(DateTimeInterface::class);
+        $fullyQualified = new FullyQualified('DateTimeInterface');
         if ($param->type instanceof NullableType) {
             $param->type = new NullableType($fullyQualified);
             return;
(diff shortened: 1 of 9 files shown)


ruudk:
Can we somehow prevent this from happening again? Because nobody thinks about prefixing when contributing new rectors.

TomasVotruba:
Thanks :+1: 

The first choice would be PHPStan rule

ruudk:
Do you know if there is such a rule?

TomasVotruba:
If it would, I'd look for it here: https://github.com/symplify/phpstan-rules/blob/master/docs/rules_overview.md

There is similar for method calls: https://github.com/symplify/phpstan-rules/blob/master/docs/rules_overview.md#requirestringargumentinmethodcallrule

It can be duplicated and instead of `MethodCall` the `New_` node should be used.

PR #5598, opened 2021-02-17
[DeadCode] Add Extra files test for RemoveEmptyAbstractClassRector

samsonasik:
ref https://github.com/rectorphp/rector/pull/5558#discussion_r577073223


samsonasik:
@TomasVotruba I am not sure with the assertion process, it seems whatever file contents added to the extra files, it always correct. Could you verify? Thank you.

rules/autodiscovery/tests/Rector/FileNode/MoveInterfacesToContractNamespaceDirectoryRector/MoveInterfacesToContractNamespaceDirectoryRectorTest.php
```diff
@@ -16,59 +15,30 @@ final class MoveInterfacesToContractNamespaceDirectoryRectorTest extends Abstrac
...
         } else {
             $this->assertFileWasNotChanged($this->originalTempFileInfo);
         }
-
-        $expectedAddedFilesWithContent = [];
-        foreach ($extraFiles as $extraFile) {
-            $expectedAddedFilesWithContent[] = $extraFile->getAddedFileWithContent();
-        }
-
-        $this->assertFilesWereAdded($expectedAddedFilesWithContent);
```
TomasVotruba, line 39: This testing was useless, as it was testing that input fixture file exists.

rules/dead-code/tests/Rector/Class_/RemoveEmptyAbstractClassRector/Fixture/multi_extends.php.inc
```diff
@@ -36,7 +36,7 @@ abstract class AnotherAbstractClassMulti
      }
 }
 
-class MultiExtends extends AnotherAbstractClassMulti
+class MultiExtends extends \Rector\DeadCode\Tests\Rector\Class_\RemoveEmptyAbstractClassRector\Fixture\AnotherAbstractClassMulti
```
TomasVotruba, line 39: By convention any changed class must be in FQN format, so the output is consistent

TomasVotruba:
The original was super confusing, even for me and I wrote it :face_with_head_bandage: I'm sorry about that 

I fixed the naming, so it coresspons with what it should do :)

samsonasik:
Thank you 👍

Issue #5593, opened 2021-02-17
How does Rector interfere/require with my project's libraries and/or PHP version?


mpdude:
👋🏼 This is about my struggles trying to use Rector for the first time. Sorry for the long post 😕.

I would like to use Rector to upgrade – step by step – various things in a project that is currently on PHP 7.2.

Since the latest released version of Rector requires at least PHP 7.3, and since I learned that it is generally useful to install _tools_ not as Composer requirements but somehow as standalone entities (to prevent messing up project and tooling dependency requirements/restrictions, for example), my naïve assumption was that [running Rector in Docker](https://github.com/rectorphp/rector/blob/master/docs/how_to_run_rector_in_docker.md) would be the way to go.

```bash
$ docker run --rm -v $(pwd):/project rector/rector:latest process tests --dry-run

 [ERROR] strlen(): Argument #1 ($str) must be of type string, array given
```

A helpless attempt of mine was to add `-vvv`, hoping to elicit a bit more context, but no... 

I then found #5059, suggesting to try 

`docker run --rm -v $(pwd):/project -w /rector rector/rector:latest process /project/tests --config /project/rector.php --autoload-file=/project/vendor/autoload.php --dry-run`

This, at least, gives a new error message: 

`Fatal error: Array and string offset access syntax with curly braces is no longer supported in /project/vendor/mikey179/vfsstream/src/main/php/org/bovigo/vfs/vfsStream.php on line 309`

So, obviously there must be a problem because some project code (a dependency, actually) uses PHP <8 syntax; and the Docker image has been updated to use PHP 8.0 just a few days ago (#5536).

Which leads me to the first question: 

Why is there a conflict between the PHP version used by the _tool_ itself and the version of code (i. e. input, data?) the tool _works on_? How useful is an upgrade tool if it requires me to be PHP 8-compatible already? 🤔 

But, secondly, the #5059 PR also contains [this comment](https://github.com/rectorphp/rector/pull/5059#issuecomment-753975182), stating that this is way of running the Docker image is not a good/recommended default approach, since it changes the working directory from the project to the Rector installation inside the container. 

It then refers to [another comment](https://github.com/rectorphp/rector/issues/5003#issuecomment-752733167) that deals with how this change of the `cwd` changes autoload priorities to use Rector's dependencies over project dependencies, how this can be used to sidestep version conflicts between the version of Symfony components required by Rector and the project, and that in general, this would be "unexplored area".

So, second question: 

[...]

PR #5599, opened 2021-02-17
Add a Docker build target for a "scoped" variant

mpdude:
As an attempt to improve on #5593, here is an extension of the `Dockerfile` to include a "scoped Rector" build stage. Additionally, the PHP version to build upon can be passed as a build  argument.

```
docker build --target rector -t rector-php8.0 .
docker build --target rector-scoped -t rector-scoped-php8.0 .
docker build --target rector --build-arg PHP_VERSION=7.4 -t rector-php7.4 .
docker build --target rector-scoped --build-arg PHP_VERSION=7.4 -t rector-scoped-php7.4 .
```

A full, maintainable solution would need to integrate with GH Actions and possibly remove redundancy: Right now, there is some code from `./build-rector-scoped.sh` duplicated in the Dockerfile. To me, it seemed to be more natural this way, keeping the build process entirely in Docker and the Dockerfile.

After the Docker image has been built, the "scoped" Rector version can be copied out of it:

```
docker create -ti --name dummy rector-scoped-php8.0 bash ; docker cp dummy:/rector ./rector-scoped ; docker rm -f dummy
```

I'd be glad if this at least could serve as a starting point for a discussion on how to approach the issue of having a "scoped" Docker image.


Dockerfile
```diff
@@ -1,5 +1,14 @@
-FROM php:8.0-cli as rector
-WORKDIR /rector
+ARG PHP_VERSION=8.0
+FROM php:${PHP_VERSION}-cli as base
+
+RUN apt-get update && apt-get install -y \
+    libzip4 \
+    libicu63 \
```
mpdude, line 6: I had the impression we need those as runtime requirements; it that really the case?

mpdude:
@JanMikes if you think that in general this is heading in the right direction, I could have a look at the GHA workflows and see if I get it all integrated.

JanMikes:
Hi @mpdude there is little refactoring planned: creating separate repository for base images, so we save time in CI + make Dockerfiles a little cleaner. 

We had a call with @TomasVotruba and we agreed on following:
- Official docker images will contain scoped (prefixed) version only
- We will build images for multiple php versions - 7.3, 7.4 and 8.0

mpdude:
That's great news!

What exactly do you mean with "separate repository"?

A repo just for the Docker-related things? Can/should I try to help elsewhere?

mpdude:
Oh, and while we’re at it, maybe use `-alpine` base images for much smaller Docker images? 🌱

JanMikes:
Regards alpine - im not sure if anything changes, see this comment https://github.com/rectorphp/rector/pull/3374#issuecomment-653822476

> A repo just for the Docker-related things? Can/should I try to help elsewhere?

Exactly, but not for everything, just for base images, that will contain system dependencies, php extension and so on. Specific builds will remain in this repository, so we will have something like:
```
FROM rector/php:${PHP_VERSION}
```

> Can/should I try to help elsewhere?

Ofc you can, currently im afraid it will get blocked by me for 2 more days before i get to it :(.

mpdude:
@JanMikes ok, let's put Alpine aside for the time being.

Since I wanted to find out for myself anyway how builds could best be cached in GHA, I've opened #5667. This uses the GitHub Container Registry and the `buildx` builder to cache builds, including intermediate layers for multistage builds.

I don't know if this supersedes the idea of having dedicated base images, but we could this discuss over there.

Looking forward to your feedback!

Dockerfile
JanMikes: Honestly, im not sure, they must be there for reason 😄

Dockerfile
```diff
@@ -1,56 +1,90 @@
...
+
+WORKDIR /build
 
 # Install php extensions
 RUN apt-get update && apt-get install -y \
-        git \
-        unzip \
         g++ \
-        libzip-dev \
+        git \
```
JanMikes, line 16: Lets keep it like this, but in future it is always great to keep refactoring for separate MR than adding new feature so it is easier to review.

Dockerfile
```diff
@@ -1,56 +1,90 @@
...
+
+COPY --from=build-scoped /scoped /rector
+RUN chmod +x /rector/bin/rector
+
+RUN mkdir -p /tmp/opcache \
+    && /rector/bin/rector list > /dev/null \
+    && chmod 777 -R /tmp
+
 ## Used for getrector.org/demo
 FROM rector as rector-secured
```
JanMikes, line 89: Secured rector can use scoped version as well.

Dockerfile
```diff
@@ -1,56 +1,90 @@
...
+RUN rm -f "phpstan-for-rector.neon" \
+    && php -d memory_limit=-1 php-scoper.phar add-prefix bin config packages rules src templates vendor composer.json --output-dir /scoped --config scoper.php \
+    && composer dump-autoload --optimize --classmap-authoritative --no-dev --working-dir /scoped
 
-# To warmup opcache a little
-RUN bin/rector list
+COPY ./scoped /scoped/
 
-RUN chmod 777 -R /tmp
+FROM base as runtime
```
JanMikes, line 57: Im not sure why do we need this extra `runtime` target, why it is not directly `FROM base as rector` ?

Dockerfile
```diff
@@ -1,56 +1,90 @@
...
+FROM runtime as rector
+
+COPY --from=build /build /rector
+RUN chmod +x /rector/bin/rector
+
+RUN mkdir -p /tmp/opcache \
+    && /rector/bin/rector list > /dev/null \
+    && chmod 777 -R /tmp
+
+FROM runtime as rector-scoped
```
JanMikes, line 79: We can use scoped only in docker

JanMikes (changes requested):
Please update dockerfile so there is always and only scoped version used

JanMikes:
> To me, it seemed to be more natural this way, keeping the build process entirely in Docker and the Dockerfile.

I like the idea, though i am almost 100 % sure @TomasVotruba will disagree as he runs it locally without docker 😄 so i would reverse it -> in docker run the bash script instead of copy paste could do the job.

TomasVotruba:
@JanMikes If this output is preserved - https://github.com/rectorphp/rector-prefixed - I don't mind :)

mpdude:
#5667 addresses a few of your comments already, so let's wait here until/if that gets merged and proceed with the remaining items then.

As to running the scoping thing inside Docker: I did this way and without the extra shell script because that contains some steps not really necessary in Docker; for example, copying the files around using `rsync` to yet another workdir, scope from there to some other directory, only to copy back once again... And also, we'd have to install rsync in the `build` layer just to copy the files.

But if you'd like to have an integrated approach with the same shell script used stand-alone and inside the Dockerfile, I am sure I can come up with a solution.

Dockerfile
mpdude: Done in #5667

Dockerfile
mpdude: Done in #5667

Dockerfile
mpdude: Done in #5667

mpdude:
I think #5667 solved the original comment.

So – shall we create other/new issues for the questions:
- Can the "scoped" build process remain as-is (without Docker)?
- Is it worthwhile to use the same build script when building the "scoped" variant locally, in GHA and for the Docker images?

JanMikes:
Let's create new issue/pr please, this is getting too messy and i believe this issue was solved via #5667 

> Can the "scoped" build process remain as-is (without Docker)?
> Is it worthwhile to use the same build script when building the "scoped" variant locally, in GHA and for the Docker images?>

From what Tomas said, i believe if the process will be working and will bring some benefits (etc no code duplication, simplicity, performance,...) there is no reason to reject it 😊

PR #5602, opened 2021-02-17
Tweak build-rector-scoped.sh so it "works on my machine"(TM)

mpdude:
* Prevent an  endless loop when `rsync` copies the `rector-nested` and `rector-scoped` directories
* Work around a possible PHP memory limit


TomasVotruba:
Could you add those comment to code, so it's obvious for future generations?

mpdude:
Updated.

TomasVotruba:
Thank you :+1:

PR #5606, opened 2021-02-18
[EarlyReturn] Register PreparedValueToEarlyReturnRector to early-return config set

samsonasik:


samsonasik:
I added handling indirect variable definition before if and skip for var used in if cond.

samsonasik:

TomasVotruba:
Thank you :+1:

PR #5611, opened 2021-02-18
[TypedPropertyFromStrictConstructorRector] Keep var tag when it defines iterable type

ruudk:
# Failing Test for TypedPropertyFromStrictConstructorRector

Based on https://getrector.org/demo/96c31357-6d0e-4f31-ac9a-2df5828efed9

It's not nice that Rector removes the iterable type definition as that will cause PHPStan to complain about it.

Turns out this only happens inside a namespace 🤔 

If the type is there, leave it. Same for array I think.

### DIFF
--- a/rules/dead-doc-block/src/TagRemover/VarTagRemover.php
+++ b/rules/dead-doc-block/src/TagRemover/VarTagRemover.php
@@ -10,6 +10,8 @@ use PhpParser\Node\Stmt\Property;
 use PHPStan\PhpDocParser\Ast\PhpDoc\VarTagValueNode;
 use PHPStan\PhpDocParser\Ast\Type\ArrayTypeNode;
 use PHPStan\PhpDocParser\Ast\Type\GenericTypeNode;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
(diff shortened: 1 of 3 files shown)


ruudk:
I'm going to try to fix this myself :)

ruudk:
Oke, the problem is that the PHPDoc specifies `iterable|ApiInterface[]`.

When looping over those types in: https://github.com/rectorphp/rector/blob/4416807562645f744bba39346639c9d94f8f4b9e/rules/dead-doc-block/src/TagRemover/VarTagRemover.php#L121-L125

it passes `ApiInterface` to https://github.com/symplify/symplify/blob/893fee82ec571ae840f235ebac1ebcbff3e340ba/packages/package-builder/src/Reflection/ClassLikeExistenceChecker.php#L9

But it should take the FQCN instead of only the class name `ApiInterface`.

@samsonasik @TomasVotruba How to do this? I'm pretty sure this is already done in other places.

TomasVotruba:
On a tram... Try static type mapper service

ruudk:
@TomasVotruba 🎉 Works

ruudk:
Ready for review ✅

TomasVotruba:
Looking good :heavy_check_mark: thank you :+1:

Issue #5612, opened 2021-02-18
Incorrect behavior of SetList::TYPE_DECLARATION (AddArrayParamDocTypeRector & AddArrayReturnDocTypeRector)?


addshore:
# Bug Report

| Subject        | Details              |
| :------------- | :--------------------|
| Rector version | last dev-master      |
| Installed as   | composer dependency  |

## Minimal PHP Code Causing Issue

See https://getrector.org/demo/39b2d6ff-84cb-4f25-be48-d075cb5c090e

### Responsible rules

* `AddArrayParamDocTypeRector`
* `AddArrayReturnDocTypeRector`

Apparently....
But Setting these 2 rules doesn't cause the issue https://getrector.org/demo/1be88ca2-bad3-4fd5-a476-53a2a06db50d
Only when using the whole of `SetList::TYPE_DECLARATION`

I believe a combination of the 2 rules ends up causing the issue, or something more confusing?

- `ParamDocType` alone doesn't cause the issue. https://getrector.org/demo/d92b95ed-dba9-4fa3-9e1b-fd81afd7d1f1
- `ReturnDocType` sets a phpdoc of `mixed[][]` for the data provider, but doesnt touch the other method https://getrector.org/demo/fbd49cdc-c2c2-4fe8-ac09-8bbad050cae1
- `ParamDocType applied once this phpdoc is added also does nothing https://getrector.org/demo/edda7854-eb0b-4f6d-8bb1-a6e64cd8d2a0
- Even both rules with the phpdoc already existing doesn't cause the issue https://getrector.org/demo/bae4bc38-c79e-4f22-9aec-ef66571a0974

## Expected Behavior

`array` is not the correct type for the parameter.
This should probably be left alone entirely

TomasVotruba:
Hi,

this is a tricky one. We'll need a failing test case.
It can be added even with multiple rules or sets, like this: https://github.com/rectorphp/rector/commit/11fbc4c1c90d0405bbf8c6004a1c9374b4242b6d#diff-f9b531f4de96ebfd52019da0ee102ba24bff4f8c88a67e57ca08c67433e270fa

addshore:
Will do! :)

PR #5615, opened 2021-02-18
[TypeDeclaration] Fix array type issue in PHPUnitDataProviderParamTypeInferer

addshore:
Fixes https://github.com/rectorphp/rector/issues/5612


addshore:
The test so far checks for consistent behaviour between yield and array examples.

addshore:
Narrowed down to `Rector/FunctionLike/ParamTypeDeclarationRector` I believe

rules/type-declaration/tests/Rector/FunctionLike/ParamTypeDeclarationRector/Fixture/dataprovider_array.php.inc
```diff
@@ -0,0 +1,17 @@
...
+        return [
+            [ 'I am a string' ],
+        ];
+    }
+
+    /**
+     * @dataProvider provideThings
+     */
+    public function testGetFromId( $id ): void {}
+}
```
TomasVotruba, line 17: What should happen here? 

We use `Skip*` naming convention for classes/files that should be skipped.
Or there must be other half test fixture file for of what is expected.

```php
input
-----
expected
```

Same for file bellow

addshore: So, this could either.
- do the same as in the yield test, meaning no type is added to the parameter 
- add a type but the correct type (not array).

Seemingly either would be possible and count as a "fix".

TomasVotruba: What is a correct type to be added? Seem like an array from `provideThings()`?

addshore: Well, provideThings indeed returns an array.
But the first parameter that will be received by methods using the data provider will be a `string`

https://getrector.org/demo/f19f03fb-0731-4f6b-939f-086ebf0757b9 is a simple example of rector noticing this and setting the phpdoc @return for the provider to be correct.
Adding a second parameter there in the dataProvider seems to do something odd though.
https://getrector.org/demo/b0d16cc1-3dc9-4f3b-b3f1-6e7e9884d1a5

I'm going to dig into the code a little and see if I can figure out how this is working currently (and label the tests some more (and maybe write some more))
Any pointers appreciated!

addshore:
After digging through things I figure this is an issue with `PHPUnitDataProviderParamTypeInferer`?

addshore:
Looks like the fix for my test identifies another test that now fails and appears to be testing the wrong thing.
(I'll keep doing)

addshore:
This appears to fix the issue that was identified now.
I noticed that perhaps the situation could be improved even further for data providers that use `yield`, so I'll continue to investigate to see if I can add more typing for that case outside of this PR.

rules/type-declaration/tests/Rector/ClassMethod/AddArrayParamDocTypeRector/Fixture/data_provider.php.inc
```diff
@@ -4,38 +4,18 @@ namespace Rector\TypeDeclaration\Tests\Rector\ClassMethod\AddArrayParamDocTypeRe
...
-use PHPUnit\Framework\TestCase;
-
-class DataProvider extends TestCase
-{
-    /**
-     * @dataProvider provideData()
-     * @param string[] $names
-     */
-    public function testSomething(array $names)
+    public function testSomething( $name1, $name2 )
```
TomasVotruba: This should not be skipped, but fixed to correct types:

```suggestion
    public function testSomething(string $name1, string $name2)
```

addshore: This test only has the ParamDocType rule applied though, so only the phpdocs should be touched if anything as far as I can understand?
And It also looks like thi rule should only apply to arrays?

TomasVotruba: Ah yes, I've missed that. You're right

rules/type-declaration/tests/Rector/FunctionLike/ParamTypeDeclarationRector/Fixture/dataprovider_array.php.inc
```diff
@@ -0,0 +1,49 @@
...
+        return [
+            [ 123, true, 'I am a string', null ],
+            [ 123, true, 'I am a string', 999 ],
+        ];
+    }
+
+    /**
+     * @dataProvider provideThings
+     */
+    public function testGetFromId( int $one, bool $two, string $three, $four ): void {}
```
TomasVotruba, line 48: Very nice work!

Just one detail, looking at the data provider, this should be:

```suggestion
    public function testGetFromId( int $one, bool $two, string $three, ?int $four ): void {}
```

addshore: I agree, but I also think that could perhaps be fixed in a separate PR.
I could look at chaining that onto this one, similarly to https://github.com/rectorphp/rector/pull/5625 :)
But the fix for data providers here I think is already well worth it as it is!

TomasVotruba: > I agree, but I also think that could perhaps be fixed in a separate PR.
> I could look at chaining that onto this one, similarly to #5625 :)

Good idea :+1: I'll merge this PR as it is, thank you for your wor :)

addshore: This will end up bring fixed in https://github.com/rectorphp/rector/pull/5659

PR #5625, opened 2021-02-19
[TypeDeclaration] Add types when dataProviders that yield are used

addshore:
Leaving as a draft as this is based on https://github.com/rectorphp/rector/pull/5615 and that should be merged first.
(Thus many commits don't actually relate to this change...)
Then I'd alter this branch and make it ready for merging


TomasVotruba:
Thank you :clap:

PR #5628, opened 2021-02-19
[TEST] Add failing test for AnnotationToAttributeRector rule on Symfony constratins annotations

aless673:


rules/php80/tests/Rector/Class_/AnnotationToAttributeRector/Fixture/symfony_constraints.php.inc
```diff
@@ -0,0 +1,47 @@
...
+     */
+    private $email;
+
+    /**
+     * @\Symfony\Component\Validator\Constraints\Type("string")
+     * @\Symfony\Component\Validator\Constraints\Length(max="180")
+     * @\Symfony\Component\Validator\Constraints\Email(groups={"registration"})
+     * @\Symfony\Component\Validator\Constraints\NotNull(groups={"registration"})
+     */
+    private $email2;
```
TomasVotruba, line 23: There is too many tested case in single file. Could you separate it per class, so we see which exactly class is covered and which not right from the test case?

PR #5630, opened 2021-02-20
[Coverage] Add --testdox to phpunit command on weekly code coverage generation

samsonasik:
Last week, there is unknown error on phpunit generation:

Screen Shot 2021-02-20 at 21 15 19

I added `--testdox` to know the nearest source of error when coverage regenerated tomorrow if happen again.

### DIFF
--- a/.github/workflows/weekly_code_coverage.yaml
+++ b/.github/workflows/weekly_code_coverage.yaml
@@ -16,10 +16,10 @@ jobs:
                         run: |
                             composer require --dev nimut/phpunit-merger
 
-                            vendor/bin/phpunit --coverage-php build/logs/rules.cov rules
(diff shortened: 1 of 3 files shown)


samsonasik:
To know the detail like this:

Screen Shot 2021-02-20 at 21 16 44

TomasVotruba:
From the hindsights, the coverage is not a metric we use in any way. It's taking too much maintenance for no gain. 

> "Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away."

I already dropped it for Symplify, see https://github.com/symplify/symplify/pull/2935
Maybe we remove it here too.

samsonasik:
Ok, I create PR https://github.com/rectorphp/rector/pull/5631 to remove it.

PR #5631, opened 2021-02-20
 [Coverage] Delete weekly coverage generation 

samsonasik:
ref https://github.com/rectorphp/rector/pull/5630#issuecomment-782718287


TomasVotruba:
Thank you :+1:

PR #5645, opened 2021-02-21
[Php80] Skip ChangeSwitchToMatchRector on missed break

### LINKED ISSUE #5607: Rule ChangeSwitchToMatchRector works wrong
# Bug Report

| Rector version | 0.9.29              |

If in a `switch-case` body missed break statement the rector also converts it to short `match` construnction

```
switch ($suffix) {
    /** @noinspection PhpMissingBreakStatementInspection */
    case 'M':
        $value *= 1024;
    case 'K':
        $value *= 1024;
        break;
}
```
converts to
```
match ($suffix) {
    'M' => $value *= 1024,
    'K' => $value *= 1024,
};
```

## Expected Behaviour

Rector should skip it

https://3v4l.org/S3vFa

samsonasik:
Fixes #5607

### DIFF
--- a/rules/php80/src/NodeAnalyzer/SwitchAnalyzer.php
+++ b/rules/php80/src/NodeAnalyzer/SwitchAnalyzer.php
@@ -13,15 +13,19 @@ final class SwitchAnalyzer
     public function hasEachCaseBreak(Switch_ $switch): bool
     {
         $totalCases = count($switch->cases);
+        if ($totalCases === 1) {
+            return false;
+        }
+
         foreach ($switch->cases as $key => $case) {
+            if ($key === $totalCases - 1) {
+                return true;
+            }
+
             foreach ($case->stmts as $caseStmt) {
                 if ($caseStmt instanceof Break_) {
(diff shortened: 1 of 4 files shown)


samsonasik:

TomasVotruba:
Thank you :clap:

rules/php80/src/NodeAnalyzer/SwitchAnalyzer.php
```diff
@@ -12,17 +12,26 @@ final class SwitchAnalyzer
...
+        foreach ($switch->cases as $key => $case) {
+            if ($key === $totalCases - 1) {
                 return true;
             }
+
+            foreach ($case->stmts as $caseStmt) {
+                if ($caseStmt instanceof Break_) {
+                    continue 2;
+                }
+            }
```
roxblnfk, line 29: I'm not sure the behavior will be appropriate in this rule. However, `continue;` (without arguments) behaves the same as ' break;`

See https://github.com/rectorphp/rector/issues/5646

PR #5581, opened 2021-02-16
[Bin] Using file_exists() for check PHPStan stub ReflectionUnionType and Attribute exists

samsonasik:
@TomasVotruba In https://getrector.org/demo/3ab7d981-2708-4f30-af81-b983a7a75135 , seems got error:

```bash
Fatal error: Warning: require_once(phar://vendor/phpstan/phpstan/phpstan.phar/stubs/runtime/ReflectionUnionType.php): Failed to open stream: phar error: invalid url or non-existent phar "phar://vendor/phpstan/phpstan/phpstan.phar/stubs/runtime/ReflectionUnionType.php" in /rector/bin/rector on line 5

Fatal error: Uncaught Error: Failed opening required 'phar://vendor/phpstan/phpstan/phpstan.phar/stubs/runtime/ReflectionUnionType.php' (include_path='.:/usr/local/lib/php') in /rector/bin/rector:5
Stack trace:
#0 {main}
thrown in /rector/bin/rector on line 5
```

PR #5649, opened 2021-02-21
[Bin] Handle possible PHPStan Stubs runtime locations

samsonasik:
Point possible vendor locations by use the` __DIR__` as start to locate
Like in https://github.com/rectorphp/rector/pull/5581 that may not detected in https://getrector.org/ which uses docker to run

### DIFF
--- a/bin/rector
+++ b/bin/rector
@@ -9,9 +9,9 @@
  */
 (static function () {
     $vendorPaths = [
-        realpath(__DIR__ . '/../vendor'),          // on bin/
-        realpath(__DIR__ . '/../../vendor'),       // on vendor/rector/bin/
-        realpath(__DIR__ . '/../../../../vendor'), // on vendor/rector/rector/bin/
+        realpath(__DIR__ . '/../vendor'),          // bin/rector
+        realpath(__DIR__ . '/../../../../vendor'), // vendor/rector/rector/bin/rector
+        realpath(__DIR__ . '/../../vendor'),       // vendor/bin/rector , it may be a symlink of vendor/rector/rector/bin/rector
     ];
 
     $areStubsFound = static function (string $vendorPath) : bool {
--- a/bin/rector
+++ b/bin/rector
@@ -1,43 +1,4 @@
 #!/usr/bin/env php
 <?php
 
-/**
- * @see https://github.com/phpstan/phpstan/issues/4541#issuecomment-779434916
- *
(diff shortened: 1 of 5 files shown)


samsonasik:

bin/rector
```diff
@@ -1,14 +1,43 @@
...
-    require_once 'phar://vendor/phpstan/phpstan/phpstan.phar/stubs/runtime/ReflectionUnionType.php';
-    require_once 'phar://vendor/phpstan/phpstan/phpstan.phar/stubs/runtime/Attribute.php';
-}
+/**
+ * @see https://github.com/phpstan/phpstan/issues/4541#issuecomment-779434916
+ *
+ * Point possible vendor locations by use the __DIR__ as start to locate
+ * @see https://github.com/rectorphp/rector/pull/5581 that may not detected in https://getrector.org/ which uses docker to run
+ */
+(static function () {
```
TomasVotruba, line 10: This is look like very magically. Class with method would be better.

The `bin/rector` should be clear of any PHP code. It was my mistake to put phar includes here in the first place. Better locatio would be probably the `ContainerFactory`, so any container created is valid without any extra includes.

TomasVotruba:
What issue is this trying to solve?

samsonasik:
When `rector` bin run from non-current project directory, like via docker: 

https://github.com/rectorphp/getrector.org/blob/814a94687e29db7566fef491a120c77744c1ad43/bin/run-demo.sh#L32-L41

Currently, the `getrector.org` is using PHP 8, so everything is fine, however, when it run in PHP < 8, it will got error undefined class `ReflectionUnionType` as not exists like in the following PR https://github.com/rectorphp/rector/pull/5569

The next patch, which uses `file_exists()` : https://github.com/rectorphp/rector/pull/5581 will only patch for run rector in current directory, for non current directory, it will completely skipped, and if it run on PHP < 8, the error will appear again.

TomasVotruba:
I see. Let's add this autoload then.

Could you move it to ContainerFactory as I write above?

samsonasik:
I can't found `ContainerFactory` class, is it the `RectorContainerFactory`, or should we apply to another `StubLoader` class, eg: `Rector\Core\Stubs\PHPStanStubLoader` ?

samsonasik:
I added PHPStanStubLoader and loaded in RectorContainerFactory

samsonasik:

PR #5622, opened 2021-02-19
[Renaming] Skip RenameMethodRector on interface wildcard used

samsonasik:
Closes #5564

PR #5623, opened 2021-02-19
Use require string argument in constructor rule

ruudk:
Waiting on new tagged release for Symplify PHPStan Rules now that https://github.com/symplify/symplify/pull/2968 is merged.

This PR is a reminder for that. 

It should prevent #5596 from happening again.

PR #5652, opened 2021-02-22
Skip RenameMethodRector on interface wildcard used

ruudk:
I realized that I did not provide the test case in #5564 correctly as I was running it in my project.

By changing the existing test it fails again.

Why? Because `OtherWildcardSubscriber` does match the wildcard `*WildcardSubscriber`. That one does not implement an interface.

Because `$this->classManipulator->hasParentMethodOrInterface` just tries to find the first matching class that matches the wildcard, it checks that one always. And since that one does not implement an interface, it will allow refactoring. While it shouldn't.

That means that the implementation in #5622 is not correct (and naive). 

I'm trying to think of a way how to do this.

/cc @samsonasik


ruudk:
I think the best thing to do would be to rollback #5622 as it does not what it should do.

ruudk:
I'll work on a PR to see if I can implement it. /cc @samsonasik

ruudk:
Master is broken... partial fix here > https://github.com/rectorphp/rector/pull/5623

@samsonasik What do you think of this approach?

samsonasik:
I think new fixture is better instead of modifying existing fixture to keep existing behaviour.

ruudk:
Yeah, not sure. As the old one did not solve my use case + I reverted the PR.

ruudk:
Ran this on my project and now it finally works without issues 🎉

ruudk:
@TomasVotruba What do you think?

TomasVotruba:
I'm waiting for CI to be green, so I can review & merge.

Also, could you separate the fixture into new file? Now it's extending an existing one and it's harder to see what is going on.

TomasVotruba:
Surpassed by https://github.com/rectorphp/rector/pull/5693#issuecomment-787034952

PR #5658, opened 2021-02-22
tv better array type

TomasVotruba:


TomasVotruba:
Merged in https://github.com/rectorphp/rector/pull/5626 already

PR #5659, opened 2021-02-22
[TypeDeclaration] Fixes for array dataProviders with multiple types

addshore:


addshore:
Set as a draft until https://github.com/rectorphp/rector/pull/5625 is merged as they are touching the same set of code.

TomasVotruba:
Thank you :clap:

PR #5660, opened 2021-02-23
[EarlyReturn] Skip ReturnAfterToEarlyOnBreakRector on non-linear previous assignment before foreach

samsonasik:
It handle for example, previous foreach has assignment, but inside if, not directly.


samsonasik:

PR #5661, opened 2021-02-23
[EarlyReturn] Add ReturnBinaryOrToEarlyReturnRector

samsonasik:
To Changes Single return of || to early returns


samsonasik:

rules/early-return/src/Rector/Return_/ReturnBinaryOrToEarlyReturnRector.php
```diff
@@ -0,0 +1,145 @@
...
+    public function getRuleDefinition(): RuleDefinition
+    {
+        return new RuleDefinition('Changes Single return of || to early returns', [
+            new CodeSample(
+                <<<'CODE_SAMPLE'
+class SomeClass
+{
+    public function accept($something, $somethingelse)
+    {
+        return $something || $somethingelse;
```
TomasVotruba, line 49: Notes from call - keep simple expressions:

```php
// keep
return 1 || 2 || 5;


if (true && $value) {
}
```

Update method calls and static calls only:

```php
// changed
return $this->process(...) || $this->someMethod(...) || some::statiCall() || true;
```

samsonasik: implemented.

samsonasik:

samsonasik:
rebased.

samsonasik:

TomasVotruba:
Thank you :tada:

Issue #3490, opened 2020-06-08
Taking advantage of static reflection


ondrejmirtes:
Hi, I'm reacting to https://github.com/rectorphp/rector/pull/3488#issuecomment-640290827 which is about [PHPStan 0.12.26](https://phpstan.org/blog/zero-config-analysis-with-static-reflection).

Rector can take advantage of it, you should take these steps:

1) Use ReflectionProvider interface instead of Broker class in places where you're obtaining class/function/constant reflections. They have the same API.
2) I'm not sure how Rector's DI works. In PHPStan the current stable ReflectionProvider looks like this:

```
ChainReflectionProvider
- ClassBlacklistReflectionProvider
  - RuntimeReflectionProvider
- BetterReflectionProvider
```

The class blacklist is there so that some certain classes are skipped for the runtime reflection and go straight to static reflection.

If you enable the experimental flag `featureToggles.disableRuntimeReflectionProvider` (set to true) then the reflection provider instance is just going to be the BetterReflectionProvider.

The most important thing about BetterReflectionProvider is the source locator chain it uses. In PHPStan it's created in this class: https://github.com/phpstan/phpstan-src/blob/master/src/Reflection/BetterReflection/BetterReflectionSourceLocatorFactory.php (you might want to write your own).

It's quite difficult to construct BetterReflectionProvider by hand but fortunately there's a Nette DI-generated factory based on this interface: https://github.com/phpstan/phpstan-src/blob/master/src/Reflection/BetterReflection/BetterReflectionProviderFactory.php

Here are the services definitions: https://github.com/phpstan/phpstan-src/blob/master/conf/config.neon#L1095-L1125

Let me know if you have any questions 👍

TomasVotruba:
Hi Ondra, 

I just found out that internal errors were reported as positive in CI :man_facepalming: 

After fixin that, I've discovered that between PHPStan 0.12.25 and 0.12.26 the reflection broke.
I went through links and configs and I'm trying to fix it for last 2 hours with no success. So I ask you for help.

It breaks here:
https://github.com/rectorphp/rector/blob/5cb1947407b3aca0520561e081ea4b41baab084c/packages/node-type-resolver/src/PHPStan/Scope/PHPStanNodeScopeResolver.php#L162

Even if I make `$nodeCallback` callback empty, it's still fails.

Any ideas?

TomasVotruba:
[...]

PR #5665, opened 2021-02-23
Make use of static reflection from PHPStan

TomasVotruba:
Closes https://github.com/rectorphp/rector/issues/3490

## Todo

- [x] break rename test without autoload
- [x] replace `ClassExistenceStaticHelper::doesClassLikeExists()` with `ReflectionProvider->hasClass()`
- [x] replace `class_exists()` with `ReflectionProvider->hasClass()`
- [x] fix AnnotationReader to work with native reflections

## SourceLocatorFactory

- [x] add custom `phpstan.neon` for Rector internalls - here https://github.com/rectorphp/rector/pull/5665/commits/9071f1b6ee97d017939a5a5d8157bd742fdb4f38#diff-d4ecbf7aff4875121457666bba3371091eb5009fda360c3c51068e135075bf7c
- [x] add custom source locator to custom `phpstan.neon` config - https://github.com/rectorphp/rector/issues/3490#issuecomment-784345308
- [x] add custom source locator to Rector too, for dynamic per TestCase source adding
- [x] pass parsed input sources to `Rector\NodeTypeResolver\Reflection\BetterReflection\SourceLocatorProvider\DynamicSourceLocator`, e.g. `vendor/bin/rector process <src>`

## Tests

- [x] make test pass again
- [x] make test pass without autoload
- [x] replace `is_a()`, `instanceof`, `$objectType->hasMethod()`, `$objectType->hasProperty()` with `ClassReflection` API


packages/better-php-doc-parser/src/AnnotationReader/NodeAnnotationReader.php
```diff
@@ -87,13 +96,14 @@ public function readClassAnnotation(Class_ $class, string $annotationClassName):
...
             return null;
         }
 
         try {
             // covers cases like https://github.com/rectorphp/rector/issues/3046
 
+            // @todo this will require original reflection
             /** @var object[] $propertyAnnotations */
             $propertyAnnotations = $this->reader->getPropertyAnnotations($propertyReflection);
             return $this->matchNextAnnotation($propertyAnnotations, $annotationClassName, $property);
```
TomasVotruba, line 109: @ondrejmirtes This one is tricky :/ we're using Doctrine annotation reader that work with native `PropertyReflection`. I could not find the native reflection in `PhpPropertyReflection`. Any idea how to work around this?

packages/node-type-resolver/src/DependencyInjection/PHPStanServicesFactory.php
```diff
@@ -37,7 +37,21 @@ public function __construct(ParameterProvider $parameterProvider)
...
+            // just for loading Source directory
+            $analysedPathsFromConfig[] = __DIR__ . '/../../../../rules/renaming/tests/Rector/Name/RenameClassRector/Source';
+        }
+
+        $this->container = $containerFactory->create(
+            sys_get_temp_dir(),
+            $existingAdditionalConfigFiles,
+            [],
+            [],
+            $analysedPathsFromConfig
```
TomasVotruba, line 53: @ondrejmirtes So it seems this parameter does the trick. I passed tests fixture directory here and it works :+1: 

Now I have to replaced all `class_exists(...)` with `RelfectionProvider->hasClass(...)` :laughing:

packages/better-php-doc-parser/src/AnnotationReader/NodeAnnotationReader.php
ondrejmirtes: Does this work for you? https://github.com/phpstan/phpstan-src/commit/5fad625b7770b9c5beebb19ccc1a493839308fb4

TomasVotruba: It does! The annotation is resolved correctly :+1: 

(demo)
https://github.com/rectorphp/rector/pull/5665/commits/d896421b74d2b4b67218611c9e79667ba3edd988#diff-2dc622f75265883d5433d11b29facf32d8f3a0bc244b3f4822f6a3881ad2fa7aR212

TomasVotruba: Is there is a similar method for class and method reflection? The reader needs them:

```php
$this->reader->getClassAnnotations()
$this->reader->getMethodAnnotations()
```

packages/better-php-doc-parser/src/AnnotationReader/NodeAnnotationReader.php
```diff
@@ -82,12 +84,13 @@ public function readAnnotation(Node $node, string $annotationClass): ?object
     public function readClassAnnotation(Class_ $class, string $annotationClassName): ?object
     {
         $classReflection = $this->createClassReflectionFromNode($class);
+        $nativeClassReflection = $classReflection->getNativeReflection();
```
TomasVotruba, line 84: @ondrejmirtes Class got native reflection method :+1:

packages/better-php-doc-parser/src/AnnotationReader/NodeAnnotationReader.php
ondrejmirtes: `ClassReflection::getNativeReflection()`. You can actually get native `ReflectionMethod` on that object as well :)

TomasVotruba: Via `getNativeMethod()`. Cool :+1:

packages/better-php-doc-parser/src/AnnotationReader/NodeAnnotationReader.php
```diff
@@ -186,16 +204,45 @@ private function createPropertyReflectionFromPropertyNode(Property $property): ?
...
+            throw new ShouldNotHappenException();
+        }
+
+        $nativeMethodReflection = $builtinMethodReflection->getReflection();
+        if (! $nativeMethodReflection instanceof ReflectionMethod) {
+            throw new ShouldNotHappenException();
+        }
+
+        return $nativeMethodReflection;
+    }
```
TomasVotruba, line 247: @ondrejmirtes I might got this wrong... :laughing:  is there any easier way to get native `ReflectionMethod`?

packages/node-type-resolver/config/phpstan/static-reflection.neon
```diff
@@ -0,0 +1,15 @@
...
+        disableRuntimeReflectionProvider: false
+
+services:
+    - Rector\NodeTypeResolver\Reflection\BetterReflection\RectorBetterReflectionSourceLocatorFactory
+    - Rector\NodeTypeResolver\Reflection\BetterReflection\SourceLocator\IntermediateSourceLocator
+    - Rector\NodeTypeResolver\Reflection\BetterReflection\SourceLocatorProvider\DynamicSourceLocator
+
+    # basically decorates native PHPStan source locator with a dynamic source locator that is also available in Rector DI
+    betterReflectionSourceLocator:
+        class: _HumbugBoxfac515c46e83\Roave\BetterReflection\SourceLocator\Type\SourceLocator
```
TomasVotruba, line 13: @ondrejmirtes This is a pickle. Using your `roave\better-reflection` package version breaks here, as PHPStan requires the prefixed one. But prefix can change on any PHPStan release.

Any ideas how to avoid changing the prefix on each PHPStan release?

ondrejmirtes: @TomasVotruba This should help you, test the latest version please: https://github.com/phpstan/phpstan-src/commit/43ca4270faf7a51616111595563742501382ba8c

packages/better-php-doc-parser/src/AnnotationReader/NodeAnnotationReader.php
ondrejmirtes: @TomasVotruba This is the wrong way. I already told you: `$reflectionClass->getNativeReflection()->getMethod($methodName)` will get you an instanceof `\ReflectionMethod`.

TomasVotruba: I didn't got that :) Thanks

packages/node-type-resolver/config/phpstan/static-reflection.neon
TomasVotruba: Amazing, thanks for that :+1:

TomasVotruba: It seems the autoloading is not able to find `Roave\BetterReflection\*` classes. Do I need to include it in autoload somewhere extra?

See: https://github.com/rectorphp/rector/pull/5665/checks?check_run_id=1969350846

image

Minimal step to reproduce (any command):

```bash
bin/rector show
```

TomasVotruba: I'm using latest 3 hours old version: https://github.com/phpstan/phpstan/commit/17f9278613a868a1ad75faee5f6afa70b7e39ca9

ondrejmirtes: After this it will work for you: https://github.com/phpstan/phpstan/commit/daec3a68dca69b09ab4bb5244345fc3c8f480099

Just FYI I might rename the `Roave` namespace to `OndrejMirtes` or `PHPStan` just to avoid possible conflicts with this. I'll let you know if I do.

ondrejmirtes: Alright, so I did that: https://github.com/phpstan/phpstan-src/commit/06ef13e9c30626f657b2be10030cf0c857b1dc35

Please update your namespaces. I'm much more comfortable with this change than keeping Roave\BetterReflection unprefixed.

TomasVotruba: Thank you :+1: 

I tried to `compsoer update` and run `bin/rector` but error is the same: https://github.com/rectorphp/rector/pull/5665/checks?check_run_id=1975382136

This is the config for PHPStan:
https://github.com/rectorphp/rector/pull/5665/files#diff-208af2a2bbfa3cab247f54b61881c39c9cb3c43b358b35effed69aefc9557458

ondrejmirtes: Oh yes I forgot about one part of the puzzle, will update it shortly 👍

ondrejmirtes: Here you go, it should work now: https://github.com/phpstan/phpstan/commit/f683544424e49055afa19505a0fbecea1921a2cc

TomasVotruba: It works, thanks :+1:

mpdude:
Am I right assuming that this will remove most of the requirements like running Rector on a PHP version compatible with the project, or having to maintain Rector Prefixed?

TomasVotruba:
@mpdude Rector prefixed is due to install and version conflicts of PHP classes with same name, so it must remain. This will improve autoloading of projects PHP without need for autoload being included in Rector.

ondrejmirtes:
FYI you can now require stable PHPStan version 0.12.79: https://github.com/phpstan/phpstan/releases/tag/0.12.79

mpdude:
After reading https://phpstan.org/blog/zero-config-analysis-with-static-reflection, I thought the point of using static reflection was to be able to analyze/parse code without actually loading it, so no conflicting definitions, no language level problems, just treat the code under analysis as data?

ondrejmirtes:
@mpdude It helps partially but one of the points of `rector-prefixed` is for example that you can use Rector on codebase that uses Symfony 4 while Rector uses Symfony 5. Without `rector-prefixed` Composer wouldn't allow you to install it.

mpdude:
But as to the Rector Docker images, we might get away using the non-prefixed version of Rector, and provide just one image (maybe PHP 8) that will be able to parse also code with language features removed in PHP 8...?

TomasVotruba:
@mpdude Not sure if that works for all language changes, e.g. https://wiki.php.net/rfc/deprecate_curly_braces_array_access

mpdude:
Please forgive my tenacity on this topic.

Two tests:

```php
<?php

$foo = 'bar';
print $foo{2};
```

```php
<?php

class CustomerDTO
{
    public function __construct(
        public DateTimeImmutable $birth_date,
    ) {}
}
```

I have been able to process _both_ files successfully ✅ with the latest, PHP-8.0 based Docker image:

`docker run --init -it --rm -v $(pwd):/project rector/rector:latest process .`

I am very sure the first would at least emit a warning while parsing with PHP 7.3 in Rector versions before this change, and probably would fail with a PHP 8.0 based Rector image. And the second example would not parse with Rector on PHP 7.3.

So, the goal of this effort here was to enable Rector to treat all code that is analyzed as _data_, not having to load/parse/possibly execute it, right?

If so, am I right that – given PHPStan supports this – Rector running on PHP 8 can parse code from PHP 7.0 onwards, maybe even PHP 5.x?

If yes, would that mean that at least the Docker version does not need to provide all the different PHP variants? (Docker users don't care about the PHP version, they pull the image to run a working Rector.)

And also, if static reflection does not _load_ code to parse it, but can do that statically... for what reason do we still need a _scoped_ Rector? 

I understand that for a Composer-based install, you'd like to get out of users way by declaring as little dependencies as possible. So, you need to include "everything" in the Rector Composer package, but make it scoped so the autoloader does not trip over it.

But in the Docker image, shouldn't we be able to go without scoping now?

mpdude:
I happen to have an older project with Symfony 2.3 lying around here. I tried to run a Rector image with the changes from this PR, but _without_ scoping. 

It fails with

```
Fatal error: Declaration of Symfony\Component\Console\Style\OutputStyle::write($messages, bool $newline = false, int $type = self::OUTPUT_NORMAL) must be compatible with Symfony\Component\Console\Output\OutputInterface::write($messages, $newline = false, $type = self::OUTPUT_NORMAL) in /rector/vendor/symfony/console/Style/OutputStyle.php on line 52
```

So, proably some mess-up between my project's version of Symfony and the one included in Rector. 

I then modified `vendor/autoload.php` in my project to throw an exception:

```
Fatal error: Uncaught Exception in /project/vendor/autoload.php:3
Stack trace:
#0 /rector/bin/rector.php(115): require_once()
#1 /rector/bin/rector.php(30): AutoloadIncluder->loadIfExistsAndNotLoadedYet('/project/vendor...')
#2 /rector/bin/rector(4): require_once('/rector/bin/rec...')
#3 {main}
  thrown in /project/vendor/autoload.php on line 3
````

So, `bin/rector` includes my project's autoloader? That's because `bin/rector.php` loads `vendor/autoload.php` from the current cwd. At least in the Docker image, that's the project directory, not the `/rector` one.

With this change...

```diff
diff --git a/bin/rector.php b/bin/rector.php
index 7ae31e1762..9d1876fd44 100755
--- a/bin/rector.php
+++ b/bin/rector.php
@@ -27,7 +27,7 @@ $autoloadIncluder = new AutoloadIncluder();
 $autoloadIncluder->includeDependencyOrRepositoryVendorAutoloadIfExists();

 $autoloadIncluder->loadIfExistsAndNotLoadedYet(__DIR__ . '/../vendor/scoper-autoload.php');
-$autoloadIncluder->loadIfExistsAndNotLoadedYet(getcwd() . '/vendor/autoload.php');
+$autoloadIncluder->loadIfExistsAndNotLoadedYet(__DIR__ . '/../vendor/autoload.php');

 $autoloadIncluder->autoloadProjectAutoloaderFile();
 $autoloadIncluder->autoloadFromCommandLine();
```

... things seem to work. 

So: Doesn't `__DIR__` make more sense, at least with the changes from this PR? That would always be the autoloader including Rector, no matter if installed via Composer or in Docker?

TomasVotruba:
Is this solved by your contribution?

mpdude:
@TomasVotruba #5772 seems to solve the issue I had with running a non-scoped Rector (from Docker) on a Symfony 2.3 project (a comment I initially left here but moved to the PR).

But could you review my question above?

If it is right what I am writing, we could update the Docker build process to only create PHP 8.0 images and to omit the scoping step... /cc @JanMikes

TomasVotruba:
> But could you review my question above?

I got lost in the long text. Could you narrow it down what exactly is your issue?

mpdude:
Under which conditions would Rector still use dynamic reflection, so that code would have to be loaded/parsed by the running PHP interpreter and would possibly end up as a declared class/function/... in the current process?

(Maybe also @ondrejmirtes could answer this?)

ondrejmirtes:
It depends on how the ReflectionProvider instance looks like in Rector. In PHPStan by default it's a chain of various ReflectionProvider implementations. In layman's terms, if a class is reachable through a correctly configured Composer autoloader, a runtime reflection will be used. Otherwise static reflection is used.

TomasVotruba:
Rector is using the same ReflectionProvider service, and adding files and directories from `rector.php` and paths arguments from `vendor/bin/rector process src` to SourceLocator.

mpdude:
(How) can I run Rector with a version of PHPStan that is not packed up as a Phar?

ondrejmirtes:
Nope, PHPStan is only distributed as PHAR. What are your needs?

mpdude:
Debug into Rector, see when/how it uses `ReflectionProvider` and subclasses, being able to tinker around with that code to understand if/when/how/why runtime reflection comes into play.

ondrejmirtes:
Xdebug works even with PHARs.

mpdude:
... but changing code (commenting out things etc) does not.

mpdude:
But I'll take this as "there is no easy way to install a non-Phar source version of PHPStan" and deal with it 😉

ondrejmirtes:
You can always clone phpstan-src alongside a project and run plain PHPStan using ../phpstan-src/phpstan analyse ... on a codebase, but not in combination with Rector.

PR #5667, opened 2021-02-24
Build "scoped" Docker images in GitHub Actions

mpdude:
This follows up on #5599, in particular [this comment](https://github.com/rectorphp/rector/pull/5599#issuecomment-784370131).

It will build the "scoped" version of Rector for all pushes to `master` and for all `tags`. Images will be built for PHP 7.3, 7.4 and 8.0, based on the respective official PHP Docker images.

To keep the final images small and clean, this uses a multistage build. To allow for build caching on GitHub Actions and across all stages, the Docker `buildx` builder is used. That allows to push the resulting image and all intermediate layers (cache) separately.

According to [this research](https://dev.to/dtinth/caching-docker-builds-in-github-actions-which-approach-is-the-fastest-a-research-18ei) and the corresponding [POC repo](https://github.com/dtinth/github-actions-docker-layer-caching-poc), keeping build layers directly in the (new) [GitHub Container Registry](http://ghcr.io/) gives the fastest results. However, this happens for the build cache only. The final, "official" images are still pushed to the Docker Hub as previously.

It requires to [enable the GitHub Container Registry ghcr.io](https://docs.github.com/en/packages/guides/enabling-improved-container-support), which is currently in beta, for this repository. Also, a [Personal Access Token has to be created](https://docs.github.com/en/packages/guides/migrating-to-github-container-registry-for-docker-images#authenticating-with-the-container-registry) and setup as a GitHub Action Secret, because the new ghcr.io Registry does not (yet?) integrate with the `GITHUB_TOKEN` provided automatically.

In case you want to see this in action, have a look at the https://github.com/mpdude/rector/tree/scoped-docker-image-test branch, its [action runs](https://github.com/mpdude/rector/actions?query=branch%3Ascoped-docker-image-test) and the [Docker Hub Repo](https://hub.docker.com/r/mpdude/rector/tags?page=1&ordering=last_updated) where the test images have been pushed. From there, you can also pull the resulting images, give them a try or inspect the layers.

TODO:

- [ ] Update documentation on how to choose and run the Docker image

### DIFF
--- a/.github/workflows/publish_docker_images.yaml
+++ b/.github/workflows/publish_docker_images.yaml
@@ -36,6 +36,7 @@ jobs:
 
                   docker buildx create --name builder-php${{ matrix.php-version }} --use
                   docker buildx build \
+                        --progress plain \
                         --cache-from=$GITHUB_REPOSITORY:build-cache-php${{ matrix.php-version }} \
                         --cache-to=type=registry,ref=$GITHUB_REPOSITORY:build-cache-php${{ matrix.php-version }},mode=max,push=true \
                         --target rector \
@@ -44,6 +45,7 @@ jobs:
                         --build-arg PHP_VERSION=${{ matrix.php-version }} .
 
                   docker buildx build \
+                        --progress plain \
                         --cache-from=$GITHUB_REPOSITORY:build-cache-php${{ matrix.php-version }} \
                         --cache-to=type=registry,ref=$GITHUB_REPOSITORY:build-cache-php${{ matrix.php-version }},mode=max,push=true \
                         --target rector-secured \
--- a/.github/workflows/publish_docker_images.yaml
+++ b/.github/workflows/publish_docker_images.yaml
@@ -21,7 +21,6 @@ jobs:
             - name: Log into container registries
               run: |
                   echo "${{ secrets.DOCKER_PASSWORD }}" | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
-                  echo "${{ secrets.GHCR_PAT }}" | docker login ghcr.io -u $GITHUB_ACTOR --password-stdin
 
             - name: Build images
               run: |
@@ -36,8 +35,8 @@ jobs:
 
                   docker buildx create --name builder-php${{ matrix.php-version }} --use
                   docker buildx build \
-                        --cache-from=ghcr.io/$GITHUB_REPOSITORY:build-cache-php${{ matrix.php-version }} \
-                        --cache-to=type=registry,ref=ghcr.io/$GITHUB_REPOSITORY:build-cache-php${{ matrix.php-version }},mode=max,push=true \
+                        --cache-from=$GITHUB_REPOSITORY:build-cache-php${{ matrix.php-version }} \
+                        --cache-to=type=registry,ref=$GITHUB_REPOSITORY:build-cache-php${{ matrix.php-version }},mode=max,push=true \
                         --target rector \
                         --push \
                         --tag $GITHUB_REPOSITORY:$VERSION-php${{ matrix.php-version }} \
--- a/.github/workflows/publish_docker_images.yaml
+++ b/.github/workflows/publish_docker_images.yaml
@@ -4,6 +4,7 @@ on:
         # Publish `master` as Docker `latest` image.
         branches:
             - master
+            - experimental-scoped
 
         # Publish `v1.2.3` tags as releases.
         tags:
--- a/.github/workflows/publish_docker_images.yaml
+++ b/.github/workflows/publish_docker_images.yaml
@@ -44,11 +44,13 @@ jobs:
                         --tag $GITHUB_REPOSITORY:$VERSION-php${{ matrix.php-version }} \
                         --build-arg PHP_VERSION=${{ matrix.php-version }} .
 
+            - name: Build Rector "secured"
+              if: matrix.php-version == '8.0'
+              run: |
                   docker buildx build \
                         --progress plain \
                         --cache-from=$GITHUB_REPOSITORY:build-cache-php${{ matrix.php-version }} \
-                        --cache-to=type=registry,ref=$GITHUB_REPOSITORY:build-cache-php${{ matrix.php-version }},mode=max,push=true \
                         --target rector-secured \
                         --push \
-                        --tag $GITHUB_REPOSITORY:$VERSION-php${{ matrix.php-version }}-secured \
+                        --tag $GITHUB_REPOSITORY:secured \
                         --build-arg PHP_VERSION=${{ matrix.php-version }} .
--- a/.github/workflows/publish_docker_images.yaml
+++ b/.github/workflows/publish_docker_images.yaml
@@ -11,9 +11,17 @@ on:
 
 jobs:
     publish_images:
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-20.04
+        strategy:
+            matrix:
+                php-version: ['7.3', '7.4', '8.0']
         steps:
             - uses: actions/checkout@v2
+
+            - name: Log into container registries
+              run: |
+                  echo "${{ secrets.DOCKER_PASSWORD }}" | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
+                  echo "${{ secrets.GHCR_PAT }}" | docker login ghcr.io -u $GITHUB_ACTOR --password-stdin
 
             - name: Build images
               run: |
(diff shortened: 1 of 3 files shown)


JanMikes:
Hi, this PR is great and i really appreciate it. 

There are many things going on and i just want to read through all materials you provided.

Meanwhile i started working on base image for all rector PHP based images, including multiple php versions: https://github.com/rectorphp/docker/packages

I want to study if it is safe to use PAT in github actions in combination with docker container registry, i am little worried about security :-).

JanMikes:
I am thinking about not using github container registry, since it is in beta and requires PAT.

We might switch to github package registry (as i did in github.com/rectorphp/docker) and switch "back" to ghcr once it is stable and does not require PAT.

 What do you think? 

I have been thinking about cache for docker builds - what about disabling it? Most time we can save using base rector/php image so it will have nearly exactly the same results (i would love to see benchmark with + without cache, both using base php image 😄 ).

mpdude:
Not sure I understand all implications of what I suggested ;-).

I don't really see the benefit of having a dedicated "Rector PHP base image" as a public image. From memory, that would be the official PHP image plus the `intl`, `opcache` and `zip` extensions. You'd have to maintain those (maybe for PHP 7.2...PHP 8.0?), but why – who is going to use them directly?

See, as a Rector user using Docker, I'd just run `docker run ... rector/rector:latest-php7.3` (or an explicit version). I don't really care about the base images that's build on, at least not for building my own stuff. At best, I might want to extend `FROM rector/rector:..`.

So, as long as you don't provide something as "official Rector PHP base Docker images", you're not obliged to maintain any sort of BC or so on those.

As to Docker build caching, we can go entirely without, and also without GitHub's image registry. 

It's just that I understood you ([here](https://github.com/rectorphp/rector/pull/5599#issuecomment-784370131)) would like to have caching to speed up CI builds.

Since we're using a multistage build, there's a difference between the final image (basically, PHP baseimage + extensions + scoped Rector code copied into it) and all the layers at build time (install tools, build extensions, Composer, ...). All those layers are not (hopefully, right?) part of the resulting image and/or should not be part of the "Rector base image".

Those layers are pushed to the GitHub Registry because that's "more local" to GitHub Actions and provides the highest bandwith/fastest response.

What's the issue with the PAT? That's just another secret, like you have the Docker Registry secrets already.

What do you think? How should we proceed?

JanMikes:
> I don't really see the benefit of having a dedicated "Rector PHP base image" as a public image. From memory, that would be the official PHP image plus the intl, opcache and zip extensions. You'd have to maintain those (maybe for PHP 7.2...PHP 8.0?), but why – who is going to use them directly?

Only us - it is very common "cache" pattern that will speed up image building if layers caching is for any reason not used (etc, i would disable it completely 😄 )

> It's just that I understood you (here) would like to have caching to speed up CI builds.

Yeah, exactly that will be achieved using "base" php images with official php + extensions required by rector

> Since we're using a multistage build, there's a difference between the final image (basically, PHP baseimage + extensions + scoped Rector code copied into it) and all the layers at build time (install tools, build extensions, Composer, ...). All those layers are not (hopefully, right?) part of the resulting image and/or should not be part of the "Rector base image".

Correct 👍 


> What's the issue with the PAT? That's just another secret, like you have the Docker Registry secrets already.

I will read more, but i am afraid if my PAT leaks, since it is in open source repository and in github actions, some accident might happen - i am afraid of anyone gets access to all my private images in github container registry. I need to study this more, but afaik in later versions of ghcr it will be possible to auth with github actions token as well, thats why i am proposing waiting for that version.

I think we should not do as many steps at the same time, though i like them, lets move a little slower and more stable 😄 

My proposal to get this merged:
- Remove cache completely (this will resolve my worries about PAT and we will speed up builds using base images in next MR)


Last thing i noticed is, that it seems the `secured` version got completely removed from pushing from github actions

JanMikes:
This is sweet 🍬 

> Regarding secrecy of the PAT, there's nothing different from the Docker Hub Token currently saved in the DOCKER_SECRET Actions secret. But, of course, we can completely go without the GHCR registry for the time being 👍, so I removed that part.

Honestly, i realized that few hours later after posting my last comment 😄 

I will check what you have done here and i guess no later than tomorrow we can get this merged.

As next step, we can introduce the cache back and use ghcr it. In open source it is usually better to make smaller PRs as it will decrease complexity and time needed to review the changes.

mpdude:
Oh, in case you'd like to inspect the images:

For example, the 8.0 variant is only ~12 MB on top of the official PHP 8.0 image people might have pulled anyway:
https://hub.docker.com/layers/mpdude/rector/experimental-scoped-php8.0/images/sha256-dab7456c189d3f51060a262e2792499577a53d4fd4b648955e085097932eebe2?context=repo

Interestingly, the `-build-cache` tags are recognized by the Hub as being some kind of "layer only" (don't know the right terminology for this): https://hub.docker.com/repository/docker/mpdude/rector/tags?page=1&ordering=last_updated&name=build-cache. You cannot inspect them the same way, they also seem to lack the `linux/amd64` arch and also you cannot `docker pull mpdude/rector:build-cache-php8.0` (which is a good thing, IMHO).

JanMikes:
And btw, next next step is multiarch build. I am facing hostile environment everywhere right now after 2 weeks with my new M1 macbook 😄

JanMikes:
Can i ask you for a feedback about images naming?

```
rector/rector-8.0:latest
rector/rector-php8.0:latest
rector/rector-php-8.0:latest
```

mpdude:
Not sure what the established conventions are... 

To me it seems `rector/rector` should be the repo, and the different versions and flavors should be the tags. So, that'd be `rector/rector:some-tag-some-flavor` in general.

Regarding the `some-tag-some-flavor` part, I was trying to follow the way [PHP does it](https://hub.docker.com/_/php).

1. Take the tagged/relesed version and/or branch name (in case you'd want to publish snapshots for some branches?)
2. Separate by `-` from ...
3. the used PHP version. I chose `phpX.Y` since that's also the format used e. g. in [Ondřej Surý's PPA](https://launchpad.net/~ondrej/+archive/ubuntu/php), and since it goes without the `-`
4. ... another `-` to separate from ...
5. sub-variants, like `secured`.

Does that make sense?

mpdude:
Multiarch builds?

Never did this, but seems to be not that complicated: https://jaimyn.com.au/how-to-build-multi-architecture-docker-images-on-an-m1-mac/

JanMikes:
> seems to be not that complicated

Yeah, with buildx, it should be really easy.

> Regarding latest – don't know what that's supposed to be in the Docker world... The lastest push? The latest version? A snapshot of the master branch?

Yeah, in docker world latest is something like master/main branch, if you like to live on the edge and always use latest version, this one is for you 😄 

Docker image tags now follows git tags, so currently latest stable version of docker is 0.9.31, so latest stable image is `rector/rector:0.9.31`

> To me it seems rector/rector should be the repo

👌 agree

Here are my proposals:
```
# Here i am 100% clear
rector/rector:latest     # php 8.0 + master branch
rector/rector:0.9.31    # php 8.0 + 0.9.31 git ref

rector/rector:secured     # we need only one version of secured for online demo

rector/rector:0.9.31-php7.4    # specific git ref + specific php version
rector/rector:php7.4                # (same as latest-php7.4 but without latest keyword)

rector/rector:0.9.31-php8.0   
rector/rector:php8.0                # basically just alias of :latest

rector/rector:dev-<branch-name>    # thinking about dev- prefix, exactly of how composer does it 
rector/rector:dev-<branch-name>-php7.4
```

What do you think?

mpdude:
I like that! 💪🏻

Do you happen to have the logic already somewhere around to come up with all those tags in the workflow?

mpdude:
Although... that gives the possibly unstable/broken `master`vif you just use `rector/rector`, won’t it?

Is that really a good thing support-wise?

mpdude:
I have updated the workflow to just build one `secured` version, based on PHP 8.0.

JanMikes:
> Although... that gives the possibly unstable/broken mastervif you just use rector/rector, won’t it?

You mean with :secured or in general?

It has been like this since now, so we are not making any bc breaks :-) 

for :secured it is important to run on latest version because of bugfixes in master, which are not get tagged yet.

For general usage, user can always use image with stable tag. Overall i believe master is stable enough as there is extensive test suite :-) 

> Do you happen to have the logic already somewhere around to come up with all those tags in the workflow?

https://github.com/rectorphp/rector/blob/master/.github/workflows/publish_docker_images.yaml#L21-L27 this could be good starting point

JanMikes:
FYI im gonna merge at 2 PM CET

mpdude:
So is the way tags are assigned right now working the way you described? You're going to merge anyway?

JanMikes:
Yep, here https://github.com/rectorphp/rector/blob/master/.github/workflows/publish_docker_images.yaml#L21-L27

- master = `:latest` 
- v0.9.0 = ':0.9.0'
- 0.9.0 = ':0.9.0'
- some-branch = does not tag image

Im gonna merge anyways and tweak it so it work as described in this comment https://github.com/rectorphp/rector/pull/5667#issuecomment-788387982

.github/workflows/publish_docker_images.yaml
```diff
@@ -26,14 +34,23 @@ jobs:
...
-
-            - name: Push images
+                  docker buildx create --name builder-php${{ matrix.php-version }} --use
+                  docker buildx build \
+                        --progress plain \
+                        --cache-from=$GITHUB_REPOSITORY:build-cache-php${{ matrix.php-version }} \
+                        --cache-to=type=registry,ref=$GITHUB_REPOSITORY:build-cache-php${{ matrix.php-version }},mode=max,push=true \
+                        --target rector \
+                        --push \
+                        --tag $GITHUB_REPOSITORY:$VERSION-php${{ matrix.php-version }} \
```
JanMikes, line 44: Mismatch github repo = rectorphp/rector, docker hub image = rector/rector 😄

JanMikes: https://github.com/rectorphp/rector/pull/5726

mpdude:
@JanMikes 🎉

Improved tagging is in #5732.

PR #5669, opened 2021-02-24
[Meta] Add Monorepo Builder

samsonasik:


samsonasik:
@TomasVotruba I applied monorepo builder, and tried run locally with `--dry-run`, it got the following output:

Screen Shot 2021-02-24 at 20 52 01

Please verify if that is correct. Thank you.

TomasVotruba:
Thanks! 
The first steps seem not useful here, only in monorepo.
What do you think?

samsonasik:
Yes, it seems, I removed the first step so it got the following output:

Screen Shot 2021-02-24 at 22 29 39

samsonasik:
rebased.

samsonasik:

TomasVotruba:
Looks good, thank you :+1:

PR #5678, opened 2021-02-25
Document current limitations

### LINKED ISSUE #5593: How does Rector interfere/require with my project's libraries and/or PHP version?

mpdude:
There are some current limitations on how Rector can be used that made me struggle as a first-time user (see #5593).

Here is an attempt to write them up in a way that I think would have saved me some problems on my first attempts.

Please review them carefully to make sure the technical aspects are correct – my understanding is still limited.

I'd say this closes #5593.

### DIFF
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -1,0 +1,69 @@
+# Current Limitations and Installation Alternatives
+
+At least until [#3490](https://github.com/rectorphp/rector/issues/3490) is resolved, Rector relies on _runtime PHP reflection_ to analyze your source code. That means the code Rector works on cannot merely be treated as _data_, but has to be loaded (through the autoloader) into memory to be analyzed by the PHP interpreter.
+
+> Read the [PHPStan 0.12.26 release announcement](https://phpstan.org/blog/zero-config-analysis-with-static-reflection) to learn more about runtime and static reflection. PHPStan is the static analysis tool used by Rector.
+
+This brings a few limitations and affects your choices of how you can run Rector.
+
+## PHP Language Level
+
+The PHP process that runs Rector code also needs to load and parse your source code. So, you need to run a version of PHP that is both compatible with your project code as well as Rector.
+
+This means you cannot run Rector on PHP 8 to work on a project that uses language features limited to previous PHP versions.
+
+> One example for this is [array and string offset access using curly braces](https://www.php.net/manual/de/migration74.deprecated.php#migration74.deprecated.core.array-string-access-curly-brace).
+
+But also the other way round, you cannot run Rector on PHP 7 to analyze a project with PHP 8 language features, like [constructor property promotion](https://stitcher.io/blog/constructor-promotion-in-php-8).
+
+Together with Rector's minimum PHP 7.3 version requirement, this might limit your options of how to run Rector.
+
+Note that using an older version of Rector might allow you to run a version of PHP <7.3.
+
+## Code Collisions
+
+Since Rector needs to load the code it works on in addition to the code it needs to run itself, conflicts may occur.
+
+One practical example of this is when you're running Rector on a project that uses an older version of Symfony. Since Rector uses a few Symfony Components itself, you may either run into problems when classes are re-declared, or when Rector happens to load and later (by accident) run parts of Symfony from your project installation. In the later case, the mix-up of different versions may lead to generally unpredictable results.
+
+To remedy this situation, [Rector Prefixed](https://github.com/rectorphp/rector-prefixed) has been created. Basically, it takes Rector and all of its dependencies and moves _all_ that code under a unique namespace prefix.
+
+(Add a note here: Drawbacks? Why isn't Rector-prefixed _the_ recommended approach?)
+
+## Installation Options
+
+With these limitations in mind, here are your installation options.
+
+### 1. Direct Composer install
+
+```bash
+composer require rector/rector --dev
+```
+
+Composer will try to find a version of Rector compatible with your current version of PHP. Also, since it will resolve a _single_ version for every dependency your project and Rector have in common, code collisions will not occur.
+
+The downside is that having Rector installed this way might affect the dependency versions your project will be able to use.
+
+### 2. Use Composer to install Rector Prefixed
+
+```bash
+composer require rector/rector-prefixed --dev
+```
+
+Since Rector Prefixed brings along most of the code it needs in a separate namespace, the impact and resulting limitations for your project's requirements are minimal.
+
+As of writing, the only additional dependency requirement is on `phpstan/phpstan`. But apart from that, Composer should be able to install it with whatever packages your project requires.
+
+If you need Rector `^0.9.0`, however, you need to run at least PHP 7.3.
+
+### 3. Use Docker to run Rector
+
+When you're using a different version of PHP than the desired version of Rector supports, you can run one of the Docker images we provide.
+
+See the [documentation on running in Docker](/docs/how_to_run_rector_in_docker.md) on how to do this.
+
+There is one caveat, though: As of writing, the [official Rector Docker images](https://hub.docker.com/r/rector/rector/tags?page=1&ordering=last_updated) are built with PHP 8.0 and use the non-prefixed version of Rector.
+
+So, you might run into problems with these images when your code uses language features not compatible with PHP 8, or when there is a collision between your project's and Rector's dependencies.
+
+[Work is underway](https://github.com/rectorphp/rector/pull/5667) to improve this situation and to provide Docker images with alternative PHP versions and/or a prefixed version of Rector.
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ It supports all versions of PHP from 5.3 and major open-source projects:
 
 - [Explore 660+ Rector Rules](/docs/rector_rules_overview.md)
 - [How Does Rector Work?](/docs/how_it_works.md)
+- [Current Limitations](/docs/limitations.md)
 - [PHP Parser Nodes](https://github.com/rectorphp/php-parser-nodes-docs/blob/master/docs/nodes_overview.md)
 
 ### Advanced
@@ -81,14 +82,19 @@ It supports all versions of PHP from 5.3 and major open-source projects:
 
 ## Install
 
+The currently recommended way of installing Rector is as a Composer `dev` dependency:
+
 ```bash
 composer require rector/rector --dev
 ```
 
-- Having conflicts during `composer require`? → Use the [Rector Prefixed](https://github.com/rectorphp/rector-prefixed)
-- Using a different PHP version than Rector supports? → Use the [Docker image](/docs/how_to_run_rector_in_docker.md)
+If that does not work for you
+* due to version conflicts,
+* minimum PHP requirements,
+* does not give you a suitably recent version of Rector or
+* you do not want to install tools like Rector as Composer dependencies,
 
-<br>
+please make sure to review the [current limitations and installation alternatives](/docs/limitations.md).
 
 ## Running Rector
 
--- a/docs/how_to_run_rector_in_docker.md
+++ b/docs/how_to_run_rector_in_docker.md
@@ -3,6 +3,8 @@
 You can run Rector on your project using Docker.
 To make sure you are running latest version, use `docker pull rector/rector`.
 
+In case of problems, please make sure to review the [current limitations and installation alternatives](/docs/limitations.md).
+
 *Note that Rector inside Docker expects your application in `/project` directory - it is mounted via volume from the current directory (`$pwd`).*
 
 ```bash


docs/limitations.md
```diff
@@ -0,0 +1,69 @@
...
+
+### 2. Use Composer to install Rector Prefixed
+
+```bash
+composer require rector/rector-prefixed --dev
+```
+
+Since Rector Prefixed brings along most of the code it needs in a separate namespace, the impact and resulting limitations for your project's requirements are minimal.
+
+As of writing, the only additional dependency requirement is on `phpstan/phpstan`. But apart from that, Composer should be able to install it with whatever packages your project requires.
```
samsonasik, line 55: for string operation, it may require nette/utils as well.

mpdude: Shouldn’t that be declared as a dependency then?

https://github.com/rectorphp/rector-prefixed/blob/master/composer.json

samsonasik: probably need to be added to :

https://github.com/rectorphp/rector/blob/a9ca776ddc7da6dd000c7070681f2c6b64fccac2/scoped/composer.json#L10

@TomasVotruba what do you think?

mpdude: https://github.com/rectorphp/rector-prefixed/tree/master/vendor/nette/utils

`nette/utils` is contained in the prefixed vendors, does that answer it?

docs/limitations.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Code Collisions
+
+Since Rector needs to load the code it works on in addition to the code it needs to run itself, conflicts may occur.
+
+One practical example of this is when you're running Rector on a project that uses an older version of Symfony. Since Rector uses a few Symfony Components itself, you may either run into problems when classes are re-declared, or when Rector happens to load and later (by accident) run parts of Symfony from your project installation. In the later case, the mix-up of different versions may lead to generally unpredictable results.
+
+To remedy this situation, [Rector Prefixed](https://github.com/rectorphp/rector-prefixed) has been created. Basically, it takes Rector and all of its dependencies and moves _all_ that code under a unique namespace prefix.
+
+(Add a note here: Drawbacks? Why isn't Rector-prefixed _the_ recommended approach?)
```
mpdude, line 31: Anyone?

docs/limitations.md
samsonasik: oh, probably I was trying `rector-prefixed` that doesn't contain `nette/utils`

docs/limitations.md
```diff
@@ -0,0 +1,69 @@
...
+
+When you're using a different version of PHP than the desired version of Rector supports, you can run one of the Docker images we provide.
+
+See the [documentation on running in Docker](/docs/how_to_run_rector_in_docker.md) on how to do this.
+
+There is one caveat, though: As of writing, the [official Rector Docker images](https://hub.docker.com/r/rector/rector/tags?page=1&ordering=last_updated) are built with PHP 8.0 and use the non-prefixed version of Rector.
+
+So, you might run into problems with these images when your code uses language features not compatible with PHP 8, or when there is a collision between your project's and Rector's dependencies.
+
+[Work is underway](https://github.com/rectorphp/rector/pull/5667) to improve this situation and to provide Docker images with alternative PHP versions and/or a prefixed version of Rector.
```
TomasVotruba, line 69: This should be resolved before adding this docs

docs/limitations.md
```diff
@@ -0,0 +1,69 @@
+# Current Limitations and Installation Alternatives
+
+At least until [#3490](https://github.com/rectorphp/rector/issues/3490) is resolved, Rector relies on _runtime PHP reflection_ to analyze your source code. That means the code Rector works on cannot merely be treated as _data_, but has to be loaded (through the autoloader) into memory to be analyzed by the PHP interpreter.
+
+> Read the [PHPStan 0.12.26 release announcement](https://phpstan.org/blog/zero-config-analysis-with-static-reflection) to learn more about runtime and static reflection. PHPStan is the static analysis tool used by Rector.
+
+This brings a few limitations and affects your choices of how you can run Rector.
```
TomasVotruba, line 7: This should be resolved before adding this docs

docs/limitations.md
```diff
@@ -0,0 +1,69 @@
...
+
+## PHP Language Level
+
+The PHP process that runs Rector code also needs to load and parse your source code. So, you need to run a version of PHP that is both compatible with your project code as well as Rector.
+
+This means you cannot run Rector on PHP 8 to work on a project that uses language features limited to previous PHP versions.
+
+> One example for this is [array and string offset access using curly braces](https://www.php.net/manual/de/migration74.deprecated.php#migration74.deprecated.core.array-string-access-curly-brace).
+
+But also the other way round, you cannot run Rector on PHP 7 to analyze a project with PHP 8 language features, like [constructor property promotion](https://stitcher.io/blog/constructor-promotion-in-php-8).
```
TomasVotruba, line 17: That should be doable, as promoted properties were added to php-parsed for PHP 7 too

docs/limitations.md
```diff
@@ -0,0 +1,69 @@
...
+
+Together with Rector's minimum PHP 7.3 version requirement, this might limit your options of how to run Rector.
+
+Note that using an older version of Rector might allow you to run a version of PHP <7.3.
```
TomasVotruba, line 21: That cannot be recommended, as that is older version with many missing features and invalid config loading. That could lead to random report on already fixed bugs.

Docker with PHP 7.1 downgraded and prefixed Rector should be single solution for most cases.

docs/limitations.md
TomasVotruba: That's meta information users don't need. See PHPStan for example, the release produces such version.
It should be as part of Rector release too. Fully downgraded and prefixed version, e.g. under sub tag, 0.9.50.71

mpdude: I am afraid I cannot make sense of your comment in this context 😞. 

Is this referring to just why Rector-prefixed is not the primary recommended approach? (What are the drawbacks of using it?)

Or are you suggesting to remove the explanation what Rector-prefixed is about at all?

docs/limitations.md
mpdude: Well, you decide... but that's the state of affairs right now, so it might help people to understand the situation and we can remove this any time the corresponding PR is merged.

docs/limitations.md
mpdude: As above.

docs/limitations.md
mpdude: I'll just remove the remark then.

docs/limitations.md
TomasVotruba: > Or are you suggesting to remove the explanation what Rector-prefixed is about at all?

:+1: 

Unless Docker, the install command for Rector should be always the same:

```bash
composer install rector/rector
```

Composer itself should resolve if the version is prefixed (because of conflicts), or PHP 7.1 (because of lower PHP version). We should handle the release and tagging, instead of telling user the information mentioned above.

docs/limitations.md
TomasVotruba: I'm affraid this would create much more ways to install Rector and add more salt to the injure.
Our goal should be the other ways around:

- have 1 way to install with composer
- have 1 way to install with Docker

Anything else is wrong on our side and putting burden on the users.

docs/limitations.md
TomasVotruba: I don't use Docker. Could you verify it in a GitHub Action?

docs/limitations.md
mpdude: `composer require --dev rector/rector` is the recommended, primary approach and the only one written directly in the README. This will – without further notice or warning – install an older version of Rector if that meets your requirements. 

For example, with PHP 7.2, you'll just get `0.8.56`.

So, if you really want to encourage people not to work with older versions (rather give Docker a try), we should state that more clearly.

By the way, what is "PHP 7.1 downgraded"? I just found https://github.com/rectorphp/rector-php71, but as far as I can see, its mentioned nowhere in the README or `docs`.

docs/limitations.md
mpdude: https://github.com/rectorphp/getrector.org/issues/269 contains an example of PHP 8 code that could not be processed with the PHP 7.3 Docker image, and so the Docker image was upgraded to use PHP 8.

TomasVotruba: I see. I forgot to add this must be merged first, as code is now included manually.https://github.com/rectorphp/rector/pull/5665

docs/limitations.md
TomasVotruba: That is current state, yes. But that's wrong. Our goal is that this command:

```php
composer require --dev rector/rector
```

Installs 0.9.x even for PHP 7.1 and also for version conflicts.
The `rector-prefixed` and `rector-php71` won't exist as they do now, but only as subtag of 0.9.x-y

docs/limitations.md
mpdude: Hm, I'll have to let that sink in a bit... 

I've tried such thing (offer several versions with "smart" tags/versions, let Composer find the "best" one) with internal packages in the past, and it turned out not to be possible in all cases, required a lot of effort to get it right and resulted in unintuitive version numbers (not what you'd expect from SemVer). Not to say it's not doable, just ...

But back to the initial question – what's the drawback of using `-prefixed`, at least for the average user? Is it just that it is less tested, does it bring other problems (for users, not developers)?

docs/limitations.md
mpdude: Agree.

mpdude:
We've had some interesting exchanges in the annotations/comments above, thanks for that 🙏🏻.

I guess the question now is:

Do you want to resolve other pending issues first, and then come back here and see how much of the things still apply? I'm fine with putting this PR on the back burner for the time being.

Or would it be fine to finalize this here, and maybe come back later for updates when the issues are resolved? Of course, things that are just plain wrong (technically) should be fixed.

Keep in mind that most of what's documented here is meant as an additional, background information. It's the information I wish I had known when I ran into #5593 as a first-time user.

mpdude:
#5784 shows that the issues described here have been resolved, so I am closing this.

TomasVotruba:
Thank you :+1:

PR #5680, opened 2021-02-25
[Naming] Register RenameForeachValueVariableToMatchExprVariableRector  to naming config set

samsonasik:


packages/better-php-doc-parser/src/AnnotationReader/AnnotationReaderFactory.php
```diff
@@ -52,8 +52,8 @@ public function create(): Reader
         // without this the reader will try to resolve them and fails with an exception
         // don't forget to add it to "stubs/Doctrine/Empty" directory, because the class needs to exists
         // and run "composer dump-autoload", because the directory is loaded by classmap
-        foreach (self::IGNORED_NAMES as $ignoredName) {
-            $annotationReader::addGlobalIgnoredName($ignoredName);
+        foreach (self::IGNORED_NAMES as $self::IGNORED_NAME) {
```
samsonasik, line 55: this need to be skipped.

PR #5686, opened 2021-02-26
[NodeTypeResolver] Handle object type or nullable object type from Parameter as Identifier

### LINKED ISSUE #5673: Process stuck while analysing lambda functions with parameter names present in the same scope
# Bug Report

| Rector version | v0.9.31             |

Process stuck while analysing lambda functions with parameter names present in the same scope.

I'm putting the sample code here, because https://getrector.org/demo don't stop processing it.

### Input Code:
```php
<?php

final class ThingHelper
{
    public static function generateThing(string $name): string
    {
        $thingMaker = fn ($name) => $name . ' thing';

        return $thingMaker($name);
    }
}
```

### Config (rector.php)
```php
<?php

use Rector\Core\Configuration\Option;
use Rector\Set\ValueObject\SetList;

    $parameters->set(Option::SETS, [SetList::CODE_QUALITY]);
};

```

## Expected Behaviour

Process finishing normally, with no code changes.

## Additional info

It works when i change the lambda's parameter name:

```php
$thingMaker = fn ($someName) => $someName . ' thing';
```

samsonasik:
Fixes #5673

### DIFF
--- a/packages/node-type-resolver/src/NodeTypeResolver.php
+++ b/packages/node-type-resolver/src/NodeTypeResolver.php
@@ -11,6 +11,9 @@ use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\New_;
 use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Name;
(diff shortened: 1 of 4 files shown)


samsonasik:

TomasVotruba:
Thank you :+1:

PR #5693, opened 2021-02-27
[DX] Narrow wobly strings to strict ObjectType()

TomasVotruba:
Cherry picks from static reflection PRs: https://github.com/rectorphp/rector/pull/5665


ruudk:
@TomasVotruba I see that this PR removes the support for wildcard renames. Does that mean it's not possible anymore or is there a different way to do this? 

I find the wildcard support super useful as it allows me to rename methods that are part of objects that are in specific namespaces (like `CommandHandler` ) and don't implement an interface (to target them by). 

Does it mean we need to close #5652?

TomasVotruba:
Indeed. The reason is to have more consistent type system. We basically inherit the robust one from PHPStan. 

Now I see it was a mistake to go with wildcards on class type resolution. Your use case should be solved with PHPStorm refactoring or custom Rector rule instead.

This change also prevents prefixed class bugs like these: https://github.com/rectorphp/rector-prefixed/commit/26b740aa3f45558bab16a07448bb888d5f4bf4a2#diff-986b232e60ade0f6607d242741abdfc3d7de1cb51beb1f2078d322783f7cdc56

Yes, the #5652 should be closed.

ruudk:
It could still be possible if there was another thing in place that would search for all classes that match wildcard and then feed them in as strings to the rename method config. Would that make sense?

TomasVotruba:
It would have to be special kind of `ObjectType`, e.g. `WildcardObjectType`, implemented across whole code base.

PR #5700, opened 2021-02-28
[Rector-CI] Update EndBug/add-and-commit to v7.0.0

samsonasik:


samsonasik:
`rector-ci (rules)` seems still hanging even update ` EndBug/add-and-commit` to v7.

PR #5704, opened 2021-02-28
Add failing test fixture for ChangeOrIfReturnToEarlyReturnRector

staabm:
# Failing Test for ChangeOrIfReturnToEarlyReturnRector

Based on https://getrector.org/demo/ce6c9301-ea30-4f20-a367-adb82b5933f7

got similar cases with `bool`  types

grafik

### DIFF
--- a/rules/early-return/tests/Rector/If_/ChangeOrIfReturnToEarlyReturnRector/Fixture/rex_formatter.php.inc
+++ b/rules/early-return/tests/Rector/If_/ChangeOrIfReturnToEarlyReturnRector/Fixture/rex_formatter.php.inc
@@ -1,0 +1,33 @@
+<?php
+namespace Rector\EarlyReturn\Tests\Rector\If_\ChangeOrIfReturnToEarlyReturnRector\Fixture;
+
+abstract class rex_formatter
+{
+    private static function getTimestamp($value)
+    {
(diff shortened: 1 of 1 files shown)


rules/early-return/tests/Rector/If_/ChangeOrIfReturnToEarlyReturnRector/Fixture/rex_formatter.php.inc
```diff
@@ -0,0 +1,33 @@
...
+-----
+<?php
+namespace Rector\EarlyReturn\Tests\Rector\If_\ChangeOrIfReturnToEarlyReturnRector\Fixture;
+
+abstract class rex_formatter
+{
+    private static function getTimestamp($value)
+    {
+        if (is_int($value)) {
+            return $value;
```
staabm, line 25: as is, rector inserts a `(int)` case here, which is redundat, regarding the `if` expression was transformed into `is_int()`

samsonasik: The code example itself already always redundant cast to `(int)`:

```php
        if (is_int($value) || ctype_digit($value)) {
            return (int) $value;
        }
```

the rule is about use the statement in if statement, not changing casting.

PR #5719, opened 2021-03-02
[Changelog] Re-generate Changelog

samsonasik:


samsonasik:
@TomasVotruba the Latest `0.9.*` release header generated seems only `0.9.30` and `0.9.31`, is it ok?

CHANGELOG.md
```diff
@@ -12,6 +11,71 @@ PRs and issues are linked, so you can find more about it. Thanks to [ChangelogLi
...
+- [#5695] Fix url to nodes-overview, Thanks to [@staabm]
+
+### Removed
+
+- [#5709] [DX] Removing more parent abstract circular classes
+- [#5717] [Downgrade] Remove excessive abstract class inheritance, use single service
+- [#5714] Removing parent classes
+- [#5713] remove unmatched errors + require string in object types to avoid prefixing of generics objects
+
 ### Added
```
TomasVotruba, line 79: There is "added" section twice now, it should be just once per release

TomasVotruba:
@samsonasik Are the previous included too?

samsonasik:
The not included is the "header tag", eg: "## [0.9.29] - yyyy-mm-dd", the changelog list seems generated.

CHANGELOG.md
samsonasik: Yes, that's seems the limitation of current changelog merge

TomasVotruba: The current state of merge seems quite problematic. We have to find out what is the last tag, if it is relased or not and then somehow put merge two lists. It leads mostly to bugs and broken output.

Instead the whole changelog could be generated instead.
The only limite we can add is `--since-tag`, so we can create changelog for specific version, e.g. Symplify 8+.

That would save us lot of content juggling :) 

What do you think?

samsonasik: that seems the way to go, is the `since-tag` exists?

TomasVotruba: Not yet, both is just theory

TomasVotruba: Could you handle this in Symplify?

samsonasik: I will try

samsonasik:
closing due no longer used.

PR #5722, opened 2021-03-02
Point to the new parser nodes docs URL

matthiasnoback:
This would make it easier to include the interface in the book, @TomasVotruba Would you agree to do it like this?

### DIFF
--- a/src/Contract/Rector/PhpRectorInterface.php
+++ b/src/Contract/Rector/PhpRectorInterface.php
@@ -11,7 +11,7 @@ interface PhpRectorInterface extends NodeVisitor, RectorInterface
 {
     /**
      * List of nodes this class checks, classes that implements \PhpParser\Node
-     * See beautiful map of all nodes https://github.com/rectorphp/rector/blob/master/docs/nodes_overview.md
+     * See beautiful map of all nodes https://github.com/rectorphp/php-parser-nodes-docs
      *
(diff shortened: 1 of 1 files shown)


TomasVotruba:
The correct place now is actually:
https://github.com/rectorphp/php-parser-nodes-docs

I would not put shortened urls into code, as it remove the trust from the source. I wan't to know what I'm clicking at.

matthiasnoback:
Okay, no problem. The new URL is shorter anyway, so we just might be lucky :)

matthiasnoback:
I've updated this PR to change the URL instead

TomasVotruba:
Pesonally, I would extend the width for code in book. Modify format if needed.
The native code we use on Github and codebases have now around 120 chars.

E.g. the line here before change has only 108 chars:

image


Reading anything
shorted is pretty
annoying and hard
to read :D 

And can drag 5 to
10 pages.

TomasVotruba:
Thanks :+1:

matthiasnoback:
For sure, we didn't talk about it, but for me it's important that the book also works in print format. So we can actually provide print copies ;) But let's discuss this next time we meet!

TomasVotruba:
I see, but print format should not force non-printed code to be unreadable. Maybe other way around... Yea, we should discuss it :)

PR #5725, opened 2021-03-02
Create skip_symfony_route_attribute.php.inc

### LINKED ISSUE #5724: Incorrect behavior of PrivatizeLocalOnlyMethodRector: Symfony Route with PHP 8 Attributes
# Bug Report

See https://getrector.org/demo/cfcba653-8f4b-4fcb-9163-22293a12a43d

```php
<?php
use Symfony\Component\Routing\Annotation\Route;

#[Route(path: '/')]
final class FrontController
{
    #[Route(path: '', name: 'homepage')]
    public function homeAction() : Response
    {
    }
}
```

### Responsible rules

* `PrivatizeLocalOnlyMethodRector`

## Expected Behavior

See #5670 and #5671.

stephanvierkant:
Fixes #5724.

### DIFF
--- a/rules/privatization/tests/Rector/ClassMethod/PrivatizeLocalOnlyMethodRector/Fixture/skip_symfony_route_attribute.php.inc
+++ b/rules/privatization/tests/Rector/ClassMethod/PrivatizeLocalOnlyMethodRector/Fixture/skip_symfony_route_attribute.php.inc
@@ -1,0 +1,13 @@
+<?php
+
+namespace Rector\Privatization\Tests\Rector\ClassMethod\PrivatizeLocalOnlyMethodRector\Fixture;
+
+use Symfony\Component\Routing\Annotation\Route;
+
+final class FrontController
+{
+    #[Route(path: '', name: 'homepage')]
+    public function homeAction() : Response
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Thank you :+1: 

The issue is that docblock does not care about attributes.
The fix would be to place attribute check here: https://github.com/rectorphp/rector/blob/0ffabe49ad0aa1bfb8b925c1e6f0873782f3bf0d/rules/privatization/src/Rector/ClassMethod/PrivatizeLocalOnlyMethodRector.php#L145

`ClassMethod` node has `attrGroups` property, that contains all the attributes. See: https://github.com/rectorphp/php-parser-nodes-docs/#public-properties-53

Something like this would help:

```php
foreach ($classMethod->attrGroups as $attrGroup)
{
    /** @var AttributeGroup $attrGroup */
    if ($attrGroups->attrs as $phpAttribute) {
       // check if $phpAttribute is route, and skip if so
    }
}
```

samsonasik:
@stephanvierkant  I cherry-picked your commit at PR https://github.com/rectorphp/rector/pull/5737

PR #5727, opened 2021-03-02
[StaticReflection] Add OptimizedDirectorySourceLocator to speedup directory analysis

TomasVotruba:
Rector is failing in some weird combination of directories, after #5665 was merged.
I'm trying to figure out what combination on GitHub Actoins is causing this.

The `process` command is being stuck after 2 minutes, where usually is couple of hundred files:

image

With `--debug`, it gets stuck on 1st file:

```bash
➜  rector git:(multi-rector-matrix) bin/rector p rules -n --debug
[parsing] rules/arguments/config/config.php
[parsing] rules/arguments/src/NodeAnalyzer/ArgumentAddingScope.php
```

<br><br>

@ondrejmirtes Hi Ondra, do you have any tips on how to improve performance with static reflection? The parsing of each file is super slow after the #5665 gets merged.

Right now we're adding adding all the files found in processed directory right to SourceLocator: https://github.com/rectorphp/rector/blob/c6bf799da4a1f532f5bdb9f90f9ff236b2ce0756/src/Application/RectorApplication.php#L199 

Maybe it should be just the directory?

### DIFF
--- a/.github/workflows/rector.yaml
+++ b/.github/workflows/rector.yaml
@@ -22,13 +22,11 @@ jobs:
             matrix:
                 directories:
                     #- rules
-                    - rules/arguments rules/autodiscovery rules/cakephp rules/carbon rules/code-quality rules/code-quality-strict rules/coding-style rules/composer rules/dead-code rules/dead-doc-block rules/defluent rules/dependency-injection rules/doctrine rules/doctrine-code-quality rules/doctrine-gedmo-to-knplabs rules/downgrade-php70  rules/downgrade-php71 rules/downgrade-php72 rules/downgrade-php73
+                    - rules/arguments rules/autodiscovery rules/cakephp rules/carbon rules/code-quality rules/code-quality-strict rules/coding-style rules/composer rules/dead-code rules/dead-doc-block rules/defluent rules/dependency-injection rules/doctrine rules/doctrine-code-quality rules/doctrine-gedmo-to-knplabs rules/downgrade-php70  rules/downgrade-php71 rules/downgrade-php72 rules/downgrade-php73 rules/naming rules/nette rules/nette-code-quality rules/nette-kdyby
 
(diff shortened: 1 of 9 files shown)


TomasVotruba:
When Rector runs on whole `rules` directory, it gets stuck. But running only on `rules/arguments`, it works correctly :thinking: 

It looks like all files are pre-parsed in `SourceLocator`

Peek 2021-03-02 14-54

ondrejmirtes:
I'd suspect some of the source locators. Make sure to use [OptimizedDirectorySourceLocator](https://github.com/phpstan/phpstan-src/blob/master/src/Reflection/BetterReflection/SourceLocator/OptimizedDirectorySourceLocator.php) whenever possible - when the user gives you a directory to process.

Use the single file source locator (https://github.com/phpstan/phpstan-src/blob/master/src/Reflection/BetterReflection/SourceLocator/OptimizedSingleFileSourceLocator.php) only when you have a single file to process.

The OptimizedDirectorySourceLocator has a convenient factory with `createByDirectory` and `createByFiles` method: https://github.com/phpstan/phpstan-src/blob/master/src/Reflection/BetterReflection/SourceLocator/OptimizedDirectorySourceLocatorFactory.php

There are slight differences between OptimizedDirectorySourceLocator and OptimizedSingleFileSourceLocator - for example, OptimizedDirectorySourceLocator will not see global constant definitions whereas OptimizedSingleFileSourceLocator will. But that shouldn't be a problem in practice.

TomasVotruba:
@ondrejmirtes It helped to use the `DynamicSourceLocatorDecorator`, thanks!

Peek 2021-03-03 00-34


Could you give a brief review if there is something to improve in current PR?

TomasVotruba:
This is the important file that adds files + directories to SourceLocator: https://github.com/rectorphp/rector/pull/5727/files#diff-58f98550514a1830d7d2cf4604aa53c8edb5431301eda03c7d33920b1a0ff01a

PR #5732, opened 2021-03-02
Improve Docker image tagging

mpdude:
As discussed in https://github.com/rectorphp/rector/pull/5667#issuecomment-788387982.

```
rector/rector:latest     # php 8.0 + master branch
rector/rector:0.9.31    # php 8.0 + 0.9.31 git ref

rector/rector:0.9.31-php7.4    # specific git ref + specific php version
rector/rector:php7.4                # (same as latest-php7.4 but without latest keyword)

rector/rector:0.9.31-php8.0    # same as :0.9.31
rector/rector:php8.0                # same as :latest
```

Not modified/affected:

```
rector/rector:secured     # we need only one version of secured for online demo, can be master + php 8 
```

Not sure how to reasonably test this, though...

### DIFF
--- a/.github/workflows/publish_docker_images.yaml
+++ b/.github/workflows/publish_docker_images.yaml
@@ -24,14 +24,32 @@ jobs:
 
             - name: Build images
               run: |
-                  # Strip git ref prefix from version
-                  VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
+                  DOCKER_TAGS=""
 
-                  # Strip "v" prefix from tag name
-                  [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
+                  # Pull tag and branch from github.ref, which is either refs/heads/... or refs/tags/...
+                  TAG=$(echo "${{ github.ref }}" | sed -e '/refs\/tags\//!d; s,refs/.*/v\?\(.*\),\1,')
+                  BRANCH=$(echo "${{ github.ref }}" | sed -e '/refs\/heads\//!d; s,refs/.*/\(.*\),\1,')
 
-                  # Use Docker `latest` tag convention
-                  [ "$VERSION" == "master" ] && VERSION=latest
+                  if [ "master" == "$BRANCH" ]; then
+                      DOCKER_TAGS="$DOCKER_TAGS php${{ matrix.php-version }}"
+                      if [ "8.0" == "${{ matrix.php-version }}" ]; then
+                          DOCKER_TAGS="$DOCKER_TAGS latest"
+                      fi
+                  fi
+
+                  if [ ! -z "$TAG" ]; then
+                      DOCKER_TAGS="$DOCKER_TAGS $TAG-php${{ matrix.php-version }}"
+                      if [ "8.0" ==  "${{ matrix.php-version }}" ]; then
+                          DOCKER_TAGS="$DOCKER_TAGS $TAG"
+                      fi
+                  fi
+
+                  echo "Image will be tagged with: $DOCKER_TAGS"
+
+                  DOCKER_TAG_ARG=""
+                  for ARG in $DOCKER_TAGS; do
+                      DOCKER_TAG_ARG="$DOCKER_TAG_ARG --tag rector/rector:$ARG"
+                  done
 
                   docker buildx create --name builder-php${{ matrix.php-version }} --use
                   docker buildx build \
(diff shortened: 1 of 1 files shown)


JanMikes:
Hi, is it rebased? I am afraid it will cause conflicts. We can then check if it does exactly the same thing, with just different syntax or not 😄

mpdude:
It’s not rebased, but has no merge conflicts either...?

I think this fixes some of the tags/variants that are not created right now.

mpdude:
The current solution only does `latest-phpX.Y` and `0.9.31-phpX.Y`.

This PR adds `latest`, `phpX.Y` and `0.9.31` (PHP 8.0) tags.

JanMikes:
> The current solution only does latest-phpX.Y and 0.9.31-phpX.Y.
> This PR adds latest, phpX.Y and 0.9.31 (PHP 8.0) tags.

Nono, they are already covered:

https://github.com/rectorphp/rector/blob/master/.github/workflows/publish_docker_images.yaml#L53 this covers `latest` and `0.9.31` 

https://github.com/rectorphp/rector/blob/master/.github/workflows/publish_docker_images.yaml#L50 this covers `phpX.Y` :-)

So it seems it is only different syntax. Though i am not 100% happy with my solution. Because currently buildx pushes only to remote registry and not locally i had to run extra `docker pull rector/rector:<built-tag>`, add tags and again `docker push` though this operation is quite fast, it could be done in a little more efficient way - either make buildx to build image for local machine as well so we do not have to run the `pull` or the way you proposed -> before running `buildx build` have all tags already defined and pass it to the command.

mpdude:
Ouch, I missed the commits where you added that :+1:.

So yes, then this PR does not add missing tags, but maybe you like it anyway because it replaces the tagging commands you commented as workaround.

I try to come up with all the necessary tags by looking at the `BRANCH` and/or `TAG` and then pass all the necessary tag names to `docker buildx build ...` in one step.

mpdude:
And it's rebased now 😉

.github/workflows/publish_docker_images.yaml
```diff
@@ -24,14 +24,32 @@ jobs:
...
             - name: Build images
               run: |
-                  # Strip git ref prefix from version
-                  VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
+                  DOCKER_TAGS=""
 
-                  # Strip "v" prefix from tag name
-                  [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
+                  # Pull tag and branch from github.ref, which is either refs/heads/... or refs/tags/...
+                  TAG=$(echo "${{ github.ref }}" | sed -e '/refs\/tags\//!d; s,refs/.*/v\?\(.*\),\1,')
```
mpdude, line 30: This `sed` expression removes = skips the (only) line if it does not match `refs/tags/`, and otherwise keeps only the part from after `refs/tags/`. It will also remove a leading `v` there.

.github/workflows/publish_docker_images.yaml
```diff
@@ -24,14 +24,32 @@ jobs:
...
+                  BRANCH=$(echo "${{ github.ref }}" | sed -e '/refs\/heads\//!d; s,refs/.*/\(.*\),\1,')
```
mpdude, line 31: As before, but this time just outputs the branch name if the ref starts with `refs/heads/`

JanMikes:
Hi, we should definitely delete the original code which is supposed to be replaced with this PR so we do not tag+push twice:
https://github.com/rectorphp/rector/blob/master/.github/workflows/publish_docker_images.yaml#L46-L55

It seems tome that it is not up to date with current master - is your fork up to date with upstream (rectorphp/rector repository)?

JanMikes (changes requested):
Hi, we should definitely delete the original code which is supposed to be replaced with this PR so we do not tag+push twice:
https://github.com/rectorphp/rector/blob/master/.github/workflows/publish_docker_images.yaml#L46-L55

It seems to me that it is not up to date with current master - is your fork up to date with upstream (rectorphp/rector repository)? 

Or maybe only the original code has to be deleted, then it will be okay :-)

mpdude:
Yeah, my local `master` branch was not up to date when I did the rebase. Should be fine now.

Btw, glad to see the `-secured` variant became obsolete since now static reflection is available 🎉

JanMikes:
> Btw, glad to see the -secured variant became obsolete since now static reflection is available 🎉

Cant express how happy i am myself about that 😄

JanMikes:
So pls delete duplicate tagging now as proposed - just delete these lines: https://github.com/rectorphp/rector/blob/master/.github/workflows/publish_docker_images.yaml#L46-L55

Then i will merge and see what happens :-)

mpdude:
I did, but messed up the rebase commit without noticing. 😖

Rinse & repeat.

JanMikes (approved):

JanMikes:
Previously it generated as well `:latest-php7.4`, im not 100% sure if we want to keep it or not 🤷

mpdude:
> Previously it generated as well `:latest-php7.4`, im not 100% sure if we want to keep it or not 🤷

Rather have a look at https://github.com/rectorphp/rector/pull/5665#issuecomment-790490168 and tell me if we need different PHP version builds at all 😉

JanMikes:
Honestly, i do not know and im unsure 😄 

I had this discussion with @TomasVotruba and last info was that it is always best to run rector in php version as close as possible to your project version. Tomas, does anything change with the better reflection or is it still valid? The discussion now is if it brings any benefits of supporting multiple php versions in docker or not.

TomasVotruba:
I depends how people use PHP-version-based Docker images in practise. I would go with standard approach :+1:

PR #5747, opened 2021-03-03
[Symfony4] Handle Ternary with null if in ConsoleExecuteReturnIntRector

samsonasik:


TomasVotruba:
You were reading my mind :hugs:  nice!

Issue #5739, opened 2021-03-03
Incorrect behavior of UnderscoreToCamelCaseLocalVariableNameRector


sanmai:
# Bug Report

| Subject        | Details              |
| :------------- | :--------------------|
| Rector version | last dev-master      |
| Installed as   | composer dependency  |

## Minimal PHP Code Causing Issue

See https://getrector.org/demo/6744bfdc-bfbc-4f79-9d30-7c48ec15b334

```php
<?php

final class DoNotOverwriteExistingVariables
{
    public function foo()
    {
       $foo_bar[] = 4;
        
       $fooBar = 2;

       $testExample = 5;
       
       $test_example[] = 7;
    }
}
```

### Responsible rules

* `UnderscoreToCamelCaseLocalVariableNameRector`

## Expected Behavior

Fix for this issue previously reported as #5715 didn't fix the problem correctly. Not only should variable be not renamed into something else if it was assigned, but if it was used too. If a name was used before, Rector should not rename things into that name.

Issue #5715, opened 2021-03-01
Incorrect behavior of UnderscoreToCamelCaseLocalVariableNameRector


sanmai:
# Bug Report

| Subject        | Details              |
| :------------- | :--------------------|
| Rector version | last dev-master      |
| Installed as   | composer dependency  |

## Minimal PHP Code Causing Issue

See https://getrector.org/demo/34fcf41f-eef0-4ff3-aad9-ce6bf3abfbb1

```php
<?php

final class DoNotOverwriteExistingVariables
{
    public function foo()
    {
       $foo_bar = 4;
       $fooBar = 2;

       $testExample = 5;
       $test_example = 7;
    }
}
```

Diff:
```diff
     public function foo()
     {
-       $foo_bar = 4;
+       $fooBar = 4;
        $fooBar = 2;
 
        $testExample = 5;
-       $test_example = 7;
+       $testExample = 7;
     }
```

### Responsible rules

* `UnderscoreToCamelCaseLocalVariableNameRector`

## Expected Behavior

Rector should not rename variables if there's another variable with the same name in current scope. Otherwise put, this example should not be changed at all.

TomasVotruba:
Thank you for your report and demo link!

Could you send a failing test case in a pull-request, so we have it covered in Rector?
Here is step by step tutorial how to add it: https://github.com/rectorphp/rector/blob/master/docs/how_to_add_test_for_rector_rule.md

sanmai:
Sure thing #5716

PR #5748, opened 2021-03-03
Add failing test fixture for UnderscoreToCamelCaseLocalVariableNameRector

### LINKED ISSUE #5750: Incorrect behavior of UnderscoreToCamelCaseLocalVariableNameRector
# Bug Report

See https://getrector.org/demo/754d498e-ac9d-4f72-9d8c-d695f340318b

```php
<?php

final class DoNotOverwriteExistingVariables
{
    public function foo()
    {
       foreach ([[4]] as list ($foo_bar)) {}

       foreach ([4] as $foo_bar) {}
        
       yield function () use (&$foo_bar) {
           list ($foo_bar) = [$foo_bar];
           return $foo_bar === 1;
       };

       $fooBar = 2;
       
       if ($foo_bar === 4) {
           $test_example = 1;

           $testExample = 5;

           yield function () use (&$test_example) {
               return $test_example === 1;
           };
       }
        

    }
}
```

### Responsible rules

* `UnderscoreToCamelCaseLocalVariableNameRector`

## Expected Behavior

Another addition to #5739 and #5715. Neither fix got to the root of the issue.

sanmai:
# Failing Test for UnderscoreToCamelCaseLocalVariableNameRector

Based on https://getrector.org/demo/754d498e-ac9d-4f72-9d8c-d695f340318b

Another addition to #5739 and #5715. 

Shall we fix #5750 this once and for all, or shall I invent more cases?

### DIFF
--- a/rules/naming/tests/Rector/Variable/UnderscoreToCamelCaseLocalVariableNameRector/Fixture/do_not_overwrite_existing_variables.php.inc
+++ b/rules/naming/tests/Rector/Variable/UnderscoreToCamelCaseLocalVariableNameRector/Fixture/do_not_overwrite_existing_variables.php.inc
@@ -1,0 +1,34 @@
+<?php
+
+namespace Rector\Naming\Tests\Rector\Variable\UnderscoreToCamelCaseLocalVariableNameRector\Fixture;
+
+final class DoNotOverwriteExistingVariables
+{
+    public function foo()
+    {
+       foreach ([[4]] as list ($foo_bar)) {}
+
+       foreach ([4] as $foo_bar) {}
+        
+       yield function () use (&$foo_bar) {
+           list ($foo_bar) = [$foo_bar];
+           return $foo_bar === 1;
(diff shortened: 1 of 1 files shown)


samsonasik:
Latest getrector.org seems deployed 11 hours ago, so using  getrector.org may not be accurate. Using failing test case like this is the way we can detect the bug.

I think one by one is ok as that will be easier to review. If you can provide a patch too, please do, thank you.

TomasVotruba:
@samsonasik I tried to fix it, but it seems there is more than meets the eye. Could you look into this?
Service `BreakingVariableRenameGuard` might help, but it also report few false positives.

samsonasik:
@TomasVotruba yes, I tried as well, and it is very complex. I will try re-check on `BreakingVariableRenameGuard`

TomasVotruba:
See https://github.com/rectorphp/rector/issues/5750#issuecomment-799813551

PR #5754, opened 2021-03-03
Update README.md

javaDeveloperKid:
I find the current comment confusing.
1. `skip` in comment and `IMPORT...`  in constant name do not fit here together. Btw. `AUTO_IMPORT_NAMES` constant name is not clear to me. What "names"? It doesn't have to be changed but let's improve the comment here.
2. added very verbose and common term i.e. `... use statements` , because `import` word in context of what Rector does here can also be understood as `add type hint for root namespace classes`  
3. I don't know IMPORT_DOC_BLOCKS option yet, but I may require the same reword. I will have a look at this later or you can propose you own based on my 2 rewords.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,10 +166,16 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     // is your PHP version different from the one your refactor to? [default: your PHP version], uses PHP_VERSION_ID format
     $parameters->set(Option::PHP_VERSION_FEATURES, PhpVersion::PHP_72);
 
-    // auto import fully qualified class names? [default: false]
+    // whether Rector can leave FQCN or automatically add them to use statements and remove from code occurences [default: false]
+    // e.g. 
+    // private \My\Namespace\MyService $field;
+    // vs
+    // use My\Namespace\MyService;
+    // ...
+    // private MyService $field;
     $parameters->set(Option::AUTO_IMPORT_NAMES, true);
 
-    // skip root namespace classes, like \DateTime or \Exception [default: true]
+    // whether to add use statements for root namespace classes, like \DateTime or \Exception [default: true]
     $parameters->set(Option::IMPORT_SHORT_CLASSES, false);
 
     // skip classes used in PHP DocBlocks, like in /** @var \Some\Class */ [default: true]


README.md
```diff
@@ -166,10 +166,16 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     // is your PHP version different from the one your refactor to? [default: your PHP version], uses PHP_VERSION_ID format
     $parameters->set(Option::PHP_VERSION_FEATURES, PhpVersion::PHP_72);
 
-    // auto import fully qualified class names? [default: false]
+    // whether Rector can leave FQCN or automatically add them to use statements and remove from code occurences [default: false]
```
TomasVotruba, line 169: A or B option with true or false,  is like "do you have pizza or cake", yes or no.
Better leave one option.

README.md
```diff
@@ -166,10 +166,16 @@ return static function (ContainerConfigurator $containerConfigurator): void {
...
+    // e.g. 
+    // private \My\Namespace\MyService $field;
+    // vs
+    // use My\Namespace\MyService;
+    // ...
+    // private MyService $field;
```
TomasVotruba, line 175: This is confusing. Which is before after? Why partial?
One clear diff example would be better.

README.md
javaDeveloperKid: I'm affraid I don't understand :)

TomasVotruba: There is a binary option with binary answer: `// whether Rector can A or B`

javaDeveloperKid: I think I know what's your point. I used `or` an then `and`. i think a comma would be helpful or as we are developers some parentheses what to read first :)

TomasVotruba:
I think this would deserve own page in the docs, where this would be linked and explained better.
There 3 options get sometimes confused, so it would be nice to have dedicated source that we can link next time.

javaDeveloperKid:
Ok, I will give it a try this week and we will figure out the best description.

TomasVotruba:
Closing to avoid staling, with this output https://github.com/rectorphp/rector/pull/5754#issuecomment-789713612

Feel free to re-open with PR that adds new docs file :+1:

Issue #5472, opened 2021-02-09
[Php7.1 Scoped] Make it prefixed


TomasVotruba:
Run prefixing on PHP 7.1 as well: https://github.com/rectorphp/rector/blob/9f6834c46ea7599446dce808d21a069a4a6a9189/.github/workflows/build_scoped_rector.yaml#L25

- The prefix script: https://github.com/rectorphp/rector/blob/9f6834c46ea7599446dce808d21a069a4a6a9189/build-rector-scoped.sh

- The prefixing line: https://github.com/rectorphp/rector/blob/9f6834c46ea7599446dce808d21a069a4a6a9189/build-rector-scoped.sh#L47

- Running locally - this will output `rector-scoped` directory with prefixed code:

```bash
sh build-rector-scoped.sh
```

TomasVotruba:
See https://github.com/rectorphp/rector/issues/5083#issuecomment-754180262

TomasVotruba:
Resolved in https://github.com/rectorphp/rector/pull/5880

PR #5780, opened 2021-03-05
[CI] Downgrade Rector to PHP 7.1 + prefix it

TomasVotruba:
Closes https://github.com/rectorphp/rector/issues/5472

---

The former bash script is extremely complicated and hard to debug. After static reflection something broke and it's impossible to find out where and what. Each vendor package has it's on matrix of configs, so there is probbaly around 100-200 small little jobs that are being run inside the bash script. 

We need simple approach as Rectror CI has - one run, one dir. That way we can see what file broke on which rule, and ale to fix it.

This is attempt to make it KISS again :slightly_smiling_face: 

## Acceptance Criteria

- [x] single build with scoping + PHP 7.1 downgraded job passing
- [ ] https://github.com/rectorphp/rector-prefixed running on PHP 7.1


leoloso:
Since I built the script, I'll be happy to help you clarify how/why I did anything, if you need to. Just let me know.

Btw, I later on simplified the downgrade script for my own repo (check [here](https://github.com/leoloso/PoP/blob/master/layers/Engine/packages/root/ci/downgrade_code.sh)). So if you need to understand what it does, checking there might be easier.

TomasVotruba:
Thanks! I get the idea, just need 1 config-approach for Rector, so we can maintain it easily.

TomasVotruba:
Closes https://github.com/rectorphp/rector/issues/5472

TomasVotruba:
Looking good :eyes:

TomasVotruba:
Few bugs on PHP 7.1

image

image

TomasVotruba:
There is an issue with covariance, here the `string` must be completed:

image

TomasVotruba:
Surpassed by https://github.com/rectorphp/rector/pull/5880

PR #5782, opened 2021-03-06
[TypeDeclaration] Add AddParamTypeFromCallersRector

TomasVotruba:


staabm:
Wouldnt it make sense to reduce this rector (optional?) to only process private methods?

that way the public api of the classes won‘t get touched and therefore the transformation can be done easily

TomasVotruba:
For private projects it does not make sense.
But for opened source, it would. There is option `PROJECT_TYPE` could active it: https://github.com/symplify/symplify/blob/0f796431bc04adad1db35f0af212da72aa6e5635/rector.php#L67

staabm:
My comment was more regarding private vs public visibility of methods, rather then projects

TomasVotruba:
Understood, I was reacting to that. Public and private methods are the same in scope of private project.  Only current code base is using them, so all passed types are known. 

In matter of open-source project, where input data are missing, it's valid point.

Issue #5003, opened 2020-12-26
Rector in Docker autoload issue


andrey-mostovoy:
For those who faced with error
`[ERROR] Call to undefined method Symfony\Component\DependencyInjection\ContainerBuilder::autowire()`
even doing command like this
`docker run --rm -v $(pwd):/project rector/rector:latest list`

Try to add `--workdir "/rector"`, so the command will be
`docker run --rm -v $(pwd):/project --workdir "/rector" rector/rector:latest list`

And for command `process` you should add `--autoload-file=/project/vendor/autoload.php` for autoloading your project files.

Notice, /project - its a volume inside the docker container, where you mounted your code

TomasVotruba:
Hi,

thanks for the feedback. 
Could you update the README.md so other can use this knowledge?

JanMikes:
Hi. I think that this is a "feature, not a bug" situation 😄 

Can you Share what symfony version does your application use? I am specifically asking if `Symfony\Component\DependencyInjection\ContainerBuilder::autowire()` exists in your codebase. 

It makes me think, that you just faced version mismatch - rector uses newer version of symfony, than your application requires in composer.json - this is usually solved using prefixed rector, unfortunately docker image does not use prefixed, but "classic" rector dependency, just installed globally.

Your workaround works, because if you change workdir to rector, it gets higher priority for autoloading and rector files are autoloaded first and same class but from different directory (your project's vendor) will not get autoloaded anymore, because it is loaded already.

Your solution might work for someone and maybe for someone else not - it is unexplored area 😄

andrey-mostovoy:
Hi. This one "feature" not on symfony application... But we have in project some symfony components....
```
$ composer info | grep depen
php-di/php-di                            6.0.11             The dependency injection container for humans
symfony/dependency-injection             v3.1.10            Symfony DependencyInjection Component
```
this one `symfony/dependency-injection` installed as a dependency of someone other package...

You are right about unexplored area ))

TomasVotruba:
Continue in far developed issue https://github.com/rectorphp/rector/issues/5593

PR #5784, opened 2021-03-06
Add first "system level" (full-scale integration) tests

mpdude:
This adds three "system" level tests for problem areas I am aware of:

* Analyzing a project with PHP 7-specific code with Rector running under PHP 8
* Analyzing PHP 8 code with a Rector running PHP 7
* Autoloading or even accidentally executing project code due to autoloader mess-ups between Rector and the project's codebase.

These tests do a full-scale integration, i. e. they will run `bin/rector` as separate processes for all fixture projects below `tests/system-tests`. 

`tests/system-tests/run-tests.sh` is provided to run all these tests in succession.


mpdude:
@TomasVotruba regarding https://github.com/rectorphp/rector/pull/5775#issuecomment-791909495 – is this heading in the right direction?

mpdude:
I would be grateful for hints how to fix the remaining checks.

tests/system-tests/parse-php8-code/vendor/autoload.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+// autoload.php @generated by Composer
+
+require_once __DIR__ . '/composer/autoload_real.php';
+
+return ComposerAutoloaderInita53ca8a540a790451fc389ced0b5860d::getLoader();
```
TomasVotruba, line 7: Not sure what is the goal here, but vendor should not be part of repository. We won't maintain it

.github/workflows/tests.yaml
```diff
@@ -30,3 +30,5 @@ jobs:
                 if: "matrix.dependencies != 'lowest'"
 
             -   run: vendor/bin/phpunit
+
+            -   run: tests/system-tests/run-tests.sh
```
TomasVotruba, line 34: This is nesting workflow to a workflow. Now GitHub Actions can't be used to and we're left with bare bash.
It should be standalone Workflow file with isolated setup

tests/system-tests/run-tests.sh
```diff
@@ -0,0 +1,22 @@
...
+    $TESTS_DIR/../../bin/rector process --dry-run --clear-cache
+    if [ $? -eq 0 ]; then 
+        echo "-----> Result: OK <-----"
+    else
+        echo "-----> Result: FAILED <-----"
+        HAS_FAILURES=1
+    fi
+done
+
+exit $HAS_FAILURES
```
TomasVotruba, line 22: This file should be a workflow file basically

TomasVotruba:
> @TomasVotruba regarding #5775 (comment) – is this heading in the right direction?

I thought the problem was about Docker and mix of code version vs Docker version. But I don't see it here at all

TomasVotruba:
> This adds three "system" level tests for problem areas I am aware of:

Better start with 1 problem, so it's easier to review and finish. Then we can move to another. Otherwise it can be too complex to review and become stale. Then closed few weeks later.
See: https://matthiasnoback.nl/2021/02/refactoring-prepare-to-stop/

mpdude:
Ok, let's try to sort out things step by step.

In [this comment](https://github.com/rectorphp/rector/pull/5775#issuecomment-791033829), you wrote:

> Scoping is still needed. There are some rare cases, like annotation parsing or native method reflection, that require native class to be loaded.
> But that's only guess... It would be better to have this covered by test case right here, e.g. installing Symfony Console 2.8 on PHP 5.6 with one simple Command implementation and running PHP 8 Docker image on it.

This is what this PR is about: Let's have a minimalistic "project" or codebase that uses Symfony Console 2.8, or that uses PHP 7-specific language syntax, or that uses PHP 8-specific language syntax.

And then let's run pure (unscoped) Rector on that code, using different PHP versions (7.3, 8.0, ?), to find out what happens.

Do you agree so far?

TomasVotruba:
> And then let's run pure (unscoped) Rector on that code, using different PHP versions (7.3, 8.0, ?), to find out what happens.

That would not work, as scoped version is also to avoid composer install conflicts. E.g. this would not work:

```bash
composer require symfony/console:2.8
composer require rector/rector:0.9
```

mpdude:
Seems I need to work on my explanation skills, I cannot get the point across 😔. Yes, there was room for misinterpretation and you found that weak spot...

Let me try again.

Yes, when you install Rector via Composer directly into your project, Composer will resolve package versions that work for both the project and for Rector. Or Composer will abort with an unresolvable conflict. So we cannot install Rector + a conflicting version e. g. of `symfony/console` directly this way through Composer.

But lets look at another usage scenario where Composer resolution does _not_ happen that way, and that's when running the Docker image. In this case, Rector in the Docker image brings along its own set of dependencies (i. e. `vendor`). And the project has a `vendor` folder that was built without taking Rector into consideration. Now when Rector also activates/used the project's autoloader, there may be surprising side effects or crashes, for example #5003.

The problem here is not due to Docker: We have the same situation if you clone the `rectorphp/rector` repo somewhere, run `composer install`, and then you change into your project's directory and run `path/to/rector/bin/rector ...`. It's the same that happens when you run the Docker image, but without Docker :wink:. (By the way, my guess would be `.phar` brings very similar challenges, but let's leave that aside, ok?)

You agree so far?

TomasVotruba:
I'm not sure I see an issue. So this would not happen in Docker, but there is no Docker test in this PR. What is the point of such test that does not test a Docker then?

The failing GitHub Action would tell it more clearly :)

mpdude:
Let's put it this way: This PR is not about an issue, but it adds a few tests that document and guard a very important result you achieved with the static reflection PR (#5665).

To better illustrate this, I have ported the tests from this PR here to the commit immediately before you merged #5665. You can see the result in [this action run](https://github.com/mpdude/rector/actions/runs/631465595) in my fork. They all fail, in part for different reasons.

* On all PHP versions, Rector was not able to clearly separate its own autoloader (and thus, its own code it needs to run) from the project's autoloader. It might happen that it would _run_ code from a directory that should be treated entirely as _data_ (input) for analysis/processing. ([This is the code](https://github.com/mpdude/rector/blob/a92118dd607a1b450007fd84abce81b5aad7d7bc/tests/system-tests/dont-execute-code/src/OutputInterface.php#L7), [here is an example failure](https://github.com/mpdude/rector/runs/2054365505?check_suite_focus=true#step:6:14)). 
* Rector running on PHP 7.3 [failed](https://github.com/mpdude/rector/runs/2054365542?check_suite_focus=true#step:6:37) when analyzing [PHP 8 code](https://github.com/mpdude/rector/blob/a92118dd607a1b450007fd84abce81b5aad7d7bc/tests/system-tests/parse-php8-code/src/Foo.php#L8)
* Rector running on PHP 8 [failed](https://github.com/mpdude/rector/runs/2054365520?check_suite_focus=true#step:6:5) when analyzing [PHP 7 code](https://github.com/mpdude/rector/blob/a92118dd607a1b450007fd84abce81b5aad7d7bc/tests/system-tests/parse-php7-code/src/Foo.php#L10).

But now, with #5665 (and I believe also #5772), this [is no longer the case](https://github.com/rectorphp/rector/actions/runs/628154687). All of those tests pass, [here is the PHP 8 example](https://github.com/rectorphp/rector/runs/2047984180?check_suite_focus=true#step:7:2). 🥳 

So, what can we learn from all that?

1. Running Rector from a source checkout (`git clone ...Rector...`, `composer install`, `cd ~/your-project`, `../path/to/rector/bin/rector ...`) and running from a Rector Docker image will no longer mix up autoloading from Rector's `vendor` with autoloading from the codebase being analyzed. Rector will not load or even execute project code ("data") in its own PHP process.
2. So it was probably a safe decision to drop the `secured` Docker image variant (#5759)
3. Scoping/prefixing is no longer necessary for Rector to put its own dependencies into a safe namespace (so to say) when running from Docker.
4. We do not anymore need the PHP (major) version running Rector to be the same as the PHP (major) version being used in the project.

(Unless, of course, you happen to know edge cases where 1. is _not_ true? 🤞🏻)

If we can agree that these points 1. – 4. are correct, then

* We might have addressed [this remark](https://github.com/rectorphp/rector/pull/5775#issuecomment-791033829)?
* Many (if not all?) caveats I documented in #5678 have been resolved; maybe I can close that PR entirely (I'll see).
* Through #5775, we can
  * simplify the Docker build process a bit, 
  * remove redundancy ([keep this](https://github.com/rectorphp/rector/blob/master/build/build-rector-scoped.sh), [drop that](https://github.com/rectorphp/rector/blob/896ca733dcb82384476ca971f7e0ca317d0c1d33/Dockerfile#L48-L50))
  * make end users' life easier: No need to think about which Docker image variant to use if there is only one (PHP 8.0),
  * move contents of the Docker image closer to a pristine "source" checkout: Maybe that might ease debugging/support or avoid potential problem areas.
  
What do you think?

TomasVotruba:
I'm sorry, I have troubles with long texts with dozens of links to understand. I'm unable to process this.

Could you process the comment I made so we have the failing output in CI?

- https://github.com/rectorphp/rector/pull/5784#discussion_r588943531
- https://github.com/rectorphp/rector/pull/5784#discussion_r588943496
- https://github.com/rectorphp/rector/pull/5784#discussion_r588943414

.github/workflows/tests.yaml
mpdude: Not really nesting a workflow...

See: `tests/system-tests` contains subdirectories with "mini-projects". After the unit tests have passed, `bin/rector` is run on all of these (like fixtures).

`run-tests.sh` is a script to run Rector on all these directories. Like PHPUnit runs on all `*Test` classes, but this script runs the entire Rector application.

If you like, I can make those system tests a dedicated workflow, but it will 99% be like `tests.yaml` (only this shell script instead of `vendor/bin/phpunit`).

tests/system-tests/run-tests.sh
mpdude: Not sure, please see my comment above.

This is a helper script to run Rector on all subdirectories. If any of them has failures, the script's final return will be a failure too.

tests/system-tests/parse-php8-code/vendor/autoload.php
mpdude: I understand it looks confusing. But please have a look here: https://github.com/mpdude/rector/tree/system-level-tests/tests/system-tests/parse-php8-code

This directory is a fixture. It feels like a project/codebase. You can run `/path/to/your/bin/rector process` in it. 

The `vendor` is not for real vendors, but the result of having a `composer.json` file and running `composer dump-autoload`. This way, Rector finds everything it expects from a project.

We will never have to run `composer update` or similar in this directory. It's just a "fake" to run a particular test.

mpdude:
Sorry, how impolite of me 🙈.

I tried to answer your comments as briefly as possible.

There is no failing test here. This PR adds a few ✅ tests for important results of your #5665.

If we take the same tests, but run them just before you merged #5665, look at https://github.com/mpdude/rector/actions/runs/631465595.

Shall we go through step by step why I find the results important?

(I sent you a direct message in the Symfony Devs Slack channel, in case you prefer other communication channels?)

tests/system-tests/parse-php8-code/vendor/autoload.php
TomasVotruba: > The vendor is not for real vendors, but the result of having a composer.json file and running composer dump-autoload. This way, Rector finds everything it expects from a project.

That should be part the build script.

.github/workflows/tests.yaml
TomasVotruba: It is nesting, as it depends on all previous steps and tests.
This must be in standalone workflow. There is no way around it.

TomasVotruba:
Before any further review from my side, the commens above must be resolved. It does not make sense to me to give the same feedback again and drag in circles.

mpdude:
Agree, I will try to work on that. 

Let‘s 🤞🏻I understand what you mean.

mpdude:
@TomasVotruba I think finally understood your remarks above. System tests are now a dedicated workflow, `vendor/` inside the fixture directories is set up during the action run. That makes it much simpler and fewer files in this PR.

Result is here: https://github.com/rectorphp/rector/runs/2059903058?check_suite_focus=true

.github/workflows/php_linter.yaml
```diff
@@ -11,11 +11,11 @@ jobs:
...
...tests/system-tests --exclude packages/rector-generator/templates --exclude rules/psr4/tests/Rector/Namespace_/MultipleClassFileToPsr4ClassesRector/Source --exclude rules/autodiscovery/tests/Rector/FileNode/MoveInterfacesToContractNamespaceDirectoryRector/Expected --exclude packages/node-type-resolver/tests/PerNodeTypeResolver/PropertyFetchTypeResolver/Source/ClassWithNativeProps.php --exclude packages/node-type-resolver/tests/PerNodeTypeResolver/PropertyFetchTypeResolver/Source/ClassWithNativePropsPhp80.php --exclude packages/node-type-resolver/tests/PerNodeTypeResolver/PropertyFetchTypeResolver/Source/ClassWithTypedPropertyTypes.php --exclude rules/nette-kdyby/tests/Rector/MethodCall/ReplaceEventManagerWithEventSubscriberRector/Source/ExpectedSomeClassCopyEvent.php --exclude rules/nette-kdyby/tests/Rector/MethodCall/ReplaceMagicPropertyEventWithEventClassRector/Source/ExpectedFileManagerUploadEvent.php --exclude rules/nette-kdyby/tests/Rector/MethodCall/ReplaceMagicPropertyEventWithEventClassRector/Source/ExpectedDuplicatedEventParamsUploadEvent.php --exclude rules/type-declaration/tests/Rector/ClassMethod/ParamTypeFromStrictTypedPropertyRector/Source/OutOfControlExternalClass.php
```
mpdude, line 14: This is to `--exclude tests/system-tests` from linting. Is there a better way to do it?

mpdude:
Rebased, all checks are now ok.

@TomasVotruba did I resolve your comments appropriately?

TomasVotruba:
Thanks for simplification :+1: 

Now it's much clearer what is the goal of this. It seems like e2e tests - we definitely need that :+1: 

<br>

This repository is split-only after last weekend. The development is moving to `rector-src`, see https://getrector.org/blog/prefixed-rector-by-default for more.

I'll move this PR manually there and test it locally, to save your the booring work. I think you've invested already enough time and energy. Thank you for that :bow:

TomasVotruba:
It's up and running: https://github.com/rectorphp/rector/runs/2555050237

image

:clap:

PR #5785, opened 2021-03-06
Rename phpunit.xml to phpunit.xml.dist

mpdude:
To my knowledge, it is an established practice to name the PHPUnit configuration file `phpunit.xml.dist`, so that a per-developer configuration can override it in `phpunit.xml`.


TomasVotruba:
It's optional naming, I would like to keep it simple instead of disting everything that can.

What do you need override for?

mpdude:
Had to make changes while #5784 was in a preliminary state...

I mean the `.dist` practice is there for a reason, isn't it? At least 90% of the projects I know or work with follow it, so... 🤷🏻

TomasVotruba:
If there is no reason to add, there is no need to add extra complexity.

PR #5795, opened 2021-03-07
[AbstractRector] Point getStaticType() to call NodeTypeResolver->resolve()

samsonasik:
As `NodeTypeResolver-> getStaticType()` is deprecated.


TomasVotruba:
This will require more changes, as they're not 1:1. 
More refactoring is needed on these. The goal is to have just 1 method in the end to work with types

PR #5796, opened 2021-03-08
[TypeDeclarationg] Add Tokens iterator to AddArrayReturnDocTypeRector and AddArrayParamDocTypeRector

TomasVotruba:


samsonasik:
@TomasVotruba it will keep adding `@return` tag:

```php

     * @return array<string, SmartFileInfo[]>&mixed[]|SmartFileInfo[]
     * @return array<string, SmartFileInfo[]>&mixed[]|SmartFileInfo[]
     * @return array<string, SmartFileInfo[]>&mixed[]|SmartFileInfo[]
     * @return array<string, SmartFileInfo[]>&mixed[]|SmartFileInfo[]
```

probably need to be closed first?

TomasVotruba:
This is WIP

samsonasik:
The issue is the rectify seems keep adding: 

```php
@return array<string, SmartFileInfo[]>&mixed[]|SmartFileInfo[]
```

while it already added, it probably will be infinite. 

I think we can close first, if any change, re-open with force push to the PR.

TomasVotruba:
This is WIP. Wait for CI to be green before any review. Thanks

PR #5798, opened 2021-03-08
[Phpunit] Prevent MethodCall instances of being processed

### LINKED ISSUE #5797: [Phpunit] Some rectors throw exception when encountering method calls
# Bug Report

| Rector version | Rector 0.9.29             |

A few phpunit rectors do not seem to be able to handle method calls and instead throw an `\Rector\Core\Exception\ShouldNotHappenException` with the message `Name called on "PhpParser\Node\Expr\MethodCall" is not possible. Use $this->getName($node->name) instead`
These include but may not be limited to:
- AssertTrueFalseToSpecificMethodRector
- AssertFalseStrposToContainsRector
- AssertPropertyExistsRector

https://getrector.org/demo/e3abd896-bd98-4f5a-ad71-dc097a1cbdd3

## Expected Behaviour
The rector does not throw an exception and also does not change the code (similar to how the named rectors already skip static methods)

Bl00D4NGEL:
Fix for https://github.com/rectorphp/rector/issues/5797

---

Edit @TomasVotruba  Fixes #5797

### DIFF
--- a/rules/phpunit/src/Rector/MethodCall/AssertFalseStrposToContainsRector.php
+++ b/rules/phpunit/src/Rector/MethodCall/AssertFalseStrposToContainsRector.php
@@ -79,7 +79,7 @@ final class AssertFalseStrposToContainsRector extends AbstractRector
         }
 
         $firstArgumentValue = $node->args[0]->value;
-        if ($firstArgumentValue instanceof StaticCall) {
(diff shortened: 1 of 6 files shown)


TomasVotruba:
Fixes #5797

TomasVotruba:
Awesome, thank you!

Issue #5646, opened 2021-02-21
Replace `continue;` by `break;` in a switch body


roxblnfk:
# Feature Request

 
## Diff 

```diff
$value = 1;
$suffix = 'M';
switch ($suffix) {
    case 'M':
        $value *= 1024 * 1024;
-       continue;
+       break;
    case 'K':
        $value *= 1024;
}
```

`continue` without argument also throws warning:
 `Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"?`

samsonasik:
it can be for php 7.3 set, ref https://3v4l.org/1Nnti

samsonasik:
already exists, ref https://github.com/rectorphp/rector/pull/5816#issuecomment-795387828

roxblnfk:
> already exists, ref [#5816 (comment)](https://github.com/rectorphp/rector/pull/5816#issuecomment-795387828)

Why is this rule not active in the PHP_70 set?
https://getrector.org/demo/2768f207-688b-4fa6-9641-0f6ea07714e2

samsonasik:
You can register `Rector\Php52\Rector\Switch_\ContinueToBreakInSwitchRector::class` into services.

see https://getrector.org/demo/716f6a89-f2f9-4fc9-8b12-5dc7792b668c

See PR that register set/move to to php73 set:
 
https://github.com/rectorphp/rector/pull/5820 

and 

https://github.com/rectorphp/rector/pull/5821

PR #5816, opened 2021-03-09
added ConvertContinueToBreakRector

Bl00D4NGEL:
refs https://github.com/rectorphp/rector/issues/5646

### DIFF
--- a/rules/php73/src/Rector/Case_/ConvertContinueToBreakRector.php
+++ b/rules/php73/src/Rector/Case_/ConvertContinueToBreakRector.php
@@ -5,8 +5,8 @@ declare(strict_types=1);
 namespace Rector\Php73\Rector\Case_;
 
 use PhpParser\Node;
-use PhpParser\Node\Expr;
 use PhpParser\Node\Scalar\LNumber;
+use PhpParser\Node\Stmt;
 use PhpParser\Node\Stmt\Break_;
 use PhpParser\Node\Stmt\Case_;
 use PhpParser\Node\Stmt\Continue_;
@@ -14,6 +14,9 @@ use Rector\Core\Rector\AbstractRector;
 use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
 use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
 
+/**
+ * @see \Rector\Php73\Tests\Rector\Case_\ConvertContinueToBreakRector\ConvertContinueToBreakRectorTest
+ */
 final class ConvertContinueToBreakRector extends AbstractRector
 {
     public function getRuleDefinition(): RuleDefinition
@@ -64,7 +67,7 @@ BAD_SWITCH,
         return $node;
     }
 
-    private function shouldReplaceWithBreak(Expr $node): bool
+    private function shouldReplaceWithBreak(Stmt $node): bool
     {
         if (!$node instanceof Continue_) {
             return false;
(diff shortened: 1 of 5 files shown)


rules/php73/src/Rector/Case_/ConvertContinueToBreakRector.php
```diff
@@ -0,0 +1,86 @@
...
+    public function getNodeTypes(): array
+    {
+        return [
+            Case_::class
+        ];
+    }
+
+    public function refactor(Node $node): ?Node
+    {
+        if (!$node instanceof Case_) {
```
samsonasik, line 54: not needed, it is always `Case_` as defined in `getNodeTypes()`

rules/php73/src/Rector/Case_/ConvertContinueToBreakRector.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * @param Case_ $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        $statements = $node->stmts;
+
+        foreach ($statements as $i => $statement) {
```
samsonasik, line 59: I've not tried, but it should can be simplied with:

**Option 1: Reference value var**

```php
foreach ($node->stmts as &$statement) {
    if ($this->shouldReplaceWithBreak($statement)) {
         $statement = new Break_();
    }
}

return $node;
```

**Option 2: Using `traverseNodesWithCallable()`**

```php
        $this->traverseNodesWithCallable($node->stmts, function (Node $node) : void {
            if (! $this->shouldReplaceWithBreak($node)) {
                return;
            }

            $node = new Break_();
        });
        
        return $node;
```

TomasVotruba: Better the 2nd option, references can go wrong very quickly

Bl00D4NGEL: Option 1 works but I personally prefer to avoid references at all cost to prevent (unwanted) side effects. Not sure what hte "standard" is for rector but if generally speaking everybody is fine/wants to use references I'm fine with changing it. (Is there maybe a benefit of reference over value that I'm not aware of?)
Option 2 didn't work for me, tried tinkering a bit with it but couldn't get it to work.

samsonasik: I just tried with `traverseNodesWithCallable()` and it seems not working for re-apply node with new node. I think it is ok as is 👍

rules/php73/src/Rector/Case_/ConvertContinueToBreakRector.php
```diff
@@ -0,0 +1,85 @@
...
+ */
+final class ConvertContinueToBreakRector extends AbstractRector
+{
+    public function getRuleDefinition(): RuleDefinition
+    {
+        return new RuleDefinition(
+            'Replaces continue statement with break',
+            [
+                new CodeSample(
+                    <<<'GOOD_SWITCH'
```
samsonasik, line 28: one more thing, it should be use `CODE_SAMPLE` for consistency

TomasVotruba:
I'm curious, how does it differ from this rule?
https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#continuetobreakinswitchrector

samsonasik:
oh, that seems same

Bl00D4NGEL:
The rector Tomas linked is even better than mine (given it also considers variables) so I guess this already existed in the first place. Guess we can close the PR then? And also close https://github.com/rectorphp/rector/issues/5646?

samsonasik:
Yes, I think it can be closed

PR #5820, opened 2021-03-10
[PHP 7.3] Register existing ContinueToBreakInSwitchRector in PHP 5.2 set to PHP 7.3 config set

samsonasik:
As it actually still working in php 5.2, while in php 7.3, it already error https://3v4l.org/1Nnti. Another solution is move the `ContinueToBreakInSwitchRector` to `Php73` set.

PR #5821, opened 2021-03-10
[PHP 7.3] Move ContinueToBreakInSwitchRector from PHP 5.2 to PHP 7.3

samsonasik:
Alternative of https://github.com/rectorphp/rector/pull/5820


samsonasik:
PHPStan error caused by rectify seems unrelated.

samsonasik:
closed as use #5820 instead.

Issue #5819, opened 2021-03-10
[PHPUnit] Replace "$this->at" by "$this->withConsecutive" and "$this->willReturnOnConsecutiveCalls" 


Bl00D4NGEL:
# Feature Request

[PHPUnit 9 deprecated the `at` function and will remove it in PHPUnit 10](https://github.com/sebastianbergmann/phpunit/issues/4297). I would love a rector that can automatically convert the `at(x)` to the suggested `withConsecutive` and `willReturnOnConsecutiveCalls`

## Diff 

Before:
```php     
$this->pageBuilder
    ->expects($this->exactly(3))
    ->method('create')
    ->withConsecutive([$this->collection[0]], [$this->collection[1]], [$this->collection[2]]);

$this->pageBuilder
    ->expects($this->at(1)) // the middle creation will error
    ->method('create')
    ->willThrowException($exception);
```

After:
```php
$this->pageBuilder
    ->expects($this->exactly(3))
    ->method('create')
    ->withConsecutive([$this->collection[0]], [$this->collection[1]], [$this->collection[2]]);
    ->willReturnOnConsecutiveCalls(
        null,
        $this->throwException(new RuntimeException('Some message')),
        null
    );
```

TomasVotruba:
Closes https://github.com/rectorphp/rector/pull/5822

PR #5822, opened 2021-03-11
Migrate at to with consecutive and will return on consecutive calls

Bl00D4NGEL:
refs https://github.com/rectorphp/rector/issues/5819

Checklist:
- [X] Keep expressions from existing `willX(xxx)` and add them to the `willReturnOnConsecutiveCalls`
- [X] If no `willX` statements exist no `willReturnOnConsecutiveCalls` is added
- [X] If no `with` statements exist no `withConsecutive` is added
- [X] Replaces all previous expectation nodes and only replaces the last node (read: `\PhpParser\Node\Stmt`)
- [x] Can differentiate between different variable expectations
- [x] Assertions that expect different `method`s for the same variable (read: mock) are skipped 
- [x] Assertions that have 1 or more `withX` statements but not all are declared the expectation is skipped (e.g. if a `willX` statements exist for `at(0)` and `at(2)` but not for `at(1)` the assertion is skipped)
- [x] Assertions that have no `willX` but missing `with` statements are refactored to `withConsecutive` and an empty array is put in place of the missing index.

Notes:
It is possible to refactor a missing `willX` to a simple `null` value as this will break existing tests/code.
One could circumvent this by not returning null but instead an anonymous class that implements `PHPUnit\Framework\MockObject\Stub\Stub` like this:
```php
        $someClass = new class() implements Stub {
            public function toString(): string
            {
                return '';
            }

            public function invoke(Invocation $invocation)
            {
                return $invocation->generateReturnValue();
            }
        };
```
but this is, in my opinion, out of scope of this rector / pull request.
It seems that not the same is true for missing `with` statements as it seems to be totally valid to replace
```php
       $mock
            ->expects($this->at(0))
            ->method('someMethodThatReturnsAString');
        $mock
            ->expects($this->at(1))
            ->method('someMethodThatReturnsAString')
            ->with('1');
````

with

```php
        $mock
            ->method('someMethodThatReturnsAString')
            ->withConsecutive([], ['1']);
```


Bl00D4NGEL:
@TomasVotruba not sure what's happening but it can't seem to find my newly added classes, any idea what's happening? Is it maybe cause I'm developing this on Windows with the files being on my WSL? Any help would be highly appreciated! :)

TomasVotruba:
Hi,

that's because in  https://github.com/rectorphp/rector/pull/5826 we simplifed the structure to improve PSR-4 autoloading.

You need to update your paths to match it:

```diff
-rules/phpunit/src/...
+rules/PHPUnit/...
```

```diff
-rules/phpunit/tests/...
+rules-tests/PHPUnit/...
```

Bl00D4NGEL:
@TomasVotruba thanks! That fixed it :)
PR is now ready for review

TomasVotruba:
Wow, that's a huge feature. Thank you for it :+1: 

Looks good! Let's shipt it :)

PR #5828, opened 2021-03-12
remove runnable interface, to lower complexity

TomasVotruba:


TomasVotruba:
Surpassed by https://github.com/rectorphp/rector/pull/5829

PR #5831, opened 2021-03-13
[CI] Drop fixture checks, that are now covered by static reflection

TomasVotruba:
After https://github.com/rectorphp/rector/pull/5665, each fixture file is parsed as unique per-test file content.
Before we needed unique fixture, to avoid autoloading crashes. Now the static reflection removes this problem.

It was useful in time before though :+1:  cc @samsonasik


rules-tests/TypeDeclaration/Rector/ClassMethod/AddMethodCallBasedStrictParamTypeRector/Fixture/improve_nullable.php.inc
```diff
@@ -36,7 +36,7 @@ use PhpParser\Node;
 use PhpParser\Node\Expr\Array_;
 use PhpParser\Node\Stmt\Nop;
 
-final class ImproveNullable
+final class DoubleCallArray
```
TomasVotruba, line 39: This class is now duplicated, but tests still work

samsonasik:
@TomasVotruba Thank you 👍

PR #5833, opened 2021-03-13
[TypeDeclaration] Skip null type on non-private property and not used on CompleteVarDocTypePropertyRector

samsonasik:
Skip null default value on non-private property, eg: public can be used by outside class.


samsonasik:

TomasVotruba:
:+1:

PR #5845, opened 2021-03-14
[CodeQuality] Skip ChangeArrayPushToArrayAssignRector not in expression

### LINKED ISSUE #5844: Incorrect behavior of ChangeArrayPushToArrayAssignRector
# Bug Report

See https://getrector.org/demo/77ca42cb-c843-4fde-a45f-befd62bd03f3

```php
<?php
function force_balance_tags( $text ) {
	$tagstack  = array();
	$stacksize = 0;
	$tag = 'a';
	$stacksize = array_push( $tagstack, $tag );
}
```

### Responsible rules

* `ChangeArrayPushToArrayAssignRector`

## Expected Behavior

Rector should not apply this rule if the return value is going to be used.
Rector does shows this error when it's run:
```
[ERROR] Could not process  file, due to:                    
         "Node "PhpParser\Node\Expr\FuncCall":7 is child of "PhpParser\Node\Expr\Assign", so it cannot be removed as it would break PHP code. Change or remove the parent node instead.". 
```
The demo environment doesn't show this error and it would be nice if it did.
Also it'd be nice if errors like this would automattically skip rules so things could continue.

samsonasik:
Fixes #5844

### DIFF
--- a/rules-tests/CodeQuality/Rector/FuncCall/ChangeArrayPushToArrayAssignRector/Fixture/skip_array_push_not_in_expression.php.inc
+++ b/rules-tests/CodeQuality/Rector/FuncCall/ChangeArrayPushToArrayAssignRector/Fixture/skip_array_push_not_in_expression.php.inc
@@ -1,0 +1,13 @@
+<?php
+
+namespace Rector\Tests\CodeQuality\Rector\FuncCall\ChangeArrayPushToArrayAssignRector\Fixture;
+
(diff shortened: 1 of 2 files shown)


samsonasik:

superdav42:
I can confirm this PR resolves the issue in the simple test and when applied to the legacy codebase I'm working with.

TomasVotruba:
Thank you 👍

PR #5856, opened 2021-03-14
[Scoper] Try only replace in PrettyVersions only

samsonasik:


TomasVotruba:
Amazing work, CI is full green :heavy_check_mark:  :checkered_flag:  Thanks a lot!

image

PR #5857, opened 2021-03-14
[scoped] rector now runs without command name too

TomasVotruba:
Rector now runs also without `process` command.

This PR fixes exception if no config is around https://github.com/rectorphp/rector-prefixed/runs/2107855629

/cc @samsonasik


samsonasik:
@TomasVotruba thank you 👍

PR #5858, opened 2021-03-14
Bugfix for AbsolutizeRequireAndIncludePathRector

Fonata:
Absolute paths should not be changed, dots need a leading slash.

Here is the original code I found this bug with:
- `include('../../prepend_light.php'); // unmodified`

After running Rector:
- `include(__DIR__ . '../../prepend_light.php'); // incorrectly modified (without this PR)`
- `include(__DIR__ . '/../../prepend_light.php'); // correctly modified (with this PR)`

See https://www.php.net/manual/en/function.include.php for how PHP treats the string passed to `include` and the related language constructs.

I am a bit confused how the broken PHPUnit fixtures were checked. What am I forgetting? 🤔


TomasVotruba:
Looks good :+1:  Thanks you

Ready to merge?

Fonata:
Yes, from my perspective this is ready to be merged.

TomasVotruba:
Thank you :+1:

PR #5859, opened 2021-03-14
Bugfix for SimplifyForeachToArrayFilterRector

Fonata:
Bugfix: leave the loop as it is if the array looped over is modified within the loop (rather then filled from scratch).

I have added an example to the PHPUnit tests which this Rector rule should not touch. The example is odd, but harmless. If the rule applies it would break.


TomasVotruba:
Thanks :+1:

PR #5882, opened 2021-03-17
Replace StaticRectorStrings with Stringy class, remove CamelCaseFunctionNamingToUnderscoreRector to avoid naming rules

TomasVotruba:


shaffe-fr:
Hi @TomasVotruba, 
I found the `CamelCaseFunctionNamingToUnderscoreRector` rule pretty interesting and I don't find an equivalent.
I'm curious about the reason for its removal? What to you mean by "to avoid naming rules"?
Thank you!

PR #5888, opened 2021-03-17
[DeadCode] Remove RemoveUnusedParameterRector

TomasVotruba:
This rule can be very unreliable without knowing full code context and deep project analysis.
We don't have it even actived here, so better remove it to avoid causing fatal removals


TomasVotruba:
Yet it can be used on private methods :)

Issue #5891, opened 2021-03-18
Add ability to extends custom SetList


samsonasik:
# Feature Request

Currently, it seems there is no way to have custom `SetList` as the configuration is a constant and the config is early read:

https://github.com/rectorphp/rector/blob/c815334bed781c555f08dff82ce7861d758e9e7a/packages/Set/RectorSetProvider.php#L30

ref https://github.com/laminas/laminas-servicemanager-migration/pull/2#issuecomment-801530436

TomasVotruba:
How can we abstract this without hardcoding the values in `RectorSetProvider`?

samsonasik:
probably check at `RectorConfigsResolver` ?

https://github.com/rectorphp/rector/blob/a389839c5cb0dbcbfc480012bee31762e336011b/src/Bootstrap/RectorConfigsResolver.php#L58-L87

TomasVotruba:
How would it look like without hardcoding `DowngradeSetList` and `SetList`?

samsonasik:
I am not sure, how about create static property and update if not found?

samsonasik:
I found the solution for it with import via `$containerConfigurator->import()`:

```php
<?php
// rector.php
declare(strict_types=1);

use Rector\Core\Configuration\Option;
use Laminas\ServiceManager\Migration\Rector\Set\ValueObject\SetList;
use Rector\CodingStyle\Rector\FuncCall\CountArrayToEmptyArrayComparisonRector;
use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;

return static function (ContainerConfigurator $containerConfigurator): void {
    $parameters = $containerConfigurator->parameters();
    $services = $containerConfigurator->services();
    $containerConfigurator->import(SetList::LAMINAS_SERVICEMANGER_40);

    $parameters->set(Option::PATHS, [__DIR__ . '/module']);
    $services->set(CountArrayToEmptyArrayComparisonRector::class);
};
```

![Screen Shot 2021-03-20 at 18 04 03](https://user-images.githubusercontent.com/459648/111867464-ac1b9f80-89a6-11eb-867e-d9ad97afacec.png)

I am closing it.

TomasVotruba:
Great approach, I didn't think of this :+1: 
Set constant is basically config file, so this makes 100 % sense.

PR #5894, opened 2021-03-18
[Set] Handling possible custom set list

samsonasik:
Fixes https://github.com/rectorphp/rector/issues/5891

### DIFF
--- a/bin/rector.php
+++ b/bin/rector.php
@@ -8,10 +8,12 @@ use Rector\Core\Console\ConsoleApplication;
 use Rector\Core\Console\Style\SymfonyStyleFactory;
 use Rector\Core\DependencyInjection\RectorContainerFactory;
 use Rector\Core\HttpKernel\RectorKernel;
+use Rector\Core\ValueObject\Bootstrap\BootstrapConfigs;
 use Symplify\PackageBuilder\Console\ShellCode;
 use Symplify\PackageBuilder\Reflection\PrivatesCaller;
 use Symplify\SetConfigResolver\Bootstrap\InvalidSetReporter;
 use Symplify\SetConfigResolver\Exception\SetNotFoundException;
+use Symplify\SmartFileSystem\SmartFileInfo;
 use Tracy\Debugger;
 
 // @ intentionally: continue anyway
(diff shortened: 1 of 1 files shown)


bin/rector.php
```diff
@@ -48,6 +49,16 @@
...
 } catch (SetNotFoundException $setNotFoundException) {
+    // possible custom set list
+    $match = Strings::match($setNotFoundException->getMessage(), '#Set (".*") was not found#');
+    if ($match) {
+        $path = realpath(trim($match[1], '"'));
+        if ($path) {
+            // eg : /Users/samsonasik/www/laminas-mvc-skeleton/vendor/laminas/laminas-servicemanager-migration/config/rector/set/laminas-servicemanager-40.php
+            echo $path;die;
+        }
+    }
```
samsonasik, line 60: Hi @TomasVotruba , this is where the path is catched when we define custom `SetList` class, do you have idea to make it work to read `$path` as load config from it ?

ref https://github.com/rectorphp/rector/issues/5891
ref https://github.com/laminas/laminas-servicemanager-migration/pull/2#issuecomment-801530436

samsonasik: @TomasVotruba it works 🎉 , I updated to try re-create container with path found https://github.com/rectorphp/rector/pull/5894/commits/d96afeca8d4d8570ee341eddef88288649b6194b

bin/rector.php
```diff
@@ -47,9 +50,26 @@
...
+                $invalidSetReporter = new InvalidSetReporter();
+                $invalidSetReporter->report($setNotFoundException);
+                exit(ShellCode::ERROR);
+            }
+        }
+    } else {
+        $invalidSetReporter = new InvalidSetReporter();
+        $invalidSetReporter->report($setNotFoundException);
+        exit(ShellCode::ERROR);
+    }
```
TomasVotruba, line 72: This should not require a container to boot. Purpose of this is to keep thin bin file.

Better approach might something `MissingRectorRulesReporter`, only focused on sets:
https://github.com/rectorphp/rector/blob/main/src/Reporting/MissingRectorRulesReporter.php

samsonasik: I don't understand, as at this point, the custom `Laminas\ServiceManager\Migration\Rector\Set\ValueObject\SetList` is never catched as not registered at 

https://github.com/rectorphp/rector/blob/c815334bed781c555f08dff82ce7861d758e9e7a/packages/Set/RectorSetProvider.php#L30

The get by error message is got a full path of set list constant value.

The `MissingRectorRulesReporter` seems only for reporting, not make the custom `SetList` works.

How to make it work if not re-create the `$container` ? is

samsonasik: Tried add both `parameters` and `services`. The `services` part not works, only the parameters part.

samsonasik:
I am closing it as the solution can use `$containerConfigurator->import()` like commented at https://github.com/rectorphp/rector/issues/5891#issuecomment-803291383

PR #5895, opened 2021-03-18
Improve static reflection speed

TomasVotruba:


packages/NodeTypeResolver/Reflection/BetterReflection/SourceLocatorProvider/DynamicSourceLocatorProvider.php
```diff
@@ -49,6 +54,10 @@ public function addFiles(array $files): void
 
     public function provide(): SourceLocator
     {
+        if ($this->cachedSourceLocator) {
+            return $this->cachedSourceLocator;
+        }
```
TomasVotruba, line 59: Without this this dynamic source locator was created on every identifier call

PR #5899, opened 2021-03-18
Fix for PhpSpec method shouldEqual

laurent35240:
Hi

Here is a quick fix for handling PhpSpec method `shouldEqual` when using rector suite PhpSpecToPHPUnit


TomasVotruba:
Thank you :+1:

PR #5902, opened 2021-03-18
Forward compatibility with PHPStan 0.12.82

ondrejmirtes:
ObjectTypeMethodReflection is just one of the implementations of MethodReflection interface, its existence was never part of a public API promise. Now a different implementation is returned, it depends on the situation. I guess this change is sufficient.


TomasVotruba:
Thanks :+1:

PR #5910, opened 2021-03-19
docs: fix path to tests and namespace in "how to add a test"

staabm:
it seems tests were moved to `rules-tests` lately and namespaces therefore changed a bit.

this PR tries to adjust the docs for the changed conditions.


samsonasik:
Thank you @staabm

PR #5932, opened 2021-03-21
[DeadCode] Skip RemoveUnusedPrivateMethodRector when class has magic __call

samsonasik:


samsonasik:

rules-tests/DeadCode/Rector/ClassMethod/RemoveUnusedPrivateMethodRector/Fixture/skip_magic_call.php.inc
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Rector\Tests\DeadCode\Rector\ClassMethod\RemoveUnusedPrivateMethodRector\Fixture;
+
+/**
+ * @see https://3v4l.org/Lh5O5
+ */
+class SkipMagicCall
+{
+    private function test($params)
```
staabm, line 10: Why should „test“ not be removed.. its still unused?

samsonasik: I've updated fixture test case to proof it https://github.com/rectorphp/rector/pull/5932/commits/bafe6d097c890ffa2c7f92444b9f8f709936de91 . It called with `$this->__call('test', []);`

samsonasik:

rules/DeadCode/Rector/ClassMethod/RemoveUnusedPrivateMethodRector.php
```diff
@@ -108,6 +108,10 @@ private function shouldSkip(ClassMethod $classMethod): bool
         }
 
         // skip magic methods - @see https://www.php.net/manual/en/language.oop5.magic.php
-        return $classMethod->isMagic();
+        if ($classMethod->isMagic()) {
+            return true;
+        }
+
+        return $classReflection->hasMethod('__call');
```
staabm, line 115: Should this checks for invocations of __call() then?

samsonasik: `__call` is common used, and running the check can be on specific directory, not the caller, I think currently, this is the safest way.

TomasVotruba:
Thank you :+1:

PR #5934, opened 2021-03-21
[DeadCode] Skip RemoveAssignOfVoidReturnFunctionRector on variable assign used

samsonasik:


samsonasik:
Also update `BetterNodeFinder->findFirstNext()` to not lookup next when parent node is FunctionLike as it out of scope.

samsonasik:
reverted `findFirstNext()` not lookup outside for now FunctionLike_ parent as cause rectify running to non-related code.

samsonasik:
The rectify seems unrelated with the patch. I am closing for now and will try re-open it.

Issue #5938, opened 2021-03-21
Register invalid rule in services in config skipped, should throw exception


samsonasik:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | v0.9.33             |
| Installed as   | composer dependency |

Register invalid services in rector config seems skipped and not show any error, eg:

```php
<?php

use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;

return static function (ContainerConfigurator $containerConfigurator): void {
    $services = $containerConfigurator->services();
    $services->set(Foo\Bar\Baz::class);
};
```

ref https://getrector.org/demo/3b3215ad-2086-4f2d-8a47-66253d989b95

## Expected Behaviour

It should throw an exception that the service not found, to avoid user not know that service already deleted/moved, eg `DeadDocBlock` moved to `DeadCode` in https://github.com/rectorphp/rector/pull/5936 , the user will not know if service already moved.

TomasVotruba:
Closing as this should be handled in the project itself:

- by Symfony itself - https://github.com/rectorphp/rector/pull/5940#issuecomment-803630739
- by PHPStan rule for checking class existance - https://github.com/rectorphp/rector/runs/2160177167

PR #5940, opened 2021-03-21
[Services] Failing test register invalid service

samsonasik:


samsonasik:
@TomasVotruba this should throw exception and inform user that service is not exists, ref https://github.com/rectorphp/rector/issues/5938

TomasVotruba:
How would you recommend to check this in Symfony? I assume this should be handled in the framework itself

samsonasik:
if we can get the `container->has()` check somewhere, probably we can override via event listener

TomasVotruba:
That might create extra dependency on event listener, that we don't need here. 

But similar operation can do compiler pass on container compilation:
https://symfony.com/doc/current/service_container/compiler_passes.html

TomasVotruba:
Yet, I think this should be resolved in Symfony core level, as any Symfony project will suffer this problem, not just Rector.

samsonasik:
Ok, I will check if I can do something with Compiler Pass for now.

I am closing for now.

Issue #2971, opened 2020-02-29
Parallel run speedup feature


TomasVotruba:
# Inspiration

- https://github.com/phpstan/phpstan-src/commit/9124c66dcc55a222e21b1717ba5f60771f7dda92
- https://github.com/phpstan/phpstan-src/pull/125
- https://github.com/symplify/symplify/pull/3325

## Notes

* https://tomasvotruba.com/blog/introducing-up-to-16-times-faster-easy-coding-standard/

TomasVotruba:
@JanMikes this is starting point with few resources

TomasVotruba:
@JanMikes Here are the 3 processed I talked about yesterday:
https://github.com/rectorphp/rector/blob/220e2c9daf444bd194e167f11d3e92f743aed248/src/Application/RectorApplication.php#L140-L169

Each of them needs to be parallelized:

- run step 1 in parallel
- finish
- run step 2 in parallel
- finish
- run step 3 in parallel
- finish

JanMikes:
I considered following packages:
- AMP
- ReactPHP
- Swoole
- Spatie/async

AMP is the winning choice for me and i will implement parallel running with.

I dont like ReactPHP and Swoole bcoz they both use sockets and AMP has better performance than Spatie/async + spatie/async probably does not work on windows (due to how it handles signals in unix vs windows.

vladyslavstartsev:
@JanMikes @TomasVotruba any progress on this?

vladyslavstartsev:
this wil be SUPER useful

TomasVotruba:
> @JanMikes @TomasVotruba any progress on this?

Not really. 

> this will be SUPER useful

You can't imagine how useful this would be.
This challange is looking for its hero :)

JanMikes:
Hi, unfortunately this issue is currently blocked - i have not found a way how to serialize+deserialize nodes from php parser.

Because of how parallelism works and the fact, that parent process must be able to communicate (exchange data) with children processes in some format (for example json) and since we need to pass nodes between processes and not able to, i could not get it working.

vladyslavstartsev:
@JanMikes just an idea. Maybe it's worth a look at how psalm/phpstan/php_codesniffer does this?

ruudk:
I'm trying to run rector on a very large project and every run takes ±60 minutes on a 3,6 GHz 8-Core Intel Core i9. 

[...]

PR #5958, opened 2021-03-22
Remove buggy rules that depend on false condition of all class and calls analysis, to improve static reflection

TomasVotruba:
Due to static reflection in Rector 0.10 #3490, the autoloading of whole project is not require anymore. On the other hand, some rules required full composer autoload of all classes.

These rules attempted to detect unused public methods, yet it's hard to determine if that is true or not. E.g. in case of magic methods without parent contract interface, Doctrine entities, controller/presenter methods, template calls of objects etc. Magical limit is too high. 

This caused false positives on removed methods, privatized constants od method that were actually used.
Another bad side effect was slow performance, as whole project had to be analysed with clear-cache. In the end we didn't even used this rules on Rector itself.

One more problem: these rules cripled parallel run idea, as big amount of metadata had to be passed between runs, compared to other rules that were independent on global context of the project.

Saying that, this is out of Rector scope. It's a place for advanced too, another package or even private service.

This is concerned ~10 rules. To avoid more problems caused to Rector users, we remove them.

Feel free to re-use them, making own packages or wrapper tool. It has a potential.

<br>

Similar tool doing similar work phpdcd was deprecated couple of years go, possibly due to similar "too-magical" reason: https://github.com/sebastianbergmann/phpdcd

<br>

**PHPStorm does much better job in this matter**, suggesting dead methods right in the code. It's also easier to verify it there by looking for a string name of the method.

image

## See Related Sources

- https://github.com/sebastianbergmann/phpdcd
- https://stackoverflow.com/questions/11532/how-can-i-find-unused-functions-in-a-php-project


TomasVotruba:
cc @JanMikes This will make paralel much easier to implement :)

packages/Caching/Contract/Rector/ZeroCacheRectorInterface.php
```diff
@@ -1,13 +0,0 @@
...
-
-namespace Rector\Caching\Contract\Rector;
-
-/**
- * Rectors implementing this interface require to run with --clear-cache, so full application is analysed.
- * Such rules can be remove unused public method, remove unused class etc. They need full app to decide correctly.
- */
-interface ZeroCacheRectorInterface
-{
-}
```
TomasVotruba, line 13: @mssimi Hey, we've finally cleanup the broken dead-code rules :) thanks for the interface, it's easier to find them now

TomasVotruba:
Ref #2971

Issue #5948, opened 2021-03-22
Reduce package size


olivernybroe:
# Feature Request
There are a lot of folders which are included in the composer package which could be removed and reduce the package size + increase user's local indexing, as there would be fewer files.

This could be done via `.gitattributes` file  or `archive.exclude` in `composer.json`.

TomasVotruba:
Could you send a PR?

olivernybroe:
What would you prefer, `.gitattributes` or `composer.json` solution?

TomasVotruba:
What do you recommend based on your experience?
Including building prefixed Rector

olivernybroe:
Hmm I normally see other projects going with `.gitattributes`, so I think that would be the easiest to maintain. 

Just quickly read up on how the prefixed rector works, and unfortunately this approach won't affect it. 

I think the best solution would be to write a script that deletes the files in the prefixed rector after building based on the `.gitattributes` information.

TomasVotruba:
Sounds good :+1:

olivernybroe:
I can see a lot of the tests classes were actually already fixed when upgrading to a newer version, as all tests were moved inside `rules-test` instead.

TomasVotruba:
Yes, I realized we have too complicated namespace imports :) 

So I made refactoring to 2 dirs:

- https://github.com/rectorphp/rector/pull/5826
- https://github.com/rectorphp/rector/pull/5824

olivernybroe:
That's soo awesome! 😄 
Those changes also made such big impact on the package size, so really a win-win.

TomasVotruba:
Btw, could you compare 0.9 size and 0.10 size after those 2 refactoring? I'm curious how much it helped :)

olivernybroe:
Rector 0.10.3 = 5.5 MB
Rector 0.9 = 10.6 MB

So about half the size 👍 That's pretty good so far

TomasVotruba:
That's very good :+1:

PR #5960, opened 2021-03-23
Add missing paths to .gitattributes

olivernybroe:
I have added a bunch of files to being ignored in exports. 

I think the biggest difference here is adding docs folder as it contains pictures. 

I have also re-added back the license file, as I don't think the license should be excluded when I include the package 👍 

I was unsure if the `phpstan-for-rector.neon` file should be excluded or not, so I left it in.

closes: #5948

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,7 +12,6 @@
 .gitignore export-ignore
 .phpstorm.meta.php export-ignore
 CODE_OF_CONDUCT.md export-ignore
-LICENSE export-ignore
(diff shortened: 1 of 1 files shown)


TomasVotruba:
> I was unsure if the phpstan-for-rector.neon file should be excluded or not, so I left it in.

That is used only locally in this repository, so should be excluded too.

TomasVotruba:
Mergin as it is to speed this up

PR #5973, opened 2021-03-24
Readme and docs should be included in rector installation

samsonasik:
I think Readme and docs files should be included in rector installation so user doesn't need to look at github repo to read it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -36,7 +36,6 @@ scoper.php export-ignore
 .github export-ignore
 .dockerignore export-ignore
 CONTRIBUTING.md export-ignore
-README.md export-ignore
 CHANGELOG.md export-ignore
 
 # testing Windows spaces - https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings
--- a/.gitattributes
+++ b/.gitattributes
@@ -30,7 +30,6 @@ full_build.sh export-ignore
 Dockerfile export-ignore
 UPGRADE_09.md export-ignore
 rule-doc-generator.php export-ignore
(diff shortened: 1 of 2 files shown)


samsonasik:

olivernybroe:
Hi @samsonasik,

I think we should consider at least not including `docs` folder or at least not `docs/images`, as images are often a big sin when talking filesize.

samsonasik:
@olivernybroe the readme is consuming `docs/images` for displaying logos, so probably `docs/images` still needed. 

About `phpstan-for-rector.neon`, @TomasVotruba is the `phpstan-for-rector.neon` a fallback when there is no `Option::PHPSTAN_FOR_RECTOR_PATH` defined?

olivernybroe:
@samsonasik Yes, we use it for images in the readme. However I don't think it hurts the readme that much that we don't display the images when browsing the readme through vendor folder.
I think that's better than adding 1MB of extra package size, for logos. 

We could potentially also just change the images to be url's to main branch instead? That way they will show locally if you have a internet connection.

samsonasik:
Thank you. I've updated to use URL for images location, re-add back `docs` to `.gitattributes`.

samsonasik:

TomasVotruba:
Thank you both :+1: 

> About phpstan-for-rector.neon, @TomasVotruba is the phpstan-for-rector.neon a fallback when there is no Option::PHPSTAN_FOR_RECTOR_PATH defined?

This config is needed only in our repository here. Each project can to defined their own config if needed.

Issue #5454, opened 2021-02-08
CorrectDefaultTypesOnEntityPropertyRector throws error with null default value for integer doctrine entity field


runawaycoin:
# Question
Not sure if this is a bug / not catered for yet or I'm doing something wrong?

When using the `CorrectDefaultTypesOnEntityPropertyRector` i'm hitting the `NotImplementedYetException` inside `refactorToIntType` with the following Doctrine Entity code. Its due to the null default value.

```
 /**
     * @ORM\Column(type="integer", nullable=true)
     */
    private ?int $myInt = null;
```

The error i get is:

```
[ERROR] Could not process "src/Entity/MyEntity.php" file by                                                            
         "Rector\DoctrineCodeQuality\Rector\Property\CorrectDefaultTypesOnEntityPropertyRector", due to:                
         "".                                                                                                            
                 
```

TomasVotruba:
Hi, we'll need a failing demo link to have context: https://getrector.org/demo

runawaycoin:
Hi,

Great tool, sorry i didn't use this in my initial post.

I think this is reproducing it:
https://getrector.org/demo/3b637639-ed29-4fd9-941c-caf89cb9bf22

Thanks.

TomasVotruba:
Thank you for your report and demo link!

Could you send a failing test case in a pull-request, so we have it covered in Rector?
Here is step by step tutorial how to add it: https://github.com/rectorphp/rector/blob/master/docs/how_to_add_test_for_rector_rule.md

TomasVotruba:
We've just switch to a custom static doctrine annotation parser, where this issues should be resolved.

https://github.com/rectorphp/rector/pull/5974

Issue #4334, opened 2020-09-30
[FPDP] Format preserving doc printer


TomasVotruba:
Mirror of https://github.com/nikic/PHP-Parser/issues/344, just for phpdoc

It might be worth to create https://github.com/phpstan/phpdoc-parser which actually preserver format

Here is knowledge snippet from php-parser: https://github.com/nikic/PHP-Parser/issues/487

## Related Issues

- https://github.com/rectorphp/rector/issues/3932

staabm:
> It might be worth to create [phpstan/phpdoc-parser](https://github.com/phpstan/phpdoc-parser) which actually preserver format

what does this mean? do you have a plan/a idea of what needs to be done?

TomasVotruba:
We need a new package that builds on https://github.com/phpstan/phpdoc-parser,
but also add printing the docblock back.

I don't have any ETA on that.

staabm:
Whats the job of this package? What should it provide on top of what phpstan/phpdoc-parser already can?

TomasVotruba:
> Whats the job of this package? 

Current job of existing package is getting docs block to types.

E.g.

```php
/**
 * @var int
 */
```

↓

Very rougly:

```php
new VarTagNode(new IntegerType())
```

The problem is, all these cases:

```php
/**
 * @var int
 */

/**
 * @var integer
 */

/**
 *  @var     int
 */
```

Are printed back as:

```php
/**
 * @var int
 */
```

<br><br>

> What should it provide on top of what phpstan/phpdoc-parser already can?

Keep the format in it's original form.

TomasVotruba:
Work in progress https://github.com/rectorphp/phpdoc-parser-printer

TomasVotruba:
Big jump forward via:

- https://github.com/rectorphp/rector/pull/5841
- and native attribute support in phpdoc-parser https://github.com/phpstan/phpdoc-parser/pull/65 :tada:

TomasVotruba:
We've just added another jump with static doctrine annotatoin parsing https://github.com/rectorphp/rector/pull/6040

Before, to parse class-based annotations the https://github.com/doctrine/annotations was needed:

```php
use Doctrine\ORM\Mapping as ORM;

/**
 * @ORM\Entity
 */
class Product
{
}
```

[...]

PR #5974, opened 2021-03-24
Refactor doctrine/annotation from dynamic to own static-reflection parser

TomasVotruba:
Follow up to #5665 

Now the Doctrine annotations need classes to be autoloaded, because they use https://github.com/doctrine/annotations

```php
use Symfony\Component\Routing\Annotation\Route;

final class SomeController
{
    /**
     * @Route(path="/some_path", name="some_name")
     */
    public function some()
    {
    }
}
```

This PR replaced doctrine/annotations package with custom parser that mimics the behavior, but uses phpstan/phpdoc-parser with static reflection :+1: 

<br>

That means, the doctrine annotation classes don't have to be autoloaded anymore. It's alo simplified the doc node printing with much less code.

image


TomasVotruba:
This is big jump for annotation format preserving https://github.com/rectorphp/rector/issues/4334

Before, to parse class-based annotations the https://github.com/doctrine/annotations was needed:

```php
use Doctrine\ORM\Mapping as ORM;

/**
 * @ORM\Entity
 */
class Product
{
}
```

That required the `Doctrine\ORM\Mapping\Entity` class to be present and autoloaded. That kind-of kills the advantage of [static reflection added in Rector 0.10](https://getrector.org/blog/2021/03/15/legacy-refactoring-made-easy-with-static-reflection) for annotations.

Further more, this approach required every such annotation to have own php doc node class and factory to make it. In total, Rector contained ~50 such classes + factories. 

<br>

The https://github.com/rectorphp/rector/pull/5974 PR **solves this** using https://github.com/phpstan/phpdoc-parser and **copying the doctrine/annotation behavior 1:1 without autoloading used classes and constants**.

Now, Rector is now using static for annotations making it more accessible to wide PHP comunity :hugs: :+1:

TomasVotruba:
Closes https://github.com/rectorphp/rector/issues/5454

PR #5982, opened 2021-03-24
[DeadCode] Handle RemoveDefaultArgumentValueRector on removal default argument on not last argument

### LINKED ISSUE #5975: RemoveDefaultArgumentValueRector dropping arguments that aren't last
# Bug Report

`RemoveDefaultArgumentValueRector` drops arguments that have the default value and aren't last.

E.g. the minimal code below has different behaviour when the `$b` and `$c` arguments to `doStuff()` are true, however `RemoveDefaultArgumentValueRector` drops the argument to `$b` (replacing it with the argument to `$c`) if it finds a call to `doStuff()` with the default value for `$b`.

See https://getrector.org/demo/298459d2-590f-4fc3-af21-ae1370efa305

```php
<?php

function doStuff($a = false, $b = false, $c = false, $d = false) {
    $level = getLevelFromSomewhere();
    
    if ($level == "a" && !$a) return false;
    if ($level == "b" && !$b) return false;
    if ($level == "c" && !$c) return false;
    if ($level == "d" && !$d) return false;
}

doStuff(true, false, true, false);
```

### Responsible rules

* `RemoveDefaultArgumentValueRector`

## Expected Behavior

```php
<?php

function doStuff($a = false, $b = false, $c = false, $d = false) {
    $level = getLevelFromSomewhere();
    
    if ($level == "a" && !$a) return false;
    if ($level == "b" && !$b) return false;
    if ($level == "c" && !$c) return false;
    if ($level == "d" && !$d) return false;
}

doStuff(true, false, true);
```

samsonasik:
Fixes #5975

### DIFF
--- a/rules-tests/DeadCode/Rector/MethodCall/RemoveDefaultArgumentValueRector/Fixture/not_last_argument.php.inc
+++ b/rules-tests/DeadCode/Rector/MethodCall/RemoveDefaultArgumentValueRector/Fixture/not_last_argument.php.inc
@@ -1,0 +1,33 @@
+<?php
+
+namespace Rector\Tests\DeadCode\Rector\MethodCall\RemoveDefaultArgumentValueRector\Fixture;
+
+function notLastArgument($a = false, $b = false, $c = false, $d = false) {
+    $level = getLevelFromSomewhere();
+
+    if ($level == "a" && !$a) return false;
+    if ($level == "b" && !$b) return false;
+    if ($level == "c" && !$c) return false;
+    if ($level == "d" && !$d) return false;
+}
+
+notLastArgument(true, false, true, false);
+
+?>
+-----
+<?php
+
+namespace Rector\Tests\DeadCode\Rector\MethodCall\RemoveDefaultArgumentValueRector\Fixture;
+
+function notLastArgument($a = false, $b = false, $c = false, $d = false) {
+    $level = getLevelFromSomewhere();
+
+    if ($level == "a" && !$a) return false;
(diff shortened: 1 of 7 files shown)


samsonasik:

JulianCBC:
I hadn't looked at the code before now, but this seems a bit over complicated.

(Also, I must complement you on sorting this out so fast!)

Would something like this work:

```php
<?php

for ($key = count($node->args) - 1; $key--; $key >= 0) {
    if (! isset($defaultValues[$key])) {
        break;
    }

    if ($this->nodeComparator->areNodesEqual($defaultValues[$key], $arg->value) {
        $keysToRemove[] = $key;
    } else {
        break;
    }
}

return $keysToRemove;
```

samsonasik:
@JulianCBC your approach is not working, I tried that and cause error in other fixtures tests. 

Please provide another pull request if you have another approach, thank you.

samsonasik:
rebased.

samsonasik:

JulianCBC:
Ok, fair enough then. I'm not objecting to your approach - I just think it could be simpler.

samsonasik:
got another failing test case not in last very first set equal default https://github.com/rectorphp/rector/pull/5982/commits/27847a4c2ee07bbbb89c84678d1ca727de85d485

samsonasik:
fixed 🎉

samsonasik:

TomasVotruba:
Thank you :+1:

PR #5984, opened 2021-03-24
Add failing test fixture for RemoveUselessReturnTagRector

OskarStark:
# Failing Test for RemoveUselessReturnTagRector

Based on https://getrector.org/demo/507ba36e-e7da-4f4f-9d7f-a34a239606c2

IMHO this should be kept like it is


TomasVotruba:
Here we respect the PHPStan type structure, where self type is specific type of object:

image


image

PR #5986, opened 2021-03-25
[GitAttributes] Add phpstan-for-rector.neon to .gitattributes

samsonasik:
Based on https://github.com/rectorphp/rector/pull/5973#issuecomment-806228914


samsonasik:

Issue #5875, opened 2021-03-16
only written local variable in dead-code detection


staabm:
# Feature Request

local variables which are only written but never read should be removed
 
https://getrector.org/demo/cfb7fc79-d018-4fdc-8683-226fcc918406

## Diff 

before
```php
<?php

final class DemoFile
{
    public function run()
    {
        $fromDate = null;
        if (isset($params['todate'])) {
            $fromDate = 123;
            $z = true;
        } 
        
        if (isset($params['abc'])) {
            $fromDate = 456;
        }
        
        if ($z) {
            echo "yes";
        }
    }
}
```

after
```php
<?php

final class DemoFile
{
    public function run()
    {
        if (isset($params['todate'])) {
            $z = true;
        } 
                
        if ($z) {
            echo "yes";
        }
    }
}
```

_I will provide a failing test-case within this week_

Bl00D4NGEL:
Sounds more like a potential bug for the [RemoveUnusedVariableAssignRector](https://github.com/rectorphp/rector/blob/main/rules/DeadCode/Rector/Assign/RemoveUnusedVariableAssignRector.php). I guess it doesn't detect that this can be removed because it is written multiple times and not just once and therefore considered "used"

TomasVotruba:
This can be quite complicated to implement right. Nested if/foreaches
But if you would add this feature and the tests, we can merge it :+1: 

Closing as accepted if contributed

PR #5991, opened 2021-03-25
[DEAD-CODE] add failing testcase for only written local variable

fbecker-complex:
Testcase for https://github.com/rectorphp/rector/issues/5875


TomasVotruba:
Full feature is needed, see https://github.com/rectorphp/rector/issues/5875#issuecomment-816742328

Issue #5877, opened 2021-03-16
only read private properties should be removed in dead-code detection


staabm:
# Feature Request

the private property which is only read but never written to, should be removed.

https://getrector.org/demo/29847930-1490-4fee-b34b-96f8e21da596 

_I will provide a failing test-case within this week_

TomasVotruba:
> I will provide a failing test-case within this week

:+1:

staabm:
My fellow worker sent the failing test in the linked PR

TomasVotruba:
Full contribution is needed to accept this

PR #5992, opened 2021-03-25
[DEAD-CODE] add failing test case for only read private properties

fbecker-complex:
Testcase for https://github.com/rectorphp/rector/issues/5877


TomasVotruba:
Could you provide fix as well? We can merge it then

rules-tests/DeadCode/Rector/Property/RemoveUnusedPrivatePropertyRector/Fixture/only_read.php.inc
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Rector\Tests\DeadCode\Rector\Property\RemoveUnusedPrivatePropertyRector\Fixture;
+
+class OnlyRead
+{
+    private $unused;
```
staabm, line 7: beware: when the property has a default value, the code within the if is not dead and should be left untouched

TomasVotruba:
Full feature with passing tests is needed

PR #5999, opened 2021-03-26
[CodeQualityStrict] Allow MoveVariableDeclarationNearReferenceRector to pass static call on Throwable instance

samsonasik:
Throwable instance with static call mostly used for named constructor call, eg:

```php
throw MyException::notFound();
```

which usually not change external data. This allow to pass it to move variable after it.


samsonasik:

TomasVotruba:
Thank you :+1:

PR #6000, opened 2021-03-26
[Scoped] Add templates on build rector-prefixed-downgraded

### LINKED ISSUE #4636: "file does not exist" error when running rector-prefixed
# Bug Report

I am experiencing this error when installing rector-prefixed as a composer dependency, or when downloading the phar file from the rector-prefixed github page.

I run `$ php rector.phar init` or `app/Vendor/bin/rector init` and get:

```
Rector 0.8.52

In Filesystem.php line 38:

  Failed to copy "phar://C:/inetpub/<project>/rector.phar/src/Console/Command/../../../templates/rector.php.dist" because file does not exist.init
```

The version installed with composer is `"rector/rector-prefixed": "^0.8.52"`.

I am on Windows 10 and PHP 7.3 I've tried running it in cmd.exe, cygwin, and in vagrant, and get the same error in every case.

samsonasik:
Fixes #4636


samsonasik:
Let's try it to fix #4636

PR #6002, opened 2021-03-27
[TypeDeclaration] Skip null type on non-private property and not used on PropertyTypeDeclarationRector

samsonasik:
like in https://github.com/rectorphp/rector/pull/5833


samsonasik:

TomasVotruba:
Thank you :+1:

PR #6025, opened 2021-03-31
[CodeQuality] Handle colon in arg value for create variable or MoveOutMethodCallInsideIfConditionRector

### LINKED ISSUE #6020: Incorrect behavior of MoveOutMethodCallInsideIfConditionRector: colon in variable name
# Bug Report

When having the `MoveOutMethodCallInsideIfConditionRector` active the code in example get invalid. It has a `colon` in the varibable name.

See https://getrector.org/demo/cd295637-3264-4f40-8f9c-42a1ee8c6f38

```php
<?php
use DateTime;
final class DemoFile
{
    public function run()
    {
        $dt = new DateTime();
        if ($dt->format('H:i:s') == '00:00:00') {
            echo "test".PHP_EOL;
        }
    }
}
```

### Responsible rules

* `MoveOutMethodCallInsideIfConditionRector`

## Expected Behavior

If it should be skipped or not I cannot say, but if not skipped the `$H:i:sDtFormat` variable should not contain a `colon` in the variable name.

samsonasik:
Fixes #6020


samsonasik:

samsonasik:
@TomasVotruba I am merging it.

PR #6029, opened 2021-04-02
League event: change dispatch string events to objects

lulco:
Updated league event upgrade:
- changed AddInterfaceByParentRector
- added DispatchStringToObjectRector


rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
```diff
@@ -0,0 +1,103 @@
...
+
+        if (! $this->isObjectType($node->var, new ObjectType('League\Event\EventDispatcher'))) {
+            return null;
+        }
+
+        if (! $this->getStaticType($node->args[0]->value) instanceof StringType) {
+            return null;
+        }
+
+        $node->args[0] = new Arg(new New_(new Class_(null, [
```
samsonasik, line 86: please create new variables for the new statements to handle PHPStan new nest notice, there is 6 new nested here.

lulco: I did, and it throws another error. Something like array with keys is not allowed, use Value object. Any idea how to fix it?

samsonasik: That need one by one verify, you possibly need to create value object for keys in: 

```php
                new ClassMethod('eventName', [
                     'flags' => Class_::MODIFIER_PUBLIC,
                     'returnType' => 'string',
                     'stmts' => [
                         new Return_($node->args[0]->value),
                     ],
                 ]),
```

or move to other variables except above object instantiation.

lulco: I've extracted implements and statements to variables, please check it now. PHPStan is green

lulco:
Error in Packages Tests doesn't seem to be related to this PR

rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
samsonasik: probably need to wait green PR from @TomasVotruba at https://github.com/rectorphp/rector/pull/5974 to be merged and rebase after that.

samsonasik:
@lulco please rebase

lulco:
> @lulco please rebase

Sorry, I am not near computer. Maybe at night

lulco:
@samsonasik I've merge master into my branch, but the build is broken again because of this commit https://github.com/rectorphp/rector/commit/393031b0275f45cb53b3491b1ae9408c9628be62

lulco:
@samsonasik @TomasVotruba build is finally green, ready for review and merge O:-)

samsonasik (approved):

rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * @param MethodCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        $methodName = $this->getName($node->name);
+        if ($methodName !== 'dispatch') {
+            return null;
+        }
```
TomasVotruba, line 76: Here `isName()` should be used

rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
```diff
@@ -0,0 +1,105 @@
...
+        $statements = [
+            new ClassMethod('eventName', [
+                'flags' => Class_::MODIFIER_PUBLIC,
+                'returnType' => 'string',
+                'stmts' => [
+                    new Return_($node->args[0]->value),
+                ],
+            ]),
+        ];
+        $node->args[0] = new Arg(new New_(new Class_(null, [
```
TomasVotruba, line 98: This should be extracted to a method or new factory service to improve readability

rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
```diff
@@ -0,0 +1,105 @@
...
+                ,
+                <<<'CODE_SAMPLE'
+final class SomeClass
+{
+    /** @var \League\Event\EventDispatcher */
+    private $dispatcher;
+
+    public function run()
+    {
+        $this->dispatcher->dispatch(new class implements \League\Event\HasEventName
```
TomasVotruba, line 49: This seems very confusing. Could you link original upgrade source that recommend this?

It should be at least own variable...

```suggestion
-        $this->dispatcher->dispatch(new class implements \League\Event\HasEventName
+        $myEvent = new class implements \League\Event\HasEventName 
+        $this->dispatcher->dispatch($myEvent);
```

...or better standalone event class, as the event can be dispatched used on multiple places and it anonymous class makes event class duplicated per each instance. This is standard way in Symfony and Nette, that allow argument reusability and validation via value object ctor.

TomasVotruba:
Thank you for the PR :+1: I've added few comments

rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
lulco: There is no recommendation in original package, I just fall into this when upgrading one of my projects. Old version used to support string events. New doesn’t. I found this solution as the easiest way how to upgrade package without additional configuration. Of course the best way would be create some standalone event class, but what will be its name? What namespace will it use? Where it will be placed? Does Rector support creating new file at all?

TomasVotruba: There is quite similar rule for converting magic Nette `on*` properties to Symfony-like event classes:

https://github.com/rectorphp/rector-nette/tree/main/tests/Kdyby/Rector/MethodCall/ReplaceMagicPropertyEventWithEventClassRector

rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
lulco: Maybe you should create some phpstan rule for it? I didn't know I should use method isName instead of getName :)

rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
lulco: Thanks, I went through this and I still believe that anonymous class is the smartest and simplest solution for this (not very used use case). If somebody will want to create standalone class, he will create it before/after running rector. This anonymous class will not create new class(es) in the target project and will work like old version.

lulco: > It should be at least own variable...
> 
Can I change parent node in rector? If I want to create variable $myEvent before method call, like you suggested, I cannot modify method call but some other type of node right? Again, it will overcomplicate this simple solution

TomasVotruba:
Let's add this as it is, LGTM :+1:

PR #6035, opened 2021-04-04
Failing test: Node type is not updated after executing rector

lulco:
Running rename class rector and then rename method rector with expected new class name not working because node type is still the old class name


TomasVotruba:
This feature would be great to include. Now the old node types/name are applied here:
https://github.com/rectorphp/rector/blob/fb1e74615780a4bb196e8dca96dd0eaff35748e2/src/Rector/AbstractRector.php#L310

TomasVotruba:
Could you rebase on `main`? It might have changed

lulco:
I screwed something here O:-) new PR here https://github.com/rectorphp/rector/pull/6070, but it seems it still not working :(

TomasVotruba:
It seems you struggle with rebase of your fork. It a bit tricky, since you need to work with `origin` (your fork) and `upstream` (this repository) at the same time. I'll share how I handle it, in case you find it useful. If you don't need it, just ignore it :)

<br>

This will update your fork with this repository:

```bash
git remote add upstream git@github.com:rectorphp/rector.git
git fetch upstream
git merge upstream/main 
```

This will update your branch to main:

```bash
git checkout my-branch
git rebase -i main
```

This is needed to update commits and keep only yours:

```bash
git push --force
```

samsonasik:
I wrote a blog post about git rebase and in case there is git conflict https://samsonasik.wordpress.com/2015/09/16/practical-git-4-rebasing-conflicted-task-branch-against-primary-branch/

lulco:
Thank you both :) For now I will probably stay with merging main into my-branch :) it worked all the time before O:-)

PR #6051, opened 2021-04-07
Add failing test fixture for RemoveUnusedVariableAssignRector

zingimmick:
Variable used in property fetch and method call should be skipped.


samsonasik:
@zingimmick I cherry-picked your commit at PR https://github.com/rectorphp/rector/pull/6053

Issue #6059, opened 2021-04-08
Add generic processor interface to handle other file formats than php


sabbelasichon:
Rector is focusing on PHP-Files but already has some processing of non php files. In order to make this happen in a more generic and customizable approach it would be nice define custom processors implementing an interface into to process command. 

PR is coming soon.

sabbelasichon:
Closing the issue because it is done by #6060

Issue #5282, opened 2021-01-22
Change config (and other non-php) files too


lulco:
As we discussed on twitter, I am creating issue...

Since I live in Nette world I was focused on how to migrate configs of Nette applications (Neon files) 2.4 to new versions. I tried several approaches, which (maybe) will help you to do the right decision or at least they will show you what is not the right way:
1) Neon::decode -> Neon:encode
	pros:
		- changes on/off to true/false (because on/off are deprecated in 3.0)
 	cons:
 		- changed formatting (tabs instead of spaces)
		- adds numeric keys to anonymous services (`- Service1` is changed to `0: Service1`) - this is understandable since neon is just an array
		- ignores that class should be changed to factory - it is not its responsibility
		- ignores schema of extensions (for example: database › default › password should be string, so if your database password was 123, it will be unchanged)

2) Configurator + Compiler (idea)
Based on findings with approach 1) I was trying to build Container using Configurator + Compiler in while cycle and in each step I tried to fix one error (which was thrown by the Compiler). I didn't finish work on it.
	pros:
		- if it would work, it would be great self-correcting mechanism
	cons:
		- it would be extremely complicated and probably will not work anyway :)
			- we would need all config files, in same order as they are used in app
			- we have to clear cache after each change
		- it was slow

Approaches 1) and 2) are not generic for all frameworks and file types, so I went back to my very first choice: Regular expressions. I already use them when upgrading apps from Nette 2.4 to Nette 3.0 and they work just fine.
	pros:
		- they are generic (can be used also in another file types - latte, yaml, xml, twig, even in php for some "unrectorable" situations...)
		- I believe you can handle anything with good regex
		- it is already used in Rector with NonPhpFileClassRenamer
		- with --dry-run you can try it over and over again
	cons:
		- possible false positives
		- it is regex :) it is not so cool like the rest of Rector

After too many hours spend with this issue, I am convinced that replace strings in non-php files with regular expressions is the right and only possible way how to upgrade any application.

Aerendir:
I don't understand why use regular expressions.

Nette has a decoder for neon files: https://github.com/nette/neon

Why would you use reg exes instead of a simple array get using the package from Nette?

lulco:
@Aerendir Like I wrote, we need some generic solution. Nette::decode just create array from string. That is not bad, but we need to do some changes required by new versions of framework. Also there are some problems with formatting when we encode array back to string. See example below:

```
Original:
----------
# Global comment
parameters:
[...]

PR #6060, opened 2021-04-08
[FEATURE] Use generic processor interface for non php files

sabbelasichon:
Relates: #6059


rules/Composer/Processor/ComposerProcessor.php
```diff
@@ -32,75 +31,64 @@ final class ComposerProcessor
...
         $this->composerJsonFactory = $composerJsonFactory;
         $this->composerJsonPrinter = $composerJsonPrinter;
         $this->configuration = $configuration;
-        $this->errorAndDiffCollector = $errorAndDiffCollector;
-        $this->smartFileSystem = $smartFileSystem;
         $this->composerModifier = $composerModifier;
     }
 
-    public function process(string $composerJsonFilePath): void
+    public function process(SmartFileInfo $smartFileInfo): ?string
```
TomasVotruba, line 55: Decoupling saving file content from this method is good :+1:

rules/Composer/Processor/ComposerProcessor.php
```diff
@@ -32,75 +31,64 @@ final class ComposerProcessor
...
 
-    private function addComposerJsonFileDiff(
-        ComposerJson $oldComposerJson,
-        ComposerJson $newComposerJson,
-        SmartFileInfo $smartFileInfo
-    ): void {
-        $newContents = $this->composerJsonPrinter->printToString($newComposerJson);
-        $oldContents = $this->composerJsonPrinter->printToString($oldComposerJson);
-        $this->errorAndDiffCollector->addFileDiff($smartFileInfo, $newContents, $oldContents);
+    public function allowedFileExtensions(): array
```
TomasVotruba, line 82: This might be better name as it's about supported extensions. What do you think?

```suggestion
    public function getSupportedFileExtensions(): array
```

src/Console/Command/ProcessCommand.php
```diff
@@ -312,4 +307,53 @@ private function includeBootstrapFiles(): void
...
+
+        $nonPhpFileInfos = $this->filesFinder->findInDirectoriesAndFiles($paths, $nonPhpFileExtensions);
+
+        $composerJsonFilePath = getcwd() . '/composer.json';
+        if ($this->smartFileSystem->exists($composerJsonFilePath)) {
+            $nonPhpFileInfos[] = new SmartFileInfo($composerJsonFilePath);
+        }
+
+        return $nonPhpFileInfos;
+    }
```
TomasVotruba, line 358: There seems lot of new logic in here. The `ProcessCommand` should be as thin as possible, as it's only command line wrapper. The application logic should be somewhere else, so it's useable also without command.

Saying that, could you decopule these 2 methods to own service?

src/Contract/Processor/NonPhpFileProcessorInterface.php
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Rector\Core\Contract\Processor;
+
+use Symplify\SmartFileSystem\SmartFileInfo;
+
+interface NonPhpFileProcessorInterface
+{
+    public function process(SmartFileInfo $smartFileInfo): ?string;
+
+    public function canProcess(SmartFileInfo $smartFileInfo): bool;
```
TomasVotruba, line 12: Borrowing naming from PHP CS Fixer: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/cbde9be1d04d46ca351e24cf39067a63326b7adf/src/Fixer/FixerInterface.php#L76

```suggestion
    public function supports(SmartFileInfo $smartFileInfo): bool;
```

src/NonPhpFile/NonPhpFileProcessor.php
```diff
@@ -4,23 +4,17 @@
...
+use Rector\Core\ValueObject\StaticNonPhpFileSuffixes;
 use Rector\PSR4\Collector\RenamedClassesCollector;
 use Symplify\SmartFileSystem\SmartFileInfo;
-use Symplify\SmartFileSystem\SmartFileSystem;
 
 /**
  * @see \Rector\Tests\Renaming\Rector\Name\RenameClassRector\RenameNonPhpTest
  */
-final class NonPhpFileProcessor
+final class NonPhpFileProcessor implements NonPhpFileProcessorInterface
```
TomasVotruba, line 16: This service should be probably the one to put those 2 methods in.

It should accept array of `NonPhpFileProcessorInterface[]` services in constructor.

TomasVotruba:
Thanks :+1: , I've dropped few comments about location of methods

src/Contract/Processor/NonPhpFileProcessorInterface.php
```diff
@@ -0,0 +1,20 @@
...
+
+    public function transformOldContent(SmartFileInfo $smartFileInfo): string;
```
TomasVotruba, line 14: What is difference to process () method?
It seems there should be just one

sabbelasichon: The transformOldContent method is especially a workaround for the ComposerProcessor because it outputs the content differently.

src/NonPhpFile/NonPhpFileProcessor.php
sabbelasichon: I see. But renaming the old class seems hard, because it is a container for everything. So what should be the name of it?

TomasVotruba: Not sure, as I don't know the wholesome architecture yet. Any ideas?

src/Contract/Processor/NonPhpFileProcessorInterface.php
TomasVotruba: It should be merged into 1 method even in ComposerProcessor, to have clear API

src/NonPhpFile/NonPhpFileProcessor.php
sabbelasichon: No ideas, i just call it NonPhpFileProcessorService for now. Because i don´t really know what is the NonPhpFileProcessor is for.

TomasVotruba:
I'll have to tidy up the naming but current prototype works well.
Let's try test it in the wild :)

sabbelasichon:
You know. Naming things is hard. I am sorry not to came up with a better name. But the old was somehow very unspecific for me and i could not come up with a better one for now.

TomasVotruba:
No worries mate, it's like a new baby just born.... We have to get to know them better to know what name feels right for them :)

lulco:
Great, it is very close to my idea I wanted to implement for https://github.com/rectorphp/rector/issues/5282 :)

I found one bug here:
--dry-run keeps only last change for each file so last change wins

I think it could be fixed with adding parameter $content to NonPhpFileProcessorInterface::process() method and some update of this variable in NonPhpFileProcessorService foreach.

@sabbelasichon @TomasVotruba What do you think?

@TomasVotruba how it is handled in php files?

PR #6069, opened 2021-04-09
[NodeTypeResolver] Fixes #6065 Use existing NodeTypeResolver object on Auto Imports short name

### LINKED ISSUE #6065: @param/@throws space after type should be ignored by rector
# Bug Report

| Rector version | e.g. dev-main             |
| Installed as   | composer dependency                          |

Hi @TomasVotruba I tried rector `dev-main` in CodeIgniter4 project, as current `1.10.3` is hang. The `dev-main` is working ok, but it cause removal of spaces, like the following:

```diff
-	 * @throws             Exception
+	 * @throws Exception
```

ref https://github.com/codeigniter4/CodeIgniter4/pull/4538#issuecomment-816759159

which the multi space should be ignored. On debugging, I found that the auto import option:

```php
```

cause it. I tried to create failing fixture for it, but it showing ok:

Screen Shot 2021-04-09 at 22 40 16

Screen Shot 2021-04-09 at 22 39 07

It possibly due to post rector ?

samsonasik:
Use existing NodeTypeResolver object and change public name property instead of re-create new object when name is equal. Fixes #6065


samsonasik:
All checks have passed 🎉. I can't reproduce in fixture, but I tried `CodeIgniter4` project and it seems works.

@TomasVotruba it is ready for review

TomasVotruba:
Thanks :+1:

PR #6070, opened 2021-04-09
Failing test: Node type is not updated after executing rector

lulco:


TomasVotruba:
Solution would be similar to parent type on new nodes: https://github.com/rectorphp/rector/pull/6057/files#diff-1926ef5eec07704744c3f56345a1755c5863d2672344add93d58f81c21c2b24fR310-R322

lulco:
I tried to do some steps, but no success. This part of rector is magic for me, so I don't feel to be the one who will fix it :(

Issue #6072, opened 2021-04-09
Output changelog url in OutputFormatterInterface


sabbelasichon:
# Feature Request

In TYPO3-Rector we have an option to output the url to the changelog https://github.com/sabbelasichon/typo3-rector/blob/master/src/Configuration/Typo3Option.php#L12

The url is defined in the Rector via @see docblock. Should be better an @link docblock.

In order to achieve this i have created a rather nasty Decorator for the ConsoleOutputFormatter.php https://github.com/sabbelasichon/typo3-rector/blob/master/src/Console/Output/DecoratedConsoleOutputFormatter.php

I would like to have this in rector itself. I know it is only my itch, but maybe it is useful for others. 

I am gonna create a PR for that even if it gets rejected.

lulco:
Can you provide some screenshot how it would look like?

sabbelasichon:
@lulco Uploaded an image here: https://github.com/rectorphp/rector/pull/6073

PR #6073, opened 2021-04-09
[FEATURE] Output changelog url in OutputFormatterInterface

sabbelasichon:
Resolves: #6072

### DIFF
--- a/packages/ChangesReporting/ValueObject/RectorWithFileAndLineChange.php
+++ b/packages/ChangesReporting/ValueObject/RectorWithFileAndLineChange.php
@@ -4,8 +4,8 @@ declare(strict_types=1);
 
 namespace Rector\ChangesReporting\ValueObject;
 
+use Rector\ChangesReporting\Annotation\AnnotationExtractor;
 use Rector\Core\Contract\Rector\RectorInterface;
-use ReflectionClass;
 
 final class RectorWithFileAndLineChange
 {
@@ -42,10 +42,10 @@ final class RectorWithFileAndLineChange
         return get_class($this->rector);
     }
 
-    public function getRectorClassWithChangelogUrl(): string
+    public function getRectorClassWithChangelogUrl(AnnotationExtractor $annotationExtractor): string
     {
         $rectorClass = get_class($this->rector);
-        $changeLogUrl = $this->getChangelogUrl();
+        $changeLogUrl = $this->getChangelogUrl($annotationExtractor);
 
         if ($changeLogUrl === null) {
             return $rectorClass;
@@ -54,30 +54,11 @@ final class RectorWithFileAndLineChange
         return sprintf('%s (%s)', $rectorClass, $changeLogUrl);
     }
 
-    public function getChangelogUrl(): ?string
+    public function getChangelogUrl(AnnotationExtractor $annotationExtractor): ?string
     {
         $rectorClass = get_class($this->rector);
 
-        $rectorReflection = new ReflectionClass($rectorClass);
-
-        $docComment = $rectorReflection->getDocComment();
-
-        if (! is_string($docComment)) {
-            return null;
-        }
-
-        $pattern = "#@link\s*(?<url>[a-zA-Z0-9, ()_].*)#";
-        preg_match($pattern, $docComment, $matches);
-
-        if (! array_key_exists('url', $matches)) {
-            return null;
-        }
-
-        if (! filter_var($matches['url'], FILTER_VALIDATE_URL)) {
-            return null;
-        }
-
-        return trim((string) $matches['url']);
+        return $annotationExtractor->extractAnnotationFromClass($rectorClass, '@changelog');
     }
 
     public function getLine(): int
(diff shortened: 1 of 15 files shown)


sabbelasichon:
I am cheating here. I don´t know why or how i should use the phpstan ClassReflection here. Any ideas?

packages/ChangesReporting/ValueObject/RectorWithFileAndLineChange.php
```diff
@@ -41,6 +42,33 @@ public function getRectorClass(): string
         return get_class($this->rector);
     }
 
+    public function getRectorClassWithChangelogUrl(): string
+    {
+        $rectorClass = get_class($this->rector);
+
+        /* @phpstan-ignore-next-line */
```
TomasVotruba, line 49: We put ignores to `phsptan.neon`, to keep track on them and avoid code polution with tool ignores. Could you move it there?

TomasVotruba:
Thanks for the PR! 

I have troubles understanding, because "changelog" reminds me `CHANGELOG.md`, but I guess this is probably just showing `@link` in the output after list of affeted rules?

<br>

Could you provide some real life example how you use it and what happens? Gif is worth thousand words :)


```bash
vendor/bin/rector ...
↓
output (screenshot)
```

sabbelasichon:
Screenshot 2021-04-10 at 12 19 29

sabbelasichon:
Changelog has nothing to do in our case with the Changelog.md file. We are creating for every change in TYPO3 a really nice changelog file in the rst format transformed to html. Example: https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.4/Deprecation-85793-SeveralConstantsFromSystemEnvironmentBuilder.html. In TYPO3 Rector i add for every rector a nice @link block to reference to this url. This is really nice for the consumer to see why a certain rules is applied.

sabbelasichon:
Screenshot 2021-04-10 at 12 30 32

That´s the json output

TomasVotruba:
I see. So this way we can learn more detailed desription of the change in particular framework?

We actually do something like this in implicit way:

https://github.com/rectorphp/rector/blob/66a39a1ac903e4e090b7cd55914437669df3a192/rules/Php55/Rector/Class_/ClassConstantToSelfClassRector.php#L15

<br>

We could enable this by default, as there is no reason to hide this information.

<br>

One more thing. Currently the JSON output is working with classes only in "applied_rectors". That way it can be used in [Rector demo page](https://getrector.org/demo) to link exact rule: 

image

To keep the original value and add a new one, there should be a new value:

```json
{
    "applied_rules": ["SomeRule"],
    "applied_rules_with_changelog": [
        "SomeRule": "https://project/change/link.md"
    ]
}
```

sabbelasichon:
@TomasVotruba I am not 100% sure about the tests. I have introduced some tests to verify the behaviour. Am i right, that no tests exists before to ensure the desired behaviour? Do i miss something?

packages/ChangesReporting/ValueObject/RectorWithFileAndLineChange.php
```diff
@@ -41,6 +42,44 @@ public function getRectorClass(): string
...
+
+        if (! array_key_exists('url', $matches)) {
+            return null;
+        }
+
+        if (! filter_var($matches['url'], FILTER_VALIDATE_URL)) {
+            return null;
+        }
+
+        return trim((string) $matches['url']);
```
TomasVotruba, line 80: This seems quite a complext logic for simple value object.
Could you extract this to own service and cover it with simple test?

```php
$extractedValue = $this->extractAnnotationFromClass('SomeRector::class', 'link');
$this->assertSame('...', $extractedValue);
```

TomasVotruba:
These are the first test to covert this part, so I think it's ok :+1: 

It's getting into good shape :+1: 

I wonder about last think - the annotation name, @see, @link and @source seems a bit unclear. Now that you explained the purpose, maybe the name `@changelog` would be better name. In most of Symfony/Nette/Laravel/CakePHP rules it would actually link a headline in changelog where the change is described. What do you think?

This feature would be a big addition to understand what rules do and why :+1:

sabbelasichon:
Is @changelog an de-facto official docblock? But, maybe we should not care about it and invent our own.

packages/ChangesReporting/ValueObject/RectorWithFileAndLineChange.php
sabbelasichon: Where should this service be located? Is there not already something similar we could use?

packages/ChangesReporting/ValueObject/RectorWithFileAndLineChange.php
TomasVotruba: Fixed by: https://github.com/rectorphp/rector/pull/6078

TomasVotruba:
> But, maybe we should not care about it and invent our own.

:+1: I think it's good name for what is does

packages/ChangesReporting/ValueObject/RectorWithFileAndLineChange.php
TomasVotruba: I'll trust your choice :+1:

sabbelasichon:
If i run the phpstan analysis locally everything is fine. Weird.

TomasVotruba:
That's because I've just added this rule into `main` branch. Just ignore it in `phpstan.neon`, I'll handle it later.

TomasVotruba:
Thank you :+1: Let's :ship: it :)

PR #6075, opened 2021-04-10
[DowngradePhp72] Skip DowngradePregUnmatchedAsNullConstantRector on different flags in BitwiseOr

### LINKED ISSUE #6074: Incorrect behavior in DowngradePregUnmatchedAsNullConstantRector
# Bug Report

| Rector version | e.g. v0.10.4              |
| Installed as   | composer dependency                          |

Rule `DowngradePregUnmatchedAsNullConstantRector` is downgrading the code below, but it should not.

I tried to create the issue via getrector.org/demo, but I couldn't, see: https://github.com/rectorphp/getrector.org/issues/304

This code is being downgraded, but it should not, since the flag in `preg_match_all` is `PREG_OFFSET_CAPTURE | PREG_SET_ORDER`, and the rule should be applied for `PREG_UNMATCHED_AS_NULL` only:

```diff
/**
 * Parses a route string that does not contain optional segments.
 */
private function parsePlaceholders($route) {
    if (!preg_match_all(
        '~' . self::VARIABLE_REGEX . '~x', $route, $matches,
        PREG_OFFSET_CAPTURE | PREG_SET_ORDER
    )) {
        return [$route];
    }
+    array_walk_recursive($matches, function (&$value) {
+        if ($value === '') {
+            $value = null;
+        }
+    });
    $offset = 0;
    $routeData = [];
    foreach ($matches as $set) {
        if ($set[0][1] > $offset) {
            $routeData[] = substr($route, $offset, $set[0][1] - $offset);
        }
        $routeData[] = [
            $set[1][0],
            isset($set[2]) ? trim($set[2][0]) : self::DEFAULT_DISPATCH_REGEX
        ];
        $offset = $set[0][1] + strlen($set[0][0]);
    }

    if ($offset != strlen($route)) {
        $routeData[] = substr($route, $offset);
    }

    return $routeData;
}
```

## Expected Behaviour

The rule should be skipped.

samsonasik:
Fixes #6074

### DIFF
--- a/rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
+++ b/rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
@@ -65,6 +65,9 @@ CODE_SAMPLE
             ]);
     }
 
+    /**
+     * @return array<class-string<Expr>>
+     */
     public function getNodeTypes(): array
     {
         return [MethodCall::class];
@@ -94,12 +97,7 @@ CODE_SAMPLE
         ])) {
             return false;
         }
-
-        if (! $this->getStaticType($methodCall->args[0]->value) instanceof StringType) {
-            return true;
-        }
-
-        return true;
+        return $this->getStaticType($methodCall->args[0]->value) instanceof StringType;
     }
 
     private function updateNode(MethodCall $methodCall): MethodCall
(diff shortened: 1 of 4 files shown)


samsonasik:

rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
```diff
@@ -94,12 +97,7 @@ private function shouldSkip(MethodCall $methodCall): bool
         ])) {
             return false;
         }
-
-        if (! $this->getStaticType($methodCall->args[0]->value) instanceof StringType) {
-            return true;
-        }
-
-        return true;
+        return $this->getStaticType($methodCall->args[0]->value) instanceof StringType;
```
samsonasik, line 100: @lulco this seems always return true, should just return `! $this->nodeTypeResolver->isObjectTypes()` above cond.

TomasVotruba:
Thank you :+1:

rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
samsonasik: ref the original code that always return true: https://github.com/rectorphp/rector/blob/4b71eaa53d16c9c8cf4a6a2eb1fe4bf6d8006c58/rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php#L96-L100

leoloso:
Thank you 👍

PR #6076, opened 2021-04-10
[LeagueEvent] Remove unnecessary check instanceof StringType on DispatchStringToObjectRector

samsonasik:
Ref https://github.com/rectorphp/rector/pull/6075#discussion_r611030217 /cc @lulco

### DIFF
--- a/rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
+++ b/rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
@@ -91,13 +91,10 @@ CODE_SAMPLE
             return true;
         }
 
-        if ($this->nodeTypeResolver->isObjectTypes($methodCall->var, [
+        return ! $this->nodeTypeResolver->isObjectTypes($methodCall->var, [
             new ObjectType('League\Event\EventDispatcher'),
             new ObjectType('League\Event\Emitter'),
-        ])) {
-            return false;
-        }
-        return $this->getStaticType($methodCall->args[0]->value) instanceof StringType;
+        ]);
     }
 
     private function updateNode(MethodCall $methodCall): MethodCall


rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php
```diff
@@ -91,13 +91,10 @@ private function shouldSkip(MethodCall $methodCall): bool
...
         }
 
-        if ($this->nodeTypeResolver->isObjectTypes($methodCall->var, [
+        return ! $this->nodeTypeResolver->isObjectTypes($methodCall->var, [
             new ObjectType('League\Event\EventDispatcher'),
             new ObjectType('League\Event\Emitter'),
-        ])) {
-            return false;
-        }
-        return $this->getStaticType($methodCall->args[0]->value) instanceof StringType;
```
lulco, line 100: I am not sure about this. If the value is variable, you have to pass it somehow to anonymous class or not?

samsonasik: it needs more fixture to proof that, it was always return true, which the check can be removed. see https://github.com/rectorphp/rector/blob/4b71eaa53d16c9c8cf4a6a2eb1fe4bf6d8006c58/rules/LeagueEvent/Rector/MethodCall/DispatchStringToObjectRector.php#L96-L100

lulco: Ok I will create some fixtures with different approaches

lulco: @samsonasik Failing test added https://github.com/rectorphp/rector/pull/6120
I'll try to add fix

PR #6085, opened 2021-04-10
Refactor file processors towards universal collector

TomasVotruba:
Follow up to https://github.com/rectorphp/rector/pull/6060/files


src/Application/ApplicationFileProcessor.php
```diff
@@ -0,0 +1,121 @@
...
+    public function run(SmartFileInfo $smartFileInfo): string
+    {
+        $newContent = null;
+
+        foreach ($this->fileProcessors as $fileProcessor) {
+            if (! $fileProcessor->supports($smartFileInfo)) {
+                continue;
+            }
+
+            $newContent = $fileProcessor->process($smartFileInfo);
```
lulco, line 70: Imho we need content here as a parameter because diff in dry run will not work when one file is changed by more processors - first change content but it is not saved, then the second read the content from unchanged file

TomasVotruba: This is WIP, not working :) Wait for review till CI is green.

lulco: Ok ok, I just wanted to inform you about the situation :)

TomasVotruba: Thanks! :) That's where I got stuck yesterday.

It will probably require value object holding both content and file info reference. Like PHP_CodeSniffer has.

TomasVotruba:
@lulco There is few more details, but the concept is ready to review :+1:

lulco:
> @lulco There is few more details to finish, but the concept is ready to review 👍 What do you think?

Too many changes :D I have to touch it more. I will try to merge it into https://github.com/rectorphp/rector/pull/6100 and check if everything works as expected

rules/Arguments/Rector/ClassMethod/ArgumentDefaultValueReplacerRector.php
```diff
@@ -145,7 +145,7 @@ private function processArgs(Expr $expr, ArgumentDefaultValueReplacer $argumentD
     /**
      * @param mixed $value
      */
-    private function normalizeValueToArgument($value): Arg
+    private function normalizeValueToArgument($value): \Arg
```
samsonasik, line 148: this add `\` prefix seems  incorrect.

TomasVotruba:
This should be called in the end of `process()` method:

```php
$this->fileDiffFileDecorator->decorate([$file]);
```

lulco:
In all processors? I didn’t see that in existing processors. I found it in ApplicationFileProcessor::run. It should be enough, right?

TomasVotruba:
Not sure. I'll try to debug...

lulco:
I can prepare some failing small application later tonight if you will need

TomasVotruba:
So the answer is "yes", because every processor has its own printer. The best inspiration would be in `ComposerFileProcessor`:

https://github.com/rectorphp/rector/blob/4d4fb3611ffb8e78e3f1aa9c18dd8b353b765b3a/rules/Composer/Application/FileProcessor/ComposerFileProcessor.php#L89-L90

TomasVotruba:
>  can prepare some failing small application later tonight if you will need

Failing CI is always the best :+1:

lulco:
> The best inspiration would be in the smaller one `ComposerFileProcessor`:

OK I don't get it :) printer just creates new content as a string, then method changeFileContent is called and that's all...

How is it possible that it works with composer, but not with NonPhpFileProcessor and NetteDINeonMethodCallRenamer too?

lulco:
> > can prepare some failing small application later tonight if you will need
> 
> Failing CI is always the best 👍

I'm sure it is :) but this is a test for whole application, because all parts are working standalone, but not together :)

PR #6100, opened 2021-04-11
Neon rename method

lulco:
Renaming method calls in Neon files. Similar classes could be created also for symfony yaml and other frameworks using non-php config files


src/NonPhpFile/NeonMethodCallRenamer.php
```diff
@@ -0,0 +1,52 @@
...
+declare(strict_types=1);
+
+namespace Rector\Core\NonPhpFile;
+
+use Rector\Core\Contract\Processor\FileProcessorInterface;
+use Rector\Core\ValueObject\NonPhpFile\NonPhpFileChange;
+use Rector\Renaming\Configuration\MethodCallRenameCollector;
+use Symplify\SmartFileSystem\SmartFileInfo;
+
+class NeonMethodCallRenamer implements FileProcessorInterface
```
lulco, line 11: maybe it should be placed in rectorphp/rector-nette package ?

src/NonPhpFile/NeonMethodCallRenamer.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public function process(SmartFileInfo $smartFileInfo): ?NonPhpFileChange
+    {
+        $oldContent = $newContent = $smartFileInfo->getContents();
+
+        foreach ($this->methodCallRenameCollector->getMethodCallRenames() as $methodCallRename) {
+            $oldObjectType = $methodCallRename->getOldObjectType();
+
+            $className = str_replace('\\', '\\\\', $oldObjectType->getClassName());
```
lulco, line 30: Is it possible to find all children of given class / interface? In rename method rector there could be some interface, but in config.neon we will have its implementation, so the following pattern find nothing

samsonasik: you may can use NodeRepository, see https://github.com/rectorphp/rector/blob/1e4a0745a140c59346f5a119f974a76bc083c58e/packages/NodeCollector/NodeCollector/NodeRepository.php#L365

src/NonPhpFile/NeonMethodCallRenamer.php
TomasVotruba: Makes sense :+1:

src/NonPhpFile/NeonMethodCallRenamer.php
lulco: Unfortunately it is not working (at least in tests), I also tried get_declared_classes() or ExistingClassesProvider but those load only some classes not those which are in Source directory of tests

samsonasik: Yes, it seems that can't collect class/interface in `Source` directory right now, you probably need to use in one fixture, eg:

```php
class AnotherClass
{}

class Fixture
{
   public function run()
  {
         new AnotherClass();
  }
}

```

lulco: that can be problem :) My fixture is neon file (config similar to yaml). I don't have php here

src/NonPhpFile/NeonMethodCallRenamer.php
lulco: And what about the name? Maybe it is not neon feature but rather NetteDI... -> NetteDINeonMethodCallRenamer? :)

src/NonPhpFile/NeonMethodCallRenamer.php
TomasVotruba: You'd have to add such feature then

lulco: this is not neon-related, I know class / interface name (as string) and want to get list of its children (including it self). I cannot believe you never need it :)

btw I tried NodeRepository::findClassesAndInterfacesByType(AbstractRector::class); and it found nothing too. So I don't know if I used it wrong or it didn't work at all

src/NonPhpFile/NeonMethodCallRenamer.php
TomasVotruba: What do you consider better?

src/NonPhpFile/NeonMethodCallRenamer.php
TomasVotruba: This requires to load classes with robot loader or using `composer dump-autoload -ao`. That's why we dropped removing unused private classes and similar :)

Personally I would skip it in MVP to make merge easier.

src/NonPhpFile/NeonMethodCallRenamer.php
lulco: Now I like NetteDINeonMethodCallRenamer

lulco: name changed

lulco:
MVP is ready to review

lulco:
@TomasVotruba @samsonasik any issues / comments here?

rules/Renaming/Configuration/MethodCallRenameCollector.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Rector\Renaming\Configuration;
+
+use Rector\Renaming\Contract\MethodCallRenameInterface;
+
+final class MethodCallRenameCollector
```
samsonasik, line 7: I think `*Collector` should be in a "Collector" namespace instead of `Configuration`

lulco: OK I agree, this was inspired by [RenamedClassesDataCollector](https://github.com/rectorphp/rector/blob/28e1e1f3f4b0796b34105d1e2e26a3d66e74682a/src/Configuration/RenamedClassesDataCollector.php)

lulco: done

tests/NonPhpFile/NetteDINeonMethodCallRenamer/NetteDINeonMethodCallRenamerTest.php
```diff
@@ -0,0 +1,148 @@
...
+        setup:
+            - addConfig('key1', 'value1')
+            - addConfig('key2', 'value2')
+
+    -
+        class: Rector\Core\Tests\NonPhpFile\NetteDINeonMethodCallRenamer\Source\SecondService('service-name')
+        setup:
+            - addConfig('first-key', 'first-value')
+            - addConfig('second-key', 'second-value')
+";
```
TomasVotruba, line 138: This should be decopuled in fixture-style as other tests.

```
a
-----
b
```

so its reasier to extend.

TomasVotruba:
Could you move this to Nette package? As NEON package belongs there :)
https://github.com/rectorphp/rector-nette

Thanks

lulco:
@TomasVotruba sure, here is first part we need in rector core https://github.com/rectorphp/rector/pull/6149 then I can create PR for rector-nette

PR #6101, opened 2021-04-11
Fix nullsafe transpile

withinboredom:
It looks like the current transpilation of the PHP 8 null safe operator results in multiple calls to the method which may be undesired and cause subtle bugs. This fixes that by storing the result of the null-check in a temporary variable and relying on PHP's order-of-operations to reuse the same variable name. `$_` may result in a collision with other code, so if there's reason to fear that, perhaps a hash of some kind could be used instead; I'm open to any suggestion.

### DIFF
--- a/rules/DowngradePhp80/Rector/NullsafeMethodCall/DowngradeNullsafeToTernaryOperatorRector.php
+++ b/rules/DowngradePhp80/Rector/NullsafeMethodCall/DowngradeNullsafeToTernaryOperatorRector.php
@@ -50,11 +50,11 @@ CODE_SAMPLE
      */
     public function refactor(Node $node): ?Node
     {
-        $temp_var = new Node\Expr\Variable('_');
+        $tempVar = new Node\Expr\Variable('_');
         $called = $node instanceof NullsafeMethodCall
-            ? new MethodCall($temp_var, $node->name, $node->args)
-            : new PropertyFetch($temp_var, $node->name);
+            ? new MethodCall($tempVar, $node->name, $node->args)
+            : new PropertyFetch($tempVar, $node->name);
 
-        return new Ternary(new Assign($temp_var, $node->var), $called, $this->nodeFactory->createNull());
+        return new Ternary(new Assign($tempVar, $node->var), $called, $this->nodeFactory->createNull());
     }
 }
--- a/rules/DowngradePhp80/Rector/NullsafeMethodCall/DowngradeNullsafeToTernaryOperatorRector.php
+++ b/rules/DowngradePhp80/Rector/NullsafeMethodCall/DowngradeNullsafeToTernaryOperatorRector.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Rector\DowngradePhp80\Rector\NullsafeMethodCall;
 
 use PhpParser\Node;
+use PhpParser\Node\Expr\Assign;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\NullsafeMethodCall;
 use PhpParser\Node\Expr\NullsafePropertyFetch;
@@ -29,8 +30,8 @@ $dateAsString = $booking->startDate?->dateTimeString;
 CODE_SAMPLE
 ,
                 <<<'CODE_SAMPLE'
-$dateAsString = $booking->getStartDate() ? $booking->getStartDate()->asDateTimeString() : null;
-$dateAsString = $booking->startDate ? $booking->startDate->dateTimeString : null;
+$dateAsString = ($_ = $booking->getStartDate()) ? $_->asDateTimeString() : null;
+$dateAsString = ($_ = $booking->startDate) ? $_->dateTimeString : null;
 CODE_SAMPLE
             ),
         ]);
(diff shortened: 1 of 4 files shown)


withinboredom:
The test failure doesn't look related to the actual changes.

samsonasik:
@withinboredom I restarted the github action on test failure.

rules/DowngradePhp80/Rector/NullsafeMethodCall/DowngradeNullsafeToTernaryOperatorRector.php
```diff
@@ -49,10 +50,11 @@ public function getNodeTypes(): array
      */
     public function refactor(Node $node): ?Node
     {
+        $temp_var = new Node\Expr\Variable('_');
```
samsonasik, line 53: please use camel case

withinboredom: Done!

leoloso:
Wonderful PR, thanks @withinboredom! I actually had this issue on my mind, as I wrote that blog post

TomasVotruba:
Hi, thanks for contribution.

What about already existing `$_` variables above?
The produced code is long chunk of human-unreadable code.

What could be changed to have better readability with same result?

leoloso:
@withinboredom to avoid the issue with `$_` already existing, you can use the same code from the `DowngradeStripTagsCallWithArrayRector` rule, [here](https://github.com/rectorphp/rector/blob/21db1e9e383de26cbbc9602a04ab77294da9224d/rules/DowngradePhp74/Rector/FuncCall/DowngradeStripTagsCallWithArrayRector.php#L119:L124):

```php
$variableName = $this->variableNaming->resolveFromFuncCallFirstArgumentWithSuffix(
    $node,
    'AllowableTags',
    'allowableTags',
    $node->getAttribute(AttributeKey::SCOPE)
);
```

(If I'm not wrong) This service, if the variable already exists, will attach a counter to it and iterate, until finding a variable name that is not used.

withinboredom:
> The produced code is long chunk of human-unreadable code.
> -- @TomasVotruba 

I find it much more readable than what was produced before this PR, as you only need to get a hang of the pattern to read it. Previously, you'd end up with some very long lines while this seems to produce shorter lines for complex statements. I'm not sure how to make it more readable than it already is, TBH, but I'm open to some suggestions!

> if the variable already exists, will attach a counter to it and iterate, until finding a variable name that is not used.

Oh perfect! I'll give that a try!

rules/DowngradePhp80/Rector/NullsafeMethodCall/DowngradeNullsafeToTernaryOperatorRector.php
```diff
@@ -29,8 +42,8 @@ public function getRuleDefinition(): RuleDefinition
 CODE_SAMPLE
 ,
                 <<<'CODE_SAMPLE'
-$dateAsString = $booking->getStartDate() ? $booking->getStartDate()->asDateTimeString() : null;
-$dateAsString = $booking->startDate ? $booking->startDate->dateTimeString : null;
+$dateAsString = ($_ = $booking->getStartDate()) ? $_->asDateTimeString() : null;
+$dateAsString = ($_ = $booking->startDate) ? $_->dateTimeString : null;
```
withinboredom, line 46: I'm partial to leaving this as-is just so people don't rely on variable names (for some reason 🤔) but happy to change it to reflect what it actually transforms it into. Thoughts?

withinboredom:
The renaming as suggested by @leoloso makes it much better. Thanks for the tip!

rules/DowngradePhp80/Rector/NullsafeMethodCall/DowngradeNullsafeToTernaryOperatorRector.php
samsonasik: the example need to have same behaviour with one of fixtures

rules/DowngradePhp80/Rector/NullsafeMethodCall/DowngradeNullsafeToTernaryOperatorRector.php
```diff
@@ -29,8 +42,8 @@ public function getRuleDefinition(): RuleDefinition
 CODE_SAMPLE
 ,
                 <<<'CODE_SAMPLE'
-$dateAsString = $booking->getStartDate() ? $booking->getStartDate()->asDateTimeString() : null;
-$dateAsString = $booking->startDate ? $booking->startDate->dateTimeString : null;
+$dateAsString = ($_ = $booking->getStartDate()) ? $_->asDateTimeString() : null;
```
TomasVotruba, line 45:
```suggestion
$dateAsString = ($bookingGetStartDate = $booking->getStartDate()) ? $_->asDateTimeString() : null;
```

rules/DowngradePhp80/Rector/NullsafeMethodCall/DowngradeNullsafeToTernaryOperatorRector.php
TomasVotruba, line 46:
```suggestion
$dateAsString = ($bookingGetStartDate = $booking->startDate) ? $_->dateTimeString : null;
```

TomasVotruba:
@withinboredom Thank you :+1:

Issue #1238, opened 2019-03-19
[BUG] False positive with RemoveUnusedParameterRector


ejunker:
It is not detecting the `compact()` where the `$group` variable is used

```
-    public function groupSponsorship(Group $group)
+    public function groupSponsorship()
     {
         return $this->render('layouts.partials.sidebar.groups._sponsorship', compact('group'));
     }
    ----------- end diff -----------

Applied rectors:

 * Rector\DeadCode\Rector\ClassMethod\RemoveOverriddenValuesRector
 * Rector\DeadCode\Rector\FunctionLike\RemoveCodeAfterReturnRector
 * Rector\DeadCode\Rector\ClassMethod\RemoveUnusedParameterRector
```

TomasVotruba:
Thanks for reporting :+1: 

Are there any other "stringy" functions that should be checked?

ejunker:
The only other thing I could think of would be `extract()` http://php.net/manual/en/function.extract.php

TomasVotruba:
Thanks for the input and research.
 I'll look on this today

repat:
@TomasVotruba  This doesn't seem fixed? I only used the `SetList::DEAD_CODE` list.

![image](https://user-images.githubusercontent.com/516807/114292730-b8e26d80-9ae4-11eb-8bb4-fc9ad007bc03.png)

Interestingly, it left the `compact('clients')` call but removed the object creation 🤔 

![image](https://user-images.githubusercontent.com/516807/114292723-a2d4ad00-9ae4-11eb-99cf-bdef2cda624b.png)

![image](https://user-images.githubusercontent.com/516807/114292727-ac5e1500-9ae4-11eb-8166-20832f63f17f.png)

This is afaik quite a common way to pass data to the views in Laravel, so it'd be grand if it could be fixed! :) 

Thank you

TomasVotruba:
Thank you for an example!

Could you send a failing test case in a pull-request, so we have it covered in Rector?
Here is step by step tutorial how to add it: https://github.com/rectorphp/rector/blob/master/docs/how_to_add_test_for_rector_rule.md

PR #6102, opened 2021-04-12
Failing test for removing "unused" code in compact

repat:
As requested in https://github.com/rectorphp/rector/issues/1238#issuecomment-817266195, a failing test case.

Rector removes a supposedly unused variable, but it's actually used in PHPs `compact()` call, e.g. returning values to a Laravel view like so: `return view('home', compact('value'))` or returning an array (JSON) `return compact('json')`


samsonasik:
@repat I cherry-picked your commit at PR #6103

PR #6111, opened 2021-04-12
Make ComposerFileProcessor run on non-root paths of composer.json

TomasVotruba:


rules/Composer/Application/FileProcessor/ComposerFileProcessor.php
```diff
@@ -57,7 +57,7 @@ public function supports(File $file): bool
             return true;
         }
 
-        return $fileInfo->getRealPath() === getcwd() . '/composer.json';
+        return $fileInfo->getFilename() === 'composer.json';
```
lulco, line 60: this is not working for me, because I usually (always) run rector with path "src" or "app" but composer is one level up, so it is not loaded for processing

TomasVotruba: What would work then?

lulco: 1) add composer.json file to path list hardcoded (as it was before): getcwd() . '/composer.json';
2) let it on user to run rector with two paths, in my case
bin/rector process /var/www/league-event-app/app/ /var/www/league-event-app/composer.json

lulco: 3. add option "exclude-paths" and let user to run rector on app root:
```
bin/rector process /var/www/league-event-app/ --exclude-paths="vendor"
```

TomasVotruba: Ah, I though it's a `getBasename()` :D that would do?

lulco: not at this place :)

composer.json will never go to this method if rector is executed on app / src dir

TomasVotruba: That's a different place :)

This method only checks passing value, not loading it.

lulco: that's what I am talking about :)

PR #6115, opened 2021-04-12
[CodeQuality] Avoid double negation on UnnecessaryTernaryExpressionRector

samsonasik:
This patch remove double negation for `BooleanNot` got flipped `false ? true`, eg:

```php
return !! $data
```

as refactor result for the following code:

```php
return ! $data ? false : true;
```

to be:

```php
return $data;
```

or 

```php
return (bool) $data;
```

based on the condition type.


samsonasik:

TomasVotruba:
Thank you :+1:

PR #6118, opened 2021-04-12
Allow execution of process command without registered Rectors, but with NonPhpFileProcessors

### LINKED ISSUE #6087: Running without a rule but a NonPhpFileProcessor is not possible
As the new baby is out i tried it and it actually works like a charm in our context.
But one thing popped up. If i try to run the process command without any set or rector but only with the FilesProcessor it failed due to a guard clause in rector. Maybe we could also take the FilesProcessor into account before stopping the process.

lulco:
Resolves #6087 

Warning is still there if no Rectors are registered, but NonPhpFileProcessors are executed. Only difference is that tha output of command is still 0 (null) and command prints message 
```
[OK] Rector is done!
```
or
```
[OK] X files has been changed by Rector
```

### DIFF
--- a/src/Reporting/MissingRectorRulesReporter.php
+++ b/src/Reporting/MissingRectorRulesReporter.php
@@ -36,13 +36,11 @@ final class MissingRectorRulesReporter
             return ! $rector instanceof PostRectorInterface;
         });
 
-        if ($activeRectors !== []) {
-            return null;
+        if ($activeRectors === []) {
+            $this->report();
         }
 
-        $this->report();
-
-        return ShellCode::ERROR;
+        return null;
     }
 
     public function report(): void


lulco:
cc @sabbelasichon @TomasVotruba

src/Reporting/MissingRectorRulesReporter.php
```diff
@@ -36,13 +36,11 @@ public function reportIfMissing(): ?int
...
-        if ($activeRectors !== []) {
-            return null;
+        if ($activeRectors === []) {
+            $this->report();
         }
 
-        $this->report();
-
-        return ShellCode::ERROR;
+        return null;
```
TomasVotruba, line 43: This now won't fail CI if no rule is registered. Not a way to go.

The `$activeRectors` should include non-PHP Rectors rules instead.

lulco: that's what I wrote in description - NonPhpFileProcessors are always registered, you can't check them if they are active because it is always true, so what you suggest?

sabbelasichon: You should also inject all FileProcessorInterface in the Reporter the same way as the RectorInterface. Renaming it to MissingRulesReporter and check if any Rector or any FileProcessor is registered.

sabbelasichon:
@lulco Thanks for taking care.

src/Reporting/MissingRectorRulesReporter.php
TomasVotruba: @lulco The processors are not relevant here, the `RectorInterface` should be checked.

lulco: I did it, but they (FileProcessorInterface) were always registered, how should I check if they are active?

lulco:
> @lulco Thanks for taking care.

no problem, I found this issue months ago, but there was only class rename in non php files processor :) now they can grow... as it is your baby, I would like to be a good uncle :D

src/Reporting/MissingRectorRulesReporter.php
lulco: Issue is about: I have no rectors registered, but I want to run rector because I want to change non-php files e.g. configs etc. It is not possible now, because of this check.

@sabbelasichon probably we have to register at least one random Rector in config :)

TomasVotruba: Yes. The FileProcessor should only invoke Rector rules, like EventDispatcher calls subscribers. If there are no subscribers, event dispatcher is dead.

Without these boundaries, there is no reason to have 2 different classes.

lulco: Hm, not sure how @sabbelasichon wants to use this, but in my case there are several "non php files changes" which are not invoked by Rector rules.

For example now I need to do some replaces in configurations out of rector scope and I don't like it. I would be happy if I can configure it in rector.php with some "SimpleReplace" service.

TomasVotruba:
Closing as this needs different implementation on `RectorInterface` level: https://github.com/rectorphp/rector/pull/6118/files#r612008845

src/Reporting/MissingRectorRulesReporter.php
sabbelasichon: @lulco I am probably using it the same as you. But as you already figured out the FileProcessorInterface is always registered. Maybe we could add an additional Marker Interface just for that purpose. Will think about it and try to suggest something.

sabbelasichon: Or maybe we should just use the RectorInterface as the Marker interface and implement the useful getRuleDefinition method anyways. See: https://github.com/sabbelasichon/typo3-rector/blob/master/src/Yaml/Form/Transformer/FormYamlTransformer.php

PR #6119, opened 2021-04-12
Make ComposerFileProcessor run on composer.json in root automatically

lulco:
In huge refactoring, it was somehow removed so I would like to add it back O:-)

### DIFF
--- a/src/FileSystem/FilesFinder.php
+++ b/src/FileSystem/FilesFinder.php
@@ -86,6 +86,11 @@ final class FilesFinder
             $smartFileInfos[] = new SmartFileInfo($file);
         }
 
+        $composerPath = getcwd() . '/composer.json';
+        if (file_exists($composerPath) && in_array('json', $suffixes, true)) {
+            $smartFileInfos[] = new SmartFileInfo($composerPath);
+        }
+
         $smartFileInfos = array_merge($smartFileInfos, $this->findInDirectories($directories, $suffixes));
         return $this->fileInfosBySourceAndSuffixes[$cacheKey] = $smartFileInfos;
     }
--- a/tests/Application/ApplicationFileProcessor/ApplicationFileProcessorTest.php
+++ b/tests/Application/ApplicationFileProcessor/ApplicationFileProcessorTest.php
@@ -44,7 +44,7 @@ final class ApplicationFileProcessorTest extends AbstractKernelTestCase
     public function test(): void
     {
         $files = $this->fileFactory->createFromPaths([__DIR__ . '/Fixture']);
-        $this->assertCount(2, $files);
+        $this->assertCount(3, $files);
 
         $this->applicationFileProcessor->run($files);
 


src/FileSystem/FilesFinder.php
```diff
@@ -86,6 +86,11 @@ public function findInDirectoriesAndFiles(array $source, array $suffixes): array
             $smartFileInfos[] = new SmartFileInfo($file);
         }
 
+        $composerPath = getcwd() . '/composer.json';
```
sabbelasichon, line 89: Would move this to the FileFactory https://github.com/rectorphp/rector/blob/main/src/ValueObjectFactory/Application/FileFactory.php#L39

TomasVotruba:
It was not removed, but delegated to finder :).

Current application finds `*.json` files thanks to https://github.com/rectorphp/rector/blob/b0dc20a4625d41953097029e9434bdce9e4eddfc/src/ValueObjectFactory/Application/FileFactory.php#L62

lulco:
> It was not removed, but delegated to finder :).
> 
> Current application finds `*.json` files thanks to
> 
> https://github.com/rectorphp/rector/blob/b0dc20a4625d41953097029e9434bdce9e4eddfc/src/ValueObjectFactory/Application/FileFactory.php#L62

If it would I am not creating this PR :)
Application finds *.json only in defined paths e.g. src / app or something user pass as paths to process command... composer.json is usually on the same level as these paths, not inside them

src/FileSystem/FilesFinder.php
lulco: Maybe, it is one level up. What would be the difference?

TomasVotruba:
That is expected. Any magic addition of files would be confusing.

lulco:
> That is expected. Any magic addition of files would be confusing.
> Imagine running phpstan on `/src` and getting resutls from `/config` too

Well composer is somehow special :) And it works before... so it is breaking change and not very user friendly imho.

So, now we have to define composer.json in paths (source argument) when running process command? Do you consider my suggestion with exclude paths https://github.com/rectorphp/rector/pull/6111#discussion_r611920105?

TomasVotruba:
> So, now we have to define composer.json in paths (source argument) when running process command?

Yes, just add it to `rector.php` to other paths.

> Do you consider my suggestion with exclude paths #6111 (comment)?

Running any tool on root directory usually leads to bugs. Here is a good opt-in standard in tools like PHPUnit, CodingStandards or PHPStan. It might not seem for us as tool contributors, but explicit paths work better for end-users, as they feel they're in control.

lulco:
> Running any tool on root directory usually leads to bugs. Here is a good opt-in standard in tools like PHPUnit, CodingStandards or PHPStan. It might not seem for us as tool contributors, but explicit paths work better for end-users, as they feel they're in control.

All mentioned tools have (or used to have) an option to exclude paths.
phpcs:
```
-- ignore [dir]
```
phpunit:
```
<exclude> in xml config
```
phpstan:
```
parameters:
	excludePaths:
```

But it is your library, your decision :)

sabbelasichon:
@lulco I agree with Tomas here. I was quite confused after running rector and suddenly the composer.json was updated accidentally and just removed the version key. But i agree with you it is a breaking behaviour. But the library is still in some kind of beta stadium it is okish? What do you think?

lulco:
@sabbelasichon I am reconciled to any decision of @TomasVotruba :) Adding composer.json to paths is ok for me because I know how it works. I just think it should be documented - composer update behavior and also changes in non php files. Imho they are great features and we are close to upgrade whole project only with rector.

And if Tomas don't want to implement exclude option, it is up to him.

TomasVotruba:
Exclude options is already implemented in `SKIP` parameter

sabbelasichon:
@lulco I think if we can join forces and the new release is out we should definitely add it to the documentation. I have the same problem, all the features are hidden gems. I am the only one who knows about it. Let´s change this.

lulco:
> Exclude option is already implemented in `SKIP` parameter

Oh great :) didn't know it can be used also for paths. Is it relative to rector.php file?

lulco:
> @lulco I think if we can join forces and the new release is out we should definitely add it to the documentation. I have the same problem, all the features are hidden gems. I am the only one who knows about it. Let´s change this.

Sure, I'm in!

TomasVotruba:
> Oh great :) didn't know it can be used also for paths. Is it relative to rector.php file?

Absolute, fnmatch mask and relative paths are supported :+1: 

This is the package: https://github.com/symplify/skipper#1-configure-with-optionskip-parameter

lulco:
sorry about next question but it is connected to composer again... after changing composer.json it is not running composer update as before... is it a bug or expected behavior? And if expected, then why?

sabbelasichon:
@lulco I hope we are not getting enemies, but i think this new behaviour is good. Didn´t like it, that through the Process a composer update was triggered.

TomasVotruba:
>  And if expected, then why?

This is intentional - the `composer update` is something devs run on purpose. I run Rector on one project and it updated the `composer.lock` and created lot of work I had to revert. Took a couple of commits till I noticed :face_with_head_bandage:

lulco:
> @lulco I hope we are not getting enemies, but i think this new behaviour is good. Didn´t like it, that through the Process a composer update was triggered.

Why enemies? :) just because we like / don't like different things? :)

It was triggered only if you use composer rectors for adding / changing composer packages - so you know about composer.json will be changed. It was a feature that should help you as a developer to upgrade project without any manual work...

lulco:
> > And if expected, then why?
> 
> This is intentional - the `composer update` is something devs run on purpose. I run Rector on one project and it updated the `composer.lock` and created lot of work I had to revert. Took a couple of commits till I noticed 🤕

Then maybe I was using rector wrong way :D

I run rector first time and if app not working, I revert all changes, run composer install (to rollback all composer changes), change rector's config and try it again while app is not working

sabbelasichon:
Of course not. Was just saying because i disagreed with you the second time today ;-)

lulco:
No problem :) I just thought @TomasVotruba likes it when he wrote a blog about it back in january :) https://getrector.org/blog/2021/01/25/introducing-composer-json-rector-rules

I was happy and I removed several lines from my upgrade script and start thinking about how to run rector on other non-php files to remove other lines from upgrade script. So now I am little bit disappointed that I have to put composer update back :)

PR #6120, opened 2021-04-12
[League event] Dispatch only strings to anonymous class

lulco:


lulco:
@samsonasik ready for review. I had to rewrite it a little bit because we can pass value with several approaches

TomasVotruba:
Looks good

Issue #6079, opened 2021-04-10
[Meta] [PHP 7.1 do 5.6] Plan to downgrade PHP


TomasVotruba:
After [Rector 0.10 with PHP 7.1 downgraded code](https://getrector.org/blog/2021/03/22/rector-010-released-with-php71-support), we want to go even deeper. Next stop is PHP 5.6, as that's where most legacy projects are stuck. Right before the huge wall of PHP 7.0 :)

This is meta issue to reference downgrade rules. To kick of with few.

# From PHP 7.1 to PHP 7.0 - 

RFC links: https://wiki.php.net/rfc#php_71

- [x] Downgrade Symmetric array destructuring to list() function ( https://github.com/rectorphp/rector/blob/main/rules/DowngradePhp71/Rector/Array_/SymmetricArrayDestructuringToListRector.php  )
- [x] Downgrade class constant visibility ( https://github.com/rectorphp/rector/blob/main/rules/DowngradePhp71/Rector/ClassConst/DowngradeClassConstantVisibilityRector.php ) 
- [x] Remove the iterable pseudo type params and returns, add @param and @return tags instead ( https://github.com/rectorphp/rector/blob/main/rules/DowngradePhp71/Rector/FunctionLike/DowngradeIterablePseudoTypeDeclarationRector.php )
- [x] Remove the nullable type params, add @param tags instead ( https://github.com/rectorphp/rector/blob/main/rules/DowngradePhp71/Rector/FunctionLike/DowngradeNullableTypeDeclarationRector.php )
- [x] Remove `void` type - https://wiki.php.net/rfc/void_return_type ( https://github.com/rectorphp/rector/blob/main/rules/DowngradePhp71/Rector/FunctionLike/DowngradeVoidTypeDeclarationRector.php )
- [x] Downgrade negative string offset to strlen ( https://github.com/rectorphp/rector/blob/main/rules/DowngradePhp71/Rector/String_/DowngradeNegativeStringOffsetToStrlenRector.php  ) 
- [x] Downgrade single one | separated to multi catch exception(https://github.com/rectorphp/rector/blob/main/rules/DowngradePhp71/Rector/TryCatch/DowngradePipeToMultiCatchExceptionRector.php)
- [x] Downgrade keys in list ( https://github.com/rectorphp/rector/pull/6170 )

# From PHP 7.0 to PHP 5.6

RFC link: https://wiki.php.net/rfc#php_70

[...]

PR #6122, opened 2021-04-13
[DowngradePhp70] Add DowngradeAnonymousClassRector

samsonasik:
part of https://github.com/rectorphp/rector/issues/6079

### DIFF
--- a/rules-tests/DowngradePhp70/Rector/Class_/DowngradeAnonymousClassRector/Fixture/with_args.php.inc
+++ b/rules-tests/DowngradePhp70/Rector/Class_/DowngradeAnonymousClassRector/Fixture/with_args.php.inc
@@ -1,0 +1,37 @@
+<?php
+
+namespace Rector\Tests\DowngradePhp70\Rector\Class_\DowngradeAnonymousClassRector\Fixture;
+
+class WithArgs
+{
+    public function run($arg1, $arg2)
+    {
+        return new class ($arg1, $arg2) {
+            public function execute()
+            {
+            }
+        };
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\Tests\DowngradePhp70\Rector\Class_\DowngradeAnonymousClassRector\Fixture;
+
+class WithArgs
+{
+    public function run($arg1, $arg2)
+    {
+        return new Anonymous($arg1, $arg2);
+    }
+}
+class Anonymous
+{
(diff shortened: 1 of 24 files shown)


rules/DowngradePhp70/Rector/Class_/DowngradeAnonymousClassRector.php
```diff
@@ -0,0 +1,70 @@
...
+            public function execute()
+            {
+            }
+        };
+    }
+}
+CODE_SAMPLE
+                    ,
+                    <<<'CODE_SAMPLE'
+class Anonymous
```
vladyslavstartsev, line 47: this one should be prefixed|postfixed with some hash like https://github.com/box-project/box does. Since there may be more that one anonymous class in this name space.
Also this is a good catch for test

samsonasik: I am planning to add class name with increasing number when found, eg Anonymous1, Anonymous2, etc

rules-tests/DowngradePhp70/Rector/Class_/DowngradeAnonymousClassRector/Fixture/with_args.php.inc
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Rector\Tests\DowngradePhp70\Rector\Class_\DowngradeAnonymousClassRector\Fixture;
+
+class WithArgs
+{
+    public function run($arg1, $arg2)
+    {
+        return new class ($arg1, $arg2) {
```
samsonasik, line 9: @TomasVotruba anonymous class with arguments seems not updateable with : 

```php
new Name(sprintf('Anonymous(%s)', $argsString))
```

Do you have idea how to replace parent node (always a `New_` instance) like the following:

```php
        $parent = $class->getAttribute(AttributeKey::PARENT_NODE);
        $parent = new New_(new Name('Anonymous'), $args);
```

samsonasik: @TomasVotruba finally works with `New_` for node type 🎉

samsonasik:

TomasVotruba (approved):

TomasVotruba:
Thank you

PR #6141, opened 2021-04-14
Newly created nodes have no scope

lulco:
When we create new node with rector, they don't have any scope which is very often used in other rectors this way:

```
/** @var Scope $scope */
$scope = $node->getAttribute(AttributeKey::SCOPE);
$classReflection = $scope->getClassReflection();
```
which causes error: `Error: Call to a member function getClassReflection() on null`

If we run rector second time (when nodes are already created), scope is present and no error is thrown


TomasVotruba:
This is based on PHPStan, where traits don't have scope, only classes. Trait can have different types in different classes.

TomasVotruba:
This is logical duplicate of https://github.com/rectorphp/rector/pull/6070

lulco:
The new node isn't trait but anonymous class. But I agree that probably it is duplicate of #6070 as creating new node is also running some rector rule

PR #6143, opened 2021-04-15
[Php73] Refactor JsonThrowOnErrorRector: change 2nd parameter of json_decode to null and skip on json_last_error, json_last_error_msg

samsonasik:
- change 2nd parameter of json_decode to null  as default value is null, ref https://www.php.net/manual/en/function.json-decode.php 

- skip when next func call is/are `json_last_error()`, `json_last_error_msg()` as may be used next ref https://3v4l.org/3R5fM


samsonasik:

samsonasik:
Also handling when next statement is `Return_`

samsonasik:

TomasVotruba:
Thank you :clap:

PR #6144, opened 2021-04-15
[Php73] Skip RegexDashEscapeRector on 3 backslash with next to be escaped

samsonasik:
It previously produce invalid regex ref https://regex101.com/r/QkRzuF/1 , also, if passed to `preg_match()`, it will result different data ref https://github.com/codeigniter4/CodeIgniter4/runs/2345676568#step:11:177

### DIFF
--- a/rules/Php73/Rector/FuncCall/RegexDashEscapeRector.php
+++ b/rules/Php73/Rector/FuncCall/RegexDashEscapeRector.php
@@ -21,6 +21,13 @@ use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
  */
 final class RegexDashEscapeRector extends AbstractRector
 {
+
+    /**
+     * @var string
+     * @see https://regex101.com/r/BjKZFg/1
+     */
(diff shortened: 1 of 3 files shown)


rules/Php73/Rector/FuncCall/RegexDashEscapeRector.php
```diff
@@ -21,6 +21,13 @@
  */
 final class RegexDashEscapeRector extends AbstractRector
 {
+
+    /**
+     * @var string
+     * @see https://regex101.com/r/1e3hoF/1
+     */
+    private const THREE_BACKSLASH_FOR_ESCAPE_NEXT_REGEX = '#(?<=[^\\\\])\\\\{3}(?=[^\\\\])#';
```
samsonasik, line 29: it seems working on regex101.com https://regex101.com/r/iQbGgZ/1/ but somehow not working on tests, possibly need escape multiple `\` as `\\\\`

samsonasik: fixed, updated to use `{2}` as detected only 2 after `$this->regexPatternArgumentManipulator->matchCallArgumentWithRegexPattern()` call

samsonasik:

PR #6145, opened 2021-04-15
Make use of inline parser over extreme node construction

TomasVotruba:


packages/NodeCollector/NodeCollector/ParsedPropertyFetchNodeCollector.php
```diff
@@ -55,7 +54,7 @@ public function collect(Node $node): void
         }
 
         // make sure name is valid
-        if (StaticNodeInstanceOf::isOneOf($node->name, [StaticCall::class, MethodCall::class])) {
+        if ($node->name instanceof StaticCall || $node->name instanceof MethodCall) {
             return;
         }
```
TomasVotruba, line 59: This revert will help PHPStan and IDE to recognize the types. It was not possible with the static call

PR #6148, opened 2021-04-15
[DeadCode] Remove RemoveDefaultArgumentValueRector

TomasVotruba:
This rule is kind of opinionated - it actually decreases readability, as default variables can become explicit. It also more clear what happens if they're included, e.g.:

```php
Strings::replace($content, '#find#');
```

vs

```php
Strings::replace($content, '#find#', '');
```

In first example - what is going to happen? 
In second example, we see the found pattern is replaced with empty string.


samsonasik:
I think we can unregister from the config set instead of remove the rule. By this, user that think it is usefull will still can use it. What do you think?

TomasVotruba:
Out in the wild most rules are used only as part of sets. Unless configurable and used on purpose.
We've had some feedback on dead-code being to strict, so I'm trying to clean it up a bit to be more available.

It can still be implemented & maintained by community though.

PR #6149, opened 2021-04-15
Collect method call renames for future usage

lulco:
Splitted PR to core and rector-nette

@TomasVotruba this part we need here in core


TomasVotruba:
I seem the target branch did not exist. It's not part of `main` after merge

TomasVotruba:
The `rectorphp:main` should be `main` instead

TomasVotruba:
My bad, it's there :) 
image


I probablly pulled too fast

Issue #6135, opened 2021-04-14
using static reflection from within a custom rector rule


clxmstaab:
# Question

I want to use phpstans static reflection from within rector.
therefore I am using the `ReflectionProvider`.

I try to use static reflection on a class which is known to rector because of the `Option::AUTOLOAD_PATHS` setting.
I have the feeling that the `ReflectionProvider` does not know about the extended autoloading path and therefore doesn't know about my class. when using regular composer autoloading the `ReflectionProvider` works like expected.

my custom rector rule wors like expected on the master branch, without the above changes from the pull request.

see https://github.com/staabm/rector-view-scope/pull/19 for a small repro

TomasVotruba:
Closing as dicussion continues in linked PR.

clxmstaab:
underlying problem was fixed in https://github.com/rectorphp/rector/pull/6239

PR #6158, opened 2021-04-16
show that classes doesn't get autoloaded

clxmstaab:
this PR tries to show, that ATM rector does not trigger custom autoloaders when the ReflectionProvider is used on unknown classes.

I was not able to reproduce this in a separate test and also did not know how to structure a possible Test to show the problem

running the modified test produces the following output:
```
$ vendor/bin/phpunit rules-tests/CodeQuality/Rector/Assign/CombinedAssignRector/CombinedAssignRectorTest.php
PHPUnit 9.5.4 by Sebastian Bergmann and contributors.

.                                                                   1 / 1 (100%)string(145) "Class ClassWhichNeedsShouldTriggerAutoloading was not found while trying to analyse it - discovering symbols is probably not configured properly."


Time: 00:00.826, Memory: 36.00 MB
```

actually rector should trigger the autoloading which defines the class which is missing here.

sorry for the confusing PR. I tried several ways to properly test the autoloading but was not successfull.

I hope this PR is usefull for someone more familiar with the code-base then I am.

refs https://github.com/rectorphp/rector/issues/6135


rules/CodeQuality/Rector/Assign/CombinedAssignRector.php
```diff
@@ -48,6 +55,14 @@ public function getNodeTypes(): array
      */
     public function refactor(Node $node): ?Node
     {
+        try {
+            $class = $this->reflectionProvider->getClass("ClassWhichNeedsShouldTriggerAutoloading");
```
clxmstaab, line 59: here we try to use the reflectionProvider on a class which does not exist.
my understand is that this should invoke the configured autoloader in https://github.com/rectorphp/rector/pull/6158/files#diff-ee5a9cb644ef1a427edfa91f0d8656e988995c358f98656f9d83e5aefccd8f6dR13

rules-tests/CodeQuality/Rector/Assign/CombinedAssignRector/Fixture/fixture.php.inc
```diff
@@ -15,25 +15,3 @@ function combinedAssignFunction()
     $baz = $baz ** 2;
     $quux = $quux | Something::FOO;
 }
-
```
clxmstaab, line 18: here I am just deleting files to cleanup the output when invoking the actual test.
this does not make sense of course .. its just for repro sake.

staabm:
I am still digging arround and cannot find the reason why the ReflectionProvider is not able to find my class, even if I have configured a autoloader within rector.

I am still running in `Class \AdmgrpController was not found while trying to analyse it - discovering symbols is probably not configured properly.` errors.

I somehow have the feeling that rector should register autoloaders defined with `autoload-file` or `autoload_paths` within the phpstan ReflectionProvider.

Can someone point me to the lines where rector configures phpstans' ReflectionProvider?

rules-tests/CodeQuality/Rector/Assign/CombinedAssignRector/config/configured_rule.php
```diff
@@ -3,10 +3,14 @@
...
 
 use Rector\CodeQuality\Rector\Assign\CombinedAssignRector;
+use Rector\Core\Configuration\Option;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 
 return static function (ContainerConfigurator $containerConfigurator): void {
     $services = $containerConfigurator->services();
+    $parameters = $containerConfigurator->parameters();
+
+    $parameters->set(Option::AUTOLOAD_PATHS, [__DIR__ . '/autoloader.php']);
```
staabm, line 13: I did more testing.. I guess the whole point of my issue is, that neither `Option::AUTOLOAD_PATHS` nor `Option::OPTION_AUTOLOAD_FILE` include the autoloader in a way that rector makes use of it.

it does not gets triggered when files cannot be found at runtime.

adding a manual `require_once __DIR__ . '/autoloader.php'';` inside my rector rule makes the autoloader work like expected

TomasVotruba: I see. The includes files were moved to `BOOTSTRAP_FILES`:

I've updated the release to include it: https://github.com/rectorphp/rector/releases/tag/0.10.0

> adding a manual require_once __DIR__ . '/autoloader.php''; inside my rector rule makes the autoloader work like expected. In your case:

```php
$parameters->set(Option::BOOTSTRAP_FILES, [
    __DIR__ . '/autoload.php',
]);
```


Thanks for feedback

TomasVotruba: Ref https://github.com/rectorphp/getrector.org/commit/03661372fbbb9e6711446fad1df424a0b49cc0b7

staabm: thx for the feedback. it doesn't even work with `Option::BOOTSTRAP_FILES`

TomasVotruba: That's weird :thinking: I'll try to run this PR locally

TomasVotruba: When I run tests, it's passing :+1: 

```bash
vendor/bin/phpunit rules-tests/CodeQuality/Rector/Assign/CombinedAssignRector/CombinedAssignRectorTest.php
```

image

staabm: you can see the actual problem in the class not found error, inside your screenshot

TomasVotruba: Foudn it... the tests were missing autoload/bootstrap feature

TomasVotruba: See https://github.com/rectorphp/rector/pull/6239/files

TomasVotruba:
Thank you for simple enough demo :+1:

Issue #6146, opened 2021-04-15
General pretty printing guidance


jdart:
# Question
I saw the part of the readme with a link to your ecs.php and tried it out, but it didn't fix most of the unwanted changes and added more. Do you know of a fixer that will remove leading slashes on global functions and things like parent, and true? Do you know of a fixer that will replace a fqn with a class name if it's imported? It would be really great to provide a minimal ecs config that fixes common things php parser changes, maybe that's impossible.

```
        \parent::setUp();
        \app();
        // not using imported class, instead using fqn
        \true

```

TomasVotruba:
Hi, thanks for reporting.

The `parent` and `true` should not be prefixed with `\\` by Rector.

Could you provide minimal failing code snippet that is causing it? https://getrector.org/demo

jdart:
I'm not able to reproduce there but I should mention I'm using the prefixed build (RectorPrefix20210414\).

jdart:
I tried reproducing with a fresh composer.json with the prefixed build and was again unable to reproduce so there must be something funny with my dependencies. Any ideas for troubleshooting?

TomasVotruba:
Hi, I would copy paste the latest patch version from your test repository to you main repository.

E.g. 

```diff
-rector/rector:^0.10
+rector/rector:^0.10.5
```

for all other packages too

jdart:
I believe I'm able to reproduce now...

composer.json
```
{
    "name": "jdart/rekt",
    "authors": [
    ],
    "require": {},
    "require-dev": {
        "rector/rector-prefixed": "^0.10.6"
    },
    "autoload": {
        "psr-4": {
          "App\\": "app/"
        }
    }
}
```

app/ThingTest.php
```
<?php

declare(strict_types=1);

class BaseC {
    public function setup(): void
    {
    }
}

class Extender extends BaseC
{
    public function setup(): void
    {
        parent::setup();
    }
}
```

rector.php
```
<?php

declare(strict_types=1);

return static function (\RectorPrefix20210414\Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator $containerConfigurator): void {
    $services = $containerConfigurator->services();
    $services->set(\Rector\PSR4\Rector\FileWithoutNamespace\NormalizeNamespaceByPSR4ComposerAutoloadRector::class);
};
```

Output:
```
❯ ./vendor/bin/rector process --debug --clear-cache app/ThingTest.php
[parsing] app/ThingTest.php
[...]

PR #6159, opened 2021-04-17
Do not prefix "parent", "self", "static" and native constants

TomasVotruba:
Closes https://github.com/rectorphp/rector/issues/6146

### DIFF
--- a/rules/PSR4/NodeManipulator/FullyQualifyStmtsAnalyzer.php
+++ b/rules/PSR4/NodeManipulator/FullyQualifyStmtsAnalyzer.php
@@ -72,18 +72,27 @@ final class FullyQualifyStmtsAnalyzer
                 return null;
             }
 
-            $parent = $node->getAttribute(AttributeKey::PARENT_NODE);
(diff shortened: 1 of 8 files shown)


rules/PSR4/NodeManipulator/FullyQualifyStmtsAnalyzer.php
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        // FQNize all class names
+        $this->simpleCallableNodeTraverser->traverseNodesWithCallable($nodes, function (Node $node): ?FullyQualified {
+            if (! $node instanceof Name) {
+                return null;
+            }
+
+            $fullyQualifiedName = $this->nodeNameResolver->getName($node);
+            if (in_array($fullyQualifiedName, ['self', 'parent'], true)) {
```
samsonasik, line 71: is `static` need to be included as well?

TomasVotruba: Should be :+1: thanks

PR #6160, opened 2021-04-17
[NodeAnalyzer] Remove ClassAnalyzer

samsonasik:
Remove ClassAnalyzer for detect anonymous class as there is `isAnonymous()` for `Class_` node.


samsonasik:
It seems not works on anonymous class with arg. I think we need to make consistent to use `ClassAnalyzer-> isAnonymousClass()` instead as there is both `node->isAnonymous()` and `classAnalyzer->isAnonymousClass()` usages.

Issue #6129, opened 2021-04-13
[DX] Add convention-based config loading to allow services from extension to be loaded


TomasVotruba:
**Problem:** If you create and Rector extension, to run Rector with it the `rector.php` must include it explicitly. Without it, the extension does not load its services. 

The workaround is to add config manually e.g. in `bin/rector.php`: https://github.com/sabbelasichon/typo3-rector/blob/b7a6b8a67adf0c2e84974eb1be6f8469de1b920e/bin/rector.php#L48

<br>

Solution would be to include config in `composer.json` configuration, like PHPStan composer pluging does: https://github.com/phpstan/extension-installer

The generated class is used during DI creation: https://github.com/phpstan/phpstan-src/blob/c471c7b050e0929daf432288770de673b394a983/src/Command/CommandHelper.php#L193-L217

<br>

It would also help with magical loading we currently have:

https://github.com/rectorphp/rector/blob/021ffc7ada2bd57772481c067aeeca158f6bdc68/config/config.php#L15-L29

<br>

Based on call with cc @sabbelasichon

sabbelasichon:
Thanks for taking care. I would like to create something here. Maybe not so fast as you do, but it is nothing with time pressure, right?

TomasVotruba:
You can go for it 👍

sabbelasichon:
Basically i am done. Would you create a dedicated rector/rector-installer package or is it just part of rector itself?

TomasVotruba:
> is it just part of rector itself

:+1: I would do it as MVP and see how it works.

sabbelasichon:
Sorry, but i think it should be a standalone package because it must be of type composer-plugin to register it correctly. I don´t like to give the whole rector package the type of composer-plugin.

sabbelasichon:
If you like you could create a rector/rector-installer repository and i am gonna fork it and provide my solution.

TomasVotruba:
Yes, plugin is standalone and imlementation in Rector is part of core.

There you go: https://github.com/rectorphp/rector-installer, you should have write access so no fork is needed :+1:

sabbelasichon:
The installer package is ready, https://github.com/rectorphp/rector-installer/
We have to register the package at packagist.org still.

sabbelasichon:
If you like as a follow up i could also add the configurations to rector/rector-*. Just tell me. I would be happy to help.

TomasVotruba:
[...]

PR #6174, opened 2021-04-20
[TASK] Use rector-installer to load config of extensions

sabbelasichon:
- Relates: #6129

I am using dev-main for the packages here. Maybe we should create new releases here.


TomasVotruba:
> Maybe we should create new releases here.

:+1:

sabbelasichon:
@TomasVotruba I have moved the loading of the configs into the container configuration itself. Because the bootstrapping happens only by invoking the bin/rector.php

config/config.php
```diff
@@ -12,32 +12,10 @@
...
-            __DIR__ . '/../vendor/rector/rector-generator/config/config.php',
-            null,
-            'not_found'
-        );
-        // 3rd party dependency
-        $containerConfigurator->import(__DIR__ . '/../../rector-generator/config/config.php', null, 'not_found');
+    $extensionConfigResolver = new ExtensionConfigResolver();
+    $configFiles = $extensionConfigResolver->provide();
+    foreach ($configFiles as $configFile) {
+        $containerConfigurator->import($configFile->getRealPath(), null, 'not_found');
```
sabbelasichon, line 18: Ignoring errors should not be necessary, but it is a safety net, if someone / something deletes moves the config file accidentally and not redefining it in the composer.json extra section.

TomasVotruba: This might hide unwanted bugs. How would you report missing config to package owners?

src/Bootstrap/ExtensionConfigResolver.php
```diff
@@ -11,11 +11,12 @@
 final class ExtensionConfigResolver
 {
     /**
-     * @param SmartFileInfo[] $configFileInfos
      * @return SmartFileInfo[]
      */
-    public function appendExtensionsConfig(array $configFileInfos): array
+    public function provide(): array
```
TomasVotruba, line 16: :+1:

src/Bootstrap/ExtensionConfigResolver.php
```diff
@@ -11,11 +11,12 @@
 final class ExtensionConfigResolver
 {
     /**
-     * @param SmartFileInfo[] $configFileInfos
      * @return SmartFileInfo[]
```
TomasVotruba, line 14: Btw, the `SmartFileInfo` can return only existing files. If the file is missing, the exception is thrown. This is to avoid false positive on incorrect file location.

That means the `not_found` you've commened on is already checked here :+1:

sabbelasichon: I just wanted to say the same thing right now.

composer.json
```diff
@@ -4,6 +4,8 @@
     "bin": [
         "bin/rector"
     ],
+    "minimum-stability": "dev",
+    "prefer-stable": true,
     "license": "MIT",
```
TomasVotruba, line 9: Could you move these 2 in the end of the file? Just personal prefference I have in all `composer.json` not to clutter main composer json info :)

config/config.php
```diff
@@ -12,32 +12,10 @@
...
-            null,
-            'not_found'
-        );
-        // 3rd party dependency
-        $containerConfigurator->import(__DIR__ . '/../../rector-generator/config/config.php', null, 'not_found');
+    $extensionConfigResolver = new ExtensionConfigResolver();
+    $configFiles = $extensionConfigResolver->provide();
+    foreach ($configFiles as $configFile) {
+        $containerConfigurator->import($configFile->getRealPath());
     }
```
TomasVotruba, line 20:
```suggestion
    $extensionConfigFiles = $extensionConfigResolver->provide();
    foreach ($extensionConfigFiles as $extensionConfigFile) {
        $containerConfigurator->import($xtensionConfigFile->getRealPath());
    }

```

This naming would make more clear its only extension configs here, nothing else

composer.json
```diff
@@ -37,15 +37,16 @@
...
-        "rector/rector-cakephp": "^0.10.3",
-        "rector/rector-phpunit": "^0.10.6",
-        "rector/rector-doctrine": "^0.10.2",
+        "rector/rector-cakephp": "dev-main",
+        "rector/rector-doctrine": "dev-main",
+        "rector/rector-installer": "dev-main",
+        "rector/rector-laravel": "dev-main",
+        "rector/rector-nette": "dev-main",
+        "rector/rector-phpunit": "dev-main",
+        "rector/rector-symfony": "dev-main",
```
TomasVotruba, line 49: I've just released ned version for these packages. Could you use them instead?

composer.json
```diff
@@ -73,7 +74,7 @@
         "nette/forms": "^3.0",
         "phpstan/phpstan-nette": "^0.12.16",
         "phpunit/phpunit": "^9.5",
-        "rector/rector-generator": "^0.1.5",
+        "rector/rector-generator": "dev-main",
```
TomasVotruba, line 77: Same for this package

composer.json
```diff
@@ -37,15 +37,16 @@
...
+        "phpstan/phpstan": "^0.12.83",
         "phpstan/phpstan-phpunit": "^0.12.18",
-        "rector/rector-symfony": "^0.10.3",
-        "rector/rector-nette": "^0.10.5",
-        "rector/rector-laravel": "^0.10.1",
-        "rector/rector-cakephp": "^0.10.3",
-        "rector/rector-phpunit": "^0.10.6",
+        "rector/rector-cakephp": "^0.10.4",
         "rector/rector-doctrine": "^0.10.2",
+        "rector/rector-installer": "dev-main",
```
TomasVotruba, line 45: Could you tag this one too, so we can remove min-stability dev?

sabbelasichon: I have also bumped it up to PHP Version 7.3

TomasVotruba:
Thank you :clap:

TomasVotruba:
I've just learned the:

```bash
composer update
```

does not invoke plugin installation, while 

```bash
composer install
```

does :)

Issue #6180, opened 2021-04-20
Running process with `--debug` gives: `Uncaught Error: Call to undefined method PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocNode::setAttribute`


ruudk:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.10.6             |
| Installed as   | prefixed                           |

```
vendor/bin/rector process --debug src/Core/EventBundle/Event/Event/LocationRemovedEvent.php
[parsing] src/Core/EventBundle/Event/Event/LocationRemovedEvent.php
[refactoring] src/Core/EventBundle/Event/Event/LocationRemovedEvent.php
PHP Fatal error:  Uncaught Error: Call to undefined method PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocNode::setAttribute() in /Volumes/CS/www/website/vendor/rector/rector-prefixed/packages/BetterPhpDocParser/PhpDocInfo/PhpDocInfoFactory.php:129
Stack trace:
#0 /Volumes/CS/www/website/vendor/rector/rector-prefixed/packages/BetterPhpDocParser/PhpDocInfo/PhpDocInfoFactory.php(101): Rector\BetterPhpDocParser\PhpDocInfo\PhpDocInfoFactory->setPositionOfLastToken(Object(PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocNode))
#1 /Volumes/CS/www/website/vendor/rector/rector-prefixed/packages/BetterPhpDocParser/PhpDocInfo/PhpDocInfoFactory.php(73): Rector\BetterPhpDocParser\PhpDocInfo\PhpDocInfoFactory->createFromNode(Object(PhpParser\Node\Stmt\Property))
#2 /Volumes/CS/www/website/vendor/rector/rector-prefixed/src/Exclusion/ExclusionManager.php(50): Rector\BetterPhpDocParser\PhpDocInfo\PhpDocInfoFactory->createFromNodeOrEmpty(Object(PhpParser\Node\Stmt\Property))
#3 /Volumes/CS/www/website/vendor/rector/rector-prefixed/src/Exclusion/ExclusionManager.p in /Volumes/CS/www/website/vendor/rector/rector-prefixed/packages/BetterPhpDocParser/PhpDocInfo/PhpDocInfoFactory.php on line 129
[...]

PR #6181, opened 2021-04-20
[prefixed] require compatible PHPStan with attribute phpdoc-parser

TomasVotruba:
Closes https://github.com/rectorphp/rector/issues/6180

### DIFF
--- a/build/target-repository/composer.json
+++ b/build/target-repository/composer.json
@@ -7,11 +7,15 @@
     ],
     "require": {
         "php": "^7.0|^8.0",
-        "phpstan/phpstan": "^0.9.3 || ^0.12.82"
+        "phpstan/phpstan": "^0.12.83"
     },
     "extra": {
         "branch-alias": {
             "dev-main": "0.10-dev"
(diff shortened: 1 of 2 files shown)


build/target-repository/composer.json
```diff
@@ -7,11 +7,15 @@
     ],
     "require": {
         "php": "^7.0|^8.0",
-        "phpstan/phpstan": "^0.9.3 || ^0.12.82"
+        "phpstan/phpstan": "^0.12.83"
```
samsonasik, line 10: @TomasVotruba it seems will make not installable by require prefixed from php 7

samsonasik: after installed on php 7.3, the phpstan should be removed from composer.json to make it installable from php 7

TomasVotruba: Yes, that's correct.

The phpstan for PHP 7.0 must be used from here https://github.com/phpstan/phpstan-src and downgraded to PHP 7.0

PR #6190, opened 2021-04-21
[PHP 7.2] Add covariant param support for trait methods

TomasVotruba:
Covers trait types https://3v4l.org/EgHFm

Fixes

image


samsonasik:
it seems not applied on rector-prefixed yet

PR #6194, opened 2021-04-22
[UnionTypesRector] Do not refactor to `DateTime|object` property type

ruudk:
# Failing Test for UnionTypesRector

Based on https://getrector.org/demo/ade5aba8-d7bb-4f15-9afc-f83fd9a1b3f9

The change done by Rector gives a fatal error:
```
Fatal error:  Type DateTime|object contains both object and a class type, which is redundant
```


ruudk:
Maybe it should just leave it as it was.

samsonasik:
@ruudk I cherry-picked your commit at PR https://github.com/rectorphp/rector/pull/6237

PR #6195, opened 2021-04-22
[TypedPropertyRector] Handle generic types

ruudk:

Based on https://getrector.org/demo/675190c7-7853-4f31-812b-0e8249b6d594

It produces `Fatal error:  Duplicate type object is redundant`

### DIFF
--- a/rules-tests/Php74/Rector/Property/TypedPropertyRector/Fixture/generic_object_type.php.inc
+++ b/rules-tests/Php74/Rector/Property/TypedPropertyRector/Fixture/generic_object_type.php.inc
@@ -1,0 +1,47 @@
+<?php
+
+/**
+ * @template T of object
+ */
+namespace Rector\Php74\Tests\Rector\Property\TypedPropertyRector\Fixture;
+
+final class GenericObjectType
+{
+    /**
+     * @var T
+     */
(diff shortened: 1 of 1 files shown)


rules-tests/Php74/Rector/Property/TypedPropertyRector/Fixture/generic_object_type.php.inc
```diff
@@ -0,0 +1,47 @@
+<?php
+
+/**
```
staabm, line 3: this doc should be before the class definition, jot the namespace

ruudk: You're right, getrector website did this wrong, will fix

TomasVotruba: The fixture code is generated here https://github.com/rectorphp/getrector.org/tree/main/packages/github-magic-link/src/BodyFactory

samsonasik:
@ruudk I cherry-picked your commit at PR https://github.com/rectorphp/rector/pull/6378

PR #6196, opened 2021-04-22
[TypedPropertyRector] Do not refactor to `iterable` property type

ruudk:

Based on https://getrector.org/demo/3a8fc4f5-898d-4fb4-8ed3-99dc8312df77

```
Fatal error: Cannot use iterable as iterable because 'iterable' is a special class name
```


ruudk:
Hmm, why does the test not fail?

samsonasik:
The current getrector.org seems not using latest dev-main

TomasVotruba:
Same as previous

PR #6197, opened 2021-04-22
[TypedPropertyRector] Do not refactor to `callable` property type

ruudk:

Based on https://getrector.org/demo/e8724b53-a00c-4f0c-b10f-e4e54be0a664

```
PHP Fatal error:  Property SkipCallablePropertyType::$startDate cannot have type DateTimeImmutable|callable
```


ruudk:
Hmm, why does the test not fail?

samsonasik:
The current getrector.org seems not using latest dev-main

ruudk:
@samsonasik Damnz. I think that should be more visible.

TomasVotruba:
Merging so we have it covered.

> The current getrector.org seems not using latest dev-main

Daily CRON builds would solve it.

samsonasik:
@TomasVotruba I created PR https://github.com/rectorphp/getrector.org/pull/305 for it

PR #6202, opened 2021-04-22
[DowngradePhp71] Add DowngradeIsIterableRector

samsonasik:
To handle prefixed undefined function is_iterable notice:

Screen Shot 2021-04-22 at 19 18 13


rules/DowngradePhp71/Rector/FuncCall/DowngradeIsIterableRector.php
```diff
@@ -0,0 +1,71 @@
...
+     * @param FuncCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        /** @var mixed $arg */
+        $arg = $node->args[0]->value;
+        $funcCall = $this->nodeFactory->createFuncCall('is_array', [$arg]);
+        $instanceOf = new Instanceof_($arg, new FullyQualified('Traversable'));
+
+        return new BooleanOr($funcCall, $instanceOf);
```
samsonasik, line 69: @TomasVotruba direct return `BooleanOr` seems cause hang in tests, do you have idea how to check it? it working ok on `print_node` check.

samsonasik: fixed with `traverseNodesWithCallable()` 🎉

samsonasik: I just re-check that it need `isName()` check first to skip other function call

TomasVotruba:
:+1:

PR #6204, opened 2021-04-22
[DeadCode] Skip expr check from statement on RemoveUnusedVariableAssignRector for Nop stmt

samsonasik:
Nop stmt doesn't has expr property on next check.


samsonasik:
Added Expression check like in If stmt check.

PR #6209, opened 2021-04-23
[DeadCode] [WIP] Refactor RemoveUnusedVariableAssignRector: use NextVariableUsageNodeFinder service

samsonasik:


samsonasik:
it needs re-check agains re-use assignment.

Issue #6211, opened 2021-04-23
Skip FileProcessors


sabbelasichon:
Non php FileProcessors are not skippable for now

sabbelasichon:
Gonna provide a PR for that.

sabbelasichon:
I am going to close this in favour of https://github.com/rectorphp/rector/pull/6251

PR #6212, opened 2021-04-23
[TASK] Make FileProcessors skippable

sabbelasichon:
Resolves: #6211

### DIFF
--- a/rules/Composer/Application/FileProcessor/ComposerFileProcessor.php
+++ b/rules/Composer/Application/FileProcessor/ComposerFileProcessor.php
@@ -10,6 +10,7 @@ use Rector\Core\ValueObject\Application\File;
 use Rector\Testing\PHPUnit\StaticPHPUnitEnvironment;
 use Symplify\ComposerJsonManipulator\ComposerJsonFactory;
 use Symplify\ComposerJsonManipulator\Printer\ComposerJsonPrinter;
+use Symplify\Skipper\Skipper\Skipper;
 use Symplify\SmartFileSystem\SmartFileInfo;
 
 final class ComposerFileProcessor implements FileProcessorInterface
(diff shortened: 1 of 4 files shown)


src/Application/ActiveFileProcessorsProvider.php
```diff
@@ -0,0 +1,35 @@
...
+     */
+    public function __construct(Skipper $skipper, array $fileProcessors)
+    {
+        $this->fileProcessors = array_filter($fileProcessors, function (FileProcessorInterface $fileProcessor) use (
+            $skipper
+        ) {
+            return ! $skipper->shouldSkipElement($fileProcessor);
+        });
+    }
+
```
TomasVotruba, line 27: This should be done on rule-level. 
The same way all PHP rules are not skipped not, just specific ones.

src/NonPhpFile/NonPhpFileProcessor.php
```diff
@@ -46,6 +54,9 @@ public function __construct(
     public function process(array $files): void
     {
         foreach ($files as $file) {
+            if ($this->skipper->shouldSkipFileInfo($file->getSmartFileInfo())) {
+                continue;
+            }
             $this->processFile($file);
```
TomasVotruba, line 60: :+1:

rules/Composer/Application/FileProcessor/ComposerFileProcessor.php
```diff
@@ -45,6 +53,9 @@ public function __construct(
     public function process(array $files): void
     {
         foreach ($files as $file) {
+            if ($this->skipper->shouldSkipFileInfo($file->getSmartFileInfo())) {
+                continue;
```
TomasVotruba, line 57: :+1:

src/Application/ActiveFileProcessorsProvider.php
sabbelasichon: But, how could someone just ignore a complete FileProcessor which is part a set? In my case this would make sense, but i agree, it is different from current approach.

src/NonPhpFile/NonPhpFileProcessor.php
sabbelasichon: After looking at the code again, i see a duplication of this pattern in every FileProcessor. I am trying to move this one level up into the ApplicationFileProcessor.

sabbelasichon:
I am going to close this PR in favour of the https://github.com/rectorphp/rector/pull/6251

PR #6213, opened 2021-04-23
[RemoveUnusedVariableInCatchRector] Skip when variable is used outside catch

ruudk:
# Failing Test for RemoveUnusedVariableInCatchRector

Based on https://getrector.org/demo/194e0c75-cb30-4f02-a0de-70d32fbe152c

This broke some of my code.... you could argue that this is nasty code anyway, but it is what it is.


samsonasik:
@ruudk I chery-picked your commit at PR https://github.com/rectorphp/rector/pull/6231

PR #6084, opened 2021-04-10
[Privatization] Skip ChangeReadOnlyPropertyWithDefaultValueToConstantRector on StaticCall by Ref Param

samsonasik:
Closes #6083

PR #6049, opened 2021-04-07
[Privatization] Skip ChangeReadOnlyPropertyWithDefaultValueToConstantRector on by ref param method call

samsonasik:
Closes #6048 Fixes #6044

PR #6214, opened 2021-04-23
Add failing test fixture for ChangeReadOnlyPropertyWithDefaultValueToConstantRector

zingimmick:


samsonasik:
@zingimmick it seems placing in `Source` directory make not detected by `NodeRepository`, I actually register to composer.json as well but it still not found by `NodeRepository`, move to the current fixture solve it. 

Actually, this already handled, I cherry-picked your commit and move the class that has reference in arg method inside the fixture itself at #6234

zingimmick:
> @zingimmick it seems placing in `Source` directory make not detected by `NodeRepository`, I actually register to composer.json as well but it still not found by `NodeRepository`, move to the current fixture solve it.

Sorry for the slow response, but that is one of the steps to reproduce the problem because the code in `Source` directory is actually in `vendor` directory.

> Actually, this already handled, I cherry-picked your commit and move the class that has reference in arg method inside the fixture itself at #6234

Yes, PRs #6084, #6049 skipped `ChangeReadOnlyPropertyWithDefaultValueToConstantRector` on parameter referenced in method/static call, and [skip_referenced2.php.inc](https://github.com/rectorphp/rector/blob/dd0c1038da9d0d1da7c1ccd38f3b544801bf7cc6/rules-tests/Privatization/Rector/Property/ChangeReadOnlyPropertyWithDefaultValueToConstantRector/Fixture/skip_referenced2.php.inc) already handled this.

But since the code in `vendor` directory, move the class is not the optional way for me.

I also tried add these files to the option `autoload_paths` or `bootstrap_files` or both, but it not works. Did I miss something?

samsonasik:
@zingimmick could you create sample repository to check with github action enabled with the `ChangeReadOnlyPropertyWithDefaultValueToConstantRector ` rule enabled? Thank you.

zingimmick:
> @zingimmick could you create sample repository to check with github action enabled with the `ChangeReadOnlyPropertyWithDefaultValueToConstantRector ` rule enabled? Thank you.

Repository: https://github.com/zingimmick/rector-sample

samsonasik:
@TomasVotruba the `NodeRepository` seems can't get `ClassMethod` by `MethodCall` if the `ClassMethod` definition outside the code that call it. It seems the `addCallByType()` only register the `MethodCall`, not the `ClassMethod` 

https://github.com/rectorphp/rector/blob/a5a99317b7418c2914021d24ced9c172773465d9/packages/NodeCollector/NodeCollector/NodeRepository.php#L637-L657

samsonasik:
@zingimmick I cherry-picked your commit at PR https://github.com/rectorphp/rector/pull/6256

PR #6216, opened 2021-04-23
[ChangeSwitchToMatchRector] Add `default` case when variable is already initialized

ruudk:
# Failing Test for ChangeSwitchToMatchRector

Based on https://getrector.org/demo/f7f2ba48-3c19-4fdb-bc0a-e956e037eefd

It's important to add the default case to prevent this error when passing another code:
```
Fatal error: Uncaught UnhandledMatchError: Unhandled match value of type int
```

https://3v4l.org/E64Um


samsonasik:
@ruudk I cherry-picked your commit at PR https://github.com/rectorphp/rector/pull/6230

PR #6217, opened 2021-04-23
[ChangeSwitchToMatchRector] Skip when `switch` does not assigns a variable

ruudk:
# Failing Test for ChangeSwitchToMatchRector

Based on https://getrector.org/demo/f6b01d1b-dd26-4fcf-9260-43f850c0d2bc

It should not change this, as the docs clearly say:
```
A match expression returns a value.
```
https://www.php.net/match

This switch does not return a value.

### DIFF
--- a/rules-tests/Php80/Rector/Switch_/ChangeSwitchToMatchRector/Fixture/skip_when_switch_does_not_assign_value.php.inc
+++ b/rules-tests/Php80/Rector/Switch_/ChangeSwitchToMatchRector/Fixture/skip_when_switch_does_not_assign_value.php.inc
@@ -1,0 +1,21 @@
+<?php
+
+namespace Rector\Php80\Tests\Rector\Switch_\ChangeSwitchToMatchRector\Fixture;
+
+final class SkipWhenSwitchDoesNotAssignValue
+{
+    public function run()
+    {
+        switch ($input) {
+            case "a":
+                $ths->a();
+                break;
+            case "b":
+                $ths->b();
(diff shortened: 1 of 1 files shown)


TomasVotruba:
@tomasnorre Hi Tomas, if you're still up to help with Rector, this might be a good issue to start with :+1: 
If you don't understand anything, just ask. I'll try to guide you

tomasnorre:
I'll check it out.

tomasnorre:
The idea I have atm is to add a function to the SwitchAnalyzer, and include that into the `shouldSkipSwitch()`

```php
public function hasReturnStmt(Switch_ $switch): bool
{
    // code goes here     
}
```

But I honestly cannot get my head around how that check would  look like.

ruudk:
Maybe check if there is 1 variable assign in every switch case, and check if that is always the same variable. The cases should contain nothing else except whitespaces or comments. Then apply the rector. 

In any other case, don't do anything?

TomasVotruba:
@tomasnorre To find node inside node, you can use `NodeFinder`:

```php
foreach ($switch->cases as $case) {
	.... = $this->nodeFinder->findInstanceOf($case, Return::__class);
}
```

It might help to know how `Switch_` is build: https://github.com/rectorphp/php-parser-nodes-docs#phpparsernodestmtswitch_

samsonasik:
@ruudk I cherry-picked your commit at PR https://github.com/rectorphp/rector/pull/6257

PR #6223, opened 2021-04-23
[scoped] Build PHP 7.0 repo to own repository to avoid breaking prefixed rector

TomasVotruba:
cc @samsonasik The work on PHP 7.0 Rector repository should be done with these files :)


TomasVotruba:
After the PHP 7.0 downgrade is working, we can merge it to one :)

samsonasik:
@TomasVotruba thank you 👍

Issue #6232, opened 2021-04-24
Add template type option for init command


sabbelasichon:
In order to extend the init command with custom template types, i would like to create a PR for that.
For TYPO3-Rector we have created a custom typo3-init command, which is somehow redundant. 
https://github.com/sabbelasichon/typo3-rector/blob/master/src/Console/Command/Typo3InitCommand.php

PR #6233, opened 2021-04-24
[TASK] Add template type option for init command

sabbelasichon:
Resolves: #6232

### DIFF
--- a/docs/init_command.md
+++ b/docs/init_command.md
@@ -1,0 +1,35 @@
+# How to generate a configuration file
+
+To start quickly you can run the init command
+
+```bash
+vendor/bin/rector init
+```
+
+This will create a `rector.php` if it doesn´t already exist in your root directory with some sensitive defaults.
+
+```php
+// rector.php
+use Rector\Core\Configuration\Option;
+use Rector\Php74\Rector\Property\TypedPropertyRector;
+use Rector\Set\ValueObject\SetList;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
+    // here we can define, what sets of rules will be applied
+    $parameters = $containerConfigurator->parameters();
+    $parameters->set(Option::SETS, [SetList::CODE_QUALITY]);
+
+    // register single rule
+    $services = $containerConfigurator->services();
+    $services->set(TypedPropertyRector::class);
+};
+```
+
+The init command takes an option called --template-type or in short form -t.
+If some other Rector extension like [rector-nette](https://github.com/rectorphp/rector-nette) or [rector-doctrine](https://github.com/rectorphp/rector-doctrine) provides a custom template type you can specify it here:
+
+```bash
+vendor/bin/rector init --template-type=nette
+```
+
--- a/docs/init_command.md
+++ b/docs/init_command.md
@@ -27,9 +27,39 @@ return static function (ContainerConfigurator $containerConfigurator): void {
 ```
 
 The init command takes an option called --template-type or in short form -t.
-If some other Rector extension like [rector-nette](https://github.com/rectorphp/rector-nette) or [rector-doctrine](https://github.com/rectorphp/rector-doctrine) provides a custom template type you can specify it here:
+If some other Rector extension like [rector-nette](https://github.com/rectorphp/rector-nette) or [rector-doctrine](https://github.com/rectorphp/rector-doctrine) provides such a custom template type you can specify it here:
 
 ```bash
 vendor/bin/rector init --template-type=nette
 ```
 
+If you just want to use the default template provided by Rector you can omit the --template-type option.
+
+# How to add a template type as a developer
+In order to provide a new template type as a developer you should create a custom template class implementing the TemplateResolverInterface:
+
+```
+use Rector\Core\Contract\Template\TemplateResolverInterface;
+
+final class MyCustomTemplate implements TemplateResolverInterface
+{
+    /**
+     * @var string
+     */
+    private const TYPE = 'custom';
+
+    public function provide(): string
+    {
+        return __DIR__ . '/path/to/custom/template.php.dist';
+    }
+    public function supports(string $type): bool
+    {
+        return $type === self::TYPE;
+    }
+
+    public function __toString(): string
+    {
+        return self::TYPE;
+    }
+}
+```
--- a/docs/init_command.md
+++ b/docs/init_command.md
@@ -26,7 +26,7 @@ return static function (ContainerConfigurator $containerConfigurator): void {
 };
 ```
 
-The init command takes an option called --template-type or in short form -t.
+The init command takes an option called `--template-type`.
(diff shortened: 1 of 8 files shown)


src/Configuration/Option.php
```diff
@@ -157,4 +157,9 @@ final class Option
      * @var string
      */
     public const OPTION_NO_DIFFS = 'no-diffs';
+
+    /**
+     * @var string
+     */
+    public const TEMPLATE_TYPE = 'type';
```
TomasVotruba, line 164: Value and name should be equal. That way we can avoid conflict with e.g. `ANY_TYPE = 'type';`

TomasVotruba:
The docs file should be added as well

src/Configuration/Option.php
sabbelasichon: So it should be template-type with or without prefix OPTION?

TomasVotruba: What would you prefer?

sabbelasichon: For the value template-type and for the constant name i would use TEMPLATE_TYPE also. Not using OPTION prefix here. But sometimes it is used. Do not see a pattern where and why.

sabbelasichon:
> The docs file should be added as well
What do you mean with docs file? The only place where the init command is mentioned right now is in the README.md. 
As there is no other official template-type for now i would prefer to drop it at the beginning and add it to the docs once there are different template files.

src/Configuration/Option.php
sabbelasichon: I see. The option prefix is for the command options. So we should add it here as well. Sorry too early and yesterday too late

TomasVotruba: No troubles :)

Personally I would go for:

```php
const TEMPLATE_TYPE = 'template_type'
```

TomasVotruba:
> What do you mean with docs file?

A small documentation, that describes this feature. E.g. in `/docs/init_command.md`
The goal is to other people understand without looking at this PR and its description.

src/Configuration/Option.php
sabbelasichon: Ok. But the pattern for options within a command is different. For existing ones it is always OPTION_FOO_BAR = 'option-with-dash'

TomasVotruba: CLI is only one of possible inputs. The parameter can come from manual parameter too.

sabbelasichon:
I think we should create an issue for the FileProcessors documentation. But before documenting it i would prefer to streamline the API a little bit like skipping them or running them without a Rector should be possible.

src/Configuration/Option.php
sabbelasichon: But in cli mode i would prefer the dash. And this is visible for the user. In all other parts it is a hidden detail. Naming things is still hard and not solved by robots till now ;-)

docs/init_command.md
```diff
@@ -0,0 +1,35 @@
...
+    $services->set(TypedPropertyRector::class);
+};
+```
+
+The init command takes an option called --template-type or in short form -t.
+If some other Rector extension like [rector-nette](https://github.com/rectorphp/rector-nette) or [rector-doctrine](https://github.com/rectorphp/rector-doctrine) provides a custom template type you can specify it here:
+
+```bash
+vendor/bin/rector init --template-type=nette
+```
```
TomasVotruba, line 34: I wonder, how can I add such a template as a developer? 

When should I use this extra option and when only `init`?

sabbelasichon: I enhanced the documentation to explain this part also. Thanks for asking.

docs/init_command.md
```diff
@@ -0,0 +1,65 @@
...
+
+If you just want to use the default template provided by Rector you can omit the --template-type option.
+
+# How to add a template type as a developer
+In order to provide a new template type as a developer you should create a custom template class implementing the TemplateResolverInterface:
+
+```
```
TomasVotruba, line 41:
```suggestion
```php
```

docs/init_command.md
```diff
@@ -0,0 +1,65 @@
...
+The init command takes an option called --template-type or in short form -t.
+If some other Rector extension like [rector-nette](https://github.com/rectorphp/rector-nette) or [rector-doctrine](https://github.com/rectorphp/rector-doctrine) provides such a custom template type you can specify it here:
+
+```bash
+vendor/bin/rector init --template-type=nette
+```
+
+If you just want to use the default template provided by Rector you can omit the --template-type option.
+
+# How to add a template type as a developer
```
TomasVotruba, line 38: :+1:

docs/init_command.md
```diff
@@ -0,0 +1,65 @@
...
+    $parameters = $containerConfigurator->parameters();
+    $parameters->set(Option::SETS, [SetList::CODE_QUALITY]);
+
+    // register single rule
+    $services = $containerConfigurator->services();
+    $services->set(TypedPropertyRector::class);
+};
+```
+
+The init command takes an option called --template-type or in short form -t.
```
TomasVotruba, line 29:
```suggestion
The init command takes an option called `--template-type`.
```

TomasVotruba: Adding short options makes command less readable, so I would prefer avoid it untill it's something generic across CLI apps e.g. like `-n` for `--dry-run`.

docs/init_command.md
```diff
@@ -0,0 +1,65 @@
...
+If some other Rector extension like [rector-nette](https://github.com/rectorphp/rector-nette) or [rector-doctrine](https://github.com/rectorphp/rector-doctrine) provides such a custom template type you can specify it here:
+
+```bash
+vendor/bin/rector init --template-type=nette
+```
```
TomasVotruba, line 34: Thanks for completing the answers!

One last detail - what is the motivation to use a "nette" template for `rector.php`? How does it differ from the main `rector.php`? So people understand, why they should consider using the `nette` one over generic one.

src/Console/Command/InitCommand.php
```diff
@@ -29,26 +33,45 @@ final class InitCommand extends Command
...
+        $this->templateResolvers = $templateResolvers;
         parent::__construct();
     }
 
     protected function configure(): void
     {
         $this->setDescription('Generate rector.php configuration file');
+        $this->addOption(
+            Option::TEMPLATE_TYPE,
+            't',
```
TomasVotruba, line 62:
```suggestion
            null,
```

docs/init_command.md
sabbelasichon: Could we answer this later? Because i don´t know yet. I only know my personal motivation for TYPO3.

sabbelasichon: I would have to come up with a contrived example because no template for nette or any other extension exists yet.

sabbelasichon: I promise i will do that later.

TomasVotruba: > I only know my personal motivation for TYPO3

Real motivation is the best. Put it there :+1:

sabbelasichon: Done

TomasVotruba: Awesome, now I get it! Thanks :+1:

PR #6234, opened 2021-04-24
[Privatization] Skip ChangeReadOnlyPropertyWithDefaultValueToConstantRector on call reference by another class

samsonasik:
@zingimmick this already handled, just can't be detected by `NodeRepository` as it was placed in `Source` directory, even registered in composer.json autoload-dev.

The "another class" call with reference moved to fixture class itself solve it.

Closes #6214


samsonasik:

samsonasik:
Temporary closing it as the issue happen for `NodeRepository` can't detect "Source" as vendor ref https://github.com/rectorphp/rector/pull/6214#issuecomment-826241071

PR #6235, opened 2021-04-24
[Scoped] Scoped PHP 7.0

samsonasik:


.github/workflows/build_scoped_rector_php70.yaml
```diff
@@ -34,17 +33,20 @@ jobs:
                     COMPOSER_TOKEN: ${{ secrets.ACCESS_TOKEN }}
 
             # install only prod dependencies - do not use ramsey, it uses cache including "dev", we want to avoid it here
-            -   run: composer install --no-dev --ansi
+            -   run: |
+                    composer install --no-dev --ansi
+                    composer remove phpstan/phpstan
+                    composer require phpstan/phpstan-src
```
samsonasik, line 39: @TomasVotruba to handle dependency conflict, I think we can do this, remove `phpstan/phpstan` and require `phpstan/phpstan-src`, with `phpstan` dependencies then removed from composer.json target https://github.com/rectorphp/rector/pull/6235/commits/c884f78ea066a56f0a4f86b272885edae30c57da

samsonasik: The `phpstan/phpstan-src` is not in the packagist, seems need to be added to "repositories" config:

```json
    "repositories" : [
        {
            "type" : "vcs",
            "url" : "https://github.com/phpstan/phpstan-src",
            "no-api": true
        }
    ]
```

samsonasik: it seems not installable:

```bash
  [RuntimeException]                                                                                           
  Failed to clone 
https://github.com/phpstan/phpstan-src
 via https, ssh protocols, aborting.                   
  - 
https://github.com/phpstan/phpstan-src

    Cloning into bare repository '/home/runner/.cache/composer/vcs/https---github.com-phpstan-phpstan-src'...  
    BUG: refs/files-backend.c:2953: initial ref transaction called with existing refs                          
    Aborted (core dumped)                                                                                      
```

TomasVotruba: I used to install it like this, just without "no-api"

samsonasik: I usually use `no-api` used to allow non-owner to use, eg: in fork or PR, without no-api, it got the following error:

```bash
Error: Could not authenticate against github.com

  [Composer\Downloader\TransportException]   
  Could not authenticate against github.com  
```

samsonasik: @TomasVotruba without no-api, it got error:

```bash
Error: Could not authenticate against github.com

  [Composer\Downloader\TransportException]   
  Could not authenticate against github.com  
```

in https://github.com/rectorphp/rector/pull/6235/commits/df7769606a40667c69e36d4bc1d365377a93cdb1 .

TomasVotruba: This works for me locally:

```bash
{
	"require": {
		"phpstan/phpstan-src": "dev-master"
	},
	"repositories": [
		{
			"type": "vcs",
			"url": "https://github.com/phpstan/phpstan-src"
		}
	]
}
```

The dev packages needs to be completed too.

image

TomasVotruba: To modify `composer.json` in command line, you can use this https://getcomposer.org/doc/03-cli.md#modifying-repositories

samsonasik: @TomasVotruba that seems because of you are the repo owner so the authentication github always works.

samsonasik: undo change root composer.json

.github/workflows/build_scoped_rector_php70.yaml
```diff
@@ -34,17 +33,22 @@ jobs:
                     COMPOSER_TOKEN: ${{ secrets.ACCESS_TOKEN }}
 
             # install only prod dependencies - do not use ramsey, it uses cache including "dev", we want to avoid it here
-            -   run: composer install --no-dev --ansi
+            -   run: |
+                    composer install --no-dev --ansi
+                    composer remove phpstan/phpstan
+                    composer clear-cache
+                    composer config repositories.phpstan/phpstan-src vcs https://github.com/phpstan/phpstan-src
```
samsonasik, line 40: @TomasVotruba I updated to use `config config repositories` command to register `phpstan/phpstan-src` repo

samsonasik:
All checks have passed 🎉 @TomasVotruba let's give it a try?

samsonasik:
@TomasVotruba I am merging it to see if the build working on merge for php 7.0 scoped

PR #6237, opened 2021-04-25
[Php80] Handle UnionTypesRector on object and specific class/type in union type

samsonasik:
Closes #6194


samsonasik:

TomasVotruba:
Thanks. There should be an extra rule that removes the extra `object` keyword so change is possible.

samsonasik:
@TomasVotruba do you mean we need a new rector rule, eg: `RemoveObjectKeywordInArgs` rule? or the changed should be the following (add object type only) : 

```php
     /**
      * @param object|DateTime $message
      */
     private function getMessage(object $message)

```

TomasVotruba:
Something like this:

```diff
     /**
-      * @param object|DateTime $message
+      * @param DateTime $message
      */
     private function getMessage(object $message)
```

Then another rule can autocomplete the type

rules/Php80/Rector/FunctionLike/UnionTypesRector.php
```diff
@@ -138,6 +162,56 @@ private function refactorParamTypes(FunctionLike $functionLike, PhpDocInfo $phpD
...
+        $types = $unionType->getTypes();
+        foreach ($types as $key => $type) {
+            if ($type instanceof ObjectWithoutClassType) {
+                unset($types[$key]);
+            }
+        }
+
+        $currentType = current($types);
+        $type = count($types) === 1
+            ? new ObjectType($currentType->getClassName())
```
samsonasik, line 209: @TomasVotruba the type of `@param` with new ObjectType usage make : `DateTime` changed to `\DateTime`: 

with: 

```php
$this->phpDocTypeChanger->changeParamType($phpDocInfo, $type, $param, $paramName);
``` 

```diff
    /**
-     * @param DateTime $message
+     * @param \DateTime $message
     */
```

any idea to make it like "string" type with value "DateTime" ?

samsonasik: @TomasVotruba implemented with `IdentifierTypeNode` 🎉

samsonasik:
implemented.

samsonasik:

rules/Php80/Rector/FunctionLike/UnionTypesRector.php
```diff
@@ -129,6 +137,11 @@ private function refactorParamTypes(FunctionLike $functionLike, PhpDocInfo $phpD
                 continue;
             }
 
+            if ($this->hasObjectWithoutClassType($paramType)) {
+                $this->changeObjectWithoutClassType($paramType, $param, $phpDocInfo, $key);
+                continue;
+            }
```
TomasVotruba, line 143: I probably was not clear about this. Could you extract the "narrow `object` and type" to a separated Rector rule?

samsonasik: Do you mean this rule skip change when parameter type has object without class type?

samsonasik: @TomasVotruba your request on comment https://github.com/rectorphp/rector/pull/6237#issuecomment-826273171 actually implemented in fixture `object_and_specific_class.php.inc` (https://github.com/rectorphp/rector/blob/close-6194/rules-tests/Php80/Rector/FunctionLike/UnionTypesRector/Fixture/object_and_specific_class.php.inc) ,

am I missing something?

TomasVotruba: That's the place I was not clear enough.
This should be a new rule as UnionType should only move the type to PHP, not narrow it down.

Changing docblock by narrowing type should be a new rule. That way we can extend it with other similar rules, e.g.

```diff
-@var int|string|bool|floatarray
+@var scalar
```

Such rule can be applied without PHP 8 as well. It could be named `NarrowUnionTypeDocRector`

samsonasik: @TomasVotruba ok, implemented at https://github.com/rectorphp/rector/pull/6237/commits/3e4d3c3d09a4301274afe33b0e36bfe9d2c5876d , skip change `@param` the docblock, only change the type hint param in method.

samsonasik:

TomasVotruba:
Thanks

rules/Php80/Rector/FunctionLike/UnionTypesRector.php
TomasVotruba: Thanks :+1: 

Now we'll need a new separated rule that narrows the types.

samsonasik: I will try.

PR #6239, opened 2021-04-25
Fix AUTOLOAD_PATHS and BOOTSTRAP_FILES in tests

TomasVotruba:
Fixes https://github.com/rectorphp/rector/pull/6158


staabm:
confirmed, the fix works for me. <3

PR #6242, opened 2021-04-25
Decouple Rector PHPStan rules to own package

TomasVotruba:
See https://github.com/rectorphp/rector-phpstan-rules


TomasVotruba:
@sabbelasichon Hey, this PHPStan rules package covers lot of best practise for Rector rule development.
It will warn you about `class_exist()` and `RefletctionClass()`, where static reflection should be used instead, etc.

Be sure to add it to typo3-Rector, it will help you to develop safely :hugs:

PR #6243, opened 2021-04-26
[DowngradePhp70] Add DowngradeSpaceshipRector

samsonasik:
part of https://github.com/rectorphp/rector/issues/6079 . Use "self invoked" invokable anonymous function as it can be part of Return, echo, if, etc.


rules/DowngradePhp70/Rector/Spaceship/DowngradeSpaceshipRector.php
```diff
@@ -0,0 +1,103 @@
...
+        );
+    }
+
+    /**
+     * @param Spaceship $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        $leftVariableParam = new Variable('a');
+        $rightVariableParam = new Variable('b');
```
samsonasik, line 82: use `($a, $b)` as param is ok here as it just a named parameter, the actual call is depends on the value in the spaceship (left,right) that passed as "self invoked" anonymous parameter, eg:

**Before**:

```php
return getRand() <=> getRand();
```

**After**:

```php
        return (function ($a, $b) {
             if ($a === $b) {
                 return 0;
             }
             return $a < $b ? -1 : 1;
         })(getRand(), getRand());
```

samsonasik:

rules-tests/DowngradePhp70/Rector/Spaceship/DowngradeSpaceshipRector/Fixture/echo.php.inc
```diff
@@ -0,0 +1,32 @@
...
+class UseEcho
+{
+    public function run($a, $b)
+    {
+        echo (function ($a, $b) {
+            if ($a === $b) {
+                return 0;
+            }
+            return $a < $b ? -1 : 1;
+        })($a, $b);
```
TomasVotruba, line 28: While this might be technically correct, I find it very hard to read - definition vs call. Similar to class vs anonymous class.

Could you use an anonymous function assinged to a variable? 

```diff
+$battleShipcompare = function ($a, $b) {
-echo (function ($a, $b) {
            if ($a === $b) {
                return 0;
            }
            return $a < $b ? -1 : 1;
-        })($a, $b);
+        });
+$battleShipcompare($a, $b);
```

rules/DowngradePhp70/Rector/Spaceship/DowngradeSpaceshipRector.php
TomasVotruba: The code would not be read only by machines, but by developers too. In case of debugging working with single-letter named variables is painful, e.g. https://ocramius.github.io/blog/eliminating-visual-debt/ describes it well :) 

How could we name it better? left, right, first, second?

samsonasik: updated to use `($left, $right)` https://github.com/rectorphp/rector/pull/6243/commits/6d411371f72907b1938cfbdc318adde3b27fb3a4

rules-tests/DowngradePhp70/Rector/Spaceship/DowngradeSpaceshipRector/Fixture/echo.php.inc
samsonasik: updated to assign to variable `battleShipcompare` first at https://github.com/rectorphp/rector/pull/6243/commits/cb6e1e46bb376592f26e2c3caaa42169744265a3

samsonasik:

TomasVotruba:
Thank you 😊

PR #6251, opened 2021-04-26
[TASK] MissingRectorRulesReporter is taking all FileProcessors into account

sabbelasichon:
Resolves: #6087

### DIFF
--- a/rules/Composer/Application/FileProcessor/ComposerFileProcessor.php
+++ b/rules/Composer/Application/FileProcessor/ComposerFileProcessor.php
@@ -75,6 +75,11 @@ final class ComposerFileProcessor implements FileProcessorInterface
         return ['json'];
     }
 
+    public function isActive(): bool
+    {
+        return count($this->composerRectors) > 0;
+    }
+
     private function processFile(File $file): void
     {
         // to avoid modification of file
--- a/src/Application/FileProcessor/PhpFileProcessor.php
+++ b/src/Application/FileProcessor/PhpFileProcessor.php
@@ -12,12 +12,14 @@ use Rector\Core\Application\FileSystem\RemovedAndAddedFilesCollector;
 use Rector\Core\Application\FileSystem\RemovedAndAddedFilesProcessor;
 use Rector\Core\Configuration\Configuration;
 use Rector\Core\Contract\Processor\FileProcessorInterface;
+use Rector\Core\Contract\Rector\RectorInterface;
 use Rector\Core\Exception\ShouldNotHappenException;
 use Rector\Core\PhpParser\Printer\FormatPerservingPrinter;
 use Rector\Core\Provider\CurrentFileProvider;
 use Rector\Core\ValueObject\Application\File;
 use Rector\Core\ValueObject\Application\RectorError;
 use Rector\PostRector\Application\PostFileProcessor;
+use Rector\PostRector\Contract\Rector\PostRectorInterface;
 use Rector\Testing\PHPUnit\StaticPHPUnitEnvironment;
 use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Style\SymfonyStyle;
@@ -86,7 +88,7 @@ final class PhpFileProcessor implements FileProcessorInterface
     /**
      * @var FormatPerservingPrinter
      */
-    private $formatPerservingPrinter;
+    private $formatPreservingPrinter;
 
     /**
      * @var PostFileProcessor
@@ -98,9 +100,17 @@ final class PhpFileProcessor implements FileProcessorInterface
      */
     private $errorFactory;
 
+    /**
+     * @var RectorInterface[]
+     */
+    private $rectors;
+
+    /**
+     * @param RectorInterface[] $rectors
+     */
     public function __construct(
         Configuration $configuration,
-        FormatPerservingPrinter $formatPerservingPrinter,
+        FormatPerservingPrinter $formatPreservingPrinter,
         FileProcessor $fileProcessor,
         RemovedAndAddedFilesCollector $removedAndAddedFilesCollector,
         RemovedAndAddedFilesProcessor $removedAndAddedFilesProcessor,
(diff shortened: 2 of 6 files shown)


rules/Composer/Application/FileProcessor/ComposerFileProcessor.php
```diff
@@ -29,17 +30,14 @@ final class ComposerFileProcessor implements FileProcessorInterface
...
     private $composerRectors = [];
 
-    /**
-     * @param ComposerRectorInterface[] $composerRectors
-     */
     public function __construct(
         ComposerJsonFactory $composerJsonFactory,
         ComposerJsonPrinter $composerJsonPrinter,
-        array $composerRectors
+        ActiveRectorsProvider $activeRectorsProvider
```
TomasVotruba, line 36: I assume this service some kind of filtering based in `SKIP` parameter, right?

If so, is there a use case where we need rectors without this filtering?

It would make sense to put this filtering on [CompilerPass](https://symfony.com/doc/current/service_container/compiler_passes.html) level, so we make sure we alwasy get the same list of rules.

src/Application/FileProcessor/PhpFileProcessor.php
```diff
@@ -194,6 +203,11 @@ public function getSupportedFileExtensions(): array
         return $this->configuration->getFileExtensions();
     }
 
+    public function isActive(): bool
+    {
+        return count($this->activeRectorsProvider->provideByType(PhpRectorInterface::class)) > 0;
+    }
+
```
TomasVotruba, line 210: :+1:

rules/Composer/Application/FileProcessor/ComposerFileProcessor.php
sabbelasichon: Maybe in the future we have some kind of list command for all rules of a given interface. I would not use a CompilerPass here to only get active ones.

src/NonPhpFile/NonPhpFileProcessor.php
```diff
@@ -61,6 +70,11 @@ public function getSupportedFileExtensions(): array
         return StaticNonPhpFileSuffixes::SUFFIXES;
     }
 
+    public function isActive(): bool
+    {
+        return count($this->activeRectorsProvider->provideByType(PhpRectorInterface::class)) > 0;
```
lulco, line 75: This is the same like in PhpFileProcessor. Is it correct?

rules/Composer/Application/FileProcessor/ComposerFileProcessor.php
TomasVotruba: I'm not sure I understand. I mean now there is a space to forget `ActiveRectorsProvider`, as we use array autowiring convention across the project. We have to keep this convention to avoid [memory locks](https://tomasvotruba.com/blog/2018/08/27/why-and-how-to-avoid-the-memory-lock/).

Here the `array $composerRectors` must remain

src/NonPhpFile/NonPhpFileProcessor.php
sabbelasichon: I am not sure. But i guess so. The NonPhpFileProcessor is depending on some active Php Rector rules. But this FileProcessor is still a blackbox to me. Haven´t used it by now.

rules/Composer/Application/FileProcessor/ComposerFileProcessor.php
sabbelasichon: @TomasVotruba I understand, but as we already use this pattern with the ActiveRectorsProvider for the PhpFileProcessors i thought it is consistent to use it here again. And honestly i am not sure how to do it on CompilerPass level without introducing a lot of complexity here. I haven´t used custom CompilerPasses that much. Could you put me on the right track, maybe i can get somewhere.

TomasVotruba: I implemented the ActiveRectorsProvider very poorly. It should not require so much work to use.

This is 1-place solution with CompilerPass: https://github.com/rectorphp/rector/pull/6286

TomasVotruba: Now you can get all the active Rector rule count like this:

```php
/**
 * @param RectorInterface[]
 */
public function __construct(array $rectors)
{
     $activeRectors = array_filter($rectors, function (RectorInterface $rector) {
         // skip interface post rectors, that are always present
         return ! $rector instanceof PostRectorInterface;
     }

     $activeRectorCount = count($activeRectors);
}
```

src/NonPhpFile/NonPhpFileProcessor.php
sabbelasichon: Ok. Now we have a dedicated Marker Interface NonPhpRectorInterface. Confusion is gone.

src/Application/FileProcessor/PhpFileProcessor.php
sabbelasichon: ActiveRectorsProvider is gone.

sabbelasichon:
Ready for review

src/Application/FileProcessor/PhpFileProcessor.php
```diff
@@ -98,9 +100,17 @@ final class PhpFileProcessor implements FileProcessorInterface
...
+     */
+    private $rectors;
+
+    /**
+     * @param RectorInterface[] $rectors
+     */
     public function __construct(
         Configuration $configuration,
-        FormatPerservingPrinter $formatPerservingPrinter,
+        FormatPerservingPrinter $formatPreservingPrinter,
```
sabbelasichon, line 113: Sorry for the noise ;-)

src/Reporting/MissingRectorRulesReporter.php
```diff
@@ -17,26 +16,29 @@ final class MissingRectorRulesReporter
...
+            $this->fileProcessors,
+            function (FileProcessorInterface $fileProcessor): bool {
+                return $fileProcessor->isActive();
+            }
+        );
 
-        if ($activeRectors !== []) {
+        if ($activeFileProcessors !== []) {
             return null;
         }
```
TomasVotruba, line 43: I don't see much advantage over 3 lines above. This change requires to implement a method by every FileProcessor, that will be always the same.

sabbelasichon: True. I have missed this. As you have done the CompilerPass change i think we can close this PR here. Everything is possible now.

TomasVotruba: Agreed. Thanks!

PR #6254, opened 2021-04-27
AnnotationToAttributeRector: ensure bool param in attribute

tugmaks:
Hi, seems AnnotationToAttributeRector has bug if annotation has bool param

```
-    #[\Symfony\Component\Validator\Constraints\NotBlank(allowNull: true)]
+    #[\Symfony\Component\Validator\Constraints\NotBlank(allowNull: 'true')]
```


TomasVotruba:
Thank you for great test and the fix!

PR #6256, opened 2021-04-28
[Privatization] Handle Reference Method/Static Call from outside file on ChangeReadOnlyPropertyWithDefaultValueToConstantRector

samsonasik:
Closes #6214 by update `NodeRepository` to parse outside file if not found in the list of registered class method.

### DIFF
--- a/packages/NodeCollector/NodeCollector/NodeRepository.php
+++ b/packages/NodeCollector/NodeCollector/NodeRepository.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace Rector\NodeCollector\NodeCollector;
 
 use Nette\Utils\Arrays;
-use Nette\Utils\FileSystem;
+use Symplify\SmartFileSystem\SmartFileSystem;
 use Nette\Utils\Strings;
 use PhpParser\Node;
 use PhpParser\Node\Attribute;
@@ -242,7 +242,7 @@ final class NodeRepository
             }
 
             /** @var string $fileContent */
-            $fileContent = FileSystem::read($fileName);
+            $fileContent = SmartFileSystem::readfile($fileName);
             $nodes = $this->parser->parse($fileContent);
 
             return $this->getClassMethodFromNodes((array) $nodes, $classReflection, $className, $methodName);
--- a/packages/NodeCollector/NodeCollector/NodeRepository.php
+++ b/packages/NodeCollector/NodeCollector/NodeRepository.php
@@ -248,9 +248,8 @@ final class NodeRepository
                 return null;
             }
 
-            /** @var string $fileContent */
             $fileContent = $this->smartFileSystem->readfile($fileName);
-            $nodes = $this->parser->parse($fileContent);
+            $nodes = $this->parser->parse((string) $fileContent);
 
             return $this->getClassMethodFromNodes((array) $nodes, $classReflection, $className, $methodName);
         }
--- a/packages/NodeCollector/NodeCollector/NodeRepository.php
+++ b/packages/NodeCollector/NodeCollector/NodeRepository.php
@@ -6,6 +6,7 @@ namespace Rector\NodeCollector\NodeCollector;
 
 use Nette\Utils\Arrays;
 use Nette\Utils\Strings;
+use PhpParser\Error;
 use PhpParser\Node;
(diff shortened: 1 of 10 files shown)


samsonasik:
Also fix ReturnThisRemoveRector on check "vendor" dir

samsonasik:

packages/NodeCollector/NodeCollector/NodeRepository.php
```diff
@@ -223,6 +243,31 @@ public function findClassMethod(string $className, string $methodName): ?ClassMe
...
+        $fileContent = $this->smartFileSystem->readfile($fileName);
+        try {
+            $nodes = $this->parser->parse($fileContent);
+            return $this->getClassMethodFromNodes((array) $nodes, $classReflection, $className, $methodName);
+        } catch (Error $error) {
+            $isNotSyntaxError = ! Strings::contains($error->getMessage(), 'Syntax error');
+            if ($isNotSyntaxError) {
+                throw $error;
+            }
+        }
```
TomasVotruba, line 269: Why is this needed? It might hide unwated bugs in broken code

samsonasik: It got error on php < 8 on `FluentChainMethodCallToNormalMethodCallRector` tests: 

```bash
1) Rector\Tests\Defluent\Rector\MethodCall\FluentChainMethodCallToNormalMethodCallRector\FluentChainMethodCallToNormalMethodCallRectorTest::test with data set #17 (Symplify\SmartFileSystem\SmartFileInfo Object (...))
59
PhpParser\Error: Syntax error, unexpected T_CONSTANT_ENCAPSED_STRING on line 208
```

ref https://github.com/rectorphp/rector/runs/2454146699#step:5:58

packages/NodeCollector/NodeCollector/NodeRepository.php
```diff
@@ -499,6 +544,50 @@ public function findClassLike(string $classLikeName): ?ClassLike
...
+            }
+
+            $classMethod = $this->getClassMethod($node, $shortName, $methodName);
+            if ($classMethod instanceof ClassMethod) {
+                return $classMethod;
+            }
+        }
+
+        return null;
+    }
```
TomasVotruba, line 577: NodeRepository should only work with nodes in analysed project.

For parsing vendor/Reflection code, the `\Rector\Core\Reflection\FunctionLikeReflectionParser` should be used instead. Could you move the logic there and re-use content in that service?

samsonasik: I will try.

packages/NodeCollector/NodeCollector/NodeRepository.php
TomasVotruba: What file is that?

samsonasik: in these 2 fixture files:

1. https://github.com/rectorphp/rector/blob/main/rules-tests/Defluent/Rector/MethodCall/FluentChainMethodCallToNormalMethodCallRector/Fixture/skip_date_time.php.inc

2. https://github.com/rectorphp/rector/blob/main/rules-tests/Defluent/Rector/MethodCall/FluentChainMethodCallToNormalMethodCallRector/Fixture/skip_date_time_modify_command.php.inc

samsonasik: Looking at the fixture files, it seems the erorr is caused by try to fetch to native class DateTime's method, so it may need `isBuiltIn()` check

samsonasik: Yes, it needs `isBuiltIn()` check https://github.com/rectorphp/rector/pull/6256/commits/7abde9909df1e8eebe23f8f2b61969d0f6ac8806 🎉

packages/NodeCollector/NodeCollector/NodeRepository.php
samsonasik: implemented, moved to FunctionLikeReflectionParser https://github.com/rectorphp/rector/pull/6256/commits/677613048870fcdca38bd98fb5701cb791fea3ab

PR #6257, opened 2021-04-28
[Php80] Skip ChangeSwitchToMatchRector on no assign when not return

samsonasik:
Closes #6217 . The deleted fixture is invalid fixture which actually should be skipped.


samsonasik:
All checks have passed :tada: @TomasVotruba it is ready for review.

TomasVotruba:
Thanks :+1:

PR #6259, opened 2021-04-28
AnnotationToAttributeRector: array handling

tugmaks:
There are some issues when handling arrays:

```php
    /**
     * @Assert\Choice(choices={"foo", "bar"})
     */
    public $surname;
```
is converted into
```php
    #[\Symfony\Component\Validator\Constraints\Choice(choices: '{"foo", "bar"}')]
    public $surname;
```

This PR solves it


TomasVotruba:
Very nice work, thank you!

PR #6262, opened 2021-04-29
[CodeQuality] Add NarrowUnionTypeDocRector

samsonasik:
Based on https://github.com/rectorphp/rector/pull/6237#discussion_r619857427


samsonasik:
@TomasVotruba it is implemented for `object|DateTime` param to `DateTime` for has "object" type in union type with others are class only, any other list of types that need to be narrowed ? , eg:

```diff
-@param int|string|bool|float|array
+@param scalar 
```

TomasVotruba:
> any other list of types that need to be narrowed 

Yes, that's a good idea :+1: anything you can think of. (I tried but could not find any :smile: )

samsonasik:
I've added scalar narrow

rules/Php80/Rector/FunctionLike/UnionTypesRector.php
```diff
@@ -152,41 +158,13 @@ private function refactorParamTypes(FunctionLike $functionLike, PhpDocInfo $phpD
...
-            }
-
-            if (! $type instanceof FullyQualifiedObjectType) {
-                return false;
-            }
-        }
-
-        return true;
-    }
-
```
TomasVotruba, line 189: :+1:

samsonasik:

rules/CodeQuality/Rector/ClassMethod/NarrowUnionTypeDocRector.php
```diff
@@ -0,0 +1,127 @@
...
+                continue;
+            }
+
+            if ($this->unionTypeAnalyzer->hasObjectWithoutClassType($paramType)) {
+                $this->changeDocObjectWithoutClassType($paramType, $key, $phpDocInfo);
+                continue;
+            }
+        }
+
+        return $node;
```
TomasVotruba, line 98: In we work with docblock, the node should be returned only if the PHPDoc changed. To avoid false positive reports on rule changes:

```php
if ($phpDocInfo->hasChanged()) {
    return $node;
}

return null;
```

samsonasik: updated.

samsonasik:

TomasVotruba:
Thank you :+1:

PR #6269, opened 2021-04-30
[CodeQuality] Register NarrowUnionTypeDocRector to code-quality config set

samsonasik:


samsonasik:
handle skipped partial scalar types

samsonasik:

PR #6270, opened 2021-04-30
[DowngradePhp70] Add DowngradeDefineArrayConstantRector

samsonasik:
part of https://github.com/rectorphp/rector/issues/6079


rules/DowngradePhp70/Rector/FuncCall/DowngradeDefineArrayConstantRector.php
```diff
@@ -0,0 +1,94 @@
...
+            [
+                new Const_(
+                    new Name(
+                        $node->args[0]->value->value
+                    ),
+                    $node->args[1]->value
+                )
+            ]
+        );
+        return $parent->consts[0];
```
samsonasik, line 82: @TomasVotruba do you have idea to create `const` assign definition? The code above define constant but without `const ` start, eg:

```diff
-const ANIMALS = [
+ANIMALS = [
     'dog',
     'cat',
     'bird'
```

samsonasik: it works with string `const ` prefix on const name definition https://github.com/rectorphp/rector/pull/6270/commits/73001fa5d1657f8a43cc7c4b56871b926224f47e :tada:

samsonasik:

rules/DowngradePhp70/Rector/FuncCall/DowngradeDefineArrayConstantRector.php
samsonasik: that's not add "const " prefix definiton, that's solved with add "const " on name instantiation.

TomasVotruba: Sorry, wrong link: https://github.com/rectorphp/php-parser-nodes-docs#phpparsernodestmtclassconst

samsonasik: that's class constant, not a "const NAMECONST = [...]" that is not a class constant, the define is define constant that not a class constant.

TomasVotruba: Ah, sorry. This one then: https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Stmt/Const_.php

samsonasik: I tried that previously, see https://github.com/rectorphp/rector/pull/6270/commits/dc6934f3c04bc98107b97d8a0131adcc73d93940 , the solution is add string "const " during Const_ name instantiation

TomasVotruba: I've just pushed it. Feel free to re-use

samsonasik: ok, implemented at https://github.com/rectorphp/rector/pull/6270/commits/89dcb1ea255d6ca47990857827dcc3d05cba9e0e

samsonasik:

TomasVotruba:
Thank you

Issue #6215, opened 2021-04-23
Take editorconfig into account for non php files.


sabbelasichon:
Often times many people are "complaining" about the wrong formatting output after transforming their files.
What i have done is taking the editorconfig into account for each file which is processed. 
https://github.com/sabbelasichon/typo3-rector/commit/66e7675053f91990179a808d68f7e122aea72197

What do you think? Should we add this for the composer part?

sabbelasichon:
If you like it, i could create a PR for that.

TomasVotruba:
Seems good idea, especially for multi formats. Go for it!

sabbelasichon:
Hi, i have used this little neat package here https://github.com/idiosyncratic-code/editorconfig-php. But i am thinking of taking it under my umbrella to fit our needs. Or would you like to add it under the symplify organization and i am gonna maintain it at the beginning.

TomasVotruba:
Hi, if you can take care of your umbrela, it would be the best. Other options require some maintenance cost for me.

PR #6272, opened 2021-04-30
[TASK] Take .editorconfig into account

sabbelasichon:
Resolves: #6215


sabbelasichon:
The minimum-stability and prefer-stable is still there because the dev-master of idiosyncratic-code/editorconfig-php contains  important bugfixes from my side. Stable release will be published soon.

sabbelasichon:
Still need some refactoring and hardening of the API

composer.json
```diff
@@ -32,6 +32,8 @@
         "composer/xdebug-handler": "^1.4|^2.0",
         "danielstjules/stringy": "^3.1",
         "doctrine/inflector": "^2.0",
+        "ergebnis/json-printer": "^3.1",
```
TomasVotruba, line 35: Instead we already use `Symplify\SmartFileSystem\Json\JsonFileSystem` that does the same job. Could you change it?

sabbelasichon: @TomasVotruba No, the JsonFileSystem does only a json_encode behind the scenes. And this is not really configurable hence this package exists. json_encode uses 4 spaces to print the file. Or do i miss something here?

TomasVotruba: So does the JsonFileSystem. We should use just once package instead of 2 :)

sabbelasichon: I think we have a misunderstanding. The ergebnis/json-printer package is highly configurable json printer and does more than a json_encode: https://localheinz.com/blog/2018/01/04/pretty-printing-json-with-custom-indent/

TomasVotruba: I see :+1: 

Do we need per suffix formatters? The space and indents are the same everywhere.

TomasVotruba: Could you add formatter for e.g. `*.twig` file? It should be more obvious then

sabbelasichon: > I see 👍
> 
> Why do we need per suffix formatters? The space and indents are the same everywhere.
> IMO there should be one generic formatter that works with any suffix.

I don´t know if this is possible. I guess you always need some dedicated specialized formatters for xml, json, html etc. I fear there is no one fits all solutions.

sabbelasichon: > Could you add formatter for e.g. `*.twig` file? It should be more obvious then

I am not aware of any twig formatter. What would you expect here?

TomasVotruba: Same test like this, just for twig:
https://github.com/rectorphp/rector/pull/6272/files#diff-0d87c4ff1a85d1f7b5b59e429c631725f403cd74d7f0852fd13798d00704c9ac

> I don´t know if this is possible. I guess you always need some dedicated specialized formatters for xml, json, html etc. I fear there is no one fits all solutions.

To make this feature complete we need to cover those we support already:

- json
- neon
- yaml
- twig
- latte

phpstan.neon
```diff
@@ -447,6 +447,8 @@ parameters:
         - '#Method Rector\\Core\\Tests\\DependencyInjection\\ConfigurableRectorImportConfigCallsMergeTest\:\:provideData\(\) return type has no value type specified in iterable type Iterator#'
         - '#Class with base "PhpVersion" name is already used in "PHPStan\\Php\\PhpVersion", "Rector\\Core\\ValueObject\\PhpVersion"\. Use unique name to make classes easy to recognize#'
 
+        - '#Method Rector\\Core\\Tests\\Formatter\\JsonFormatter\\JsonFormatterTest\:\:provideData\(\) return type has no value type specified in iterable type Iterator#'
+
```
TomasVotruba, line 451: This should be fixed

TomasVotruba:
Looking at this feature, there is lot of files all around `/src`. It would be better to have it in one directory together so it's easier to work with related files. 

- the `/src` code should be in `packages/FileFormatter`
- the `/tests` code should be in `packages-tests/FileFormatter`

composer.json
sabbelasichon: Yaml and neon should no be  a problem, but latte and twig. I am not sure.
The NonPhpFileProcessor is not really writing back a whole but only replacing some strings as far as i can see. 
So, no need to have a real formatter here.

TomasVotruba: I see. What format do we have to cover then?

sabbelasichon: I would say, xml, yaml, neon should be fine to start with.

sabbelasichon: I know it is some maintenance burden, if you don´t like the idea or feel resistance, we can still drop it.

TomasVotruba: I'm just trying to understand what needs to done to make it work for existing processors.

> I would say, xml, yaml, neon should be fine to start with.

OK. Could you add those?

sabbelasichon: Neon is harder than i expected. I thought that there is also a decoder with format options. But there isn´t. 

Maybe this whole formatting is a bad idea. It is really tough to cover all formats. In my case it is not only about changing tabs to spaces or indentation from 2 to 4. E.g. i am changing xml files and adding nodes, but preserve formatting is not that easy after modifying the xml structure.

TomasVotruba: I see. Lets stick what minimum you need at the moment. We'll extend it if needed in the future.

Let me know when it's ready to merge :+1:

packages-tests/FileFormatter/Formatter/XmlFileFormatter/Fixture/change_from_spaces_to_tabs.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<catalog>
+    <book id="bk101">
+<author>Gambardella, Matthew</author>
```
sabbelasichon, line 4: This is intended. Should be auto fixed.

packages/FileFormatter/FileFormatter.php
```diff
@@ -0,0 +1,68 @@
...
+        $this->fileFormatters = $fileFormatters;
+        $this->parameterProvider = $parameterProvider;
+    }
+
+    /**
+     * @param File[] $files
+     */
+    public function format(array $files): void
+    {
+        if(!$this->parameterProvider->provideBoolParameter(Option::ENABLE_EDITORCONFIG)) {
```
sabbelasichon, line 45: Should be disabled by default at the beginning. Otherwise we are annoying some people at the beginning ;-)

sabbelasichon:
Ready for review.

composer.json
```diff
@@ -32,6 +32,8 @@
...
+        "idiosyncratic/editorconfig": "dev-master",
```
TomasVotruba, line 36: I plan to do something similar as https://github.com/phpstan/phpstan-src where we could use this locked to specific commit. But untill then we need a stable tag here, so Rector can be installed

sabbelasichon: Yes, i am working on it. Sure.

TomasVotruba: Thanks :hugs:

sabbelasichon: Tagged version is out

packages/FileFormatter/FileFormatter.php
```diff
@@ -0,0 +1,68 @@
...
+            return;
+        }
+
+        foreach ($files as $file) {
+            if (! $file->hasChanged()) {
```
lulco, line 50: Why only changed files? :) If I want to replace all tabs to 4 spaces, I have to change all files first... If I enable formatting, I want to change all files

packages/FileFormatter/Formatter/JsonFileFormatter.php
```diff
@@ -0,0 +1,58 @@
...
+            $editorConfigConfiguration->getIndent(),
+            $editorConfigConfiguration->getEndOfLine()
+        );
+
+        $newFileContent .= $editorConfigConfiguration->getFinalNewline();
+
+        $file->changeFileContent($newFileContent);
+    }
+
+    public function createEditorConfigConfigurationBuilder(): EditorConfigConfigurationBuilder
```
lulco, line 48: Is this some default configuration? And it can be overriden by .editorconfig? And if have no .editorconfig file, this is used? Am I right?

packages/FileFormatter/FileFormatter.php
TomasVotruba: That's out of our scope. But phpstorm can do that, or editor config docker image ;)

lulco: Well, PHPStorm can also rename classes, methods and namespaces ;) It's just another tool you have to combine with Rector

sabbelasichon: The main reason why i implemented this feature was the fact that many people were complaining that touched files like composer.json are reformatted by rector. And i came up with the idea to take the editorconfig into account. I don´t want to touch all scanned files of the project. Only those touched by rector.

packages/FileFormatter/Formatter/JsonFileFormatter.php
sabbelasichon: Yes, exactly. This is the default configuration if no editorconfig exists and will be overridden if one exists.
But i think i should do some sniffing of the old content and take this as the default configuration. Should not be that hard to see if a line starts with spaces or tabs etc. Good catch. Thanks.

TomasVotruba:
Ready to merge?

packages/FileFormatter/FileFormatter.php
lulco: OK :)

sabbelasichon:
Not yet. I am doing the sniff part first today. This seems reasonable.

packages/FileFormatter/Formatter/JsonFileFormatter.php
lulco: Yes, I thougth about it (look to old content) too. But then I failed on mixed content e.g. 4 spaces, 2 spaces and tabs in one file :)

TomasVotruba:
All right :+1: 

I've added "ready-to-merge" label so you can mark PRs when ready.

packages/FileFormatter/Formatter/JsonFileFormatter.php
sabbelasichon: @lulco Yes, mixed content will be hard. I have chosen the "winner takes it all" approach.

sabbelasichon:
@TomasVotruba Ready to merge. I have streamlined the API and do the sniffing in the FileFormatter based on the original content. 

I cannot add a label here. Am i missing something?

TomasVotruba:
Check https://github.com/rectorphp/rector/invitations and try again. Should be fine :)

TomasVotruba:
Thank you :+1: :+1:

Issue #6276, opened 2021-05-01
[PHP 8.1] Add "never" type


TomasVotruba:
- RFC: https://wiki.php.net/rfc/noreturn_type

```diff
-function run()
+function run(): never
 {
     throw new Exception();
 }
```

PR #6283, opened 2021-05-02
[PHP 8.1] Add ReturnNeverTypeRector

TomasVotruba:
Closes https://github.com/rectorphp/rector/issues/6276

### DIFF
--- a/rules-tests/TypeDeclaration/Rector/ClassMethod/ReturnNeverTypeRector/Fixture/die_some.php.inc
+++ b/rules-tests/TypeDeclaration/Rector/ClassMethod/ReturnNeverTypeRector/Fixture/die_some.php.inc
@@ -1,0 +1,29 @@
+<?php
+
+namespace Rector\Tests\TypeDeclaration\Rector\ClassMethod\ReturnNeverTypeRector\Fixture;
+
+final class DieSome
+{
+    public function run()
+    {
+        echo 100;
+        die;
+    }
+}
+
+?>
+-----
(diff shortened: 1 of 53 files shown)


rules-tests/TypeDeclaration/Rector/ClassMethod/ReturnNeverTypeRector/Fixture/die_some.php.inc
```diff
@@ -0,0 +1,29 @@
...
+
+?>
+-----
+<?php
+
+namespace Rector\Tests\TypeDeclaration\Rector\ClassMethod\ReturnNeverTypeRector\Fixture;
+
+final class DieSome
+{
+    public function run(): never
```
staabm, line 22: when this rector would use `@return never` instead of the native return type, it would ease usage with php 7.x (without the need for a additional downgrading step) - would you accept a PR changing this rector from `:never` to `@return never` ?

TomasVotruba: I think other rules also use `@x` alternative, unless the type is nativelly supported in PHP, so yes, go for it :+1:

staabm: here we go https://github.com/rectorphp/rector-src/pull/296

PR #6286, opened 2021-05-02
Refactor ActiveRectorsProvider to autowired array

TomasVotruba:
cc @sabbelasichon Thanks to compiler pass the Rector rule are now excluded in 1 place :+1:


src/DependencyInjection/CompilerPass/RemoveSkippedRectorsCompilerPass.php
```diff
@@ -0,0 +1,58 @@
...
+        foreach ($containerBuilder->getDefinitions() as $id => $definition) {
+            if ($definition->getClass() === null) {
+                continue;
+            }
+
+            if (! in_array($definition->getClass(), $skippedRectorClasses, true)) {
+                continue;
+            }
+
+            $containerBuilder->removeDefinition($id);
```
TomasVotruba, line 31: Here the magic happens :)

src/DependencyInjection/CompilerPass/RemoveSkippedRectorsCompilerPass.php
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    /**
+     * @return string[]
+     */
+    private function resolveSkippedRectorClasses(ContainerBuilder $containerBuilder): array
+    {
+        $skipParameters = (array) $containerBuilder->getParameter(Option::SKIP);
```
TomasVotruba, line 40: This resolves all Rector classes in `SKIP` parameter in `rector.php`, e.g.:

```php
        // buggy in refactoring
        AddSeeTestAnnotationRector::class,

        StringClassNameToClassConstantRector::class,
        // some classes in config might not exist without dev dependencies
        SplitStringClassConstantToClassConstFetchRector::class,
    ]);
```

src/DependencyInjection/CompilerPass/RemoveSkippedRectorsCompilerPass.php
TomasVotruba: If the Rector class is in `SKIP` parameter, it will be excluded :+1:

PR #6287, opened 2021-05-02
[Scoped] Fixing scoped PHP 7.0

samsonasik:


samsonasik:
Got error: 

```bash
PHP Fatal error:  Uncaught TypeError: Return value of Rector\RectorInstaller\Plugin::activate() must be an instance of Rector\RectorInstaller\void, none returned in /home/runner/work/rector/rector/vendor/rector/rector-installer/src/Plugin.php:18
```

in build 7.0. It seems due return `void`

TomasVotruba:
This file need to have `void` removed:

```bash
vendor/rector/rector-installer/src/Plugin.php
```

TomasVotruba:
Try the downgrade locally on isolated case with `rector-installer`. It's much faster than wait for CI

samsonasik:
I tried running locally to `rector-build-php70` and seems not make chance for `void` return:

```bash
bin/rector process rector-build-php70/vendor/rector/rector-installer/src -n --config build/config/config-downgrade-php70.php 
 20/20 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

                                                                                                                        
 [OK] Rector is done!                                                                                    
```

it seems due to implements `Composer\Plugin\PluginInterface` which not exists in the project, that means it not updated. The composer itself seems need to require `composer/composer`

TomasVotruba:
I see. The PHP 7.1 works because composer types are accidentaly in sync.

The composer plugin package must be included manually in `composer.json` then.

PR #6291, opened 2021-05-02
[Scoper] Skip vendor/composer/composer/src/Plugin for prefixing

samsonasik:


samsonasik:
Got error:

```bash
PHP Warning:  Uncaught ErrorException: require(/home/runner/work/rector/rector/rector-build-php70/vendor/composer/../../rules-tests/Restoration/Rector/Use_/RestoreFullyQualifiedNameRector/Source/ShortClassOnly.php): failed to open stream: No such file or directory in /home/runner/work/rector/rector/rector-build-php70/vendor/composer/autoload_real.php:69
```

It seems 'tests` directories need to be removed after scoping.

PR #6295, opened 2021-05-02
[Scoper] Using different scoper.php for php 7.0

samsonasik:
revert original scoper.php


samsonasik:
@TomasVotruba if this still make php 7.1 fail, then there is something wrong with the git add process itself, as previously, it still works on php 7.1. Let's try.

.github/workflows/build_scoped_rector_php70.yaml
```diff
@@ -5,7 +5,7 @@ name: Build Scoped Rector PHP 7.0
 on:
     push:
         branches:
-            - main
+            - never
```
samsonasik, line 8: temporary disable scoped 7.0

samsonasik:
the php 7.1 scoped seems working again.

TomasVotruba:
Thanks :+1: 

https://github.com/rectorphp/rector-prefixed is passing :heavy_check_mark:

PR #6298, opened 2021-05-03
[DowngradePhp71] Downgrade :void for un-detected interface

samsonasik:
ref https://github.com/rectorphp/rector/runs/2490152390#step:10:6

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
         "php": "^7.3|^8.0",
         "ext-dom": "*",
         "ext-json": "*",
+        "composer/composer": "^2.0",
         "composer/semver": "^3.2",
         "composer/xdebug-handler": "^1.4|^2.0",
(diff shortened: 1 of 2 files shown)


samsonasik:
Got error:

```bash
1) Rector\Tests\DowngradePhp71\Rector\FunctionLike\DowngradeVoidTypeDeclarationRector\DowngradeVoidTypeDeclarationRectorTest::test with data set #0 (Symplify\SmartFileSystem\SmartFileInfo Object (...))

PHPStan\BetterReflection\Reflector\Exception\IdentifierNotFound: PHPStan\BetterReflection\Reflection\ReflectionClass "Composer\Plugin\PluginInterface" could not be found in the located source
```

samsonasik:
for downgrade `: void` that implements `Composer\Plugin\PluginInterface`, it seems need `composer require composer/composer`

PR #6310, opened 2021-05-03
[Scoper] Remove composer/composer after not used in scoped

samsonasik:


samsonasik:
Got error:

```bash
PHP Notice:  Uninitialized string offset: -1 in /home/runner/work/rector-prefixed-php70/rector-prefixed-php70/vendor/symfony/dependency-injection/ContainerBuilder.php on line 816
```

on string fetch array negative `$id[-1]`

PR #6311, opened 2021-05-03
[DowngradePhp71] Allow downgrade negative offset string on variable/PropertyFetch/StaticPropertyFetch

samsonasik:
ref https://github.com/rectorphp/rector/pull/6310#issuecomment-831321508 . It is no need `string` type check on variable as if variable is not string, it will got syntax error https://3v4l.org/TM68C


samsonasik:
Allow on PropertyFetch and StaticPropertyFetch as well.

PR #6312, opened 2021-05-03
[Scoper] Patch : void in vendor via scoper-php70 config

samsonasik:
ref https://github.com/rectorphp/rector-prefixed-php70/runs/2493424282#step:4:6 . Some vendor dependencies seems overlapped make `: void` not updated. use scoper config to patch

### DIFF
--- a/scoper-php70.php
+++ b/scoper-php70.php
@@ -159,6 +159,18 @@ return [
             );
         },
 
+        function (string $filePath, string $prefix, string $content): string {
+            if (! Strings::contains($filePath, 'vendor/')) {
+                return $content;
+            }
+
+            // @see https://regex101.com/r/r3AJFl/1
+            return Strings::replace(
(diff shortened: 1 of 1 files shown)


samsonasik:
@TomasVotruba I am not sure about the following php 7.0 error:

```bash
Run php bin/rector list --ansi

 [ERROR] Return value of                                                        
         RectorPrefix20210503\Symfony\Component\DependencyInjection\Loader\ProtectedPhpFileLoader::RectorPrefix20210503\{closure}() must be an instance of RectorPrefix20210503\void, none returned                            
```

can't find `: void` in the vendor.

samsonasik:
it seems `: void` need to be removed from `config/set` as well:

https://github.com/rectorphp/rector-prefixed-php70/blob/3e69ef128c4c4da6047ba9cc25264ec169dcb68f/config/set/action-injection-to-constructor-injection.php#L9

PR #6324, opened 2021-05-04
[Preload] Fixes root path when rector installed as dependency

### LINKED ISSUE #6323: preload.php breaks bin/rector when installed as a dependency
# Bug Report

| Rector version | v0.10.18              |

When I upgraded rector to the latest version, my build broke. It turned out that `bin/rector` requires `preload.php` which in turn requires some files from its `vendor` directory. Not **my** vendor directory (I installed rector as composer dependency) but **rector's** `vendor` directory e.g. 
```
require_once __DIR__ . '/vendor/nikic/php-parser/lib/PhpParser/Lexer.php';
```
which is then resolved as `{my_project_dir}/vendor/rector/rector/vendor/nikic/php-parser/lib/PhpParser/Lexer.php` when rector is run.

image

samsonasik:
Fixes #6323


samsonasik:
@TomasVotruba this is a patch for preload.php used on rector installed as project dependency. I tested it locally and it working ok.

 The ECS error is unrelated, due to `php-cs-fixer` dependency

TomasVotruba:
I see. The paths are correct, as it should load Rector's vendor, not project.

The preload.php should be included only on tests here or in prefixed version then

preload.php
```diff
@@ -2,240 +2,263 @@
...
+            self::$rootPath = __DIR__;
+            return self::$rootPath;
+        }
+
+        self::$rootPath = __DIR__ . '/../../..';
+        return self::$rootPath;
+    }
+}
+
+RectorVendor::getRootPath() . '/vendor/nikic/php-parser/lib/PhpParser/Lexer.php';
```
alanbem, line 28: Did you skip `require_once` on purpouse?

alanbem:
> The preload.php should be included only on tests here or in prefixed version then

What if `composer` guys will come up with some standardized support for preload files? Is rector even kind of library that should use preload functionality?

alanbem:
also should you preload files you do not own?

preload.php
samsonasik: Ooops, I missed that. As @TomasVotruba's comment https://github.com/rectorphp/rector/pull/6324#issuecomment-831855738 . It need only loaded on tests / prefixed version.

alanbem (approved):

samsonasik:
@TomasVotruba implemented, I applied to tests bootstrap https://github.com/rectorphp/rector/pull/6324/commits/fc190180068ff0dd93a11b5d6eac86f6d3e6b6a7 and rector prefixed https://github.com/rectorphp/rector/pull/6324/commits/f4ec0a274eda9529af5c1509dd5c1dc6bd738ba4

PR #6325, opened 2021-05-04
[php-parser] upgrade to 4.10.5 - add src/constants.php to mimic new PHP 8.1 tokens; update test with unicode

TomasVotruba:
Ref https://github.com/rectorphp/rector/issues/6326

### DIFF
--- a/bin/rector.php
+++ b/bin/rector.php
@@ -9,7 +9,6 @@ use Rector\Core\DependencyInjection\RectorContainerFactory;
 use Rector\Core\HttpKernel\RectorKernel;
 use Symplify\PackageBuilder\Console\ShellCode;
 use Symplify\PackageBuilder\Reflection\PrivatesCaller;
-use Tracy\Debugger;
 
 // @ intentionally: continue anyway
 @ini_set('memory_limit', '-1');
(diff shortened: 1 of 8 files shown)


bin/rector.php
```diff
@@ -21,15 +20,19 @@
...
 $autoloadIncluder = new AutoloadIncluder();
 $autoloadIncluder->includeDependencyOrRepositoryVendorAutoloadIfExists();
 
-if (file_exists(__DIR__ . '/../vendor/scoper-autoload.php')) {
-    // make local php-parser a priority to avoid conflict
+// load local php-parser only in prefixed version or development repository
+$isDevelopmentRepository = file_exists(__DIR__ . '/../phpstan-for-rector.neon');
+if (file_exists(__DIR__ . '/../vendor/scoper-autoload.php') || $isDevelopmentRepository) {
     require_once __DIR__ . '/../preload.php';
 }
```
TomasVotruba, line 32: @samsonasik The local repository detection is needed too, as we have `phpstan.phar` autoload that is loaded first and creates conflict with our `php-parser`

bin/rector.php
```diff
@@ -40,11 +43,6 @@
 
 $rectorConfigsResolver = new RectorConfigsResolver();
 
-// for simpler debugging output
-if (class_exists(Debugger::class)) {
-    Debugger::$maxDepth = 2;
-}
```
TomasVotruba, line 46: Already included in `dump_node()` function

bin/rector.php
samsonasik: @TomasVotruba thank you 👍

PR #6330, opened 2021-05-04
use php-parser 4.10.4 and stable phpstan with same version

TomasVotruba:


composer.json
```diff
@@ -36,9 +36,9 @@
         "nette/caching": "^3.1",
         "nette/robot-loader": "^3.4",
         "nette/utils": "^3.2",
-        "nikic/php-parser": "4.10.5",
+        "nikic/php-parser": "4.10.4",
         "phpstan/phpdoc-parser": "^0.5.4",
-        "phpstan/phpstan": "dev-master#7a8ebc3",
+        "phpstan/phpstan": "0.12.85",
```
TomasVotruba, line 41: @ondrejmirtes FYI, I'll revert to older PHPStan for now then.

ondrejmirtes: Yeah, you should always wait until PHPStan is on the same PHP-Parser version, I hope it's not too much trouble 👍

TomasVotruba: It's ok. I just wanted to try enum migration on stable :) 

I think we'll shift to `rector-src` approach soon to keep more stable releases.

Issue #6347, opened 2021-05-05
[0.10.19] NetteToSymfony error


Guikingone:
# Bug Report

| Subject        | Details                              |
| :------------- | :----------------------|
| Rector version |0.10.19                           |
| Installed as   | composer dependency  |

Hi everyone 👋🏻 

I'm facing an issue when installing Rector on GA with the following version in the `composer.json`: 

```json
"rector/rector": "0.10.19"
```

## Minimal PHP Code Causing Issue

Installing rector `0.10.19` on Github Actions: 

```php
  - Locking rector/rector (0.10.19)
  - Locking rector/rector-cakephp (0.10.4)
  - Locking rector/rector-doctrine (0.10.6)
  - Locking rector/rector-installer (0.10.1)
  - Locking rector/rector-laravel (0.10.2)
  - Locking rector/rector-nette (0.10.9)
  - Locking rector/rector-phpunit (0.10.8)
  - Locking rector/rector-symfony (0.10.6)
 ```
 
 When launching `vendor/bin/rector --dry-run --clear-cache`, here's the output: 
 
 ```bash
 Cannot autowire service                                                
         "Rector\NetteToSymfony\NodeAnalyzer\ClassMethodRenderAnalyzer":        
         argument "$thisTemplatePropertyFetchAnalyzer" of method "__construct()"
         references class                                                       
         "Rector\Nette\NodeAnalyzer\ThisTemplatePropertyFetchAnalyzer" but no   
         such service exists.                                                   

Error: Process completed with exit code 1.
 ```
 
 Full output: https://github.com/Guikingone/SchedulerBundle/pull/99/checks?check_run_id=2509745314

The situation is quite strange as it works locally, I've tried to install rector on previous patch versions and the error still the same 🙁 

Thanks again for the help and have a great day 🙂

sabbelasichon:
I guess you install it via --no-scripts. We have added the installer a while ago. but it requires you to run composer with scripts enabled.

sabbelasichon:
@TomasVotruba i try to come up with a fallback solution or otherwise it is very confusing

Guikingone:
The `--no-scripts` option isn't used 🤔 

```bash
Run ramsey/composer-install@v1
  with:
    dependency-versions: highest
    composer-options: --prefer-stable
Composer cache directory found at /home/runner/.cache/composer
Composer working directory set to project root.
Using highest versions of dependencies
PHP version is 7.4.18
Operating system is linux
[...]

PR #6359, opened 2021-05-05
[TASK] Add fallback mechanism to load default extensions

sabbelasichon:
- If the rector-installer failed or is not executed due to --no-scripts flag at least the old default extensions are loaded

Resolves: #6347

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -12,7 +12,7 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     $containerConfigurator->import(__DIR__ . '/services-packages.php');
     $containerConfigurator->import(__DIR__ . '/parameters.php');
 
-    $extensionConfigResolver = new ExtensionConfigResolver();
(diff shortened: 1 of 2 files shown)


TomasVotruba:
I'm not sure having 2 ways to install packages is a good approach. Devs need to learn the package is installing extension, like we did with PHPStan installer and flex installer.

IMO that reported bug is a wrong abstraction in our part. Mismatch of tags with non-existing class or something.

sabbelasichon:
@TomasVotruba It is meant as a fallback solution to circumvent this error. It is always the rector-nette package. Maybe it is sufficient to only this package here. I have never looked closely to the bug.

TomasVotruba:
That's the composer install convention on any plugin. I've decoupled NetteToSymfony to own package to remove core dependency on extension: https://github.com/rectorphp/rector/pull/6360

PR #6362, opened 2021-05-05
Config is included multiple times

lulco:


TomasVotruba:
Here is a build where the var_dump() can be seen 4times - https://github.com/rectorphp/rector/runs/2513048021

image

There should be similar test on Symfony project with some PHP config. Maybe it's Symfony defaults :)

TomasVotruba:
I gave this some debugging and this is actually expected.

That's because before loading `rector.php` to Rector DI itself, the config is analysed with small service. This services is looking for `SETS` parameter and extracts all configs from there.

https://github.com/rectorphp/rector/blob/812ff5b654c7e706220241f0b67e3f815068b3a5/src/Bootstrap/SetConfigResolver.php#L26-L31

In the future we should move to direct imports(), as that what's actually hapenning :) 


```diff
return static function (ContainerConfigurator $containerConfigurator): void {
-    $parameters = $containerConfigurator->parameters();
-    $parameters->set(Option::SETS, [SetList::DEAD_CODE]));
+    $containerConfigurator->imports(SetList::DEAD_CODE));
};

```

PR #6364, opened 2021-05-05
[DowngradePhp70] Add SplitGroupedUseImportsRector + allow to return list of stmt[] in refactor()

TomasVotruba:


u01jmg3:
Would it be worth having the opposite? A Rector that [groups use declarations](https://wiki.php.net/rfc/group_use_declarations) (as opposed to removing them).

TomasVotruba:
The trend is rather to separate items to one per line. E.g. from grouped properties, constants and variables to standalone lines.

If you prefer the opposite, you can create custom Rector rule to enforce it on your code base of course.

u01jmg3:
Just to confirm, you mean the trend (now 6 years since the RFC was approved) is not to group use declarations? e.g. one per line like you have always been able to define them

> _If you prefer the opposite, you can create custom Rector rule to enforce it on your code base of course._

Thanks, I may explore a custom rector. However, I do find it odd that the only related rector that exists **removes** newer use declaration syntax rather than adding it.

TomasVotruba:
> Just to confirm, you mean the trend (now 6 years since the RFC was approved) is not to group use declarations?

From what I see on PHP open-source and projects I work with, yes. The PSR-12 comply with this direction. It's also easier to work with those in context of AST parsing and priority assign.

> However, I do find it odd that the only related rector that exists

That's quite normal. The most of rules does not exists just because nobody added them yet :) there is no mystery.

PR #6371, opened 2021-05-06
[scoped on PHP 7.0] bin/rector run with extracted phpstan

TomasVotruba:


composer.json
```diff
@@ -96,7 +96,8 @@
         },
         "files": [
             "src/functions/node_helper.php",
-            "src/constants.php"
+            "src/constants.php",
+            "vendor/phpstan/phpstan-extracted/vendor/autoload.php"
```
TomasVotruba, line 100: This will load phpstan's autoload and its own `vendor/nikic/php-parser`.
This package is in the same version, so Rector's `preload.php` would conflict with same classes.

That's why its nulled bellow.

.github/workflows/build_scoped_rector_php70.yaml
```diff
@@ -40,8 +40,14 @@ jobs:
...
-            # original phpstan/phpstan is now duplicated, we don't need it now
-            -   run: composer remove phpstan/phpstan
+            # "composer remove phpstan/phpstan" original phpstan/phpstan is now duplicated, we don't need it now -required by phpstan extensions, cannot be removed like this
+
+            # to avoid double packing the same code
+            -   run: rm -rf vendor/bin/phpstan
+            -   run: rm -rf vendor/phpstan/phpstan/phpstan.phar
+
+            # THIS MUST WORK
+            -   run: bin/rector downgrade-paths
```
TomasVotruba, line 50: So far so good :) 

image

preload.php
```diff
@@ -2,240 +2,240 @@
 
 declare(strict_types = 1);
 
-require_once __DIR__ . '/vendor/nikic/php-parser/lib/PhpParser/Lexer.php';
```
TomasVotruba, line 5: We'll need to load own `preload.php` anyway to make sure it won't conflict with project's `php-parser`, make this CI to pass:
https://github.com/rectorphp/rector-prefixed/runs/2518375775

Not sure how to solve that now :/

composer.json
TomasVotruba: This line must be added via config CLI command: `composer config ...`

In: `.github/workflows/build_scoped_rector_php70.yaml`

Now it breaks other builds

TomasVotruba:
@samsonasik Here is a prototype of PHP 7.0 build that doesn't use on `phpstan-src` but extracts `phpstan.phar`. It solves lots of issues with installation and scoping. Feel free to inspire here

ondrejmirtes:
As I said, your job might be easier if you contribute the PHP 7.0 downgrading to PHPStan compiling itself.

TomasVotruba:
Might do later, but first I need to Rector work on PHP 7.0. Now it can handle PHP 7.1 tops. This is just theory :)

samsonasik:
@TomasVotruba thank you 👍

composer.json
```diff
@@ -25,7 +25,7 @@
         "phpstan/phpstan-phpunit": "^0.12.18",
         "rector/rector-cakephp": "^0.10.4",
         "rector/rector-doctrine": "^0.10.6",
-        "rector/rector-installer": "^0.10.1",
+        "rector/extension-installer": "^0.10.2",
```
TomasVotruba, line 28: @sabbelasichon Btw, I flipped the name to `extension-installer`, as `rector-<x>` name is used for framework-specific sets. Like `rector-nette`, `rector-symfony` etc.
This should make also clear it does not install rector, but extensions.

Issue #6361, opened 2021-05-05
Running the process command without defining a rule is still processing


sabbelasichon:
As @lulco already mentioned in one issue the RenameClassNonPhpRector is always registered and therefore running rector without defining a rule be it a Php one or from another FileProcessor the MissingRectorRulesReporter is not popping up. 

I guess the simplest solution would be to do this:
```php
$activeRectors = array_filter($this->rectors, function (RectorInterface $rector): bool {
            return ! $rector instanceof PostRectorInterface && ! $rector instanceof NonPhpRectorInterface;
        });
```

TomasVotruba:
This shoud be refactored to opt-in rule instead.

PR #6373, opened 2021-05-06
[TASK] Remove RenameClassNonPhpRector from the defaults

sabbelasichon:
Resolves: #6361

### DIFF
--- a/config/services.php
+++ b/config/services.php
@@ -29,7 +29,6 @@ use Rector\BetterPhpDocParser\PhpDocParser\BetterPhpDocParser;
 use Rector\BetterPhpDocParser\PhpDocParser\BetterTypeParser;
 use Rector\Caching\Cache\NetteCacheFactory;
 use Rector\Core\Console\ConsoleApplication;
-use Rector\Core\NonPhpFile\Rector\RenameClassNonPhpRector;
 use Rector\Core\PhpParser\Parser\NikicPhpParserFactory;
 use Rector\Core\PhpParser\Parser\PhpParserLexerFactory;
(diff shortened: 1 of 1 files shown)


sabbelasichon:
@TomasVotruba We have to be aware that this is breaking form some people. So i added it to the template file to show people that this has changed.

templates/rector.php.dist
```diff
@@ -19,6 +19,9 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     // get services (needed for register a single rule)
     // $services = $containerConfigurator->services();
 
+    // Non PHP Changes - To keep the old behaviour =< 0.10.5 add this
+    // $services->set(RenameClassNonPhpRector::class);
```
TomasVotruba, line 23: This can be removed, as template is always generated with the latest version.

TomasVotruba:
Btw, what set configs this rule should be included in?

sabbelasichon:
I don´t know. Never used this. Where does it come from?

TomasVotruba:
I think from `RenameClassRector` where the class should be renamed in framework specific non-PHP files:

- Laravel → `.blade.php`
- Nette → `.neon`
- Symfony → `.xml`, `.yaml`

sabbelasichon:
If you like we could add it to the sets of these framework packages and we should be fine. 
Btw. the whole documentation part goes here: #6374

TomasVotruba:
> add it to the sets of these framework packages and we should be fine

That would be great :+1:  :bow: 

The 3 above should be fine.

PR #6376, opened 2021-05-06
[TASK] Take glob pattern for files into account

sabbelasichon:
For now you cannot use the glob pattern on a file basis. But as we now have the FileProcessors it would be awesome to speed up the processing by selecting also files with a glob pattern

### DIFF
--- a/src/FileSystem/FilesFinder.php
+++ b/src/FileSystem/FilesFinder.php
@@ -9,6 +9,7 @@ use Symfony\Component\Finder\Finder;
 use Symfony\Component\Finder\SplFileInfo;
 use Symplify\Skipper\SkipCriteriaResolver\SkippedPathsResolver;
 use Symplify\SmartFileSystem\FileSystemFilter;
+use Symplify\SmartFileSystem\FileSystemGuard;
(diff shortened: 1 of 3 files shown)


TomasVotruba:
Hi, I like it :+1: 

This behavior should be already handled in this method:
https://github.com/rectorphp/rector/pull/6376/files#diff-c1b4d526cfa7d17990bde4cf0146bcf748e35382d13d14ca350a21820d2df256R107

Could you check it if there is some bug?

sabbelasichon:
Maybe it is a feature and not a bug ;-)

But for now only glob pattern for directories is allowed. File patterns have to be real file paths.

TomasVotruba:
I see. It should be merged in one service, now we have 2 do to the glob job.

sabbelasichon:
I know. But i was scared to change it in the other package due to side effects.

TomasVotruba:
Go for it :+1:

sabbelasichon:
I have streamlined the globbing without creating a whole new service. Not sure how to start without opening Pandora's Box which takes too much time for right now.

Issue #6380, opened 2021-05-07
Show Progressbar for every FileProcessor


sabbelasichon:
At the moment the Progressbar is only shown for the PhpFileProcessor. Would be nice to show the ProgressBar for every FileProcessor. Would like to create a PR for that.

TomasVotruba:
Hi, I wonder what would the added value that for end-user?

I mean we don't run have multiple progress bars in PHPUnit for unit tests, functional tests, integration tests...

If we clutter user with information, it musts bring them value.

sabbelasichon:
I see and agree. But now there is no feedback for user when running FileProcessors other than the Php one.

TomasVotruba:
There should be diff files in case of change.

sabbelasichon:
I am gonna think about it.

sabbelasichon:
The diff is there. That´s true. But until the diffs are shown no user feedback or application status is shown if you only run other FileProcessors other than the Php one. I know everything started with Php but now we have also other formats. Would really like to have some ProgressBar for other formats. Otherwise you gonna wait some time depending on the amount of files which are processed.

sabbelasichon:
@TomasVotruba To not spoil the list on issues here, i would like to close this one here. In practice it is not a big deal. 

I am still struggling with the part to split up the PhpFileProcessor to run the files one by one. Maybe someday we will have an idea how to do it. But for now i don´t care.

PR #6381, opened 2021-05-07
[TASK] Add progressbar for all FileProcessors

sabbelasichon:
Resolves: #6380

### DIFF
--- a/rules/Composer/Application/FileProcessor/ComposerFileProcessor.php
+++ b/rules/Composer/Application/FileProcessor/ComposerFileProcessor.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Rector\Composer\Application\FileProcessor;
 
 use Rector\Composer\Contract\Rector\ComposerRectorInterface;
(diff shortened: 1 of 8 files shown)


sabbelasichon:
We have a chicken egg problem here. Because the NeonFileProcessor must implement the new contract. What do you think about it, besides the fact that every test fails?

rules/Composer/Application/FileProcessor/ComposerFileProcessor.php
```diff
@@ -45,14 +45,14 @@ public function __construct(
     /**
      * @param File[] $files
      */
-    public function process(array $files): void
+    public function process(array $files, callable $advanceProgressBar): void
```
TomasVotruba, line 48: `callable` is very poor contract that would shoot you back with confusing error message. In general, we should code against interface.

sabbelasichon:
I am on it. I think i will come up with a good and clear solution after i revisited the current processors again

sabbelasichon:
Closing this in favour of a new try

PR #6382, opened 2021-05-07
[Scoped] Add more paths to downgrade, tweak scope configs 

samsonasik:
reduce syntax errors from : 383 to 14: 

Before:

```
Syntax error found in 383 files
```

After:

```
Syntax error found in 14 files
```

### DIFF
--- a/scoper-php70.php
+++ b/scoper-php70.php
@@ -77,6 +77,28 @@ return [
             );
         },
 
+        function (string $filePath, string $prefix, string $content): string {
+            if (! Strings::contains($filePath, 'vendor/')) {
+                return $content;
+            }
+
+            // see https://regex101.com/r/PDGN3K/1
+            $content = Strings::replace(
+                $content, '
+                #: \?\\\\.*#',
+                ''
+            );
+
+            // see https://regex101.com/r/P7nbpU/1
+            $content = Strings::replace(
+                $content, '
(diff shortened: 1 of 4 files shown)


scoper-php70.php
```diff
@@ -77,6 +77,28 @@ function (string $filePath, string $prefix, string $content): string {
...
+            );
+
+            // see https://regex101.com/r/P7nbpU/1
+            $content = Strings::replace(
+                $content, '
+                #\(\?\\.*\s+#',
+                '('
+            );
+
+            return $content;
```
TomasVotruba, line 99: This must be handled by Rector. If not, there is a bug somewhere

samsonasik: this for handilng phpstan's vendor of `ondrejmirtes/better-reflection/src` that somehow not downgrade the code.

TomasVotruba: All packages must be downgraded by Rector. `scoper.php` is only for tweaking class names for autoload edge cases.

We need to figure out why this downgrade is missed.

samsonasik: Some probably caused by exclude config : 

https://github.com/rectorphp/rector/blob/6ddb1451c8f8e6562a09c3a38d1df9492f01c85f/build/config/config-downgrade-php70.php#L51-L63

other probably need re-sort the directories.

I will check more.

PR #6383, opened 2021-05-07
downgrade covariant mix bug

TomasVotruba:


tests/Issues/Fixture/coariant_mixture.php.inc
```diff
@@ -0,0 +1,33 @@
...
+{
+    /**
+     * @param string $token
+     */
+    private function parseArgument($token, array $all)
+    {
+        reset($all);
+        if (($inputArgument = isset($all[$key = key($all)]) ? $all[$key = key($all)] : null) && 'command' === $inputArgument->getName()) {
+        }
+    }
```
TomasVotruba, line 30: @samsonasik BTW, today I find one nasty bug in Rector. 
This might be one reason why the downgrade rules were not working sometimes.

This code previously failed and missed all the removed types. Causing missmatch:

image

packages/PostRector/Collector/NodesToAddCollector.php
```diff
@@ -132,14 +142,22 @@ private function resolveNearestExpressionPosition(Node $node): string
...
+        }
+
         $parent = $node->getAttribute(AttributeKey::PARENT_NODE);
         if ($parent instanceof Return_) {
             return spl_object_hash($parent);
         }
 
         $foundNode = $this->betterNodeFinder->findParentTypes($node, [Expression::class, Stmt::class]);
-        if ($foundNode === null) {
-            $foundNode = $node;
```
TomasVotruba, line 142: Not sure why I put this here (probably not knowing difference between Stmt and Expr in that time), but it was asking for a bug :thinking:

tests/Issues/Fixture/coariant_mixture.php.inc
samsonasik: @TomasVotruba thank you 👍

PR #6417, opened 2021-05-14
[Docs] Direct remove docs/ and update image & link

samsonasik:
Alternative with direct remove docs and update image and link of https://github.com/rectorphp/rector-src/pull/35

Closes https://github.com/rectorphp/rector-src/pull/34 
Closes https://github.com/rectorphp/rector-src/pull/35


samsonasik:
alternative of https://github.com/rectorphp/rector-src/pull/35 for reference

TomasVotruba:
The content should be untouched, only the location should be changed

PR #6425, opened 2021-05-17
remove proposal of rector-prefixed installation

mente:
Since 0.11 it's not relevant


samsonasik:
@mente please create PR to https://github.com/rectorphp/rector-src instead. Thank you.

mente:
I checked rector-src first and didn't find README source there. Can you point me where is it stored there?

samsonasik:
Its here https://github.com/rectorphp/rector-src/blob/main/build/target-repository/README.md

PR #6430, opened 2021-05-18
allow latest versions of phpstan

### LINKED ISSUE #6428: phpstan is out of date
Is there a reason to pin phpstan dependency to patch version?
If this can't be changed, can you please update phpstan/phpstan to 0.12.88?

garak:
Fix #6428


composer.json
```diff
@@ -21,7 +21,7 @@
     },
     "conflict": {
         "phpstan/phpdoc-parser": "<=0.5.3",
-        "phpstan/phpstan": "<=0.12.82",
+        "phpstan/phpstan": "<=0.12.88",
```
TomasVotruba, line 24: This seems a bit confusing.
How can composer require 0.12.86 and conflict with 0.12.88 and lower at the same time?

garak: my bad, I looked in the wrong section

samsonasik:
@garak please send PR to https://github.com/rectorphp/rector-src instead. You can look the composer.json on https://github.com/rectorphp/rector-src/blob/main/build/target-repository/composer.json

garak:
> @garak please send PR to https://github.com/rectorphp/rector-src instead. You can look the composer.json on https://github.com/rectorphp/rector-src/blob/main/build/target-repository/composer.json

Maybe this is worth a mention in contributing file?

PR #6437, opened 2021-05-19
minor: Update issue templates' rector version

staabm:
Use a more recent version

### DIFF
--- a/.github/ISSUE_TEMPLATE/1_Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/1_Bug_report.md
@@ -10,7 +10,7 @@ about: Report errors and problems
 
 | Subject        | Details                                                         |
 | :------------- | :---------------------------------------------------------------|
-| Rector version | e.g. v0.9.5 (invoke `vendor/bin/rector --version`)              |
+| Rector version | e.g. v0.11.5 (invoke `vendor/bin/rector --version`)              |
 | Installed as   | composer dependency / prefixed Rector                           |
 
 <!-- Please describe your problem here. -->


.github/ISSUE_TEMPLATE/1_Bug_report.md
```diff
@@ -10,7 +10,7 @@ about: Report errors and problems
 
 | Subject        | Details                                                         |
 | :------------- | :---------------------------------------------------------------|
-| Rector version | e.g. v0.9.5 (invoke `vendor/bin/rector --version`)              |
+| Rector version | e.g. v0.11.5 (invoke `vendor/bin/rector --version`)              |
 | Installed as   | composer dependency / prefixed Rector                           |
```
staabm, line 14: this line can be dropped, because there is only a single install method as of 0.11, right?

TomasVotruba: Yes, it's from the old times :laughing:

TomasVotruba:
These files should be updated in the development repository though.
We use this as a release read-only

This is the directory that is pushed here into the root: https://github.com/rectorphp/rector-src/tree/main/build/target-repository

staabm:
Here we go https://github.com/rectorphp/rector-src/pull/77

samsonasik:
@TomasVotruba I think we can use auto-closer github action for example: 

https://github.com/laminas/laminas-db/blob/7663f838059e27da94df02fe4aa88c2c7fd7aa1c/.github/workflows/auto-close.yml

It needs organization token which seems can be saved at `.github` repository in org level, eg: https://github.com/laminas/.github

The example result is : https://github.com/laminas/laminas-db/issues/189#issuecomment-819573099

TomasVotruba:
We can give it a shot.
Could you add it?

samsonasik:
I've added `.github` repo https://github.com/rectorphp/.github  ,currently add my personal access token to `secrets`, let's try.

PR #6442, opened 2021-05-20
Test PR to be auto-closed

samsonasik:


samsonasik:
@TomasVotruba when pull request from fork, it currently got error: 

```bash
Run roots/issue-closer@v1.1
  with:
    pr-pattern: ^exact-string-will-never-match$
    pr-close-message: This repository content is generated by rectorphp/rector-src. Please open pull request to [rectorphp/rector-src](https://github.com/rectorphp/rector-src).

Error: Input required and not supplied: repo-token
```

It seems the `${{ secrets.ACCESS_TOKEN }}` not exists.

TomasVotruba:
The access token is not available to external contributors. It's a Github policy to avoid removing repository from a PR

samsonasik:
@TomasVotruba Should GITHUB_TOKEN will work then? If not, could you add organization token to https://github.com/rectorphp/.github/settings as I don't have access to add it.

118899978-1cf51e80-b93a-11eb-8391-d6ac966a82a9

Thank you.

For demo that it works on handling external contributor https://github.com/laminas/laminas-db/pull/179#issuecomment-784548256

PR #6444, opened 2021-05-20
Test PR part 2

samsonasik:
try with GITHUB_TOKEN

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # Rector - Instant Upgrades and Automated Refactoring
 
+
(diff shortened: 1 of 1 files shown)


samsonasik:
@TomasVotruba using `GITHUB_TOKEN` seems not work:

```bash
Run roots/issue-closer@v1.1
  with:
    repo-token: ***
    pr-pattern: ^exact-string-will-never-match$
    pr-close-message: This repository content is generated by rectorphp/rector-src. Please open pull request to [rectorphp/rector-src](https://github.com/rectorphp/rector-src).

Error: Resource not accessible by integration
```

It seems we need the Organization access token.

samsonasik:
It seems we can try other alternative which seems doesn't require token https://github.com/superbrothers/close-pull-request

PR #6512, opened 2021-06-17
Fix tiny rector_rules_overview.md typo

adrienbrault:


adrienbrault:
😲 https://github.com/rectorphp/rector-src/pull/239

Issue #6517, opened 2021-06-22
TwigSetList::TWIG_UNDERSCORE_TO_NAMESPACE does not affect Twig_TemplateWrapper


HenkPoley:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 5e57bb61bcae1bb015819d2e3f47bbc9bbdaab5c             |

## Minimal PHP Code Causing Issue

```php
/**
 * @param $templateFile
 * @param string $pageTitle
 * @return Twig_TemplateWrapper
 */
function getTemplate($templateFile, $pageTitle = '')
{
}
```

## Expected Behaviour

```php
/**
 * @param $templateFile
 * @param string $pageTitle
 * @return \Twig\TemplateWrapper
 */
function getTemplate($templateFile, $pageTitle = '')
{
}
```

HenkPoley:
Slightly off-topic, but it would be nice if the bot pointed me to:

https://github.com/rectorphp/rector-symfony/blob/main/config/sets/twig/twig-underscore-to-namespace.php

Or: https://github.com/rectorphp/rector-symfony/

TomasVotruba:
> Slightly off-topic, but it would be nice if the pull-request closer bot pointed me to:



Good idea! :+1: 

I don't know if the action allows it conditional text, but if you can find a way, here is the config for it:
https://github.com/rectorphp/rector/blob/main/.github/workflows/auto_closer.yaml

TomasVotruba:
Resolved by https://github.com/rectorphp/rector-symfony/pull/20

Btw, tip for automated closing on GitHub: https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

E.g.

```
Closes https://github.com/rectorphp/rector/issues/6517
```

↓

*Closes https://github.com/rectorphp/rector/issues/6517*

That would save us this manual work after the PR is merged :+1:

HenkPoley:
🤔 it appears [close-pull-request](https://github.com/superbrothers/close-pull-request) can't do that.

But maybe some [jql](https://github.com/yamafaktory/jql) on the packages.lock could surface `packages.source.{name,source.url}` (imaginary query)

TomasVotruba:
Maybe the static text could be updated to mention the split repositories. Would that help you to find them fsater? :+1:

[...]

PR #6518, opened 2021-06-22
Twig_TemplateWrapper -> \Twig\TemplateWrapper

HenkPoley:
For Issue #6517

(Maybe also put some newlines in that huuuuge array?)


HenkPoley:
It would be nice if the bot pointed me to:

https://github.com/rectorphp/rector-symfony/blob/main/config/sets/twig/twig-underscore-to-namespace.php

Or: https://github.com/rectorphp/rector-symfony/

PR #6580, opened 2021-07-19
Update CONTRIBUTING.md

NotionCommotion:
Give alternate approach to composer complete-check


TomasVotruba:
This repository is read only.

The correct file to place this is - https://github.com/rectorphp/rector-src/blob/main/CONTRIBUTING.md

The file here should be removed actually, as this repository is ready only

PR #6672, opened 2021-09-01
Allowing future versions of phpstan

ThomasLandauer:
phpstan 0.12.97 was just released today https://github.com/phpstan/phpstan/releases
So right now, rector cannot be installed :-(


ThomasLandauer:
See https://github.com/rectorphp/rector-src/pull/810

PR #6856, opened 2021-12-11
Correct link

PhilETaylor:
Fixes the correct link, without this change the link goes to 

https://github.com/rectorphp/rector/blob/main/docs/github.com/symplify/easy-coding-standard

which is incorrect.


samsonasik:
Please update on this https://github.com/rectorphp/rector-src/blob/main/build/target-repository/docs/auto_import_names.md

PR #7114, opened 2022-04-16
Update rector_rules_overview.md

javaDeveloperKid:


javaDeveloperKid:
I can't find this file inside rectorphp/rector-src.

javaDeveloperKid:
Got it. But it looks like GitHub's `tt`  keyboard shortcut does not find this file.

FYI @TomasVotruba

PR #8363, opened 2023-12-14
Update .gitattributes

JesusValera:
## 📚 Description

- Remove broken link
- Ignore `/stubs-rector` instead of `/stubs` folder (which doesn't exist in the root)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -14,13 +14,12 @@ CONTRIBUTING.md export-ignore
 CHANGELOG.md export-ignore
 CODE_OF_CONDUCT.md export-ignore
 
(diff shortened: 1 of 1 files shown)


TomasVotruba:
I think only admins can change this, as it's often used by accident.
I'll look into it :+1: 

@samsonasik Hi Abdul, btw, do you have some idea why we have the `/stubs-rector` directory here? It seems unused and can be removed

samsonasik:
`stubs-rector` is used by rector scoped, see https://github.com/rectorphp/rector-src/blob/5a1512b4c0939c1b8f9ab7a1c0eb40029acb1537/src/Autoloading/BootstrapFilesIncluder.php#L45

JesusValera:
Then there is not much to do here.

Thanks! 🙌🏼

